PR #3, opened 2020-06-16
Refactor config

kiootic:
ref #4


pkg/validation/formats.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	if !f.AllowName && addr.Name != "" {
+		return errors.New("input email must not have name")
+	}
+
+	return nil
+}
+
+// FormatURI checks if input is an absolute URI, without query/fragment.
```
louischan-oursky, line 59: Why `uri` cannot have query or fragment?

pkg/auth/config/secret_data.go
```diff
@@ -0,0 +1,195 @@
...
+`)
+
+type DatabaseCredentials struct {
+	DatabaseURL    string `json:"database_url,omitempty"`
+	DatabaseSchema string `json:"database_schema,omitempty"`
+}
+
+func (c *DatabaseCredentials) SetDefaults() {
+	if c.DatabaseSchema == "" {
+		c.DatabaseSchema = "public"
```
louischan-oursky, line 25: Let's use `app`?

pkg/auth/config/secret_data.go
```diff
@@ -0,0 +1,195 @@
...
+	Host     string   `json:"host,omitempty"`
+	Port     int      `json:"port,omitempty"`
+	Mode     SMTPMode `json:"mode,omitempty"`
+	Username string   `json:"username,omitempty"`
+	Password string   `json:"password,omitempty"`
+}
+
+func (c *SMTPServerCredentials) SetDefaults() {
+	if c.Mode == "" {
+		c.Mode = SMTPModeNormal
```
louischan-oursky, line 123: Default should be ssl

kiootic: It's the same default value as current configuration, do you want to change it?

pkg/auth/config/secret_data.go
kiootic: `public` schema is the default schema of Postgres, so developer does not need to create a new schema for quick start.

pkg/validation/formats.go
kiootic: This is `URLFullOnly` from the original format. I suppose I'll make it accepts any URI (except URI references).

Issue #4, opened 2020-06-17
Configuration refactor


kiootic:
- [x] Refactor configuration shape
  - [x] Separate configuration and secret values
  - [x] Delete unused feature configurations
- [x] Server configuration (from env)
- [x] Configuration source
  - [x] local file
- [x] Remove old JSON schema lib (#27)
- [ ] Provider -> Service

PR #5, opened 2020-06-17
Add local configuration source

kiootic:
requires #3 
ref #4

### DIFF
--- a/pkg/auth/config/server.go
+++ b/pkg/auth/config/server.go
@@ -1,0 +1,92 @@
+package config
+
+import (
+	"fmt"
+	"github.com/kelseyhightower/envconfig"
+	"github.com/skygeario/skygear-server/pkg/validation"
+	"strings"
+)
+
+type ServerConfig struct {
+	// ListenAddr sets listening address of the server
+	ListenAddr string `envconfig:"LISTEN_ADDR"`
+	// TrustProxy sets whether HTTP headers from proxy are to be trusted
+	TrustProxy bool `envconfig:"TRUST_PROXY" default:"false"`
+	// ConfigSource configures the source of app configurations
+	ConfigSource ConfigurationSourceConfig `envconfig:"CONFIG_SOURCE"`
+
+	// ReservedNameFilePath sets the file path for reserved name list
+	ReservedNameFilePath string `envconfig:"RESERVED_NAME_FILE_PATH" default:"reserved_name.txt"`
+	// StaticAsset configures serving static asset
+	StaticAsset ServerStaticAssetConfig `envconfig:"STATIC_ASSET"`
+}
+
+func LoadServerConfigFromEnv() (*ServerConfig, error) {
+	config := &ServerConfig{}
+
+	err := envconfig.Process("", config)
+	if err != nil {
+		return nil, fmt.Errorf("cannot load server config: %w", err)
+	}
+
+	err = config.Validate()
+	if err == nil {
+		return nil, fmt.Errorf("invalid server config: %w", err)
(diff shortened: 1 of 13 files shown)


pkg/auth/config/source/source.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"context"
+	"github.com/skygeario/skygear-server/pkg/auth/config"
+	"net/http"
+)
+
+type Source interface {
+	Start() error
+	Shutdown() error
```
louischan-oursky, line 11: Use the term Open and Close instead? Or can you clarify the intended lifecycle of Source?

pkg/auth/config/server.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"fmt"
+	"github.com/kelseyhightower/envconfig"
+	"github.com/skygeario/skygear-server/pkg/validation"
+	"strings"
+)
+
+type ServerConfig struct {
+	// ListenAddr sets listening address of the server
+	ListenAddr string `envconfig:"LISTEN_ADDR"`
```
louischan-oursky, line 12: Default to `0.0.0.0:3000`

pkg/auth/config/source/local_file.go
```diff
@@ -55,5 +56,16 @@ func (s *LocalFile) Shutdown() error {
...
+		return s.config, nil
+	}
+
+	host := httputil.GetHost(r, s.serverConfig.TrustProxy)
+	for _, h := range s.config.AppConfig.HTTP.Hosts {
+		if h == host {
+			return s.config, nil
+		}
+	}
+	return nil, ErrAppNotFound
```
louischan-oursky, line 70: The error message is not very helpful. Maybe generate an error at runtime

```golang
fmt.Errorf("host `%s` is not listed as valid host", host) 
```

kiootic: Most likely this error message would be written to HTTP response, and I feel uneasy to echo untrusted input in HTTP response. I suppose I'll change it to `fmt.Errorf("request host is not valid: %w", ErrAppNotFound)`, so caller can use `errors.Is(err, ErrAppNotFound)` to determine correct status code.

pkg/auth/config/source/source.go
kiootic: This is intended for watching configuration files (start/shutdown the watcher) later on. I do agree 'opening configuration source' sounds more natural. I'll change it to Open/Close.

pkg/auth/config/source/local_file.go
louischan-oursky: It would be more helpful if we include how we get the host. Of course, the getHost function have to report the header name as well.

```golang
fmt.Errorf("host found in `%s` is invalid", headerName)
```

kiootic: I'm not sure if it would be helpful, since the host can be found in the headers rather easily?

louischan-oursky: They may not know they have to set `TRUST_PROXY` to `true`. This error message may give the developer some hints.

I agree we should not write untrusted string to response. Maybe we should instead log a warning containing the header name and the value. It would be most helpful.

kiootic: Sure. WARN level would be noisy, since it can be frequently hit on Internet by bad actors. I suppose we can log down the request host value as DEBUG level, since DEBUG level should be off at production.

I'll add a TODO to update this after refactoring the DI mechanism (we need a logger in there).

PR #7, opened 2020-06-18
Refactor config

kiootic:
ref #4


pkg/redis/pool.go
```diff
@@ -0,0 +1,145 @@
...
+
+	p.cacheMutex.RLock()
+	pool, exists := p.cache[connKey]
+	p.cacheMutex.RUnlock()
+
+	if !exists {
+		p.cacheMutex.Lock()
+		pool, exists = p.cache[connKey]
+		if !exists {
+			p.cache[connKey] = openRedis(c)
```
louischan-oursky, line 44: `pool` may be `nil` in some cases.

I suggest we rewrite this as

```golang
p.cacheMutex.RLock()
pool, exists := p.cache[connKey]
p.cacheMutex.RUnlock()
if exists {
  return pool
}

p.cacheMutex.Lock()
pool, exists = p.cache[connKey]
if !exists {
  pool = openRedis(c)
  p.cache[connKey] = pool
}
p.cacheMutex.Unlock()

return pool
```

pkg/log/default.go
```diff
@@ -0,0 +1,31 @@
...
+import "github.com/sirupsen/logrus"
+
+var defaultPatterns = []MaskPattern{
+	// JWT
+	NewRegexMaskPattern(`eyJ[A-Za-z0-9-_=]+\.[A-Za-z0-9-_=]+\.?[A-Za-z0-9-_.+/=]*`),
+	// Session Tokens
+	NewRegexMaskPattern(`[A-Fa-f0-9-]{36}\.[A-Za-z0-9]+`),
+}
+
+func NewDefaultLogHook(sensitiveStrings []string) logrus.Hook {
```
louischan-oursky: This hook is for masking potentially sensitive strings. Maybe rename it to `NewMaskingLogHook`?

pkg/log/default_test.go
```diff
@@ -72,4 +60,50 @@ func TestDefaultFormatter(t *testing.T) {
...
 		})
 	})
+
+	Convey("secret log hook", t, func() {
+		h := NewSecretLogHook(&config.SecretConfig{
+			Secrets: []config.SecretItem{
+				{
+					Key: config.DatabaseCredentialsKey,
+					Data: &config.DatabaseCredentials{
+						DatabaseURL:    "postgres://postgres://user:password@localhost:5432",
```
louischan-oursky, line 70: The scheme is duplicate.

pkg/auth/handler/oauth/challenge.go
```diff
@@ -37,6 +37,14 @@ func (p *ChallengeRequest) Validate() []validation.ErrorCause {
 	return nil
 }
 
+var challengeValidator = validation.NewValidator("http://v2.skgyear.io")
+
+func init() {
+	challengeValidator.AddSchemaFragments(
+		ChallengeRequestSchema,
```
louischan-oursky, line 44: I think we need to remove `$id` from the schema fragment?

cmd/auth/main.go
```diff
@@ -1,115 +1,34 @@
...
-	router := setupRoutes(configuration, dbPool, redisPool, authDependency)
-
-	srv := &http.Server{
-		Addr:    configuration.Host,
-		Handler: router,
-	}
-	server.ListenAndServe(srv, logger, "Starting auth gear")
+	server := httputil.NewServer(rootDeps.LoggerFactory, &http.Server{
+		Addr:    serverCfg.ListenAddr,
+		Handler: nil,
```
louischan-oursky, line 45: Set the handler?

pkg/deps/context.go
```diff
@@ -0,0 +1,16 @@
+package deps
+
+import "context"
+
+type requestContainerContextKeyType struct{}
```
louischan-oursky, line 5: Let's call the object that do dependency injection "Provider"; the object that do business logic "Service"; the object that coordinate different services "Flow"; the object that access data layer "Store".

pkg/core/db/context_mock.go
```diff
@@ -0,0 +1,40 @@
+package db
```
louischan-oursky, line 1: Rename this file to `context_mock_test.go` so it is included in test only.

pkg/deps/task.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/skygeario/skygear-server/pkg/task"
+	"github.com/skygeario/skygear-server/pkg/task/executors"
+	"github.com/skygeario/skygear-server/pkg/task/queue"
+)
+
+func ProvideCaptureTaskContext(ctx context.Context, config *config.Config) queue.CaptureTaskContext {
+	return func() *task.Context {
+		return &task.Context{
+			Config:                config,
+			PreferredLanguageTags: intl.GetPreferredLanguageTags(ctx),
```
louischan-oursky, line 17: Actually the existing tasks have been refactored to not depend on Config or PreferredLanguageTags. We can remove them.

pkg/task/queue/mock_queue.go
```diff
@@ -0,0 +1,22 @@
+package queue
```
louischan-oursky, line 1: Append `_test` to the filename

pkg/auth/handler/oauth/challenge.go
kiootic: Will do it as part of removing old JSON schema library.

cmd/auth/main.go
kiootic: Will update it after fixing the dependencies.

pkg/core/db/context_mock.go
kiootic: This is also used in tests of hook provider. If it is a test file, the hook provider tests cannot use it.

I think we should use mock_gen for hook provider, then we can do this renaming.

pkg/auth/dependency/authenticator/bearertoken/deps.go
```diff
@@ -2,20 +2,21 @@ package bearertoken
 
 import (
 	"github.com/google/wire"
+
 	"github.com/skygeario/skygear-server/pkg/core/config"
 	"github.com/skygeario/skygear-server/pkg/core/db"
-	"github.com/skygeario/skygear-server/pkg/core/time"
+	"github.com/skygeario/skygear-server/pkg/time"
```
louischan-oursky, line 8: The name `time` conflicting the stdlib `time` was always annoying to me...
Maybe rename it to `datetime`, `chrono` or something else?

pkg/time/clock.go
```diff
@@ -0,0 +1,22 @@
+package time
+
+import "time"
+
+type Clock interface {
+	NowUTC() time.Time
+	Now() time.Time
```
louischan-oursky, line 7: Maybe we can take this chance to rename `Now()` to `NowLocal()`

pkg/auth/config/utils.go
```diff
@@ -44,3 +52,37 @@ func extractJWKSecrets(keys []interface{}) []string {
...
+func ToJWS(keys []interface{}) (*jwk.Set, error) {
+	set := &jwk.Set{}
+	err := set.ExtractMap(map[string]interface{}{"keys": keys})
+	if err != nil {
+		return nil, err
+	}
+	return set, nil
+}
+
+func ExtractOctetKey(keys []interface{}, id string) ([]byte, error) {
```
louischan-oursky, line 65: Why don't we first call `ToJWS` and then use its API to extract the first key?

```golang
func ExtractOctetKey(set *jwk.Set, id string) ([]byte, error) {
  for _, key := range set.Keys {
    switch key.KeyType() {
    case jwa.OctetSeq {
      if key.KeyID() == id {
         var raw []byte
         key.Raw(raw)
         return raw, nil
      }
    }
    }
  }
}
```

PR #12, opened 2020-06-22
Use new config - Part 1

louischan-oursky:
To show my progress so no duplicate efforts

@kiootic


cmd/auth/routes.go
```diff
@@ -10,19 +10,26 @@ import (
...
-	"github.com/skygeario/skygear-server/pkg/core/middleware"
+	"github.com/skygeario/skygear-server/pkg/core/sentry"
 	"github.com/skygeario/skygear-server/pkg/core/server"
 	"github.com/skygeario/skygear-server/pkg/deps"
 	"github.com/skygeario/skygear-server/pkg/middlewares"
 )
 
+func NewRouter(p *deps.RootProvider) *mux.Router {
+	rootRouter := mux.NewRouter()
+	rootRouter.Use(sentry.Middleware(sentry.DefaultClient.Hub))
```
kiootic, line 21: Actually sentry need ServerConfig (for constructing request URL), so it need to use `p.Middleware` too. We can do it in another PR.

PR #20, opened 2020-06-23
Revamp sso

louischan-oursky:


pkg/auth/config/identity.go
```diff
@@ -235,16 +245,36 @@ type OAuthSSOProviderConfig struct {
 	// KeyID and TeamID are specific to apple
 	KeyID  string `json:"key_id,omitempty"`
 	TeamID string `json:"team_id,omitempty"`
+
+	Scope string `json:"-"`
```
kiootic, line 249: I suppose we should move the scopes to the specific providers, since we're not going to allow developer to configure them?

pkg/auth/config/secret.go
```diff
@@ -98,15 +117,16 @@ var _ = SecretConfigSchema.Add("SecretKey", `{ "type": "string" }`)
...
-	SMTPServerCredentialsKey SecretKey = "mail.smtp"
-	TwilioCredentialsKey     SecretKey = "sms.twilio"
-	NexmoCredentialsKey      SecretKey = "sms.nexmo"
-	JWTKeyMaterialsKey       SecretKey = "jwt"
-	OIDCKeyMaterialsKey      SecretKey = "oidc"
-	CSRFKeyMaterialsKey      SecretKey = "csrf"
-	WebhookKeyMaterialsKey   SecretKey = "webhook"
+	DatabaseCredentialsKey    SecretKey = "db"
+	RedisCredentialsKey       SecretKey = "redis"
+	OAuthClientCredentialsKey SecretKey = "oauth.client.external"
```
kiootic, line 122: I suggest to have a 'sso' prefix to avoid confusing our clients (as IdP) and us as a client (as RP).

louischan-oursky: I added the `external` suffix for this purpose. But I agree the sso prefix makes more sense. Let me change it.

PR #24, opened 2020-06-23
Upgrade jwx

louischan-oursky:


pkg/auth/dependency/oidc/id_token.go
```diff
@@ -107,6 +80,13 @@ func (ti *IDTokenIssuer) LoadUserClaims(session auth.AuthSession) (*UserClaims,
 
 	claims := &UserClaims{
 		StandardClaims: jwt.StandardClaims{
+			// TODO(id-token): https://openid.net/specs/openid-connect-core-1_0.html#IDToken
+			// Set `aud` to `client_id`.
```
kiootic, line 84: FYI: Some of these are done in `IssueIDToken` function. `LoadUserClaims` is intended to populate user-specific claims only.

louischan-oursky: Yes. Thanks for reminding!

PR #25, opened 2020-06-24
Wire up dependencies

kiootic:


pkg/auth/dependency/webapp/template.go
```diff
@@ -50,8 +50,8 @@ var TemplateAuthUIHTMLHeadHTML = template.Spec{
 <head>
 <title>{{ .app_name }}</title>
 <meta name="viewport" content="width=device-width, initial-scale=1">
-<link rel="stylesheet" href="{{ .x_static_asset_url_prefix }}/authui/css/main.css">
-<script src="{{ .x_static_asset_url_prefix}}/authui/js/main.js"></script>
+<link rel="stylesheet" href="{{ .x_static_asset_url_prefix }}/css/main.css">
+<script src="{{ .x_static_asset_url_prefix}}/js/main.js"></script>
```
louischan-oursky, line 54: The original one is correct. What inconsistency did you encounter?

kiootic: Originally it doesn't have the `authui` prefix: https://github.com/SkygearIO/skygear-server/blob/master/pkg/auth/dependency/webapp/template.go

The `authui` segment is included in the URL prefix of static asset, so it becomes like `/static/authui/authui/js/main.js`

louischan-oursky: Changed here.

https://github.com/authgear/authgear-server/commit/cab55aa41639c56c1b7564a2bdc00c2d31b61c0e

Apply this change instead

```diff
diff --git a/pkg/auth/config/server.go b/pkg/auth/config/server.go
index 387afc7f..d8a18b57 100644
--- a/pkg/auth/config/server.go
+++ b/pkg/auth/config/server.go
@@ -70,7 +70,7 @@ type ServerStaticAssetConfig struct {
 	// Dir sets the local directory of static assets
 	Dir string `envconfig:"DIR" default:"./static"`
 	// URLPrefix sets the URL prefix for static assets
-	URLPrefix string `envconfig:"URL_PREFIX" default:"/static/authui"`
+	URLPrefix string `envconfig:"URL_PREFIX" default:"/static"`
 }
 
 type SourceType string
```

This will make everything work without setting any envvar

---

I also noticed a bug in mail sender

Please apply this diff 

```diff
diff --git a/pkg/mail/sender.go b/pkg/mail/sender.go
index 37f3a0c1..cc45a4e5 100644
--- a/pkg/mail/sender.go
+++ b/pkg/mail/sender.go
@@ -34,6 +34,7 @@ func NewGomailDialer(smtp *config.SMTPServerCredentials) *gomail.Dialer {
 		case config.SMTPModeSSL:
 			dialer.SSL = true
 		}
+		return dialer
 	}
 	return nil
 }

```

kiootic: I suppose we'll need to change the email templates as well?

louischan-oursky: > I suppose we'll need to change the email templates as well?

Yes. I also noticed that we did not include `static_asset_url_prefix` in data in `pkg/auth/dependency/authenticator/oob/provider.go`. Please fix that as well.

PR #29, opened 2020-06-25
JWX JWT

louischan-oursky:


pkg/auth/dependency/identity/anonymous/provider.go
```diff
@@ -66,80 +80,102 @@ func (p *Provider) Delete(i *Identity) error {
...
+		return
 	}
 
-	req := &Request{}
-	token, err := jwt.ParseWithClaims(requestJWT, req, keyFunc)
-	if err != nil {
-		return nil, nil, fmt.Errorf("invalid JWT signature: %w", err)
+	// The client does include alg in the JWK.
+	// Fix it by copying alg in the header.
+	if key.Algorithm() == "" {
```
kiootic, line 148: This should not happen if our SDK is used?

louischan-oursky: https://github.com/authgear/authgear-sdk-js/blob/master/packages/skygear-react-native/ios/SGSkygearReactNative.m#L417

From my testing, the SDK does not include `alg` in the `jwk`.

Issue #11, opened 2020-06-19
CLI tool


kiootic:
Sub-commands:
- `migrate`: handle db/version migration
    - `up`/`down`: apply/revert migrations
    - `verify`: verify the database schema against latest version.
- `init <type> -o <output file>`: setup initial config & generate secrets
    - common use case: `authgear init config; authgear init secret`, output file default to authgear.yaml & authgear.secret.yaml
    - in docker run: `docker run ... authgear init config -o - > authgear.yaml`
    - prompts with default value: app id, redis/database credentials, host
- `start`: start server

@louischan-oursky @chpapa Thoughts on this?

chpapa:
@kiootic looks good at the beginning, some questions:

- `migrate` feel a bit strange if we consider Authgear as a standalone daemon, is it better to just auto do the migration if the DB version is older than the current daemon? 
- Do we need `stop` if there is a `start`? Or just let people kill it?
- Speaking of which, do we need to implement SIGHUP for reload config? (no need if it is trouble , not useful at this point)

louischan-oursky:
> I assume we will show an error when the daemon to start, and ask users to run something like authgear migrate?

This is better.

> Do we need stop if there is a start? Or just let people kill it?

CTRL-C should work fine.

> do we need to implement SIGHUP for reload config? (no need if it is trouble or not useful at this point

We had an idea of watching configuration file change to reload config in the future.

chpapa:
> We had an idea of watching configuration file change to reload config in the future.

Better do it only in some sort of local mode (so with a specific flag), and listen to SIGHUP / reload in production... avoid surprisingly sysadmin :)

kiootic:
> Better do it only in some sort of local mode (so with a specific flag), and listen to SIGHUP / reload in production... avoid surprisingly sysadmin :)

I suppose we'll need a command to signal reload config (like `nginx -s reload`), and the controller would send a signal to server to reload config when config of any app changed in multi-tenant mode.

kiootic:
> I assume we will show an error when the daemon to start, and ask users to run something like authgear migrate?

One thing to note: most likely we cannot show error on database schema mismatch on startup:
- On multi-tenant mode: each tenant can has different database (for sharding, isolation, etc.), so we cannot determine database schema on startup
- On single-tenant mode: we can do it in this mode, but then it'll be a special case since we'll trigger the check only if it is single-tenant mode.

[...]

PR #31, opened 2020-06-26
Internal port

louischan-oursky:


cmd/auth/main.go
```diff
@@ -43,9 +47,62 @@ func main() {
...
-	server := httputil.NewServer(p.LoggerFactory, &http.Server{
-		Addr:    serverCfg.ListenAddr,
+	publicServer := &http.Server{
+		Addr:    serverCfg.PublicListenAddr,
 		Handler: setupRoutes(p, configSource),
-	})
-	server.ListenAndServe("starting auth gear")
+	}
+
+	internalServer := &http.Server{
```
kiootic, line 55: I think some may want to run two separate deployment for public & internal server, so may be add a flag to control which server(s) is started?

cmd/auth/routes.go
```diff
@@ -7,19 +7,34 @@ import (
...
+	r.Use(p.RootMiddleware(newSentryMiddlewareFactory(sentry.DefaultClient.Hub)))
+	r.Use(p.RootMiddleware(newRecoverMiddleware))
+	r.HandleFunc("/healthz", httputil.HealthCheckHandler)
+	return r
+}
+
+func setupInternalRoutes(p *deps.RootProvider, configSource configsource.Source) *mux.Router {
+	r := NewRouter(p)
+
+	r.Use((&deps.RequestMiddleware{
```
kiootic, line 29: This would make health check fail if app is not found. We should create a sub-router for the real routes.

cmd/auth/main.go
louischan-oursky: Maybe it should be a command argument? Delay it until we do https://github.com/authgear/authgear-server/issues/11 ?

PR #34, opened 2020-06-26
Add httproute

louischan-oursky:


pkg/auth/handler/oauth/userinfo.go
```diff
@@ -4,20 +4,19 @@ import (
...
 
-func ConfigureUserInfoHandler(router *mux.Router, h http.Handler) {
-	router.NewRoute().
-		Path("/oauth2/userinfo").
-		Methods("GET", "POST", "OPTIONS").
-		Handler(oauth.RequireScope(h))
+func ConfigureUserInfoHandler(route httproute.Route, h http.Handler) (httproute.Route, http.Handler) {
+	return route.
+		WithMethods("GET", "POST", "OPTIONS").
+		WithPathPattern("/oauth2/userinfo"), oauth.RequireScope(h)
```
kiootic, line 19: `oauth.RequireScope` is actually a middleware. Is it better to refactor its application to main package to make all `ConfigureHandler` functions have same signature?

louischan-oursky: The only usage is here and `scopes` is actually empty. So `RequireScope` does not have any effect? Am I correct?

kiootic: Sorry for the bad name. `RequireScope` requires two things:
- a valid access token; and
- the access token has the specified scopes (if provided).

So here it just requires a valid access token, without checking for scopes.

PR #39, opened 2020-06-29
Add CLI

### LINKED ISSUE #11: CLI tool

kiootic:
fix #11


cmd/authgear/config/config.go
```diff
@@ -0,0 +1,112 @@
...
+		Data: &config.RedisCredentials{
+			Host:     opts.RedisHost,
+			Port:     opts.RedisPort,
+			Password: opts.RedisPassword,
+		},
+	})
+
+	items = append(items, config.SecretItem{
+		Key:  config.JWTKeyMaterialsKey,
+		Data: &config.JWTKeyMaterials{Set: generateRSAKey()},
```
louischan-oursky, line 42: My PR is incorrect. It should be octet key.

cmd/authgear/config/config.go
```diff
@@ -0,0 +1,112 @@
...
+
+	items = append(items, config.SecretItem{
+		Key:  config.OIDCKeyMaterialsKey,
+		Data: &config.OIDCKeyMaterials{Set: generateRSAKey()},
+	})
+
+	items = append(items, config.SecretItem{
+		Key:  config.CSRFKeyMaterialsKey,
+		Data: &config.CSRFKeyMaterials{Set: generateSymmetricKey()},
+	})
```
louischan-oursky, line 53: Generate the octet key for webhook as well.

cmd/authgear/config/config.go
```diff
@@ -0,0 +1,112 @@
...
+
+func NewAppConfigFromOptions(opts *Options) *config.AppConfig {
+	return &config.AppConfig{
+		ID:   config.AppID(opts.AppID),
+		HTTP: &config.HTTPConfig{Hosts: opts.Hosts},
+	}
+}
+
+func NewSecretConfigFromOptions(opts *SecretOptions) *config.SecretConfig {
+	var items []config.SecretItem
```
louischan-oursky, line 22: Based on the generated secret.yaml, can we update the JSON schema to reflect the actual shape of the various JWK ?

.env.example
```diff
@@ -1,11 +1,2 @@
...
-VALID_HOSTS=localhost:8000
-REDIS_HOST=localhost
-INSECURE_COOKIE=true
-TEMPLATE_ENABLE_FILE_LOADER=true
-TEMPLATE_ASSET_GEAR_ENDPOINT=http://localhost:8000
-TEMPLATE_ASSET_GEAR_MASTER_KEY=master_key
-
-STATIC_ASSET_DIR=./static
-STATIC_ASSET_URL_PREFIX=/static
+DATABASE_SCHEMA=app
```
louischan-oursky, line 2: None of these are actually read. Instead we should add

```
DEV_MODE=true
LOG_LEVEL=debug
```

cmd/authgear/config/config.go
kiootic: Since JWK is an extensible standard, I think we should just add fields that we expect to see, i.e. required fields.

cmd/authgear/init.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		log.Fatalf("cannot write config: %s", err.Error())
+	}
+}
+
+func init() {
+	cmdInit.AddCommand(cmdInitConfig)
+	cmdInit.AddCommand(cmdInitSecrets)
+
+	cmdInitConfig.Flags().StringVarP(&InitConfigOutputPath, "output", "o", "authgear.yaml", "Output YAML path")
```
louischan-oursky, line 62: The `init` command should error if it would overwrite existing file.

kiootic: I think it's better to prompt user whether to overwrite existing file?

louischan-oursky: Prompt is OK but the default should be `[y/N]` so blindly pressing enter would not overwrite accidentally.

PR #41, opened 2020-06-30
Fix last login at

### LINKED ISSUE #40: Fix last login at
Last login at currently reflects the creation time of the most recent session. Most of time the value is not what developer would like to have.

- Use two field per user: `login_at` and `last_login_at`
    - `login_at`: the creation time of most recent session
    - `last_login_at`: the creation time of the second-most recent session
- When creating session:
    1. Set `last_login_at` = `login_at`
    2. Set `login_at` = session creation time = now

kiootic:
fix #40

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,33 @@
+version: "3"
+services:
+  db:
+    image: postgres:12.3
+    volumes:
+      - db_data:/var/lib/postgresql/data
+    environment:
+      POSTGRES_USER: "postgres"
+      POSTGRES_PASSWORD: "postgres"
+    ports:
+      - "5432:5432"
+
+  redis:
+    image: redis:5.0
+    volumes:
+      - redis_data:/data
+    ports:
+      - "6379:6379"
+
+  authgear:
+    build: .
+    environment:
+      DEV_MODE: "true"
+      LOG_LEVEL: "debug"
+    volumes:
+      - ./authgear.yaml:/authgear.yaml
+      - ./authgear.secrets.yaml:/authgear.secrets.yaml
+
+volumes:
+  redis_data:
+    driver: local
+  db_data:
+    driver: local
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -19,12 +19,13 @@ services:
 
   authgear:
     build: .
-    environment:
-      DEV_MODE: "true"
-      LOG_LEVEL: "debug"
+    env_file:
+      - ".env"
     volumes:
-      - ./authgear.yaml:/authgear.yaml
-      - ./authgear.secrets.yaml:/authgear.secrets.yaml
+      - ./authgear.yaml:/app/authgear.yaml
+      - ./authgear.secrets.yaml:/app/authgear.secrets.yaml
+    ports:
+      - "3000:3000"
 
 volumes:
   redis_data:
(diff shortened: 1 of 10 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,33 @@
...
+    volumes:
+      - redis_data:/redis_data
+    ports:
+      - "6379:6379"
+
+  authgear:
+    build: .
+    environment:
+      DEV_MODE: "true"
+      LOG_LEVEL: "debug"
```
louischan-oursky, line 24: docker-compose by default load `.env` so it is unnecessary.

pkg/auth/dependency/user/store.go
```diff
@@ -113,16 +116,24 @@ func (s *Store) UpdateMetadata(userID string, metadata map[string]interface{}, u
...
-func (s *Store) UpdateLoginTime(userID string, lastLoginAt time.Time) error {
+func (s *Store) UpdateLoginTime(userID string, loginAt time.Time) (*time.Time, error) {
 	builder := s.SQLBuilder.Tenant().
 		Update(s.SQLBuilder.FullTableName("user")).
-		Set("last_login_at", lastLoginAt).
-		Where("id = ?", userID)
+		Set("last_login_at", squirrel.Expr("login_at")).
+		Set("login_at", loginAt).
+		Where("id = ?", userID).
+		Suffix("RETURNING last_login_at")
```
louischan-oursky, line 125: I think we should add LoginAt and LastLoginAt to User. This function keeps returning error only. The caller calls Get to retrieve the updated user.

pkg/auth/dependency/user/commands_raw.go
```diff
@@ -55,11 +55,12 @@ func (c *RawCommands) UpdateMetadata(user *model.User, metadata map[string]inter
...
-func (c *RawCommands) UpdateLoginTime(user *model.User, lastLoginAt time.Time) error {
-	if err := c.Store.UpdateLoginTime(user.ID, lastLoginAt); err != nil {
+func (c *RawCommands) UpdateLoginTime(user *model.User, loginAt time.Time) error {
+	lastLoginAt, err := c.Store.UpdateLoginTime(user.ID, loginAt)
+	if err != nil {
 		return err
 	}
 
-	user.LastLoginAt = &lastLoginAt
+	user.LastLoginAt = lastLoginAt
```
louischan-oursky, line 64: Refetch the user here.

docker-compose.yaml
kiootic: It loads .env to substitute variable references in docker-compose.yaml, but for container environment, either `environment` or `env_file` should be used.

docker-compose.yaml
```diff
@@ -0,0 +1,33 @@
...
+    volumes:
+      - ./authgear.yaml:/authgear.yaml
+      - ./authgear.secrets.yaml:/authgear.secrets.yaml
```
louischan-oursky, line 27: The WORKDIR is `/app` so we should mount them as `/app/authgear.yaml` and `/app/authgear.secrets.yaml`.

docker-compose.yaml
```diff
@@ -0,0 +1,33 @@
...
+    environment:
+      POSTGRES_USER: "postgres"
+      POSTGRES_PASSWORD: "postgres"
+    ports:
+      - "5432:5432"
+
+  redis:
+    image: redis:5.0
+    volumes:
+      - redis_data:/redis_data
```
louischan-oursky, line 16: Should be `- redis_data:/data`

kiootic: fixed.

Issue #42, opened 2020-06-30
Rename old name


kiootic:
- Remove/Rename all reference to old name `skygear`.

- [x] `authgear-server`
- [x] `authgear-sdk-js`

PR #46, opened 2020-07-02
Rename old names

kiootic:
ref #42


pkg/template/template.go
```diff
@@ -50,7 +50,7 @@ func SetContextToURLQuery(u *url.URL, context map[string]interface{}) error {
 		return err
 	}
 	query := u.Query()
-	query.Set("x-skygear-redirect-data", encoded)
+	query.Set("x-authgear-redirect-data", encoded)
```
louischan-oursky, line 53: This function `SetContextToURLQuery` is unused now. Its friends `DecodeURLQueryParamValueToContext` and `EncodeContextToURLQueryParamValue` can also be removed.

Issue #55, opened 2020-07-06
User verification


keithtkl:
- Definition of verified Login ID
  - Satisfy one of the following condition
     - The login ID has an associated authenticator
     - The user has SSO user info with `email` claim equal to the login ID

keithtkl:
Add verification status to the setting page

kiootic:
Discussed with @louischan-oursky :
- we'll having separate templates for user verification and OTP message
- verification will be using the same OOB authenticators to share its semantics, but
   code handling would be done differently.
- for email verification:
    - the code (a long code) will be stored separately
    - message will have link (contains code & interaction id) to the web UI
    - user can verify with just the code, no need to log in
    - if interaction can be restored, the interaction will continue
- for phone verification:
    - the code (a short code) will be stored in interaction
    - message will have just the code
    - user must enter the code in same page

PR #63, opened 2020-07-08
Add user verification feature specs

kiootic:
ref #55

### DIFF
--- a/docs/specs/verification.md
+++ b/docs/specs/verification.md
@@ -17,8 +17,8 @@ An identity is verified if:
 Developer can configure the criteria used to determine verification status
 of a user. There are two possible criteria:
 
-- `any`: User has a verifiable identity and one of user's verifiable identities is verified.
-- `all`: User has a verifiable identity and all of user's verifiable identities are verified.
+- `any`: User has at least one verifiable identity and at least one verified verifiable identity.
+- `all`: User has at least one verifiable identity and all the verifiable identities are verified.
 
 By default, criteria `any` is used.
 
--- a/docs/specs/verification.md
+++ b/docs/specs/verification.md
@@ -1,0 +1,66 @@
+# User Verification
+
+## Definitions
+
+A user is verified if the user fulfill the condition specified by criteria.
+
+An identity is verifiable if:
+- it is an SSO identity; or
+- it is a login ID identity with a configured verifiable login ID key.
+
+An identity is verified if:
+- it is an SSO identity; or
+- it is a login ID identity, and the user has a matching OOB-OTP authenticator.
+
+## Criteria
+
+Developer can configure the criteria used to determine verification status
+of a user. There are two possible criteria:
+
+- `any`: User has a verifiable identity and one of user's verifiable identities is verified.
+- `all`: User has a verifiable identity and all of user's verifiable identities are verified.
+
+By default, criteria `any` is used.
+
+```yaml
+# Use 'all' criteria to determine user verification status
+verification:
+    criteria: all
+```
+
+## Requirement
+
+Developer can configure verification requirement for specific login ID keys.
+The specified login ID keys must have type `email` or `phone`.
+
+When a login ID identity with specified login ID key is created (e.g. during
+sign up), the user is required to verify the login ID using a one-time-
+password sent to the login ID before proceeding. A matching OOB-OTP
+authenticator would be created in the verification process.
+
+By default, user must verify login ID key 'email' and 'phone'.
+
+```yaml
+# Require verification for login ID key 'email' and 'phone'
+verification:
+    login_id_keys: [email, phone]
+```
+
+## Code & Message
+
+The verification code & message can be configured using OOB-OTP authenticator
+configuration.
+
+## Status Flag
+
+The verification status flag of an identity would be shown in the UI of
+account settings page.
+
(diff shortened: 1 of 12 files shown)


docs/specs/verification.md
```diff
@@ -0,0 +1,66 @@
...
+An identity is verified if:
+- it is an SSO identity; or
+- it is a login ID identity, and the user has a matching OOB-OTP authenticator.
+
+## Criteria
+
+Developer can configure the criteria used to determine verification status
+of a user. There are two possible criteria:
+
+- `any`: User has a verifiable identity and one of user's verifiable identities is verified.
```
louischan-oursky, line 20: `User has at least one verifiable identity and at least one verified verifiable identity.`

docs/specs/verification.md
```diff
@@ -0,0 +1,66 @@
...
+- `all`: User has a verifiable identity and all of user's verifiable identities are verified.
```
louischan-oursky, line 21: `User has at least one verifiable identity and all of the verifiable identities are verified.`

docs/specs/verification.md
```diff
@@ -0,0 +1,66 @@
+# User Verification
+
+## Definitions
+
+A user is verified if the user fulfill the condition specified by criteria.
+
+An identity is verifiable if:
+- it is an SSO identity; or
+- it is a login ID identity with a configured verifiable login ID key.
```
louischan-oursky, line 9: So the developer can say, this login ID key is verifiable or not.

Something like this?

```yaml
identity:
  login_id:
    keys:
    - key: email
      type: email
      verifiable: false # By default it is true
```

docs/specs/verification.md
```diff
@@ -0,0 +1,66 @@
...
+sign up), the user is required to verify the login ID using a one-time-
+password sent to the login ID before proceeding. A matching OOB-OTP
+authenticator would be created in the verification process.
+
+By default, user must verify login ID key 'email' and 'phone'.
+
+```yaml
+# Require verification for login ID key 'email' and 'phone'
+verification:
+    login_id_keys: [email, phone]
```
louischan-oursky, line 46: My configuration example that put this information in `identity.login_id.keys` seems better. If the developer changes `keys`, they do not have to change `verification.login_id_keys`.

docs/specs/verification.md
```diff
@@ -0,0 +1,66 @@
+# User Verification
+
+## Definitions
+
+A user is verified if the user fulfill the condition specified by criteria.
+
+An identity is verifiable if:
+- it is an SSO identity; or
```
chpapa, line 8: Curious, why SSO identity = verifiable?

docs/specs/verification.md
```diff
@@ -0,0 +1,66 @@
...
+
+```yaml
+# Require verification for login ID key 'email' and 'phone'
+verification:
+    login_id_keys: [email, phone]
+```
+
+## Code & Message
+
+The verification code & message can be configured using OOB-OTP authenticator
```
chpapa, line 85: Does it meant, if a user have two ways to login:

1. Email + Password
2. Email + OTP

(p.s. basically this is the case of Slack by default, after user input email address, they can choose to login manually or send a "Magic link")

The verification email after sign up, and the OTP login email content, will be the same? If yes we might have a problem.

docs/specs/verification.md
kiootic: We assumes SSO providers (e.g. Google, Facebook, etc.) are trusted: they would ensure the email returned is verified. Therefore, All SSO identities are considered verified.

Since it is more intuitive that only 'verifiable' identities can be 'verified' identities, SSO identities are considered 'verifiable' and always 'verified'.

chpapa: Noted, agreed on the assumption.

docs/specs/verification.md
kiootic: > The verification email after sign up, and the OTP login email content, will be the same?

Let me try to clarify your use case:
- User signup with email + password, no email verification is needed
- User can verify email in settings page
- User can login with email + password, or email + OTP if email is verified

If that's the case, I suppose we need to allow different content of the OTP message depending on the context.

If so, we can provide template entries for OTP messages (`otp_message_email.txt`/`otp_message_email.html`/`otp_message_sms.txt`), and put the originating operation type (`operation` = `verification` or `authentication`) in the template context. Does that sounds good to you?

chpapa: If we take Slack as the sample, the use case is more like:

1. User signup with email + password, and verify email address
2. User can login with email + password, or email + OTP

Your solution sounds ok, but I'm not sure what is the user experience will be if we do operation type vs more template entries... can you / @louischan-oursky advice?

kiootic: Discussed with @louischan-oursky , we agree that having just one template type for OTP message is better in term of developer experience, since they just need to customize single set file, versus copy & paste a dozen of template files and change just some words.

We can add some fields to indicate the context of the message. For example:
- source: signup/login/settings
- operation: primary_auth/secondary_auth/verify

Let me update the specification to reflect this.

chpapa:
Approved as I have no further comments.

Food for thought: I do have a bit worried about overloading OTP authenticator with verification, but you guys make the call.

PR #71, opened 2020-07-15
Skip running image job in forked repo

louischan-oursky:
This job requires credentials which are not available in forked repo.


kiootic:
I'm not sure if skipping the entire image building & pushing would be the best approach; some may want to build image for their own fork. How about skipping image pushing if the required secrets is not found?

louischan-oursky:
Updated to only login and push if the repo is here.

PR #77, opened 2020-07-21
Interaction Graph with Web App

louischan-oursky:


pkg/auth/dependency/webapp/state.go
```diff
@@ -1,13 +1,32 @@
...
 import (
+	"net/url"
+
 	"github.com/authgear/authgear-server/pkg/core/skyerr"
 )
 
 type State struct {
-	ID              string           `json:"id"`
+	FlowID          string           `json:"flow_id"`
+	InstanceID      string           `json:"instance_id"`
```
kiootic, line 11: Do we need these two fields?

pkg/auth/dependency/webapp/service.go
```diff
@@ -0,0 +1,251 @@
...
+	state := &State{
+		FlowID:      newFlowID(),
+		InstanceID:  newInstanceID(),
+		RedirectURI: intent.RedirectURI,
+		KeepState:   intent.KeepState,
+	}
+
+	var graph *newinteraction.Graph
+	var edges []newinteraction.Edge
+	// FIXME(webapp): How to convey the intention that we only want to
```
kiootic, line 111: In commit 4b62c07cb9f2914b20159df21d38852060a77dbf, I've fixed `WithContext` to handle the returned values better:
- nil graph, nil error -> commit without creating new graph instance, use at the end of graph.
- nil graph, non-nil error -> rollback without creating new graph instance, use when `Accept` returns a non-recognized error. the original graph should still be used.

Is that okay?
- non-nil graph, nil error -> commit AND create new graph instance, should not happen.
- non-nil graph, non-nil error -> rollback and creating new graph instance, use when `Accept`
returns `ErrRequireInput`.

louischan-oursky: In WebAppService's GetXXX, commit is never expected.

In PostXXX, commit is still rare because it only happens when the graph ends. Maybe `fn` should return one more value to tell whether commit is desired.

The commit condition should be `err == nil && commit == true`
The create-graph condition should be `graph != nil`
The rollback condition is the negation of the commit condition
WithContext must either rollback or commit.

pkg/auth/dependency/webapp/state.go
louischan-oursky: Seems unnecessary. Let me remove FlowID first.

pkg/auth/dependency/newinteraction/nodes/select_identity_oauth.go
```diff
@@ -2,46 +2,122 @@ package nodes
...
 }
 
-type EdgeSelectIdentityOAuth struct {
+type EdgeSelectIdentityOAuthProvider struct {
 	Config config.OAuthSSOProviderConfig
 }
 
-func (e *EdgeSelectIdentityOAuth) Instantiate(ctx *newinteraction.Context, graph *newinteraction.Graph, rawInput interface{}) (newinteraction.Node, error) {
-	input, ok := rawInput.(InputSelectIdentityOAuth)
+type NodeSelectIdentityOAuthProvider struct {
```
kiootic, line 36: I think it's better not to interleave the type definitions here?

pkg/auth/dependency/newinteraction/nodes/select_identity_oauth.go
```diff
@@ -2,46 +2,122 @@ package nodes
...
+	newinteraction.RegisterNode(&NodeSelectIdentityOAuthProvider{})
+	newinteraction.RegisterNode(&NodeSelectIdentityOAuthUserInfo{})
 }
 
-type InputSelectIdentityOAuth interface {
-	GetUserInfo() sso.AuthInfo
+var ErrOAuthProviderNotFound = skyerr.NotFound.WithReason("OAuthProviderNotFound").New("oauth provider not found")
+
+type InputSelectIdentityOAuthProvider interface {
+	GetProviderAlias() string
```
kiootic, line 26: I remember you suggested to bundle these fields into a plain struct `FooArgs`, do we discard that idea now?

louischan-oursky: The struct does not seem to bring any values. Interface with multiple methods should be fine.

pkg/auth/dependency/newinteraction/nodes/select_identity_oauth.go
louischan-oursky: Do you mean separating `SelectOAuthProvider` and `SelectOAuthUserInfo` into 2 files? Or something else?

kiootic: I mean the layout of file, now it is:
1. define `EdgeSelectIdentityOAuth`
2. define `NodeSelectIdentityOAuthProvider`
3. define methods of `EdgeSelectIdentityOAuth`
4. define methods of `NodeSelectIdentityOAuthProvider`

3 should be moved before 2.

Separating the `SelectOAuthProvider` and `SelectOAuthUserInfo` into 2 files also seems better, since both contains large amount of non-trivial and de-coupled logic.

louischan-oursky:
Added 2 more commits. One commit to separate the OAuth nodes. Another to move view models. Let's merge this first.

Issue #73, opened 2020-07-17
Interaction Graph


louischan-oursky:
- Change interaction.Step to InteractionNode (interface)
  - Concrete node structs
- GetInteractionGraph: retrieve the graph of interactionNode with all ancestors, current.
- DeriveImmediateChildren
- Remove Interaction
- Intent becomes a InteractionNode which is the root node.
- The graph (ancestors and current) will be persisted.
- Information directly attached to the node itself.
- The graph will be persisted so no physical cycle is allowed. But there may have logical cycles.
  - Example of logical cycles: Add two identities during signup (email + phone, with both verified)
- Able to re-play the graph into the database transaction and commit **OR** rollback
  - Nested transaction / savepoint?

kiootic:
For OOB triggers:

We'd like the interaction graph remains immutable, so rate limit time cannot be put into graph nodes (otherwise need to handle invalidation of stale graph states).

Instead, a separate rate limiter module would be used to perform rate limiting. The OOB trigger node contains the generated code and a reference (ID) to the specific rate limit (per interaction graph per address).
- OOB trigger edge would consult the rate limiter, and return errors if failed. The error would bubble up to web app layer, and interpreted into a rate limit error.
- Web app can use the rate limit ID in node to consult rate limiter for the remaining time to display in UI.

kiootic:
For OOB code rotation:

We'd like to rotate OOB code regularly (e.g. every 5 mins) in interaction. To keep nodes immutable, we can generate a secret and use it in TOTP algorithm. This implies we do not support code format for OOB codes (only number of digits is configurable).

louischan-oursky:
- Break `/oob_otp` into
  - `/verify`
  - `/setup_primary_oob_otp`
  - `/enter_primary_oob_otp`
  - `/setup_secondary_oob_otp`
  - `/enter_secondary_oob_otp`
- Break `OTPMessage` into
  - `verification_sms.txt`
  - `verification_email.txt`
  - `verification_email.html`
  - `setup_primary_oob_sms.txt`
  - `setup_primary_oob_email.txt`
  - `setup_primary_oob_email.html`
  - `setup_secondary_oob_sms.txt`
  - `setup_secondary_oob_email.txt`
  - `setup_secondary_oob_email.html`
  - `authenticate_primary_oob_sms.txt`
  - `authenticate_primary_oob_email.txt`
  - `authenticate_primary_oob_email.html`
  - `authenticate_secondary_oob_sms.txt`
[...]

PR #83, opened 2020-07-24
Interaction Graph Web App - Part 2

louischan-oursky:
ref #73

### DIFF
--- a/pkg/auth/dependency/webapp/service.go
+++ b/pkg/auth/dependency/webapp/service.go
@@ -2,14 +2,21 @@ package webapp
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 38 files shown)


pkg/auth/dependency/webapp/service.go
```diff
@@ -231,6 +249,7 @@ func (s *Service) afterPost(state *State, graph *newinteraction.Graph, edges []n
 
 	// Case: finish
 	if finished {
+		s.Logger.Debugf("afterPost finished")
```
kiootic, line 258: Actually for these logs (and also the connection open/close logging), I think `TRACE` level is more suitable, otherwise the console would be spammed with logs that aren't useful most of the time. Let's review the appropriate log levels later.

PR #88, opened 2020-07-28
Refactor bearer tokens & recovery codes

kiootic:
ref #73


docs/specs/user-model.md
```diff
@@ -238,8 +238,8 @@ Out-of-band One-time-password authenticator is either primary or secondary.
 
 OOB-OTP authenticator is bound to a recipient address. The recipient can be an email address or phone number that can receive SMS messages.
 
-An OOB-OTP authenticator may be bound to a login ID identity. If the bounded
-identity is deleted, the authenticator would be deleted as well.
+An OOB-OTP authenticator may matches a login ID identity. When a login ID
+identity is deleted, any OOB-OTP matching authenticator would be deleted as well.
```
louischan-oursky, line 242: Suppose the user signed up with OAuth. Later they added OOB-OTP authenticator as 2 factor. And then they added the login ID identity matching the OOB-OTP authenticator. When the login ID identity is removed, the authenticator should be kept.

docs/specs/user-model.md
```diff
@@ -261,18 +261,25 @@ authenticator:
...
+secondary authentication as long as it remains valid.
 
-The token is a cryptographically secure random string of 256 bits (32 bytes) in hex encoding.
+The token is a cryptographically secure random string with at least 256 bits.
 
-### Recovery Code Authenticator
+### Recovery Code
 
-Recovery code authenticator is secondary.
+Recovery codes are used to skip secondary authentication when a secondary
```
louischan-oursky, line 276: It is not accurate to say recovery code is used to skip secondary authentication.

docs/specs/user-model.md
kiootic: If we don't delete it, when user removed a login ID identity, the 'verification status' would remain, so if the same identity is re-added, it is verified immediately.

louischan-oursky: The verification status is `matched && (created_at of authenticator > created_at of identity)`?

kiootic: In that case, when user attempt to verify the re-added identity, we'll remove the existing authenticator and add a new authenticator?

louischan-oursky: Or just update the created_at column (though it seems weird)

PR #96, opened 2020-07-30
Identity settings

louischan-oursky:
ref #73


louischan-oursky:
Do not merge this.

pkg/auth/dependency/authenticator/service/service.go
```diff
@@ -73,88 +73,95 @@ func (s *Service) Get(userID string, typ authn.AuthenticatorType, id string) (*a
...
+	types := ii.Type.PrimaryAuthenticatorTypes()
+	for _, typ := range types {
+		for _, ai := range ais {
+			if ai.Type == typ {
+				switch {
+				case ii.Type == authn.IdentityTypeLoginID && ai.Type == authn.AuthenticatorTypeOOB:
+					loginID := ii.Claims[identity.IdentityClaimLoginIDValue]
+					email, _ := ai.Props[authenticator.AuthenticatorPropOOBOTPEmail].(string)
+					phone, _ := ai.Props[authenticator.AuthenticatorPropOOBOTPPhone].(string)
+					if loginID == email || loginID == phone {
```
kiootic, line 134: I suppose we should normalize the OOB email/phone before comparing with login ID? Let me fix this in verification.

pkg/auth/dependency/newinteraction/nodes/remove_identity.go
```diff
@@ -0,0 +1,72 @@
...
+func init() {
+	newinteraction.RegisterNode(&NodeRemoveIdentity{})
+}
+
+type InputRemoveIdentity interface {
+	GetIdentityType() authn.IdentityType
+	GetIdentityID() string
+}
+
+type EdgeRemoveIdentity struct{}
```
kiootic, line 18: As discussed, we need to add `IdentityType` and `IdentityID` in the edge. Let me fix this in removing anonymous identity.

pkg/auth/dependency/newinteraction/nodes/update_identity_end.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+type EdgeUpdateIdentityEnd struct {
+	IdentitySpec *identity.Spec
+}
+
+func (e *EdgeUpdateIdentityEnd) Instantiate(ctx *newinteraction.Context, graph *newinteraction.Graph, rawInput interface{}) (newinteraction.Node, error) {
+	userID := graph.MustGetUserID()
+	identityType := e.IdentitySpec.Type
+	identityID := graph.MustGetUpdateIdentityID()
```
kiootic, line 21: `UpdateIdentityID` seems to be a state local to the `UpdateIdentityBegin/End` pair of nodes. I'd prefer to pass local states through fields in edges, instead of retrieving globally from the graph.

pkg/auth/dependency/newinteraction/nodes/update_identity_end.go
```diff
@@ -0,0 +1,74 @@
...
+			return err
+		}
+
+		return nil
+	}))
+	if err != nil {
+		return err
+	}
+
+	// TODO(interaction): dispatch identity creation event if not creating user
```
kiootic, line 70: typo in TODO: identity update event

louischan-oursky:
Ready for review.

pkg/auth/dependency/newinteraction/nodes/update_identity_end.go
louischan-oursky: They are separated by a generic edge EdgeUseIdentityLoginID, which makes it non-trivial to pass local state.

louischan-oursky:
It seems that login with OAuth will give invalid credentials error.

pkg/auth/dependency/newinteraction/nodes/update_identity_end.go
kiootic: Hmm... I suppose we can review it in future whether having a generic UseIdentity node with three modes is the best way forward.

kiootic:
I suppose we'll need to special case OAuth in IntentAuthenticate, when derive edges for `NodeAuthenticationEnd` in primary authentication stage.

PR #111, opened 2020-08-03
Require number of migrations to revert

### LINKED ISSUE #108: Require number of migrations to down
Currently It simply migrate down to root which is extremely dangerous.

kiootic:
fix #108

### DIFF
--- a/cmd/authgear/migrate.go
+++ b/cmd/authgear/migrate.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/authgear/migrate.go
```diff
@@ -65,7 +66,16 @@ var cmdMigrateDown = &cobra.Command{
 			log.Fatalf("cannot load secret config: %s", err)
 		}
 
-		migrate.Down(migrate.Options{
+		if len(args) == 0 {
+			log.Fatalf("number of migrations to revert not specified; specify 0 to revert all migrations")
```
louischan-oursky, line 70: Maybe support a special value `all` to mean revert all migrations. `authgear migrate down 0` seems a no-op to me, but it is not.

PR #112, opened 2020-08-03
Add authenticator tag

### LINKED ISSUE #106: Authenticators redefined
- [x] The column type is `jsonb`. The actual JSON type is an array of strings.
- [x] Keep the invariant that primary authenticator and secondary authenticator in config are non-overlapping.
- [ ] Remove the invariant that password authenticator can only have 1.
- [x] Forgot/Reset password
  - [x] Reset primary password authenticator.
  - [x] Should NOT create password authenticator.
  - [x] Use ListIdentitiesByClaims to find the user
  - [x] Should emit error eagerly if user not found
  - [x] Should emit error eagerly if the user does not have primary password authenticator
- [x] Can act as primary authenticator: Password and OOB OTP
- [x] Can act as secondary authenticator: Password, TOTP, OOB TOP
- [x] Can act as secondary authenticator fallback: Device Token and Recovery Code
- [ ] Add tags to authenticators
  - [ ] `authentication:primary`: It is a primary authenticator
  - [ ] `authentication:secondary`: It is a secondary authenticator
  - [ ] An authenticator without the above 2 tags cannot be used in authentication but it may contribute to verification status derivation.
  - [ ] The above 2 tags are mutually exclusive.
  - [ ] `authentication:default`: When it is a secondary authenticator, it is the default one.

louischan-oursky:
fix #106

### DIFF
--- a/migrations/20200803184228-add_authenticator_tag.sql
+++ b/migrations/20200803184228-add_authenticator_tag.sql
@@ -1,0 +1,5 @@
+-- +migrate Up
+ALTER TABLE _auth_authenticator ADD COLUMN tag JSONB NOT NULL;
+
+-- +migrate Down
+ALTER TABLE _auth_authenticator DROP COLUMN tag;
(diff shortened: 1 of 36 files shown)


louischan-oursky:
Added the Filter interface.

PR #115, opened 2020-08-04
MFA with OOB OTP / password

louischan-oursky:


templates/auth_ui_translation.json
```diff
@@ -98,5 +98,9 @@
 	"settings-identity-login-id-raw": "Username",
 
 	"enter-login-id-page-title--change": "Change your {0}",
-	"enter-login-id-page-title--add": "Enter your {0}"
+	"enter-login-id-page-title--add": "Enter your {0}",
+
+	"setup-oob-otp-title": "Setup {0, select, email{email} phone{SMS} other{}} verification",
```
kiootic, line 103: I suppose we may want to separate into two messages: one for primary ('Passwordless login'), one for secondary ('Multi-factor authentication'), since the wordings would collide with user verification.

louischan-oursky: When we setup primary OOB authenticator, we actually do not land on this page. It is because the previous (`/login` or `/signup`) implements two input interfaces. So this page is for MFA. I'll adjust the copy to be `Setup email multi-factor verification` and `Setup SMS multi-factor verification`.

louischan-oursky:
@kiootic Added 3 commits.

PR #120, opened 2020-08-06
Bind web state to user agent

### LINKED ISSUE #66: Bind interaction flow to user agent
To allow multiple concurrent interaction flow in the same user agent, we put the interaction flow token in query parameters. This allow the interaction flow to be resumed using just the link.

If we support sending the interaction flow link as part of OOB OTP message, we should bind the interaction flow to the user agent to prevent unauthorized party with access to OOB factor to use the interaction flow.

kiootic:
fix #66

### DIFF
--- a/pkg/auth/dependency/webapp/service.go
+++ b/pkg/auth/dependency/webapp/service.go
@@ -6,10 +6,15 @@ import (
 	"net/http"
 	"net/url"
 
+	"github.com/authgear/authgear-server/pkg/auth/config"
 	"github.com/authgear/authgear-server/pkg/auth/dependency/newinteraction"
 	"github.com/authgear/authgear-server/pkg/auth/dependency/newinteraction/intents"
 	"github.com/authgear/authgear-server/pkg/auth/dependency/newinteraction/nodes"
+	"github.com/authgear/authgear-server/pkg/core/base32"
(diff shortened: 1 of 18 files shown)


pkg/auth/dependency/webapp/service.go
```diff
@@ -45,26 +50,61 @@ func NewServiceLogger(lf *log.Factory) ServiceLogger {
...
+		Name:     UserAgentTokenCookieName,
+		Value:    token,
+		Path:     "/",
+		Secure:   secure,
+		MaxAge:   0,                     // Use HTTP session cookie; expires when browser closes
+		SameSite: http.SameSiteNoneMode, // Ensure resume-able after redirecting from external site
+	}
+	if !samesite.ShouldSendSameSiteNone(s.Request.UserAgent(), secure) {
+		cookie.SameSite = 0
+	}
```
louischan-oursky, line 85: It seems that we now have more places to create cookie. Maybe it is a good chance to introduce a CookieMaker which take care of setting the correct Secure and SameSite flag.

My proposal

```golang
type CookieMaker interface {
  // NewCookie returns a pointer to http.Cookie with Secure and SameSite set to proper values.
  // All other values are left untouched. You must initialize them before further use.
  MakeCookie() *http.Cookie
}
```

PR #125, opened 2020-08-06
Fix not preserving OAuth parameter at internal links

### LINKED ISSUE #123: Cannot use alternative links when logging in through OAuth with existing session
When logging in through OAuth(prompt=login) with existing session, alternative links (e.g. use phone instead, create account) would lead to settings page. This is due to the OAuth parameters are not propagated in these links.

kiootic:
fix #123

### DIFF
--- a/pkg/auth/dependency/webapp/url_provider.go
+++ b/pkg/auth/dependency/webapp/url_provider.go
@@ -15,8 +15,6 @@ import (
 )
 
 type EndpointsProvider interface {
-	AuthenticateEndpointURL() *url.URL
-	PromoteUserEndpointURL() *url.URL
 	LogoutEndpointURL() *url.URL
 	SettingsEndpointURL() *url.URL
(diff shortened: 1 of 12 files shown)


pkg/auth/dependency/webapp/url_provider.go
```diff
@@ -86,38 +84,39 @@ type AuthenticateURLProvider struct {
...
-		q["prompt"] = options.Prompt
-	}
-	if options.UILocales != "" {
-		q["ui_locales"] = options.UILocales
+	inputer := func() (interface{}, error) {
+		return nil, nil
 	}
+
+	intent.RedirectURI = options.RedirectURI
+	intent.UILocales = options.UILocales
```
louischan-oursky, line 98: These two lines seem redundant. We have set the two fields when intent is constructed at line 87.

kiootic: oops forgot to delete them when moving them up...

Issue #56, opened 2020-07-06
MFA


keithtkl:
- [x] Review web app state machine and draw a state digram
- [x] In login, available authenticators are unaffected by config
- [x] Remove identity should remove its bound OOB authenticator
- [x] Populate AMR
- [x] Populate ACR
- [x] Support setup 2nd authenticator in signup
- [x] Support setup 2nd authenticator in login
- [x] Support setup 2nd authenticator in promote
- [x] Support password as 2nd factor
- [x] Add `setup_totp.html`
  - [x] Show a QR code image
  - [x] Show a text input field with digits only
  - [x] Submit button
- [x] Add `enter_totp.html`
  - [x] Show a text input field with digits only
  - [x] Submit button
- [x] Add `setup_recovery_code.html`
  - [x] Show a list of recovery code
  - [x] Offer a button to download the codes
- [x] Allow switching to use recovery code / another 2nd authenticator
- [x] Respect preferred 2nd authenticator and maximum
- [x] Support default tag
- [x] Add `enter_recovery_code.html`
  - [x] Show a text input field
  - [x] Submit button
- [x] Add remember this device checkbox conditionally to
  - [x] `enter_password.html`
  - [x] `enter_totp.html`
  - [x] `enter_oob_otp.html`
- [x] Automatically bypass 2nd authentication when device token is available

louischan-oursky:
@koskosm

PR #128, opened 2020-08-06
Setup recovery code

louischan-oursky:
ref #56

### DIFF
--- a/pkg/auth/dependency/newinteraction/intents/authenticate.go
+++ b/pkg/auth/dependency/newinteraction/intents/authenticate.go
@@ -270,21 +270,32 @@ func (i *IntentAuthenticate) DeriveEdgesForNode(ctx *newinteraction.Context, gra
 			}, nil
 
 		case newinteraction.AuthenticationStageSecondary:
-			var reason auth.SessionCreateReason
-			_, ok := graph.GetNewUserID()
-			if ok {
-				reason = auth.SessionCreateReasonSignup
-			} else {
-				reason = auth.SessionCreateReasonLogin
-			}
-
 			return []newinteraction.Edge{
-				&nodes.EdgeDoCreateSession{Reason: reason},
+				&nodes.EdgeGenerateRecoveryCodeBegin{},
 			}, nil
 
 		default:
 			panic(fmt.Errorf("interaction: unexpected authentication stage: %v", node.Stage))
 		}
+
+	case *nodes.NodeGenerateRecoveryCodeEnd:
+		return []newinteraction.Edge{
+			&nodes.EdgeDoGenerateRecoveryCode{
+				RecoveryCodes: node.RecoveryCodes,
+			},
+		}, nil
+	case *nodes.NodeDoGenerateRecoveryCode:
+		var reason auth.SessionCreateReason
+		_, ok := graph.GetNewUserID()
+		if ok {
+			reason = auth.SessionCreateReasonSignup
+		} else {
+			reason = auth.SessionCreateReasonLogin
+		}
+
+		return []newinteraction.Edge{
+			&nodes.EdgeDoCreateSession{Reason: reason},
+		}, nil
 
 	case *nodes.NodeDoCreateSession:
 		// Intent is finished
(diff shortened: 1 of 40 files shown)


pkg/auth/handler/webapp/renderer.go
```diff
@@ -51,3 +54,29 @@ func (r *HTMLRenderer) Render(w http.ResponseWriter, req *http.Request, template
...
+	).WithPreferredLanguageTags(preferredLanguageTags).RenderTemplate(
+		templateType,
+		data,
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	body := []byte(out)
+	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
```
kiootic, line 78: I suppose this should be passed in as a parameter.

pkg/auth/handler/webapp/renderer.go
```diff
@@ -51,3 +54,29 @@ func (r *HTMLRenderer) Render(w http.ResponseWriter, req *http.Request, template
...
+	w.Header().Set("Content-Length", strconv.Itoa(len(body)))
+	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s\"", filename))
```
kiootic, line 80: Let's assume `filename` does not need escaping, since we're not passing dynamic data in filename?

templates/auth_ui_setup_recovery_code.html
```diff
@@ -0,0 +1,51 @@
...
+
+<p class="description primary-txt">{{ localize "recovery-code-consumption-description" }}</p>
+
+<div class="setup-recovery-code-buttons">
+
+	<button class="btn primary-btn" type="submit" name="submit" value="">{{ localize "next-button-label" }}</button>
+
+	<a class="btn primary-btn" target="_blank" href="{{ call $.MakeURLState "" "download" "true" }}">{{ localize "download-button-label" }}</a>
+
+	<!-- FIXME(mfa): Add regenerate button -->
```
kiootic, line 41: I don't think we need regenerate button in setting up recovery code page?

pkg/auth/dependency/newinteraction/nodes/generate_recovery_code_begin.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"github.com/authgear/authgear-server/pkg/auth/dependency/authenticator"
+	"github.com/authgear/authgear-server/pkg/auth/dependency/newinteraction"
+)
+
+func init() {
+	newinteraction.RegisterNode(&NodeGenerateRecoveryCodeBegin{})
+}
+
+type EdgeGenerateRecoveryCodeBegin struct{}
```
kiootic, line 12: It's better to name it `EdgeGenerateRecoveryCode`, since the edge can go to both Begin & End node.

pkg/auth/dependency/newinteraction/nodes/do_generate_recovery_code.go
```diff
@@ -0,0 +1,43 @@
...
+	}, nil
+}
+
+type NodeDoGenerateRecoveryCode struct {
+	RecoveryCodes []string `json:"recovery_nodes"`
+}
+
+func (n *NodeDoGenerateRecoveryCode) Apply(perform func(eff newinteraction.Effect) error, graph *newinteraction.Graph) error {
+	err := perform(newinteraction.EffectRun(func(ctx *newinteraction.Context) error {
+		if len(n.RecoveryCodes) > 0 {
```
kiootic, line 27: Personally I'd like to avoid conditionally update the database in Do nodes (i.e. Do nodes always do something, otherwise it is skipped), but it does seems cleaner to just enter the Do nodes unconditionally.

What do you think about this? If you think it's better I'll update `DoVerifyIdentity` to use the same way.

pkg/auth/dependency/newinteraction/intents/authenticate.go
```diff
@@ -147,16 +147,11 @@ func (i *IntentAuthenticate) DeriveEdgesForNode(ctx *newinteraction.Context, gra
...
-			return []newinteraction.Edge{
-				&nodes.EdgeDoVerifyIdentity{
-					Identity:         node.Identity,
-					NewAuthenticator: node.NewAuthenticator,
-				},
-			}, nil
-		}
 		return []newinteraction.Edge{
-			&nodes.EdgeDoUseIdentity{Identity: node.Identity},
+			&nodes.EdgeDoVerifyIdentity{
```
kiootic, line 151: There are other places using `EdgeDoVerifyIdentity`: `add_identity`, `update_identity`, `verify_identity`, `verify_identity_resume`. Can you help to fix them too? Thanks!

PR #151, opened 2020-08-11
Select alternative authenticator

louischan-oursky:


pkg/auth/dependency/newinteraction/nodes/create_authenticator_begin.go
```diff
@@ -162,16 +162,21 @@ func (n *NodeCreateAuthenticatorBegin) deriveSecondary() (edges []newinteraction
...
+	// If yes, then no secondary authenticator is needed.
 	ais := filterAuthenticators(
 		n.Authenticators,
 		authenticator.KeepTag(authenticator.TagSecondaryAuthenticator),
 	)
+	if len(ais) > 0 {
+		return nil
+	}
+
+	// 3. Determine what secondary authenticator we allow the user to create.
```
kiootic, line 188: If we added the `len(ais) > 0` condition check above, I suppose some of the conditions can simplified: condition B is always true.

Issue #154, opened 2020-08-11
Re-organize packages


kiootic:
@louischan-oursky please review this re-organization plan and let me know what you think.

- `pkg`
    - `pkg/auth`: Authgear API server logic
        - `pkg/auth/handlers`: HTTP handlers
        - `pkg/auth/webapp`: WebApp service
        - `pkg/auth/routes.go`: Route definitions
        - `pkg/auth/endpoints.go`: Endpoint provider
        - `pkg/auth/wire_*.go`: API server DI config
    - `pkg/admin`: Admin API server logic (content TBD)
    - `pkg/resolver`: Session resolver API server logic
    - `pkg/worker`: Background task worker logic
        - `pkg/worker/tasks`: Task logic
        - `pkg/worker/worker.go`: Worker entry point
        - `pkg/worker/wire_*.go`: Task worker DI config
    - `pkg/lib`: Authgear domain logic & data persistence
        - `pkg/lib/config`: Configurations
        - `pkg/lib/authn`: Core authentication
            - `pkg/lib/authn/interaction`
            - `pkg/lib/authn/authenticator`
            - `pkg/lib/authn/identity`
            - `pkg/lib/authn/challenge`
            - `pkg/lib/authn/sso`
            - `pkg/lib/authn/user`
        - `pkg/lib/oauth`: OAuth
            - `pkg/lib/oauth/oidc`
        - `pkg/lib/session`: Session
        - `pkg/lib/hook`: Web-hook
        - `pkg/lib/api`: API definitions
            - `pkg/lib/api/models`
            - `pkg/lib/api/events`
            - `pkg/lib/api/errors`
        - `pkg/lib/features`: Additional Features
            - `pkg/lib/features/forgotpassword`
            - `pkg/lib/features/verification`
            - `pkg/lib/features/welcomemessage`
        - `pkg/lib/infra`: Supporting infrastructure
            - `pkg/lib/infra/db`
            - `pkg/lib/infra/redis`
            - `pkg/lib/infra/template`
            - `pkg/lib/infra/sms`
            - `pkg/lib/infra/mail`
            - `pkg/lib/infra/middleware`
    - `pkg/util`: Utilities (not specific to Authgear, not depending on config)
        - `pkg/util/log`
        - `pkg/util/httputil`
        - `pkg/util/jwkutil`
        - `pkg/util/image`
        - `pkg/util/httproute`
        - `pkg/util/clock`

louischan-oursky:
[...]

PR #156, opened 2020-08-12
Re-organize packages (Part 1)

kiootic:
ref #154 

Most of the packages in `pkg/core` is re-located (only `pkg/core/authn` remains for now).

I also took the chance to refactor some of the packages:
- `utils` package is too generic; refactored to `slice` package.
- Sentry middleware are grouped together with other middlewares in `lib/infra/middleware` package.
- Validation utilities are no longer aware of API errors; instead, API errors would understand & translate validation errors.


pkg/auth/dependency/authenticator/password/provider.go
```diff
@@ -5,9 +5,9 @@ import (
 
 	"github.com/authgear/authgear-server/pkg/auth/config"
 	taskspec "github.com/authgear/authgear-server/pkg/auth/task/spec"
-	"github.com/authgear/authgear-server/pkg/core/errors"
 	"github.com/authgear/authgear-server/pkg/task"
 	"github.com/authgear/authgear-server/pkg/util/clock"
+	"github.com/authgear/authgear-server/pkg/util/errors"
```
louischan-oursky, line 10: It would be great if we also remove the overlapping function (with `errors`) in `util/errors`. To avoid renaming when we need to import both, we can rename it to `util/errorutil`.

kiootic: We can remove `errors.New` & `errors.Newf`, and replace them with `errors.New` (from standard `errors` package) and `fmt.Errorf`.

I suppose we can also cleanup other under-utilized functions with little benefits?
- barriers (`errors.HandledWithMessage`)
- secondary errors (`errors.WithSecondaryError`)

louischan-oursky: Keep them please! They are useful. It is just our fault that they are not actively being used.

PR #158, opened 2020-08-13
MFA with device token

louischan-oursky:
ref #56

### DIFF
--- a/pkg/mfa/cookie.go
+++ b/pkg/mfa/cookie.go
@@ -1,0 +1,27 @@
+package mfa
+
+import (
+	"net/http"
+
+	"github.com/authgear/authgear-server/pkg/lib/config"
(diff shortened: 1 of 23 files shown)


pkg/mfa/cookie.go
```diff
@@ -0,0 +1,27 @@
...
+
+type CookieDef struct {
+	Def *httputil.CookieDef
+}
+
+func NewDeviceTokenCookieDef(r *http.Request, cfg *config.AuthenticationConfig) CookieDef {
+	def := &httputil.CookieDef{
+		Name:     CookieName,
+		Path:     "/",
+		SameSite: http.SameSiteLaxMode,
```
kiootic, line 20: I think it can be `Strict`, we expect the cookie to be used through navigation within the auth UI only.

PR #160, opened 2020-08-13
Re-organize packages (part 2)

kiootic:
ref #154


pkg/lib/infra/task/task.go
```diff
@@ -4,19 +4,23 @@ import (
...
 
+type Param interface {
+	TaskName() string
+}
+
 type Spec struct {
 	Name  string
-	Param interface{}
+	Param Param
 }
```
louischan-oursky, line 14: The struct `Spec` now seems obsolete. We can just pass the `Param` interface around.

PR #162, opened 2020-08-14
Introduce new template engine

### LINKED ISSUE #142: Support literal HTML in translation
- Improve iawaknahc/gomessageformat
  - In addition to FormatPositional, expose FormatNamed
  - In addition to FormatNamed, expose FormatHTMLTemplate
  - Add a runtime function `mf "plural" .arg_name "one"` `mf "select" .arg_name "male"` to be used along with FormatHTMLTemplate
- Update template engine to use FormatHTMLTemplate and add every translation as nested template
- Convert existing template from `{{ localize "translation-key" .a .b .c }}` to `{{ template "translation-key" (makeMap "a" .a "b" .b "c" .c) }}`

louischan-oursky:
fix #142


pkg/util/template/engine.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func (e *Engine) renderText(ctx *RenderContext, resolved *Resolved, data interface{}) (out string, err error) {
+	t := texttemplate.New(resolved.T.Type)
+
+	// Inject the funcs map before parsing any templates.
+	// This is required by the documentation.
+	t.Funcs(texttemplate.FuncMap{
+		messageformat.TemplateRuntimeFuncName: messageformat.TemplateRuntimeFunc,
+		"makemap": func(pairs ...interface{}) map[string]interface{} {
```
kiootic, line 133: Can we extract this as a standalone function? Also, this function is often used in templates supplied by developers. We may want to explicitly check for errors in arguments (e.g. odd number of arguments, non-string keys), and give a better error to developers.

louischan-oursky:
Updated.

Issue #117, opened 2020-08-05
Unify all tables common attributes


louischan-oursky:
- Common attributes
  - [x] ID
  - [x] Created At
  - [x] Updated At
  - [ ] Labels (JSON Object)
  - [x] App ID
- Entities
  - [x] User
  - [x] Identity
  - [x] Authenticator
  - [x] Session (in Redis)
  - [x] Recovery code
  - [x] OAuth authorization

louischan-oursky:
I am closing this and going to open a new issue for considering adding labels.

PR #164, opened 2020-08-17
Basic GraphQL Admin API

kiootic:


louischan-oursky:
Looks good. How is the study on dataloader integration?

kiootic:
I'll implement some connections (i.e. identities) before start investigating data loader.

By the way, we discussed on fixing the nullability of the edge & connection types. I just did a quick search and found that [nullable edges are intentional by spec](https://stackoverflow.com/a/57908071/13196616), it allows better error reporting.

kiootic:
@louischan-oursky Add authorization check for admin API.

I'll fix #117 (without labels) before continuing to add identities & authenticators.

pkg/admin/loader/user.go
```diff
@@ -13,6 +13,7 @@ type UserLoader struct {
 }
 
 func (l *UserLoader) Get(id string) (interface{}, error) {
+	println("get user " + id)
```
louischan-oursky, line 16: Remove this line.

PR #165, opened 2020-08-18
Portal client setup

louischan-oursky:


louischan-oursky:
@kiootic I extracted the server listen logic in https://github.com/authgear/authgear-server/pull/165/commits/48b6c0a80afd66a4c5f3e0bb2fdd866c37c7578d

louischan-oursky:
I discovered some problem with cookie with plain HTTP is used. Dont merge this yet.

louischan-oursky:
The problem was I have some cookies set on `https://localhost:3000`. When I switched to `http://localhost:3000`, we tried to write the same one without Secure, which is considered insecure by the user agent.

portal/src/index.scss
```diff
@@ -0,0 +1,40 @@
+html,
+body {
+  min-height: 100vh;
+}
+
+html {
```
kiootic, line 6: IIRC we're going to use fluent design, so these CSS is placeholder right?

portal/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+// This file is for typechecking only.
+// Parcel uses Babel to compile our TypeScript source files.
+// So here we only include options that are related to typechecking.
+{
+  "compilerOptions": {
```
kiootic, line 5: Since this should be used for typechecking only, let's add `noEmit` to tsconfig directly?

cmd/authgear/server/server.go
```diff
@@ -73,67 +63,31 @@ func (c *Controller) Start() {
...
 	if c.ServeMain {
-		c.startServer(cfg, serverMain, cfg.ListenAddr, auth.NewRouter(p, configSource))
-	}
-	if c.ServeResolver {
-		c.startServer(cfg, serverResolver, cfg.ResolverListenAddr, resolver.NewRouter(p, configSource))
-	}
-	if c.ServeAdmin {
-		c.startServer(cfg, serverAdminAPI, cfg.AdminListenAddr, admin.NewRouter(p, configSource))
+		specs = append(specs, server.Spec{
+			Name:          string(serverMain),
```
kiootic, line 75: I suppose the `serverType` type is no longer needed after this refactor. Let's remove it and use string literal as server name directly.

portal/src/BuildConfig.ts
```diff
@@ -0,0 +1,5 @@
+/* global process */
+// @ts-expect-error
+export const BUILD_TIME_VARIABLE: string = process.env.BUILD_TIME_VARIABLE;
```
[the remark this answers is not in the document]
louischan-oursky: Right. It's just documentation.

portal/src/index.scss
louischan-oursky: No. This rules in this file are not placeholder. Let me add more comments to explain the purpose of each rule.

louischan-oursky:
Updated.

kiootic (approved):

PR #166, opened 2020-08-19
Add basic common attributes

kiootic:
Only `created_at` and `updated_at` is added for now.

ref #117

### DIFF
--- a/pkg/lib/oauth/grant_offline.go
+++ b/pkg/lib/oauth/grant_offline.go
@@ -43,12 +43,16 @@ func (g *OfflineGrant) ToAPIModel() *model.Session {
 	amr, _ := g.Attrs.GetAMR()
 	acr, _ := g.Attrs.GetACR()
 	return &model.Session{
-		ID: g.ID,
+		Meta: model.Meta{
+			ID:        g.ID,
+			CreatedAt: g.CreatedAt,
+			// FIXME(session): when would a session be updated?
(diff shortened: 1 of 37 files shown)


pkg/lib/oauth/grant_offline.go
```diff
@@ -43,12 +43,16 @@ func (g *OfflineGrant) ToAPIModel() *model.Session {
 	amr, _ := g.Attrs.GetAMR()
 	acr, _ := g.Attrs.GetACR()
 	return &model.Session{
-		ID: g.ID,
+		Meta: model.Meta{
+			ID:        g.ID,
+			CreatedAt: g.CreatedAt,
+			// FIXME(session): when would a session be updated?
+			UpdatedAt: g.AccessInfo.LastAccess.Timestamp,
```
louischan-oursky, line 50: For example, when last access information is updated?

louischan-oursky: When that would make that field pretty useless. Let's keep this for now. Maybe we have a feature to let the user to assign a name to a session in the future.

kiootic: Yes, that's what I used for now. However I'm not sure if it is the 'correct' timestamp to put there, since the information would be duplicated in `LastAccessedAt`, so I add a FIXME to re-consider it when the session functionalities are more complete.

PR #167, opened 2020-08-20
Add identity & authenticator GraphQL nodes

### LINKED ISSUE #36: Minimal Admin API
- Separate 3 binding (ip/ports) (public, admin, proxy_auth)? Consideration:
   - Proxy_auth is intended for k8s gateway only, which should be internal network, and control access by network policies / firewall; While admin endpoint could be internal, or external system and auth by say client_credentials (replacing Master Key), and it will be easier if users can configure firewall by ip/ports
- web portal
  - Edit config
  - UI to Admin API
   - outsource to other team?
   - Use MS fluent UI

kiootic:
fix #36

### DIFF
--- a/pkg/admin/graphql/identity.go
+++ b/pkg/admin/graphql/identity.go
@@ -39,10 +39,25 @@ var nodeIdentity = entity(
 			},
 			"claims": &graphql.Field{
 				Type: graphql.NewNonNull(JSONObject),
+				Args: map[string]*graphql.ArgumentConfig{
+					"names": {Type: graphql.NewList(graphql.NewNonNull(graphql.String))},
+				},
 				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
+					names, hasNames := p.Args["names"].([]interface{})
 					info := loadIdentity(p.Context, p.Source)
 					claims := info.Map(func(value interface{}) (interface{}, error) {
-						return value.(*identity.Info).Claims, nil
+						claims := value.(*identity.Info).Claims
+						if hasNames {
+							filteredClaims := make(map[string]interface{})
+							for _, name := range names {
+								name := name.(string)
+								if value, ok := claims[name]; ok {
+									filteredClaims[name] = value
+								}
+							}
+							claims = filteredClaims
+						}
+						return claims, nil
 					})
 					return claims.Value, nil
 				},
--- a/pkg/admin/graphql/identity.go
+++ b/pkg/admin/graphql/identity.go
@@ -8,10 +8,10 @@ import (
 
 	"github.com/graphql-go/graphql"
 
-	"github.com/authgear/authgear-server/pkg/admin/utils"
 	"github.com/authgear/authgear-server/pkg/lib/api/model"
 	"github.com/authgear/authgear-server/pkg/lib/authn"
 	"github.com/authgear/authgear-server/pkg/lib/authn/identity"
+	"github.com/authgear/authgear-server/pkg/util/graphqlutil"
 )
 
 const typeIdentity = "Identity"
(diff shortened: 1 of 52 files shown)


pkg/lib/authn/user/store.go
```diff
@@ -95,6 +96,32 @@ func (s *Store) Get(userID string) (*User, error) {
 	return u, nil
 }
 
+func (s *Store) GetByIDs(userIDs []string) ([]*User, error) {
+	ids := make([]interface{}, len(userIDs))
+	for i, id := range userIDs {
+		ids[i] = id
+	}
+	builder := s.selectQuery().
+		Where("id IN ("+squirrel.Placeholders(len(ids))+")", ids...)
```
louischan-oursky, line 109: I remember I have done similar thing before. I used postgresql array that time so there is only one placeholder in the query.

pkg/admin/utils/lazy.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func NewLazyValue(value interface{}) *Lazy {
+	switch value := value.(type) {
+	case *Lazy:
+		return value
+	case func() (interface{}, error):
+		return NewLazy(value)
+	default:
+		return &Lazy{value: value}
```
louischan-oursky, line 20: Add a case to support making a lazy value of error?

pkg/admin/graphql/identity.go
```diff
@@ -0,0 +1,92 @@
...
+		Type: authn.IdentityType(parts[0]),
+		Meta: model.Meta{ID: parts[1]},
+	}, nil
+}
+
+func loadIdentity(ctx context.Context, obj interface{}) *utils.Lazy {
+	switch obj := obj.(type) {
+	case *identity.Info:
+		return utils.NewLazyValue(obj)
+	case *identity.Ref:
```
louischan-oursky, line 102: Why do we need to support both Ref and Info?

pkg/admin/graphql/identity.go
```diff
@@ -39,10 +39,25 @@ var nodeIdentity = entity(
 			},
 			"claims": &graphql.Field{
 				Type: graphql.NewNonNull(JSONObject),
+				Args: map[string]*graphql.ArgumentConfig{
```
louischan-oursky, line 42: What is the intended usage of `names`?

louischan-oursky (commented):
Some code organization ideas

1. Move `pkg/admin/utils/lazy.go` to `pkg/util/lazy/lazy.go`; `utils.Lazy` to `lazy.T`
2. Move `pkg/admin/utils/dataloader*` to `pkg/util/dataloader/`

pkg/lib/authn/user/store.go
kiootic: I copied this from the only place using it (password history). Maybe it is lost with time already?

Let me double check if we can use pg array instead.

pkg/admin/utils/lazy.go
kiootic: This function is like `Promise.resolve`, so it should not try to fill in the error field.

I've added a `NewLazyError` function before, but there's no use so I removed it.

pkg/admin/graphql/identity.go
kiootic: When querying using `node(id: "")` root field, the object is `identity.Info`. When querying using `users { identities }` connection, the object is `identity.Ref`.

`identity.Ref` can be constructed efficiently using just `identity` table, so navigating the connection would return `identity.Ref` instead of `identity.Info`, and its details is lazy loaded when accessing the fields.

pkg/admin/graphql/identity.go
kiootic: The API consumer may only care about some claims, e.g. `username`, `email`. This argument allows consumer to filter the claims, and avoid getting unneeded large claims like OAuth provider profile.

pkg/lib/authn/user/store.go
kiootic: It turns out the syntax you used is `id = ANY (?)` so I didn't found it...

I'll fix everywhere to use that instead.

kiootic:
@louischan-oursky I suppose your portal backend need to use `Lazy` and `DataLoader` as well? I propose moving them to `pkg/util/graphqlutil`, since they're supposed to be used together for GraphQL only.

pkg/admin/graphql/identity.go
louischan-oursky: I see. It seems like a workaround of the scheme-less situation here.

louischan-oursky:
Also, what about making `authgear start` starts all 3 servers by default?

kiootic:
> Also, what about making authgear start starts all 3 servers by default?

Sure, let me change it as well.

Issue #145, opened 2020-08-10
Minimal Web Portal


chpapa:
Plan / Design A minimal Web Portal that can:

- Interact with Admin API (maybe list all users?)
- Save / Export a Authgear Server config (maybe disable/enable anonymous user?)

louischan-oursky:
# Overview

The portal consists of a web frontend and a backend server.
It is distributed as a Docker image.

# Operation mode of the portal

Here are the 2 operation modes supported by the portal.

## Self-hosted Authgear and portal

- Suppose `mycompany.com` is a domain owned by our compnany. It is shared by many projects.
- `myapp.mycompany.com` is the host of our application server.
- `accounts.myapp.mycompany.com` is the host of the Authgear acting as the authentication server of our application server.
- Authgear hosted on `accounts.myapp.mycompany.com` must set `Domain` attribute of the cookies it writes to `myapp.mycompany.com`.
- `myapp-admin.mycompany.com` is the host of the portal backend.
- `accounts.myapp-admin.mycompany.com` is the host of the Authgear acting as the authentication server of the portal backend.
- Authgear hosted on `accounts.myapp-admin.mycompany.com` must set `Domain` attribute of the cookies it writes to `myapp-admin.mycompany.com`.

In this setup, the developer needs to deploy 1 instance of Authgear.
That instance of Authgear serves 2 tenants: 1 tenant for the application, 1 tenant for the portal.
Since there are 2 tenants, the set of users are **COMPLETELY** isolated.

Both Authgear and the portal backend reads the same configuration from filesystem.
On Kubernetes, it is expected that Authgear and the portal are separate Deployments, with the same ConfigMap/Secret mounted on their filesystem.
The ConfigMap/Secret are usually managed by tools like Helm, therefore, direct write to the ConfigMap/Secret is not supported.
Instead, the portal offers an option to export the current configuration as a tarball.

The exported tarball contains all configuration and resources.
The filesystem structure looks like

```
/authgear.yaml
/authgear.secrets.yaml
/templates/auth_ui_login.html
...
```

## Cloud hosted Authgear and portal

- Suppose `authgearapps.com` is the domain of the cloud hosted Authgear.
- `myapp.authgearapps.com` is the host of a particular tenant of Authgear.
- `myapp.authgearapps.com` is the host of the Authgear acting as the authentication server of some external application server.
[...]

PR #170, opened 2020-08-20
Portal authentication via Authgear

louischan-oursky:
ref #145


pkg/lib/upstreamapp/context.go
```diff
@@ -0,0 +1,35 @@
...
+
+type contextKeyType struct{}
+
+var contextKey = contextKeyType{}
+
+type contextValue struct {
+	SessionInfo *session.Info
+}
+
+func WithSessionInfo(ctx context.Context, info *session.Info) context.Context {
```
kiootic, line 17: I suppose we should move `pkg/lib/session.Info` to `pkg/lib/api`, since the session headers are supposed to be public API. Then, the middleware & context can be moved there as well.

portal/src/index.tsx
```diff
@@ -8,12 +8,27 @@ import "core-js/stable";
...
 import "./index.scss";
 
 import React from "react";
 import { render } from "react-dom";
+import authgear from "@authgear/web";
 
 import App from "./App";
 
-render(<App />, document.getElementById("react-app-root"));
+fetch("/api/runtime-config.json")
```
kiootic, line 19: I think this initialization should be done after first render? We should show loading indicator while initializing.

portal/src/App.tsx
```diff
@@ -14,7 +16,48 @@ const query = graphql`
 const ShowQueryResult: React.FC<AppQueryResponse> = function ShowQueryResult(
   props: AppQueryResponse
 ) {
-  return <div>{props.viewer?.id}</div>;
+  const { viewer } = props;
+  const onClickLogout = useCallback(() => {
+    authgear
+      .logout({
+        redirectURI: window.location.href,
```
kiootic, line 23: Since we're doing exact match validation of redirectURI at server-side, I suppose we should give a hard-coded value here (i.e. through build time variables), since the current page URL may change.

louischan-oursky: Sure. I am going to change it to `window.location.origin + "/"`.

louischan-oursky:
Updated.

Issue #169, opened 2020-08-20
Multi-tenant mode config


kiootic:
- Watch directory for tenant configurations, lazy load config on request
    - `/index.json`: a JSON object with host as key, and app ID as value
    - `/{app ID}/`: directory containing the configuration for the app
- Single tenant mode: configure a directory for app config, instead of individual config files
    - Templates are resolved relative to the `authgear.yaml` file.

PR #176, opened 2020-08-21
Hot-reload local configuration files

kiootic:
@louischan-oursky 
As discussed previously, we would like to refactor `ProvideConfig` into `ResolveAppIDByRequest` and `ResolveConfigByAppID` functions. However, I realized that resolving app ID depends on the host-appID mapping, which is strongly coupled with resolving config logic. Therefore I suppose we should keep it as-is for now.

ref #169


louischan-oursky:
Since we are going to have host-appID mapping independent of `authgear.yaml`, should we just remove `host` from `authgear.yaml`? So in local fs mode, every host is allowed. We assume the developer has setup the DNS/routing correctly.

louischan-oursky:
Anyway this PR looks good to me so I am going to merge it. Let me open a new issue to discuss the host-appid problem.

PR #177, opened 2020-08-21
List apps

louischan-oursky:
ref #145


pkg/portal/service/deps.go
```diff
@@ -0,0 +1,10 @@
+package service
+
+import (
+	"github.com/google/wire"
+)
+
+var DependencySet = wire.NewSet(
+	NewLibConfig,
```
kiootic, line 8: This means if the app config is invalid, the portal backend does not work, since the GraphQL handler cannot be constructed. Is that intentional?

portal/src/Authenticated.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+interface Empty {}
+
+interface Props {
+  children?: React.ReactElement;
+}
+
+const Authenticated: React.FC<Props> = function Authenticated(ownProps: Props) {
+  return (
+    <QueryRenderer<{ variables: Empty; response: AuthenticatedQueryResponse }>
```
kiootic, line 59: Since we never intend to pass anything as variable, is it possible to use `never` here?

louischan-oursky: QueryRenderer always requires us to pass `variable` So the query does not make use of any variables, we still need to pass an empty object. It is impossible to use `never` here.

kiootic (approved):

PR #192, opened 2020-08-25
Add kubernetes configuration source

### LINKED ISSUE #169: Multi-tenant mode config

kiootic:
fix #169


pkg/lib/config/configsource/kubernetes.go
```diff
@@ -0,0 +1,148 @@
...
+	"k8s.io/client-go/util/homedir"
+
+	"github.com/authgear/authgear-server/pkg/lib/config"
+	"github.com/authgear/authgear-server/pkg/util/log"
+)
+
+type KubernetesLogger struct{ *log.Logger }
+
+func NewKubernetesLogger(lf *log.Factory) KubernetesLogger {
+	return KubernetesLogger{lf.New("kubernetes-config")}
```
louischan-oursky, line 38: The logger name should be `configsource-kubernetes`

pkg/lib/config/configsource/kubernetes.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+type Kubernetes struct {
+	Logger LocalFSLogger
```
louischan-oursky: Should be `KubernetesLogger`

pkg/lib/config/configsource/kubernetes.go
```diff
@@ -91,6 +91,7 @@ func (k *Kubernetes) Open() error {
 	secretCtrl := k.newController(corev1.ResourceSecrets, &corev1.Secret{}, k.onUpdate, k.onDelete)
 	go configMapCtrl.Run(done)
 	go secretCtrl.Run(done)
+	go k.cleanupCache(done)
```
louischan-oursky, line 94: Consider using [this library](https://github.com/allegro/bigcache) instead? It has max number of entries, max cache size in bytes support.

pkg/lib/config/configsource/config.go
```diff
@@ -22,8 +22,8 @@ type Config struct {
 	KubeNamespace string `envconfig:"KUBE_NAMESPACE"`
 	// KubeAppHostMapName indicates the name of app host mapping ConfigMap
 	KubeAppHostMapName string `envconfig:"KUBE_APP_HOST_MAP_NAME" default:"app-hosts"`
-	// KubeAppConfigPattern sets the RegEx that matches the app ID in ConfigMap/Secret name
-	KubeAppConfigPattern string `envconfig:"KUBE_APP_CONFIG_PATTERN" default:"^app-data-(.+)$"`
+	// KubeAppConfigPrefix sets the name prefix of app ConfigMap/Secret name
+	KubeAppConfigPrefix string `envconfig:"KUBE_APP_CONFIG_PREFIX" default:"app-data-"`
```
louischan-oursky, line 26: Should we use labels instead to identity our ConfigMap and Secret?

pkg/lib/config/configsource/kubernetes.go
kiootic: It require byte slices as value, which make the current behavior (fetch at most once, even if there are concurrent requests) impossible.

pkg/lib/config/configsource/config.go
kiootic: Sure, we can use labels. So:
- A label to indicate the host mapping ConfigMap, e.g. `https://authgear.com/kubernetes/host-mapping: true`
- A label to indicate the app config ConfigMap & Secret, e.g. `https://authgear.com/kubernetes/config-app-id: 9C390E3B-9030-48B0-A350-57F51572145F`

Both label names are constants and cannot be configured. Does that sounds good to you?

louischan-oursky:
I update the config map to remove login ID key and then refresh the page. There is a log `invalidated cached config` but the login ID is still show removed login ID key.

louischan-oursky:
It turned out my configuration was correct.

pkg/lib/config/configsource/kubernetes.go
```diff
@@ -30,6 +29,11 @@ import (
 	"github.com/authgear/authgear-server/pkg/util/log"
 )
 
+const (
+	LabelHostMapping = "authgear.com/host-mapping"
```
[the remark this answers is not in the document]
kiootic: Label keys can have at most one slash: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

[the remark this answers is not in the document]
louischan-oursky: Didn't know that! 😳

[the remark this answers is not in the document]
louischan-oursky: What about `kubernetes.authgear.com/host-mapping`?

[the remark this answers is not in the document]
kiootic: Sure that's valid, but do we need to include 'kubernetes' somewhere in the label key? I think the context makes it pretty clear already.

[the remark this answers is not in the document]
louischan-oursky: Fair enough.

PR #197, opened 2020-08-26
Proxy Admin API graphql endpoint

louischan-oursky:
ref #145

### DIFF
--- a/pkg/lib/admin/authz/middleware_test.go
+++ b/pkg/lib/admin/authz/middleware_test.go
@@ -1,0 +1,67 @@
+package authz_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/lestrrat-go/jwx/jwk"
+	. "github.com/smartystreets/goconvey/convey"
+
+	adminauthz "github.com/authgear/authgear-server/pkg/lib/admin/authz"
+	"github.com/authgear/authgear-server/pkg/lib/config"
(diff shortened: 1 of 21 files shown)


pkg/portal/config/admin_api.go
```diff
@@ -0,0 +1,17 @@
+package config
+
+import (
+	"github.com/authgear/authgear-server/pkg/lib/config"
+)
+
+type AdminAPIType string
+
+const (
+	AdminAPITypeLocal AdminAPIType = "local"
```
kiootic, line 10: I suppose this name is misleading? In this mode, the admin API endpoint is configured by providing a **static** endpoint, and the endpoint may not be a *local* endpoint. I propose changing it to `static` instead.

pkg/lib/admin/authz/middleware_test.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+
+			err = authzAdder.AddAuthz(m.Auth, m.AppID, m.AuthKey, r.Header)
+			So(err, ShouldBeNil)
+
+			recorder := httptest.NewRecorder()
+			handler := m.Handle(h)
+			handler.ServeHTTP(recorder, r)
+
+			So(string(recorder.Body.Bytes()), ShouldEqual, "good")
```
kiootic, line 79: Can you add one more test cases for failed authz?

kiootic (approved):

PR #201, opened 2020-08-26
Use URL for Redis connection credentials

### LINKED ISSUE #193: Connect to redis over TLS
To improve security we shall support optionally connecting redis over tls. One use case is connecting to aws ElasticCache, which they supports TLS out of the box.

redis supported tls since version 6: https://redis.io/topics/encryption

Not sure do this require any work on authgear as the driver handled it automatically (except maybe we need set `tls-auth-clients no`).

Other optional nice to have features:
- configuration for custom CA
- mTLS (i.e. `tls-auth-clients yes`)

kiootic:
TLS can be used using `rediss://localhost`. (note the double s)

fix #193

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@
 /tls-key.pem
 
 # docker
-docker-compose.override.yml
+docker-compose.override.yaml
 
 # vs code
 .vscode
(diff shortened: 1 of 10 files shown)


rickmak:
How we specify the cert/key?

louischan-oursky:
Do you need it? We were think of adding support for custom CA cert bundle for all kind of TLS (HTTP, Redis, PostgreSQL), not just for Redis only. But we did not know if anyone is going to use it so it is not done yet.

rickmak:
I see. Sorry I misunderstand the issue scope.

Issue #198, opened 2020-08-26
Make claims verifiable


kiootic:
Currently, the minimum unit of verification is identity. However, an identity may have both email & phone number, and we want to verify them separately. We should instead make claim the unit of verification.

Logic changes:
- Add `verified_claims` property (`Set<Claim>; Claim { name: string, value: string }`) to user entity:
    - maybe normalized as `_auth_verified_claims`, or JSONB in user row
- Add verified claim to user in verification process & add OOB authenticator process
    - do not add authenticator in verification process
- For now, assume every identity may have either email or phone number, never both

Config changes:
```yaml
# remove verification config in login ID key config
verification:
    # criteria, code_expiry_seconds, etc.
    claims:
        email:
            enabled: true
            required: true
        phone_number:
            enabled: true
            required: true
````

louischan-oursky:
[Ben's comment](https://github.com/authgear/authgear-server/pull/202#issuecomment-681967774)

> @kiootic @louischan-oursky thanks for the update, I understand the reason behind it.
> 
> But I’m a bit puzzled by the change of specification and configuration.
> 
> Comment for readability of the specification:
> 
> It is not clear and hard to reason the configuration -- there are no claims for user but only for each identity (which is Login ID / OAuth Provider), I think it need to be explained in the spec for readability of future developer.
> Some questions / problems as follows:
> 
> The logic of user doesn't have claims, while each identity has claims, make the configuration hard to guess / make sense. It should be verification.[identity].claims instead of verification.claims
> The problem of 1 also make it impossible to handle the following use cases:
> 2.1. What if I want to trust Google’s email but not Wechat’s email? And both return the claims called email?
> 2.2. What if I want to have multiple login ID with email type, and only the 1st one require verification?
> Just to be sure, so if a OIDC provider return a claim called phone instead of phone_number, it will be verification.claims.phone?

louischan-oursky:
> It is not clear and hard to reason the configuration -- there are no claims for user but only for each identity (which is Login ID / OAuth Provider), I think it need to be explained in the spec for readability of future developer.

The Identity Claims is specified in `/docs/specs/user-model.md#identity-claims`, while the User Claims (ID Token) is specified in `/docs/specs/oidc.md#id-token`. I suggest we specify the User Claims in `/docs/specs/user-model.md` and the ID Token session just reference to it.

We should also clarify that there is no `email` claim in the User Claims at the moment, since we do not have primary identity.

[...]

PR #202, opened 2020-08-26
Update configuration shape for verifiable claims

kiootic:
@chpapa Discussed with @louischan-oursky that we'd like to make individual claims verifiable. The API (OIDC tokens) should not have breaking changes, but configuration need to have breaking changes, so we update the configuration shape now.

ref #198

### DIFF
--- a/docs/specs/verification.md
+++ b/docs/specs/verification.md
@@ -7,6 +7,9 @@
   * [Code &amp; Message](#code--message)
   * [Status Flag](#status-flag)
   * [Future enhancement](#future-enhancement)
+  
+
+**TODO: update for verifiable claims (#198)**
 
 ## Definitions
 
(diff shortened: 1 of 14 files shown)


chpapa:
@kiootic @louischan-oursky thanks for the update, I understand the reason behind it.

But I’m a bit puzzled by the change of specification and configuration.

The previous specification / configuration seems is about marking if each login ID require verification or not, if the type is `email` or `phone`. However it is now deleted and replaced with claims.

So my two questions are:

1. What about login ID now? Or now it is like if `verification.claims.email.required=true`, all Login ID with type `email` will require verification? All it is with key name `email` will require verification? It is not obvious from the spec / sample configuration.
1.1. If it is key name `email`, then `verification.claims` is a bit misleading, maybe should call it `verification.login_id.email`?
1.2. And does `verification.claims.email` applies to ALL OIDC provider which return a claim called `email`? What if I want to trust Google’s email but not Wechat’s email? Maybe instead we should require specifying which OIDC provider?
1.3. Just to be sure, so if a OIDC provider return a claim called `phone` instead of `phone_number`, it will be `verification.claims.phone`?

2. It seems from the change log of `docs/specs/verification.md`, it deleted all content about login ID and replaced with claims, but login ID’s verification is still important right?

chpapa:
(I made some heavy edits of last comment to make discussion easier, please read on Github and ignore if you got last comment from email)

louischan-oursky:
Let's continue the discussion [here](https://github.com/authgear/authgear-server/issues/198)

PR #206, opened 2020-08-27
Integrate fluent UI react 

louischan-oursky:
ref #145


portal/src/ScreenLayout.module.scss
```diff
@@ -0,0 +1,37 @@
+.root {
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
+  width: 100vw;
+}
+
+.body {
+  /* flex: 1 0 0% is an idiom */
+  /* If it is flex: 1 0 0, we will have this bug https://github.com/philipwalton/flexbugs#flexbug-4 */
```
kiootic, line 10: Actually, is it fine to just drop IE support?

portal/src/index.tsx
```diff
@@ -12,7 +12,13 @@ import "./index.scss";
 
 import React from "react";
 import { render } from "react-dom";
+import { initializeIcons } from "@fluentui/react";
+import { loadTheme } from "@fluentui/react/lib/Styling";
```
kiootic, line 16: I suppose we should import from `@fluentui/react` instead.

portal/src/theme.tsx
```diff
@@ -0,0 +1,65 @@
+import { createTheme } from "@fluentui/react/lib/Styling";
```
kiootic, line 1: I suppose we should import from `@fluentui/react` instead.

louischan-oursky: [This is a pattern recommended by them](https://github.com/microsoft/fluentui/tree/master/packages/styling#using-the-styling-package)

portal/src/ScreenLayout.module.scss
louischan-oursky: Fluent UI supports IE 11. I think we can support IE 11 too.

portal/src/theme.tsx
kiootic: I think they recommend against importing `@fluentui/styling` directly, not importing from `@fluentui/react`, since the [documentation](https://developer.microsoft.com/en-us/fluentui/#/controls/web/themes) is using this pattern.

portal/src/ScreenLayout.module.scss
kiootic: I read before that MS is [dropping IE support in their own product](https://news.ycombinator.com/item?id=24200864), and Fluent UI's support of IE11 & Edge Legacy seems [in legacy maintainence mode](https://github.com/microsoft/fluentui/wiki/Browser-Support). I don't think it's a good idea to continue supporting them anymore.

louischan-oursky: Dropped support for IE 11 and we only support evergreen Edge now.

kiootic (approved):

PR #230, opened 2020-09-01
Extract translations from configuration

### LINKED ISSUE #226: Extract translation from configuration
Currently, some localizable strings are set in app configuration:
- App metadata (logo URL, display name)
- Message headers (subject, sender, reply to)

For consistency, we can move these strings as an external translation JSON template (like Auth UI translation).

ref #92

kiootic:
fix #226


pkg/util/template/engine.go
```diff
@@ -27,6 +28,23 @@ type Engine struct {
...
+func (e *Engine) RenderTranslation(ctx *RenderContext, typ string, key string, data interface{}) (string, error) {
+	translations, err := e.Resolver.ResolveTranslations(
+		&ResolveContext{PreferredLanguageTags: ctx.PreferredLanguageTags},
+		typ,
+	)
+	if err != nil {
+		return "", err
+	}
+
+	return e.renderText(ctx, &Resolved{
```
louischan-oursky: One concern here is the developer may want to include HTML in the translation (e.g. HTML email)

PR #231, opened 2020-09-01
Switch from parcel to snowpack

### LINKED ISSUE #228: Replace parcel with snowpack

louischan-oursky:
fix #228


kiootic (commented):
- I suppose we can remove the exception of `:global` in stylelint config?
- I just tried to build production bundle, and it emits a runtime warning that DCE is not enabled for React. I suppose there's some problem with Babel's config? Maybe we can use some parts for the [official snowpack+react template](https://github.com/pikapkg/snowpack/tree/master/create-snowpack-app/app-template-react-typescript).

PR #237, opened 2020-09-02
Paginated Users List

### LINKED ISSUE #234: Paginated user list
- [x] Add a function to derive data for rendering pagination widget
- [x] Add pagination widget
- [x] Switch to use page number based pagination in user list

louischan-oursky:
fix #234


portal/src/util/pagination.ts
```diff
@@ -0,0 +1,122 @@
...
+  currentOffset: number;
+  offsets: number[];
+  firstPageButtonEnabled: boolean;
+  prevPageButtonEnabled: boolean;
+  nextPageButtonEnabled: boolean;
+  lastPageButtonEnabled: boolean;
+  maxOffset?: number;
+}
+
+const PAGE_OFFSET = 3;
```
kiootic, line 17: I think we should add an explanation on what it is.
If I understood correctly, it is the number of pages before/after the current page shown in the page list?

kiootic (commented):
I tested with a 10-item first page and 1-item second page, and the pagination widget jumps around when switching page. I think it should be fixed at bottom of the page?

portal/src/util/pagination.ts
louischan-oursky: Yes.

louischan-oursky:
Added 2 commits.

kiootic (approved):

PR #242, opened 2020-09-03
Enforce length limit on user inputs

### LINKED ISSUE #222: Limit password length

### LINKED ISSUE #223: Limit login ID length
- [ ] email
- [ ] username
- [ ] phone (block extension)

kiootic:
fix #222 
fix #223

### DIFF
--- a/pkg/lib/config/identity.go
+++ b/pkg/lib/config/identity.go
@@ -165,7 +165,11 @@ func (c *LoginIDKeyConfig) SetDefaults() {
 	if c.MaxLength == nil {
 		switch c.Type {
 		case LoginIDKeyTypeUsername:
-			// An arbitrary default limit.
+			// Facebook is 50.
+			// GitHub is 39.
+			// Instagram is 30.
+			// Telegram is 32.
+			// Seems average is around about ~40 characters.
 			c.MaxLength = newInt(40)
 
 		case LoginIDKeyTypePhone:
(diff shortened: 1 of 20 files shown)


pkg/lib/config/identity.go
```diff
@@ -144,21 +144,38 @@ var _ = Schema.Add("LoginIDKeyConfig", `
...
-	if c.Maximum == nil {
-		c.Maximum = newInt(1)
+	if c.MaxAmount == nil {
+		c.MaxAmount = newInt(1)
+	}
+	if c.MaxLength == nil {
+		switch c.Type {
+		case LoginIDKeyTypeUsername:
+			// An arbitrary default limit.
+			c.MaxLength = newInt(40)
```
louischan-oursky, line 169: Add the following to the comment to show we are not picking 40 randomly.

Facebook is 50.
GitHub is 39.
Instagram is 30.
Telegram is 32.

pkg/util/password/dispatch.go
```diff
@@ -43,6 +51,10 @@ func Compare(password, hash []byte) error {
 }
 
 func TryMigrate(password []byte, hash *[]byte) (migrated bool, err error) {
+	if len(password) > MaxLength {
+		password = password[:MaxLength]
+	}
+
```
louischan-oursky, line 57: We should only limit the length of newly proposed password. For existing password, we do not place any limits on it. This allows old, long password continue to work.

pkg/util/password/dispatch.go
```diff
@@ -31,10 +31,18 @@ func resolveFormat(hash []byte) (passwordFormat, error) {
 }
 
 func Hash(password []byte) ([]byte, error) {
+	if len(password) > MaxLength {
+		password = password[:MaxLength]
+	}
```
louischan-oursky, line 36: Instead of truncating the password silently, can we return error?

Issue #241, opened 2020-09-03
Add more user columns


kiootic:
Currently, the user list only displays the user ID and created at, let's display more columns according to design:
- ID (keep it for now)
- Username (hyphen if not present) (`preferred_name`)
- Email (hyphen if not present) (`email`)
- Phone (hyphen if not present) (`phone_number`)
- Signed up time
- Last login

PR #244, opened 2020-09-03
Add more user columns

axolotl0212:
Issue #241 

Fetch extra user info from claims by updating query
Added corresponding column

# Some fields empty with placeholder

Screenshot 2020-09-03 at 5 11 23 PM

# All fields have value

Screenshot 2020-09-03 at 5 25 50 PM

### DIFF
--- a/portal/package.json
+++ b/portal/package.json
@@ -37,8 +37,10 @@
     "@authgear/web": "0.1.0",
     "@fluentui/react": "7.132.0",
     "@oursky/react-messageformat": "0.10.0",
+    "@types/luxon": "^1.24.4",
     "classnames": "2.2.6",
     "history": "5.0.0",
+    "luxon": "^1.25.0",
     "react": "17.0.0-rc.0",
     "react-dom": "17.0.0-rc.0",
     "react-router-dom": "6.0.0-beta.0"
--- a/portal/src/graphql/adminapi/UsersList.tsx
+++ b/portal/src/graphql/adminapi/UsersList.tsx
@@ -20,10 +20,13 @@ import {
   UsersListQueryVariables,
   UsersListQuery_users_edges_node_identities,
 } from "./__generated__/UsersListQuery";
+
 import ShowError from "../../ShowError";
 import PaginationWidget from "../../PaginationWidget";
+
 import { encodeOffsetToCursor } from "../../util/pagination";
 import { formatDatetime } from "../../util/formatDatetime";
+
 import styles from "./UsersList.module.scss";
 
 interface Props {
@@ -41,11 +44,17 @@ interface UserInfo {
   email: string;
 }
 
+interface IdentityClaims extends GQL_JSONObject {
+  email?: string;
+  preferred_username?: string;
+  phone_number?: string;
+}
+
 function extractUserInfoFromIdentities(
   identities: UsersListQuery_users_edges_node_identities | null
 ): UserInfo {
   const placeholder = "-";
-  const claimsList: unknown[] = [];
+  const claimsList: IdentityClaims[] = [];
 
   (identities?.edges ?? []).forEach((edge) => {
     if (edge?.node?.claims == null) {
@@ -56,16 +65,13 @@ function extractUserInfoFromIdentities(
 
   // TODO: consider update UI design to allow multiple email, username and phone number
   const email =
-    claimsList.map((claims) => (claims as any).email).filter(Boolean)[0] ||
-    placeholder;
+    claimsList.map((claims) => claims.email).filter(Boolean)[0] ?? placeholder;
   const username =
-    claimsList
-      .map((claims) => (claims as any).preferred_username)
-      .filter(Boolean)[0] || placeholder;
+    claimsList.map((claims) => claims.preferred_username).filter(Boolean)[0] ??
+    placeholder;
   const phone =
-    claimsList
-      .map((claims) => (claims as any).phone_number)
-      .filter(Boolean)[0] || placeholder;
+    claimsList.map((claims) => claims.phone_number).filter(Boolean)[0] ??
+    placeholder;
 
   return { email, username, phone };
 }
--- a/portal/src/graphql/adminapi/UsersList.tsx
+++ b/portal/src/graphql/adminapi/UsersList.tsx
@@ -101,6 +101,15 @@ const query = gql`
         node {
           id
           createdAt
+          lastLoginAt
+          identities {
+            edges {
+              node {
+                id
+                claims
+              }
+            }
+          }
         }
       }
       totalCount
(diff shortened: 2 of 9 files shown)


axolotl0212:
Note: claims do not have type, haven't come up with a good way to avoid `as any`

kiootic (commented):
@axolotl0212 
Please note that the date-time format in the design is a user-friendly one.

You may try use [Luxon](https://moment.github.io/luxon/) to format the date-time value and display them according to current locale from context of `@oursky/react-messageformat`.

axolotl0212:
> @axolotl0212
> Please note that the date-time format in the design is a user-friendly one.
> 
> You may try use [Luxon](https://moment.github.io/luxon/) to format the date-time value and display them according to current locale from context of `@oursky/react-messageformat`.

You are right, need to format datetime, will update this

portal/src/graphql/adminapi/UsersList.tsx
```diff
@@ -33,6 +34,41 @@ interface Props {
...
+    claimsList.push(edge.node.claims);
+  });
+
+  // TODO: consider update UI design to allow multiple email, username and phone number
+  const email =
+    claimsList.map((claims) => (claims as any).email).filter(Boolean)[0] ||
+    placeholder;
+  const username =
+    claimsList
+      .map((claims) => (claims as any).preferred_name)
```
kiootic, line 62: Sorry there's a typo in the claim name in issue. It should be `preferred_username`.

portal/src/graphql/adminapi/UsersList.tsx
```diff
@@ -33,6 +34,41 @@ interface Props {
   onChangeOffset?: (offset: number) => void;
 }
 
+interface UserInfo {
+  username: string;
+  phone: string;
+  email: string;
+}
+
+function extractUserInfoFromIdentities(
```
kiootic, line 43: I suppose this function would be used in other pages (e.g. user details). We may want to extract it into another file. Therefore, the input arguments should not depending on a specific query. Instead, we can define a interface structural compatible with the query type, with only the fields required by the function.

louischan-oursky: If we are going to extract it to a util function, the fields should allow undefined so different caller can use their preferred placeholder string. 

```typescript
interface UserInfo {
  username?: string;
  phone?: string;
  email?: string;
}
```

and in this particular component, we should write 

```typescript
const PLACEHOLDER = "-";

<span>{userInfo.username ?? PLACEHOLDER }}</span>
```

portal/src/graphql/adminapi/UsersList.tsx
```diff
@@ -33,6 +34,41 @@ interface Props {
...
+  identities: UsersListQuery_users_edges_node_identities | null
+): UserInfo {
+  const placeholder = "-";
+  const claimsList: unknown[] = [];
```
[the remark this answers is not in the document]
axolotl0212: This is a nice idea

portal/src/graphql/adminapi/UsersList.tsx
axolotl0212: updated

portal/src/graphql/adminapi/UsersList.tsx
axolotl0212: > we can define a interface structural compatible with the query type

So we need to cast the query result to input type when using the function?

kiootic: TypeScript uses structural typing. Simply speaking, two interface with compatible fields can be assigned to each other without casting. So the `UsersListQuery_users_edges_node_identities` can be accepted into a `UserInfo` type mentioned above.

portal/package.json
```diff
@@ -37,8 +37,10 @@
     "@authgear/web": "0.1.0",
     "@fluentui/react": "7.132.0",
     "@oursky/react-messageformat": "0.10.0",
+    "@types/luxon": "^1.24.4",
     "classnames": "2.2.6",
     "history": "5.0.0",
+    "luxon": "^1.25.0",
```
louischan-oursky, line 43: Our convention is to use `--save-exact`. Next time you can install new dependency with

```typescript
npm install --save --save-exact [--save-dev] <dep>
```

Now, you can just remove the caret.

portal/src/graphql/adminapi/UsersList.tsx
axolotl0212: I see, I defined my own compatible interface and there is no warning

portal/src/util/formatDatetime.ts
```diff
@@ -0,0 +1,16 @@
+import { DateTime } from "luxon";
+
+export function formatDatetime(
+  isoTimeString: string | null,
+  locale?: string,
+  format: string = "dd/MM/yyyy HH:mm"
```
louischan-oursky, line 6: 1. The argument `locale` should be the first argument. The reason is that it is more friendly to function currying (though javascript function are not curried automatically, but this is a common practice in functional programming)
2. Instead of a hard-coded format, we should use predefined format. In https://moment.github.io/luxon/docs/manual/formatting.html we should use `DATETIME_SHORT`

So this function should look like

```typescript
function formatDate(locale: string, date: Date | string | null): string | null {
  let datetime;
  if (date instanceof Date) {
    datetime = DateTime.fromJSDate(date);
  } else if (typeof date === "string") {
    datetime = DateTime.fromISO(date);
  }

  if (datetime == null) {
    return null;
  }

  datetime = datetime.setLocale(locale);
  return datetime.toLocaleString(DateTime.DATETIME_SHORT);
}
```

axolotl0212: > more friendly to function currying

So you mean it is more convenient to define a version for specific locale by reducing it like currying in functional programming?

louischan-oursky:
```typescript
const formatDateInEnglish = formatDate.bind("en");
const formatDateInChinese = formatDate.bind("zh");
```

portal/src/util/user.ts
```diff
@@ -0,0 +1,47 @@
...
+  node: IdentitiesEdgeNode | null;
+}
+
+interface IdentitiesEdgeNode {
+  id: string;
+  claims: GQL_JSONObject;
+}
+
+export function extractUserInfoFromIdentities(
+  identities: Identities | null
```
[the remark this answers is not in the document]
axolotl0212: I see, as `IdentityClaims` has all field optional, so it can accept any claim with type `Record<string, unknown>`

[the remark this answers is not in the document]
axolotl0212: It worked

axolotl0212:
@kiootic updated again, following the signature above

Issue #243, opened 2020-09-03
Portal edit app configuration API


kiootic:
We want to allow editing app configuration in portal.

## Cluster secrets

We may want to provide cluster-wide secret values (e.g. shared database, SMTP server). These secret value should be available to Authgear main server at runtime, but should not be exposed to individual apps.

Therefore, we should redact the secret values before returning them from portal backend, and discard them when accepting updates from frontend.

A global shared secret configuration can be specified through environment variable.

A custom keyword is set in the secret config JSON schema (e.g. `x-redacted-by: database-url`) on relevant secret values. Portal backend should handle them (redaction/discard) appropriately:
- When returning from backend, the configuration is merged with the shared secret configuration and goes through the JSON schema, then a post-processing phase would redact shared secret values.
- When accepting from frontend, the input configuration is merged the shared secret configuration and goes through the JSON schema, then a post-processing phase would remove shared secret values.

> How to support optional bring-in credentials?

## Configuration overlay

User may provide configuration files, these configuration files would be merged with the default configuration, becoming the effective configuration used by the server.

A key-value entry in configuration may have states:
- Default: Omitted in user-provided configuration, default value is used.
- Specified: Specified in user-provided configuration distinct from default value.

The third state (fixed-default) becomes of concern when user changes an entry from a non-default value to a default value. Should we clear the value (i.e. default state), or keep it in user-provided configuration (i.e. specified state)?

For simplicity, we may use this behavior:
- Entry in default state originally.
- After user changing the configuration, it goes to specified state.
- The entry stays in specified state, even if default value is specified, unless user reset the entry explicitly.

> Is this behavior good enough?

We will need some updates to the UI:
- Indicator on unsaved changes on each config entry field
- Indicator on config entry fields with non-default values (i.e. specified state)
- Action to discard unsaved changes
- Action to reset to default

> Need advices on UI design.

## Configuration saving/exporting

Portal supports two modes: multi-tenant mode based on k8s, and single-tenant mode based on local FS. The backend provide two API for saving/exporting configurations:
- Save configuration API:
    - In multi-tenant mode, the configuration is persisted in k8s.
    - In single-tenant mode, the configuration is persisted to local FS if possible (e.g. mounted as R/W volume in Docker), otherwise it fails.
- Export configuration API:
    - In both mode, the configuration is loaded from k8s/local FS, then a tarball containing the configuration files (`authgear.yaml`/`authgear.secrets.yaml`) and custom template files is returned.

[...]

PR #251, opened 2020-09-04
Return raw & effective config

kiootic:
ref #243

### DIFF
--- a/pkg/portal/graphql/app.go
+++ b/pkg/portal/graphql/app.go
@@ -24,13 +24,13 @@ var nodeApp = node(
 				return obj.(*model.App).ID, nil
 			}),
 			"appConfig": &graphql.Field{
-				Type: graphql.NewNonNull(graphqlutil.JSONObject),
+				Type: graphql.NewNonNull(AppConfig),
 				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
 					return p.Source.(*model.App).Context.Config.AppConfig, nil
 				},
 			},
 			"secretConfig": &graphql.Field{
-				Type: graphql.NewNonNull(graphqlutil.JSONObject),
+				Type: graphql.NewNonNull(SecretConfig),
 				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
 					return p.Source.(*model.App).Context.Config.SecretConfig, nil
 				},
--- a/pkg/portal/graphql/app.go
+++ b/pkg/portal/graphql/app.go
@@ -5,7 +5,9 @@ import (
 
 	"github.com/graphql-go/graphql"
 	"github.com/graphql-go/relay"
+	"sigs.k8s.io/yaml"
 
+	"github.com/authgear/authgear-server/pkg/lib/config/configsource"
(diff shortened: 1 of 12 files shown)


pkg/portal/graphql/app.go
```diff
@@ -26,13 +28,37 @@ var nodeApp = node(
 			"appConfig": &graphql.Field{
```
louischan-oursky, line 28: I suggest we rename the fields from `appConfig`, `appConfigEffective` to `rawAppConfig` and `effectiveAppConfig`

pkg/portal/model/app.go
```diff
@@ -1,10 +1,27 @@
...
+
 	"github.com/authgear/authgear-server/pkg/lib/config"
 )
 
 type App struct {
 	ID      string
 	Context *config.AppContext
 }
+
+func (a *App) LoadFile(path string) ([]byte, error) {
```
louischan-oursky, line 14: I suggest we change AppContext to include effective config and raw config.

```
type AppContext struct {
  Fs fs.Fs
  RawConfig *Config
  EffectiveConfig *Config
}
```

kiootic: I'd prefer not to do this, as it is not supposed to be used directly, and having same `*Config` type as effective config may allow potential misuse.

I intended to add a `configFile(path: String!)` GraphQL field later on to get template files. In this case, the `rawAppConfig` and `rawSecretConfig` fields would be like a shorthand to that field.

louischan-oursky: You mean?

```
configFile(path: "authgear.yaml")
configFile(path: "authgear.secrets.yaml")
configFile(path: "templates/auth_ui_translation.json")
```

I just realize [alias](https://graphql.org/learn/queries/#aliases)!

louischan-oursky: Anyway we need an extra argument `type: "raw" | "effective"` to fetch raw / effective config.

kiootic: I suppose we can keep the effective app config field, the app node would look like: 
```graphql
type App implements Node {
  id: ID!

  configFile(path: String!): String!
  """ same as configFile(path: "authgear.yaml") """
  rawAppConfig: AppConfig!
  """ same as configFile(path: "authgear.secrets.yaml") """
  rawSecretConfig: SecretConfig!
  effectiveAppConfig: AppConfig!
}
```

louischan-oursky: So effectiveAppConfig cannot be fetched via `configFile()`?

I prefer we remove the shorthand.

kiootic: > So effectiveAppConfig cannot be fetched via configFile()?

Yes, `configFile` field only return the content of the specified file in the FS, without any processing.

> I prefer we remove the shorthand.

Sure, we can remove it. So the frontend need to understand which files to fetch (`authgear.yaml` and `authgear.secrets.yaml`, or other template files), and decode the YAML at browser. Does that sounds good to you?

louischan-oursky: Oh, I didn't realize they are not strictly shorthand because the type of the field is different. Let's keep them

PR #252, opened 2020-09-04
Improve recovery code UX

### LINKED ISSUE #136: Improve recovery code UX
- Generate base32 recovery code
- Normalize (e.g. O -> 0) before comparison
- Use dash to divide long code chunks for readability (display in web UI only)

louischan-oursky:
fix #136


pkg/lib/authn/mfa/recovery_code.go
```diff
@@ -24,3 +24,12 @@ func GenerateRecoveryCode() string {
 	code := rand.StringWithAlphabet(recoveryCodeLength, base32.Alphabet, rand.SecureRand)
 	return code
 }
+
+func NormalizeRecoveryCode(code string) (normalized string, err error) {
+	normalized, err = base32.Normalize(code)
+	if err != nil {
+		err = ErrRecoveryCodeNotFound
```
kiootic, line 32: It feels weird that a normalization function returns a NotFound error. Personally I'd prefer to return the error as-is, and translate any normalization error to NotFound error in `GetRecoveryCode`.

pkg/util/base32/base32.go
```diff
@@ -35,9 +35,16 @@ func Normalize(value string) (normalizedValue string, err error) {
 	normalizedDigits := make([]rune, len(value))
 	i := 0
 	for _, digit := range value {
-		if digit == Separator {
+		isSep := false
+		for _, sep := range Separators {
+			if digit == sep {
+				isSep = true
```
kiootic, line 41: We can break here.

kiootic (approved):

PR #263, opened 2020-09-07
[Portal] authentication configuration screen login ID tab

### LINKED ISSUE #246: Portal Authentication Configuration Screen
- [x] Study the shape of authentication configuration
- [x] Study the UI design of authentication configuration
- [x] Review the mapping between the configuration file and the UI
- [x] Review existing designed screens and list out some potential reusable components
- [x] Implement the authentication configuration screen with all editable fields.

axolotl0212:
fix #246 

Will define custom interface for appConfig for now, will be handled in separate PR


portal/src/util/misc.ts
```diff
@@ -0,0 +1,25 @@
+export function getListFromCommaSeparatedString(
+  commaSeparatedString: string,
+  optionEnabled: boolean = true,
+  useDefaultList: boolean = false,
+  defaultList: string[] = []
```
louischan-oursky, line 5: The optional arguments should be placed in an optional object.

```typescript
function foobar(requiredArg: string, options?: Options = DEFAULT_OPTIONS) { }
```

portal/src/util/misc.ts
```diff
@@ -0,0 +1,25 @@
...
+): string[] {
+  if (!optionEnabled) {
+    return [];
+  }
+  const list = commaSeparatedString.split(",");
+  const sanitizedList = list.map((item) => item.trim()).filter(Boolean);
+  return useDefaultList ? [...sanitizedList, ...defaultList] : sanitizedList;
+}
+
+export function setFieldIfListNonEmpty(
```
louischan-oursky, line 15: I prefer we just put these two functions in the original caller file first. When we have more users, we can move it here.

portal/src/util/validation.ts
```diff
@@ -0,0 +1,59 @@
...
+export function isErrorResult<ErrorState>(
+  result: any
+): result is ErrorResult<ErrorState> {
+  if (!(result instanceof Object)) {
+    return false;
+  }
+  return Object.keys(result).length > 0;
+}
+
+export function renderErrorMessage(
```
louischan-oursky, line 51: We are going to rely on the backend JSON schema to validation. So we do not need this module at the moment.

portal/src/graphql/portal/AuthenticationLoginIDSettings.tsx
```diff
@@ -0,0 +1,599 @@
...
+) {
+  const switchStyle = { root: { margin: "0" } };
+  return (
+    <div className={styles.widgetHeader}>
+      <Toggle
+        styles={switchStyle}
+        checked={props.enabled}
+        onChanged={props.setEnabled}
+      />
+      <span className={styles.widgetHeaderText}>{props.title}</span>
```
kiootic, line 98: This seems to be a section header, so appropriate header tag should be used.

portal/src/graphql/portal/AuthenticationLoginIDSettings.tsx
```diff
@@ -0,0 +1,599 @@
...
+    isEmailCaseSensitive,
+    isIgnoreDotLocal,
+    isAllowPlus,
+  ]);
+
+  return (
+    <div className={styles.root}>
+      <div className={styles.widgetContainer}>
+        <ExtendableWidget
+          extendable={usernameEnabled}
```
kiootic, line 407: I'm not sure if we should not let user toggle the panel when the feature is disabled. I think we can still let user see the content of panel while the feature is disabled.

portal/src/util/misc.ts
```diff
@@ -0,0 +1,25 @@
+export function getListFromCommaSeparatedString(
```
kiootic, line 1: The behavior of this function seems quite complex. Can you add some tests to demonstrate the intended behavior?

portal/src/util/validation.ts
```diff
@@ -0,0 +1,59 @@
...
+      if (!validationRule.condition(data[validationRule.inputKey])) {
+        errorResult[validationRule.errorKey] = validationRule.errorMessageId;
+      }
+      isValid = false;
+    }
+  );
+  return { errorResult, isValid };
+}
+
+export function isValidEmail(email: string): boolean {
```
kiootic, line 32: - An empty email should be invalid.
- Validating email through complex regex is frequently fragile. For example, newer TLD frequently has more than 3 characters and fails the regex. I would suggest just checking the presence of '@' character and non-empty local & domain part.

kiootic (commented):
Suggestions on localization:
- Try avoiding `renderToString` if possible, use `FormattedMessage` instead.
- Take a component node instead of string in props when appropriate (e.g. `header`, `text`, etc.)

portal/src/styles/variable.scss
```diff
@@ -1,3 +1,13 @@
...
   0px 1.5px 3.5px $box-shadow-color;
+
+@mixin checkbox-label {
+  font-size: 14px;
+  color: rgb(50, 49, 48);
+  font-weight: 400;
+  line-height: 20px;
+  font-family: "Segoe UI", "Segoe UI Web (West European)", "Segoe UI",
+    -apple-system, BlinkMacSystemFont, Roboto, "Helvetica Neue", sans-serif;
+  -webkit-font-smoothing: antialiased;
```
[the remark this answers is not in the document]
axolotl0212: This is used for matching the style of label in Checkbox component in Fluent UI, as it is only used in one place, will move it to component stylesheet

louischan-oursky:
@axolotl0212 Next time can you just add commit to fix each review comment? It is very difficult to trace which comment was addressed if you squash the commits into the old ones.

axolotl0212:
> @axolotl0212 Next time can you just add commit to fix each review comment? It is very difficult to trace which comment was addressed if you squash the commits into the old ones.

Sure, will add commit next time

axolotl0212:
pushed one more commit for fixing as any

louischan-oursky:
Some comments

1. In ExtendableWidget, make all transition `200ms ease`. 
2. In email config, we only support [three boolean flags](https://github.com/authgear/authgear-server/blob/master/pkg/lib/config/identity.go#L91). Please remove "Block reserved keywords" and "Block Domains"
3. In username config, "Block reserved username" is a [boolean flag](https://github.com/authgear/authgear-server/blob/master/pkg/lib/config/identity.go#L122) and does not support custom reserved username.
3. In username config, "Exclude keywords" does not have default list. Remove the include default list checkbox.

axolotl0212:
Updated style and removed unused fields

PR #266, opened 2020-09-08
Update /settings design

### LINKED ISSUE #262: [UI] Update /settings to match latest design
- [x] Remove `footer.html`
- [x] Update `header.html` to match latest design
- [x] Add link to `/settings/identity`

louischan-oursky:
fix #262


static/authui/css/main.css
```diff
@@ -184,35 +183,37 @@ a.btn {
...
 }
 
 .back-btn {
-  width: 36px;
-  height: 36px;
-  background-color: #f3f2f1;
-  border-radius: 18px;
-  background-image: url("../image/ic_back_arrow.png");
-  background-position: center;
+  /* background-image: url("../image/ic_back_arrow.png"); */
```
kiootic, line 215: This can be removed.

kiootic (commented):
I think reserving space for back button in every screen would look better.
Also, it seems the font weight aren't too accurate (e.g. 'Connected Identities' header in '/settings' seems to be 600 rather than bold), do we have a detailed design file on the pages?

static/authui/css/main.css
louischan-oursky: Actually we are missing the back button icon here. But I could not export the icon from the design file. Let me add a FIXME here.

louischan-oursky:
> I think reserving space for back button in every screen would look better.

Do you mean we should add `{{ template "auth_ui_nav_bar.html" . }}` in every page even the page should be an entrypoint? e.g. add it to `login.html`

> Also, it seems the font weight aren't too accurate (e.g. 'Connected Identities' header in '/settings' seems to be 600 rather than bold), do we have a detailed design file on the pages?

The font-weight is not specified in the design file. Instead the design file references a specific font with intrinsic weight `Segoe UI Semibold`, which does not seem a best practice. So I just use `h2` and rely on the user agent style to make the font weight larger.

kiootic:
> Do you mean we should add {{ template "auth_ui_nav_bar.html" . }} in every page even the page should be an entrypoint? e.g. add it to login.html

Yes, and hide the back button in entry points.

> The font-weight is not specified in the design file. Instead the design file references a specific font with intrinsic weight Segoe UI Semibold, which does not seem a best practice. So I just use h2 and rely on the user agent style to make the font weight larger.

We're using Fluent UI as the design system. The design system [discourages using bold, and recommends using semi-bold for headers](https://developer.microsoft.com/en-us/fluentui#/styles/web/typography). Let's conform to the style guide?

louischan-oursky:
Updated

kiootic (approved):

PR #268, opened 2020-09-08
Add update app config API

kiootic:
ref #243

### DIFF
--- a/pkg/portal/graphql/app.go
+++ b/pkg/portal/graphql/app.go
@@ -78,9 +78,9 @@ var appConfigFile = graphql.NewInputObject(graphql.InputObjectConfig{
 	Name:        "AppConfigFile",
 	Description: "A configuration file to update/create.",
 	Fields: graphql.InputObjectConfigFieldMap{
-		"path": &graphql.InputObjectFieldConfig{
+		"name": &graphql.InputObjectFieldConfig{
 			Type:        graphql.NewNonNull(graphql.String),
-			Description: "Path of the file.",
+			Description: "Name of the file.",
 		},
 		"content": &graphql.InputObjectFieldConfig{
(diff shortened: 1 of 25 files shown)


pkg/portal/graphql/app.go
```diff
@@ -118,10 +119,16 @@ var _ = registerMutationField(
 		},
 		Resolve: func(p graphql.ResolveParams) (interface{}, error) {
 			input := p.Args["input"].(map[string]interface{})
-			appID := input["appID"].(string)
+			appNodeID := input["appID"].(string)
```
louischan-oursky, line 122: I was also trapped before.

pkg/portal/model/app.go
```diff
@@ -27,6 +27,6 @@ func (a *App) LoadFile(path string) ([]byte, error) {
 }
 
 type AppConfigFile struct {
-	Path    string
+	Name    string
```
louischan-oursky, line 30: If we accept file name, we cannot express `/templates/zh/forgot_password_email.html`.

Maybe we need to adopt [an encoding scheme](https://github.com/facebook/flow/blob/v0.133.0/src/hack_forked/utils/sys/path.ml#L66) to encode the key so that the encoded is valid k8s configmap/secret key.

louischan-oursky:
Are we expecting the client to encode the content as YAML or JSON?

pkg/portal/model/app.go
kiootic: It'll be hard to migrate later, so I'll add encoding here now.

kiootic:
> Are we expecting the client to encode the content as YAML or JSON?
> During my test, if I give it a JSON, it works (because YAML is superset of JSON) but the file now looks like JSON forever :(

I suppose portal frontend should encode the config as YAML before passing it to backend?

PR #269, opened 2020-09-08
Update /settings/identity design 

### LINKED ISSUE #267: [UI] Update identity page design

louischan-oursky:
depend on #266
fix #267

### DIFF
--- a/static/authui/css/main.css
+++ b/static/authui/css/main.css
@@ -527,15 +527,28 @@ a.btn {
   margin: 0 6px;
 }
 
-.content-row {
-  margin: 2px;
-  padding: 2px;
+.logout-row {
+  margin: 14px 18px;
 }
 
-.content-row-layout-2x2 {
+.logout-section {
   display: grid;
-  grid-template-columns: 1fr auto;
-  grid-template-rows: auto auto;
+  grid: "title  action" auto
+      /   1fr    auto;
+}
+
+.logout-section > * {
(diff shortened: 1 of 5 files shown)


kiootic:
@louischan-oursky conflicted please rebase.

louischan-oursky:
Updated.

static/authui/css/main.css
```diff
@@ -105,7 +105,7 @@ h1, h2, h3, h4, h5, h6 {
   border: none;
   border-radius: 0;
   /* NOTE(1) */
-  font-size: 16px;
+  font-size: 1rem;
```
kiootic, line 108: We didn't set font size at root element right? Since we don't import normalizer css, let's set it to 16px explicitly at root element.

kiootic (commented):
- It seems spacing between the identity title and description is unnatural, due to size of icon. I suppose we can look into it later in #277 ?
- Are we going to follow Fluent UI design and import their web font? The font weight looks quite a bit off at the moment.

static/authui/css/main.css
louischan-oursky: I believe the default font size should be most accessible to the user. Setting it to 16px is not accessible. The intention here is to use whatever default font size the user agent has set for the user.

kiootic (approved):
Let me merge it for now and discuss remaining issues later.

PR #274, opened 2020-09-09
[Portal] User detail page

### LINKED ISSUE #272: [UI] Create user detail screen
- [x] Create route for user detail page
- [x] Create link in user list
- [x] Add graphQL query for user detail
- [x] User detail page summary
- [x] Create tabs using `Pivot` (account securities, connected identities, session)

### Account securities
- [x] research for fluent UI component for making list
- [x] get data from response `User -> authenticator`
- [x] layout the tab item (button -> primary button, all buttton dummy)

### Connected identities
- [x] get data from response `User -> identities`
- [x] Layout page using same list component as previous

### Session
- [x] no data available yet, create mock data for layout
- [x] layout the list

### Control bar
- [x] use fluent UI command bar component
- [x] remove, login as user, invalidate all session, disable
- [x] find icon in fluent UI
- [x] dummy button, mutation not available yet

axolotl0212:
fix #272


portal/src/util/types.ts
```diff
@@ -13,3 +14,12 @@ export function isPortalApiApp(value: any): value is PortalAPIApp {
   }
   return value.__typename === "App";
 }
+
+export function isUserDetails(
```
louischan-oursky, line 18: No need to have this type guard. The generated type should be correct already. Please also remove `isPortalApiApp` above.

portal/src/util/formatDatetime.ts
```diff
@@ -13,7 +13,7 @@ export function formatDatetime(
     datetime = DateTime.fromISO(date);
   }
 
-  if (datetime == null) {
+  if (!datetime?.isValid) {
```
louischan-oursky, line 16: Can you clarify in what circumstances you receive an invalid date string from the server? We should fix the server instead of working it around here.

It is better to remove this change.

portal/src/graphql/adminapi/UserDetailSummary.module.scss
```diff
@@ -0,0 +1,30 @@
...
+.userInfo {
+  margin-top: 5px;
+  margin-bottom: 20px;
+}
+
+.userInfoRow {
+  font-size: 13px;
+  font-weight: 300;
+  margin-bottom: 5px;
+}
```
louischan-oursky, line 30: We can use CSS grid. Using grid can greatly simplify the HTML markup. Technically we do not need to introduce div for layout purpose.

Let's convert the new layout in this PR to use CSS grid.

portal/src/graphql/adminapi/UserDetailSummary.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const formatedSignedUp = React.useMemo(() => {
+    return formatDatetime(locale, createdAtISO);
+  }, [locale, createdAtISO]);
+  const formatedLastLogin = React.useMemo(() => {
+    return formatDatetime(locale, lastLoginAtISO);
+  }, [locale, lastLoginAtISO]);
+
+  return (
+    <div className={cn(styles.root, className)}>
+      <img src={proPicPlaceholder} className={styles.profilePic} />
```
louischan-oursky, line 35: Use [Persona](https://developer.microsoft.com/en-us/fluentui#/controls/web/persona) for placeholder. Remove the dummy image.

portal/src/ListCellLayout.module.scss
```diff
@@ -0,0 +1,8 @@
+.cellContainer {
+  min-height: 50px;
+  border-bottom: 1px solid #c4c4c4;
+  padding-top: 20px;
+  padding-bottom: 10px;
```
louischan-oursky, line 5: I prefer we always use the shorthand property, e.g. `margin: `, `padding: `.

portal/src/graphql/adminapi/UserDetailsAccountSecurity.tsx
```diff
@@ -1,9 +1,227 @@
...
+  descriptionClassName?: string;
+  onDetailButtonClick?: () => void;
+}
+
+// authenticator type recognized by portal
+enum AuthenticatorType {
+  Password = "password",
+  OneTimePassword = "oob_otp",
+  TimeBasedOneTimePassword = "totp",
+}
```
louischan-oursky, line 36: Prefer union type.

portal/src/graphql/adminapi/UserDetailsAccountSecurity.tsx
```diff
@@ -1,9 +1,227 @@
...
+
+enum AuthenticatorKind {
+  Primary = "primary",
+  Secondary = "secondary",
+}
```
louischan-oursky, line 41: Prefer union type.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.module.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.removeButton {
+  margin-right: 3px;
+  background-color: #ff0000;
+  border: 1px solid #990000;
+  &:hover {
+    background-color: #e80000;
+    border: 1px solid #990000;
+  }
```
louischan-oursky, line 41: Can we introduce our Destructive button with a custom theme?

In `src/theme.tsx`, we have two themes. Let's add a new theme where the primary color is red.

And then we can add a component DestructiveButton which is basically a PrimaryButton with theme set to the new theme

```typescript
function DestructiveButton(props) {
  return <PrimaryButton theme={destructiveTheme} {...props} />
}
```

The theme generator is at https://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/master/theming-designer/index.html

portal/src/util/formatDatetime.ts
axolotl0212: server should return valid datetime string, will remove this

portal/src/util/types.ts
axolotl0212: without typeguard, typescript cannot decide the type of the node. In authentication screen query, there is two possible type, `__typename = "User" | "App"`, As the node implements both User and App. If we need to cast it to the correct type we will need `as any as ...` ? @louischan-oursky

portal/src/graphql/adminapi/UserDetailSummary.module.scss
axolotl0212: not sure how to use CSS grid to simplify the layout. If we put all user info into a grid, then we still need to group the rows to two group by div? Also we need to give each row a class name and assign one row in the grid by `grid-row`?

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.module.scss
axolotl0212: Nice, didn't realize that there is a theme generator

portal/src/graphql/adminapi/UserDetailsAccountSecurity.tsx
axolotl0212: AuthenticatorType is just a enum of possible `type` in authenticator while `type` has type `string`

axolotl0212:
@louischan-oursky addressed some comments, not sure how to remove type guard and use css grid

portal/src/util/types.ts
louischan-oursky: I see. Let's move this function to the file where the query is defined then. It is weird that `/src/util/types.ts` imports a generated file of a specific component.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const onRenderIdentityCell = useCallback(
+    (item?: IdentityListItem, _index?: number): React.ReactNode => {
+      return <IdentityListCell item={item} />;
+    },
+    []
+  );
+
+  return (
+    <div className={styles.root}>
+      <h1 className={styles.header}>
```
kiootic, line 97: - `h1` is already used in the breadcrumb as the page title, it may be more appropriate to use `h2` here as section title.
- I'd suggest using `Text` component to ensure more consistent styles.
- Here we're using default styling of user agent, which bolds the header. Fluent UI recommends using [semi-bold for headers](https://developer.microsoft.com/en-us/fluentui#/styles/web/typography).

portal/src/graphql/adminapi/UsersList.tsx
```diff
@@ -41,6 +44,24 @@ interface PlainUsersListProps {
...
+interface UserListItem {
+  id: string;
+  createdAt: string | null;
+  username: string;
+  phone: string;
+  email: string;
+  lastLoginAt: string | null;
+}
+
+const isUserListItem = (value: any): value is UserListItem => {
```
kiootic, line 56: Avoid using any here, accept unknown instead.

portal/src/graphql/adminapi/UserDetailsScreen.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const { loading, error, data, refetch } = useQuery<
+    UserDetailsScreenQuery,
+    UserDetailsScreenQueryVariables
+  >(query, {
+    variables: {
+      userID,
+    },
+  });
+
+  const navBreadcrumItems = React.useMemo(() => {
```
kiootic, line 139: Typo

portal/src/graphql/adminapi/UserDetailsScreen.tsx
```diff
@@ -0,0 +1,198 @@
...
+import { extractUserInfoFromIdentities } from "../../util/user";
+
+import styles from "./UserDetailsScreen.module.scss";
+
+interface UserDetailsProps {
+  data: UserDetailsScreenQuery_node_User | null;
+  loading: boolean;
+}
+
+function isUserDetails(value: any): value is UserDetailsScreenQuery_node_User {
```
kiootic, line 35: I don't think this function is needed? We can just check `node?.__typename === "User"` at call site?

kiootic (commented):
- The breadcrumb segment 'Users' in user details screen reload the page, rather than navigating within the router. Can you help fix it to use react-router's Link component instead of using raw `a` element?
- The user info rows feels too light, can you double check its font weight?
- Also, some texts are rendered using `.SF NS` in Mac Chrome, while components from Fluent UI library are rendered using `Segoe UI` web font. We should unify on the font usage.

portal/src/graphql/adminapi/UserDetailsScreen.tsx
axolotl0212: You are right, it successfully infer the type after this condition

axolotl0212:
@kiootic @louischan-oursky updated

portal/src/graphql/portal/AuthenticationConfigurationScreen.tsx
```diff
@@ -28,6 +28,16 @@ const authenticationScreenQuery = gql`
   }
 `;
 
+function isPortalApiApp(value: any): value is PortalAPIApp {
```
kiootic, line 31: I suppose this one can be replaced too?

kiootic (commented):
I'm still seeing inconsistent fonts in headers and Authentication screen, let's open an issue to fix that later.

portal/src/graphql/portal/AuthenticationConfigurationScreen.tsx
axolotl0212: This one cast the value PortalAPIApp, the condition only let typescript infer the type `Record<string, unknown>` in this case, and the overlap is not enough for casting directly. So there is 2 options

1. typeguard
2. `as any as ...`

and I found 1 looks nicer @kiootic

kiootic: It seems the `Record<string, unknown>` you got comes from the `GQL_AppConfig` scalar type. We should update it in `graphql.d.ts` to use our expected app config type `PortalAPIAppConfig`, i.e.: `declare type GQL_AppConfig = PortalAPIAppConfig;`.

axolotl0212: it works, thanks

louischan-oursky:
@axolotl0212 @kiootic I added [1 commit](https://github.com/authgear/authgear-server/pull/274/commits/cd7cf168670b081462b2444e58d8f61edb0c23cb) to make import explicit in .d.ts

PR #283, opened 2020-09-10
[Portal] Anonymous user screen

### LINKED ISSUE #280: [Portal] Anonymous user screen
- [x] review config - setting mapping
- [x] layout the page
  - [x] missing toggle button in `Enable anonymous user` design
- [x] implement function to map config <-> state
- [x] check accessibility warning

axolotl0212:
Closes #280

## Screenshot

Screenshot 2020-09-10 at 11 45 51 PM


axolotl0212:
rebased

axolotl0212:
rebased

kiootic (commented):
I'm seeing this behavior:
1. When I first enter the page, the toggle is disabled.
2. If I go to other page then come back, correct data is shown
3. If I just directly toggle it on, the dropdown is empty

My config is like this:
```yaml
authentication:
  identities: [login_id, anonymous, oauth]
identity:
  on_conflict:
    promotion: login
```
Can you try to see if you can reproduce it?

axolotl0212:
Will have a look after handling save config

axolotl0212:
This is due to setState is called before appConfig is fetched in the case of directly going into page by url

axolotl0212:
rebased, and push one more commit to fix the issue

portal/src/graphql/portal/AnonymousUsersConfigurationScreen.tsx
```diff
@@ -1,29 +1,122 @@
...
+  appConfig: PortalAPIAppConfig | null
+): AnonymousUsersConfigurationScreenState {
+  if (appConfig == null) {
+    return {
+      enabled: false,
+    };
+  }
+  const promotionConflictBehaviour = appConfig.identity?.on_conflict?.promotion;
+  return {
+    enabled: promotionConflictBehaviour != null,
```
louischan-oursky, line 46: This is incorrect. The correct way to determine whether anonymous users is enabled is to see if `appConfig.authentication.identities` contains `anonymous`.

portal/src/graphql/portal/AnonymousUsersConfigurationScreen.tsx
```diff
@@ -1,29 +1,122 @@
...
+    promotionConflictBehaviour,
+  };
+}
+
+function constructNewAppConfigFromState(
```
louischan-oursky, line 51: Similarly, if the toggle is enabled, ensure `appConfig.authentication.identities` contains `anonymous`. Otherwise `appConfig.authentication.identities` must NOT contain `anonymous`.

If the toggle is enabled, the on conflict option is saved. Otherwise, the original value should be used (i.e. keep the original value, do not touch it / delete it)

portal/src/graphql/portal/AnonymousUsersConfigurationScreen.tsx
```diff
@@ -1,29 +1,122 @@
...
+    [state]
+  );
+
+  const onConflictOptionChanged = useCallback(
+    (option: IDropdownOption) => {
+      setState({
+        ...state,
+        promotionConflictBehaviour: isPromotionConflictBehaviour(option.key)
+          ? option.key
+          : undefined,
```
louischan-oursky, line 117: Or more correctly, if the new value is invalid, do not call setState at all.

portal/src/graphql/portal/AnonymousUsersConfigurationScreen.tsx
```diff
@@ -84,6 +90,10 @@ const AnonymousUserConfigurationScreen: React.FC = function AnonymousUserConfigu
 
   const [state, setState] = useState(constructStateFromAppConfig(appConfig));
 
+  useEffect(() => {
+    setState(constructStateFromAppConfig(appConfig));
+  }, [appConfig]);
```
louischan-oursky, line 95: Can you explain more on this fix? This fix looks weird to me. In what case `appConfig` is changed?

axolotl0212: Before appConfig is loaded (receive response in query), appConfig is null, so it changes from null to the initial appConfig in response.

louischan-oursky: The screen code now looks like

```typescript
function MyScreen() {
  const { loading, error, refetch, data } = useAppConfigQuery(appID);
  if (loading) { return <Loading /> }
  if (error) { return <Error error={error} retry={refetch} /> }
  // use data, even if data may be null.
}
```

I suggest we change ALL screens (not this screen only) to the following structure.

```typescript
function MyScreen() {
  const { loading, error, refetch, data } = useAppConfigQuery(appID);
  if (loading) { return <Loading /> }
  if (error) { return <Error error={error} retry={refetch} /> }
  // data should not be null now. MyScreenContent does not need to use useEffect to refetch.
  return <MyScreenContent data={data} />
}
```

axolotl0212:
@louischan-oursky updated

louischan-oursky:
Thanks!

PR #287, opened 2020-09-11
Improve HTTP response content in case of panic

### LINKED ISSUE #199: Improve web UI error UX
We should show a friendly UI when unexpected (or expected) errors occured.

louischan-oursky:
fix #199


kiootic (commented):
We'll need to also handle DB transaction conflict error, as it is in original recovery middleware.

kiootic (approved):
Looks good to me! I suppose we'll still need to get a real design for the error page from designer later.

PR #289, opened 2020-09-11
Make claims verifiable

kiootic:
@chpapa  Updated the verification spec according to the parts we all agreed for now.

@louischan-oursky For now authenticators does not participate in verification process, since that would create a circular dependency between identity, authenticator, and verification in orphan removal process. I suppose we'll need some way to workaround it.

ref #198


pkg/lib/interaction/context.go
```diff
@@ -32,7 +32,7 @@ type IdentityService interface {
 	New(userID string, spec *identity.Spec) (*identity.Info, error)
 	UpdateWithSpec(is *identity.Info, spec *identity.Spec) (*identity.Info, error)
 	Create(is *identity.Info) error
-	Update(is *identity.Info) error
+	Update(before, afterr *identity.Info) error
```
louischan-oursky, line 35: typo

pkg/lib/authn/identity/service/service.go
```diff
@@ -65,6 +65,10 @@ type AuthenticatorService interface {
 	Delete(authenticatorInfo *authenticator.Info) error
 }
 
+type VerificationService interface {
+	RemoveOrphanedClaims(identities []*identity.Info) error
```
louischan-oursky, line 69: Should we take secondary OOB authenticator into account? They are not associated with any identity.

docs/specs/verification.md
```diff
@@ -7,29 +7,24 @@
...
 
 ## Definitions
 
 A user is verified if the user fulfill the condition specified by criteria.
 
-An identity is verifiable if:
-- it is an SSO identity; or
-- it is a login ID identity with a configured verifiable login ID key.
+An identity is verifiable is any of its claims is verifiable. An identity is
+verified is any of its claims is verified.
```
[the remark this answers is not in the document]
kiootic: I suppose there isn't a need for the concept of verified/verifiable identity, I'll remove that for now.

pkg/lib/authn/identity/service/service.go
kiootic: We should. However, as stated in the top post, it'll create a circular dependency between verification, identity, and authenticator, so I'm omitting it for now. There should be a TODO note in `verification/service.go`.

PR #293, opened 2020-09-11
[Portal] Integrate save config mutation

### LINKED ISSUE #286: [Portal] Integrate save config mutation
- [x] handle on save on each configuration screen
  - [x] check which field is modified (different from original)
  - [x] update raw config
  - [x] explore the possibility of abstracting this logic
- [x] convert the modified raw config to yaml
  - [x] install https://www.npmjs.com/package/js-yaml
- [x] integrate the mutation, check whether API works
- [x] check whether data is refetched after mutation

axolotl0212:
Closes #286


axolotl0212:
Issues encountered:

- root -> identity -> login_id -> keys -> verification field not recognized
  - schema not up to date?

- authenticator setting primary / secondary authenticator not allow to contain same key
  - eg. both primary / secondary authenticator contains key "password" leads to duplication error
  - if this is expected, need to update UI ?

portal/src/graphql/portal/AuthenticationLoginIDSettings.tsx
```diff
@@ -130,10 +131,11 @@ function setLoginIdKeyEnabled(loginIdKey: LoginIDKeyConfig, enabled: boolean) {
 }
 
 function constructAppConfigFromState(
-  appConfig: PortalAPIAppConfig,
+  rawAppConfig: PortalAPIAppConfig,
+  effectiveAppConfig: PortalAPIAppConfig,
```
louischan-oursky: `effectiveAppConfig` seems unused in this function. Remove it if possible.

portal/src/util/misc.ts
```diff
@@ -0,0 +1,17 @@
+export function isArrayEqualInOrder(arr1: unknown[], arr2: unknown[]): boolean {
```
louischan-oursky, line 1: Better to add a type argument T to this function.

```typescript
function isArrayEqualInOrder<T>(arr1: T[], arr2: T[]): boolean
```

So we will not compare apples to oranges.

portal/src/util/misc.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (arr1.length !== arr2.length) {
+    return false;
+  }
+  return arr1.every((elem, index) => elem === arr2[index]);
+}
+
+export function setFieldIfChanged(
```
louischan-oursky, line 8: A better typing for this function.

```typescript
function setFieldIfChanged<K extends string>(record: Record<K, unknown>, key: K, initialValue: unknown, value: unknown): void
```

So we will not make typo mistake of the keys of the records.

portal/src/graphql/portal/mutations/updateAppConfigMutation.ts
```diff
@@ -0,0 +1,48 @@
...
+      rawAppConfig
+      effectiveAppConfig
+    }
+  }
+`;
+
+export function useUpdateAppConfigMutation(
+  appID: string
+): {
+  updateAppConfig: (appConfigYaml: string) => Promise<PortalAPIApp | null>;
```
louischan-oursky, line 29: I guess this hook should do more and encode the given config to YAML string. It is more easy to use.

portal/src/graphql/portal/AuthenticationAuthenticatorSettings.tsx
```diff
@@ -20,7 +20,7 @@ import {
   SecondaryAuthenticatorType,
   PortalAPIApp,
 } from "../../types";
-import { isArrayEqualInOrder } from "../../util/misc";
+import { isArrayEqualInOrder, clearEmptyObject } from "../../util/misc";
```
louischan-oursky, line 24: Can you explain why do we need to do this? Is it the rawConfig returned the server containing empty object and undefined?

portal/src/LoadingModal.tsx
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  return (
+    <div className={styles.root}>
+      <ShowLoading />
+    </div>
+  );
+};
+
+export default LoadingModal;
```
louischan-oursky, line 24: Can we simply show a [Spinner](https://developer.microsoft.com/en-us/fluentui#/controls/web/spinner) next to the save button with a label telling the user that we are saving? When we are saving, set pointer events to none to prevent user interaction.

portal/src/graphql/portal/AuthenticationAuthenticatorSettings.tsx
```diff
@@ -237,7 +238,7 @@ const AuthenticationAuthenticatorSettings: React.FC<Props> = function Authentica
     [secondaryAuthenticatorState]
   );
 
-  const onSaveButtonClicked = React.useCallback(() => {
+  const onSaveButtonClicked = React.useCallback(async () => {
```
louischan-oursky, line 241: The function passed to useCallback should not be async.

portal/src/graphql/portal/AuthenticationAuthenticatorSettings.tsx
```diff
@@ -277,10 +278,13 @@ const AuthenticationAuthenticatorSettings: React.FC<Props> = function Authentica
       clearEmptyObject(draftConfig);
     });
 
-    // TODO: call mutation to save config
+    const newAppConfigYaml = yaml.safeDump(newAppConfig);
+
+    await updateAppConfig(newAppConfigYaml);
```
louischan-oursky, line 283: Add a TODO here to handle error. I think we are going to handle save error in another issue.

portal/src/hook/misc.ts
```diff
@@ -0,0 +1,34 @@
+import { useState, useCallback } from "react";
+
+export function useSimpleRPC<T>(
+  f: (...args: any[]) => Promise<T>
+): {
+  loading: boolean;
+  error: Error | null;
+  rpc: (...args: unknown[]) => Promise<T | null>;
+} {
```
louischan-oursky, line 9: We should have a better way to type this function.

```typescript
declare function useSimpleRPC<Arg extends any[], Ret>(f: (...args: Arg) => Promise<Ret>): {
    loading: boolean;
    error: unknown;
    g: (...args: Arg) => Promise<Ret>
}
```

Note that `error` should be `unknown`.

louischan-oursky (commented):
> root -> identity -> login_id -> keys -> verification field not recognized

The up-to-date schema can be found in the go source code :( The one in the documentation website is seldom updated.

> authenticator setting primary / secondary authenticator not allow to contain same key

It is intentional. Currently they cannot appear twice in the two arrays.

portal/src/graphql/portal/AuthenticationAuthenticatorSettings.tsx
axolotl0212: When constructing the new config, empty object may be generated.

1. New object may be added for optional field. However the added object is not used (eg. the whole section disabled, config not updated)
2. properties deleted, makes the parent object empty

So to avoid having to much branching (if... if ... create object), I created all the object needed, mutate the object using the screen state, the simplify the final config by removing undefined and empty object

portal/src/graphql/portal/AuthenticationLoginIDSettings.tsx
axolotl0212: removed in later commit

portal/src/LoadingModal.module.scss
```diff
@@ -0,0 +1,9 @@
+.root {
```
kiootic, line 1: Just FYI, implementing a modal is not that easy. For example, 'position: absolute' may fail when body can scroll. If we want to use modal, we can use the [`Modal` component](https://developer.microsoft.com/en-us/fluentui#/controls/web/modal) provided by Fluent UI, which *should* be more well-tested than our hand-rolled one.

portal/src/util/misc.ts
```diff
@@ -15,3 +15,24 @@ export function setFieldIfChanged(
     map[field] = value;
   }
 }
+
+export function clearEmptyObject(obj: Record<string, any>): void {
+  for (const key in obj) {
```
louischan-oursky, line 20: When using `for ... in`, we should check call to check if the enumerated key is own properties.

```
for (const key in someObject) {
  // Use this form to call this function to ensure we are calling the function we intend to.
  // The reason is that hasOwnProperty may be overridden by some malicious code. 
  if (!Object.prototype.hasOwnProperty.call(someObject, key) { continue; }
}
```

portal/src/util/misc.ts
```diff
@@ -15,3 +15,24 @@ export function setFieldIfChanged(
...
+      // must call function on child first to handle the case where obj[key]
+      // becomes empty after removing it's empty children object
+      clearEmptyObject(obj[key]);
+      // array has type "object", allow []
+      if (Object.keys(obj[key]).length === 0 && !Array.isArray(obj[key])) {
+        delete obj[key];
+      }
+    } else {
+      delete obj[key];
+    }
```
louischan-oursky, line 36: The code looks cleaner if we invert the condition and use continue.

```typescript
for (const key of obj) {
  if (!Object.prototype.hasOwnProperty.call(someObject, key) { continue; }
  if (obj[key] === undefined) {
    delete obj[key];
    continue;
  }
  // more logic here and more continue statement.
}
```

portal/src/util/misc.ts
```diff
@@ -15,3 +15,24 @@ export function setFieldIfChanged(
     map[field] = value;
   }
 }
+
+export function clearEmptyObject(obj: Record<string, any>): void {
```
louischan-oursky, line 19: Can we add unit tests for this function? This should be relatively easier to test.

portal/src/graphql/portal/mutations/updateAppConfigMutation.ts
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import { useMutation, gql } from "@apollo/client";
+
+import { client } from "../../portal/apollo";
+import { PortalAPIApp } from "../../../types";
+import { UpdateAppConfigMutation } from "./__generated__/UpdateAppConfigMutation";
+
+// relative to project root
+const UPDATE_FILE_PATH = "./authgear.yaml";
```
kiootic, line 9: Let's name it more accurately (e.g. `APP_CONFIG_PATH`), since later we would also need to edit `authgear.secrets.yaml`.

louischan-oursky:
@axolotl0212 conflicted.

axolotl0212:
@louischan-oursky resolved conflict, shall we add anonymous screen mutation to this PR as well?

louischan-oursky:
`LoadingModal` is unused now. Can we remove it?

> shall we add anonymous screen mutation to this PR as well?

Let's do it in another PR.

axolotl0212:
> LoadingModal is unused now. Can we remove it?

Sure

axolotl0212:
pushed 1 more commit to remove the loading modal component

PR #316, opened 2020-09-15
Add security section in settings screen

### LINKED ISSUE #310: [UI] Settings Screen - Security section
- [x] Show Primary Password
- [x] Show Primary OOB OTP
- [x] Show Secondary TOTP
- [x] Show Secondary OOB OTP
- [x] Show Secondary Password
- [x] Show Recovery code
- [x] Show revoke all device tokens button

louischan-oursky:
fix #310 

Most buttons are not functional. They will be implemented by later PRs.

We also discovered a problem with our localization system. We cannot localize date time (yet). ICU MessageFormat has support for datetime formatting. But our MessageFormat does not support it (yet).

### DIFF
--- a/pkg/auth/handler/webapp/settings.go
+++ b/pkg/auth/handler/webapp/settings.go
@@ -33,9 +33,9 @@ func ConfigureSettingsRoute(route httproute.Route) httproute.Route {
 type SettingsViewModel struct {
 	Authenticators           []*authenticator.Info
 	MFAActivated             bool
-	SecondaryTOTPEnabled     bool
-	SecondaryOOBOTPEnabled   bool
-	SecondaryPasswordEnabled bool
+	SecondaryTOTPAllowed     bool
+	SecondaryOOBOTPAllowed   bool
+	SecondaryPasswordAllowed bool
 }
 
 type SettingsAuthenticatorService interface {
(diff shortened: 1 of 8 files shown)


pkg/auth/handler/webapp/settings.go
```diff
@@ -26,19 +31,72 @@ func ConfigureSettingsRoute(route httproute.Route) httproute.Route {
 		WithPathPattern("/settings")
 }
 
+type SettingsViewModel struct {
+	Authenticators           []*authenticator.Info
+	RecoveryCodes            []*mfa.RecoveryCode
```
kiootic, line 36: I don't think we should load all recovery codes upon entering settings page: viewing recovery code logs a security event (which GitHub does), so it should be loaded only on the page specific to viewing recovery code.

pkg/auth/handler/webapp/settings.go
```diff
@@ -26,19 +31,72 @@ func ConfigureSettingsRoute(route httproute.Route) httproute.Route {
...
+	SecondaryTOTPEnabled     bool
```
kiootic, line 37: It seems there are used to determine whether the relevant sections should be shown. However, if user has a secondary authenticator not enabled in config (i.e. config changed), should we still show that section?

templates/auth_ui_settings.html
```diff
@@ -12,13 +12,201 @@
...
+    </p>
+    <a class="action" href="/settings/identity">
+      {{ template "details-button-label" }}
+    </a>
+  </section>
+  {{ end}}{{ end }}
+  {{ end }}
+
+  <!-- MFA title -->
+  <section class="settings-row settings-page-section-with-title">
```
kiootic, line 77: It seems it still show when no secondary authenticator is configured. We may want to hide it if the section is empty?

templates/auth_ui_settings.html
```diff
@@ -12,13 +12,201 @@
...
+      {{ template "settings-page-recovery-code-description" }}
+    </p>
+    <a class="action" href="#">
+      {{ template "show-recovery-code-button-label" }}
+    </a>
+  </section>
+  {{ end }}
+
+  <!-- Revoke all device tokens -->
+  {{ if $.RecoveryCodes }}
```
kiootic, line 192: It seems we're assuming 'have recovery codes' = 'MFA enabled'. I suppose we should lift the 'MFA enabled' flag to a field in view model to be more explicit.

louischan-oursky:
@kiootic updated.

kiootic (approved):

PR #317, opened 2020-09-15
Handle save config api validation error

### LINKED ISSUE #296: [Portal] Display validation error upon save config
Configurations from portal user could be contradicted with each other and therefore the save config API would return validation error to frontend for displaying on UI

---

- [x] Remove `useSimpleRPC`
- [x] Introduce `useAppConfigMutation`, similar to `useAppConfigQuery`
- [x] Study the shape of the error
- [x] Introduce a mechanism to show error

```typescript
type Violation = RequiredViolation | TypeViolation;

interface RequiredViolation {
	type: "required";
	location: string;
	missing: string[];
	// ...
}

type Selector = (error: unknown) => Violation[];
type Formatter = (violation: Violation[]) => string | undefined;

const errorMsg = extractError(mutationError, fieldSpecificSelector);

```

axolotl0212:
Closes #296 

will integrate the functions in single-sign-on (SSO) page PR


axolotl0212:
Screenshot 2020-09-15 at 7 28 50 PM

portal/src/util/error.ts
```diff
@@ -0,0 +1,175 @@
...
+function isViolationKind(value?: string): value is ViolationKind {
+  return value != null && violationKinds.includes(value);
+}
+
+// expected data shape of error extension from backend
+interface ErrorCauseDetails {
+  actual?: string[];
+  expected?: string[];
+  missing?: string[];
+}
```
louischan-oursky, line 27: The shape of extensions is defined in `pkg/util/graphqlutil/api_error.go`. All Authgear errors are encoded as API Error.

API Error is a JSON object roughly with the following shape

```typescript
interface APIError {
  errorName: string;
  reason: string;
  // The shape of info depends on reason.
  info: unknown;
}
```

For now, we only need to deal with the reason `ValidationFailed`.

When the reason is `ValidationFailed`, the shape of info is

```typescript
interface ValidationErrorInfo {
  causes: Cause[];
}

interface Cause {
  location: string;
  kind: string;
  // The shape of details depends on kind.
  details: unknown;
}
```

Can we update the type definitions so that the definitions can be extended to cover more `reason` and validation violation `kind` in the future?

portal/src/util/error.ts
```diff
@@ -0,0 +1,175 @@
...
+
+interface GraphQLErrorExtensions {
+  errorName: string;
+  info: ErrorExtensionInfo;
+  reason: string;
+}
+
+type ViolationSelector = (violation: Violation) => boolean;
+type ViolationSelectors<Key extends string> = Record<Key, ViolationSelector>;
```
louischan-oursky, line 46: It is better to split this file into two.

The first file should be named `error.ts`, for parsing error into recognized shapes.

The second one should be named `validation.ts`, the validation error related type definitions and functions should go here.

portal/src/util/error.ts
```diff
@@ -0,0 +1,175 @@
...
+  ) => string | undefined = defaultFormatErrorMessageList
+): string | undefined {
+  return formatErrorMessageList(
+    violations
+      .map((violation) =>
+        violationFormatter(fieldNameId, violation, renderToString)
+      )
+      .filter(nonNullable)
+  );
+}
```
louischan-oursky, line 175: Can you write tests to demonstrate the usage? With tests it is more much easier to see how everything interacts. The tests do not need to rely on React.

axolotl0212:
@louischan-oursky updated with test case.

PR #326, opened 2020-09-16
Change primary password

### LINKED ISSUE #319: [UI] Change primary password
- [x] Add a new interaction intent to change primary password
  - [x] The same intent should be reusable for changing secondary password
- [x] Unify `create_password`, `reset_password` and `change_password` to use identical HTML template.
- [x] Update the HTML template to match latest design.

louischan-oursky:
fix #319


pkg/util/password/confirm.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"crypto/subtle"
+
+	"github.com/authgear/authgear-server/pkg/api/apierrors"
+)
+
+var ErrNewPasswordTypo = apierrors.Invalid.WithReason("NewPasswordTypo").New("new password typo")
+
+func ConfirmPassword(newPassword, confirmPassword string) error {
+	typo := subtle.ConstantTimeCompare([]byte(newPassword), []byte(confirmPassword)) == 0
```
kiootic, line 12: We don't need constant time comparison here, since both sides are supplied by the same requestor.

templates/auth_ui_change_password.html
```diff
@@ -0,0 +1,57 @@
...
+	id="password"
+	data-password-policy-password=""
+	class="input text-input primary-txt"
+	type="password"
+	autocomplete="new-password"
+	name="x_password"
+	placeholder="{{ template "new-password-placeholder" }}"
+>
+
+<input
```
kiootic, line 30: We should also have `autocomplete="new-password"` for password confirmation field.

ref: https://www.chromium.org/developers/design-documents/form-styles-that-chromium-understands

pkg/lib/interaction/nodes/change_password.go
```diff
@@ -0,0 +1,128 @@
...
+		return []interaction.Edge{
+			&EdgeDoUpdateAuthenticator{
+				Stage:                     n.Stage,
+				AuthenticatorBeforeUpdate: n.OldAuthenticator,
+				AuthenticatorAfterUpdate:  n.NewAuthenticator,
+			},
+		}, nil
+	}
+	// Password was not changed, ends the interaction
+	return nil, nil
```
kiootic, line 127: The intent should control the interaction flow, but this makes the node ends the interaction. Perhaps move this logic to `DeriveEdgesForNode`?

templates/auth_ui_change_password.html
louischan-oursky: If it is auto-completed, that would defeat the purpose of checking typo.

kiootic: Since the autocomplete type is `new-password`, I suppose it would be auto-completed only if the new password is auto-generated?

louischan-oursky: OK. Let's trust chrome.

louischan-oursky: My original concern is if the new password is created by the user, and the confirm is auto-completed, typo cannot be detected.

louischan-oursky:
Updated.

kiootic (approved):

PR #327, opened 2020-09-16
Setup Awesome font 5 using CDN

### LINKED ISSUE #302: [Portal] Install Font Awesome 5
Some parts of the portal is using Font Awesome 5 Brands.

I have done this for the web UI in https://github.com/authgear/authgear-server/pull/290

Basically we should use the same CDN kit.

As far as I know, the icon of the connected identities are Font Awesome 5 Brands.

axolotl0212:
closes #302


axolotl0212:
rebased

PR #328, opened 2020-09-16
Handle block navigation and disable save button when form modified

### LINKED ISSUE #306: [Portal] Warn before leaving if the form is dirty
- [x] Investigate how we can achieve this with `react-router-dom@6`
- [x] Apply this to every form
- [x] Use [Dialog](https://developer.microsoft.com/en-us/fluentui#/controls/web/dialog) for confirmation

### LINKED ISSUE #307: [Portal] Disable save button if the form is not dirtry

axolotl0212:
Closes #306 
Closes #307 

Issue discovered: Changing tab (without navigation) causes form state to be discarded as component (`PivotItem`) is dismounted


axolotl0212:
rebased

portal/src/graphql/portal/AuthenticationAuthenticatorSettings.tsx
```diff
@@ -188,18 +192,29 @@ const AuthenticationAuthenticatorSettings: React.FC<Props> = function Authentica
...
+  const [isFormModified, setIsFormModified] = useState(false);
+
+  useEffect(() => {
+    const screenState: AuthenticationAuthenticatorScreenState = {
+      primaryAuthenticators: primaryAuthenticatorState,
+      secondaryAuthenticators: secondaryAuthenticatorState,
+    };
+    const equalInitial = deepEqual(initialState, screenState, { strict: true });
+    setIsFormModified(!equalInitial);
+  }, [initialState, primaryAuthenticatorState, secondaryAuthenticatorState]);
```
louischan-oursky, line 217: It seems that the state here is a bit complicated. Ideally we should have two copy of state. One for storing the initial state, one for storing the form state. The dirty flag should be derived instead of being a state.

```typescript
const initialState = useMemo(() => {
  return constructListData(effectiveAppConfig);
}, [effectiveAppConfig]);

const [formState, setFormState] = useState(initialState);

const isDirty = useMemo(() => {
  return !deepEqual(initialState, formState, { strict: true });
}, [initialState, formState]);
```

louischan-oursky:
How do we going to solve the pivot problem?

axolotl0212:
> How do we going to solve the pivot problem?

Haven't decided, I believe we can specify some callback on switching tab to block switching tab. But I believe this needs to be handled separately

axolotl0212:
rebased and pushed 1 more commit change isFormModified state to derived value

Issue #260, opened 2020-09-07
Custom Attributes + Claims (Primary)


chpapa:
## Metadata Feature Completion

Developer users can add custom attributes as key/value to each user.

We will need the following changes:

- Configuration: Setting the schema of custom attributes (Type, Type-Attribute, Mutatable by User)
- OAuth Endpoints: Return some Standard OIDC Attributes in Claims
- Fill in SSO scope to Standard OIDC Attributes, and sync mode = 'none' vs 'first login' vs 'force'
- Admin API: CRUD of custom attributes of User
- Management Portal: User Details, list out all custom attributes, and edit of custom attributes
- Management Portal: Configuration of the schema of custom attributes

### 2nd Release / Nice to have
- Allow developer user to change, which custom attributes to return in Claims
- Allow developer to configure, which attributes are hidden/read-only/editable in Auth UI Setting page, or via Admin API 
- More type of the custom attributes, for some auto-validation

### Future
- Allow developer to fine tune permission of each custom attributes, which app can access what custom attributes

## Claims
Currently, Authgear doesn't return much information at the claim, main reason is we support multiple identities and there are no way to choose the primary email/phone.

We should allow end user to set it in Auth UI Setting page.

kiootic:
Some notes:
- the term 'labels' is borrowed from Kubernetes.
- they are solely used by the developer, and is not read by Authgear.
- it is planned that authorization mechanism in portal would be implemented using labels.

chpapa:
> they are solely used by the developer, and is not read by Authgear.
> it is planned that authorization mechanism in portal would be implemented using labels.
Actually it smell like we actually shouldn't overload custom attributes feature implemented by labels?

kiootic:
> Actually it smell like we actually shouldn't overload custom attributes feature implemented by labels?

Portal should considered an independent component from the core Authgear (i.e. main server, resolver server, admin APIs). Therefore there're no contradictions between the two points you quoted.

Consider a Role-Based Access Control plugin implemented on top of Authgear: it may read the role labels attached to a particular user in authorization process. This kind of plugin may be developed by us in future, but it would not be considered as a core Authgear component.

chpapa:
But custom attributes will be a core Authgear feature, and I'm worried it would be not flexible enough... say if our next feature is provide end-to-end encrypted custom attributes, or if we someone want to "search" the custom attributes...?

kiootic:
I have some questions about the two use cases you mentioned:
[...]

Issue #261, opened 2020-09-07
Allow configuration of handling conflicted OAuth and Email Address


chpapa:
We should have a new configuration for the user to specify the behaviour when an OAuth Provider's email address clash with Email Address of existing identities.

Such as when user use Google login, to signup/login with an email address that's already verified by another user account

There should be two options:

* Never merge: no matter what, treat the new signup as a separate account
* Merge: it will display a UI at login, to tell users they have already got another account with the same email in this system, ask them to log in again to merge the new OAuth Provider. If `verification.claims.email.required=true`, only email address verified will trigger this merge screen.

Also, we will have an "auto merge" option for each OAuth Provider, if it is set to true, it will automatically merged with the other account if the above option is merge. Again, if `verification.claims.email.required=true`, only email address verified will trigger this merge screen. Something like this:

```yaml
identity:
  oauth:
    providers:
    - alias: google
      type: google
      client_id: "the_client_id"
      claims:
        email:
          auto_merge: true # Will auto trust and merge google's login to existing account. Default is false
```

kiootic:
Discussed with @louischan-oursky :
- When signing up with an identity, the claims of the identity will be used to match existing identity.
   If any identity has same claims, it conflicted and enter the merge flow.
- There are two stage in the merge flow: non-interactive merge/interactive merge.
   Developer can configure to always skip non-interactive merge.
- In non-interactive merge stage, developer can configure to use one of the behaviors:
    - Abort: abort the flow when there is any conflict.
    - AlwaysNew: always create a new account.
    - MergeIfPossible: merge with without user interaction is possible, otherwise go to interactive merge stage.
- Merge with without user interaction is possible if:
    - the conflicted identities have same user; and
    - some of the conflicted identities are verified; and
    - the identity to sign up is verified.
    - This case mainly occurs when user signup with a OAuth identity with a trusted email, and existing login ID/OAuth identity
       has verified email.
- After merging non-interactively, user still need to perform primary/secondary authentication if needed, as in normal authentication flow.
- In interactive merge stage, user is redirected to a page with:
    - information on the identity user attempted to sign up; and
    - inform user that existing account with same claim(email/phone number/username) exists; and
    - Provide options for user to proceed.
- Developer can configure one of the behaviors in interactive merge stage:
    - Merge: request user to authenticate with one of the conflicted identity.
    - AlwaysNew: ask user to confirm creating a new account.
    - MergeOrNew: combination of options in Merge and AlwaysNew
[...]

PR #341, opened 2020-09-18
Specification on custom attributes and claims

louischan-oursky:
ref #260


docs/specs/user-model.md
```diff
@@ -204,9 +220,9 @@ If the domain part of an Email login ID is internationalized, there are 2 ways t
 
 ## Authenticator
 
-Authgear supports various types of authenticator. Authenticator can be primary, secondary or both.
+Authgear supports various types of authenticator. Authenticator must either be primary or secondary.
 
-Authenticators have priorities. The first authenticator is the default authenticator in the UI.
+A primary or secondary authenticator can be set as default. The default authenticator is used first.
```
chpapa, line 225: I think we can turn "default" to another todo? Or it is implemented already in the backend?

docs/specs/user-model.md
```diff
@@ -260,35 +266,11 @@ configuration.
 
 #### OOB-OTP Authenticator
 
-Out-of-band One-time-password authenticator is either primary or secondary.
+Out-of-band One-time-password authenticator can be primary or secondary.
 
 OOB-OTP authenticator is bound to a recipient address. The recipient can be an email address or phone number that can receive SMS messages.
 
-An OOB-OTP authenticator may matches a login ID identity. The normalized email
-address/phone number is used to match login ID identities.
```
chpapa, line 268: These paragraphs are useful, why removed it? (Moved somewhere?)

docs/specs/user-model.md
```diff
@@ -312,3 +294,164 @@ Once used, a recovery code is invalidated.
...
+  identity_id: "google"
+  name_pointer: "#/email"
+  value_pointer: "#/email"
+```
+
+So the user will have the claim `email` immediately available in their claims.
+
+### Dangling reference in claims mapping
+
+If a deletion of identity will result in dangling reference in the claims mapping, the deletion is disallowed.
```
chpapa, line 389: Would it result in the following unexpected behaviour?
- User sign up with email
- User login with Google
- Google's email claim map to the user's email claim
- At User Setting page, the user can never remove the Google login, but he don't understand why... and could never delete his google login because the developer also don't understand why.

chpapa: or the mapping is not "generate the claims everytime claims was queried", but more like "automatically fill in according to claims change"? (random idea.... )

chpapa: or it mapping need to have some way to handle situation like the above, either:

1. If Google login was removed, the next default is the email identities
2. If Google login was removed, the email remains in the claims but unverified.

Both way seems complicated?

docs/specs/user-model.md
```diff
@@ -312,3 +294,164 @@ Once used, a recovery code is invalidated.
...
+      value_pointer: "#/profile/email"
+    - kind: "custom_attributes"
+      name_pointer: "#/zoneinfo"
+      value_pointer: "#/profile/preferred_timezone"
+    - kind: "custom_attributes"
+      name_pointer: "#/picture"
+      value_pointer: "#/profile/profile_image_url"
+```
+
+The template is copied to the claims mapping of new user.
```
chpapa, line 414: If the template is copied, does that meant when the template is updated, it need to run migration for each users?

chpapa: Or easier to just make it as template for every users... (or how useful is it for each user have its own claims map?)

chpapa (changes requested):
@louischan-oursky here are my comments.

docs/specs/user-model.md
```diff
@@ -1,71 +1,87 @@
...
+
+Custom Attributes is a JSON Object. The developer can update it via Admin API.
+
+Custom Attributes can be any valid JSON Object. There is no limitation on the nesting level. However, a sensible size limit (e.g. 10MiB) is enforced.
+
+The developer can save whatever they want in Custom Attributes.
+
+> NOTE: The initial implementation is very likely to be a single textarea for editing the whole custom attributes object.
+> Much more fancier features like generated form fields requires us to interpret the JSON schema and generate form fields.
+> That could be a very useful standalone product.
```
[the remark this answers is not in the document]
louischan-oursky: The value can be complex values like JSON object and JSON array, which does not fit well in a text field.

docs/specs/user-model.md
louischan-oursky: This issue was discussed in 2020-09-23 product meeting. As long as we allow the user to choose which authenticator to use, we do not need the concept of default authenticator.

docs/specs/user-model.md
louischan-oursky: > Would it result in the following unexpected behaviour?

The error message can explain why deletion is impossible and offer ways to resolve the error

1. Do not delete the identity at all. The information from this identity is part of the public user info (i.e. claims). Deleting the identity WILL change the public user info.
2. Remove the mapping first. It is equivalent to making the user to aware such action is affecting the claims.

---

> or the mapping is not "generate the claims everytime claims was queried", but more like "automatically fill in according to claims change"? (random idea.... )

Not sure what you meant. Can you explain in more details?

---

> If Google login was removed, the next default is the email identities

This seems surprising to me.

---

> If Google login was removed, the email remains in the claims but unverified.

This is actually an alternate design choice. So in this design, we do not have `kind: identity` claim mapping. Every claim mapping entry is just `kind: custom_attributes`. But specifically for the claim `email` and the claim `phone_number`, we will use available information to provide `email_verified` and `phone_number_verified`.

Personally I think the original design (`kind: identity`) offers a stronger guarantee. The claim `email` is very highly to be verified (at least from Authgear's point of view).

docs/specs/user-model.md
louischan-oursky: > If the template is copied, does that meant when the template is updated, it need to run migration for each users?

Yes :(

Or we can improve the design by replacing the copy behavior with a runtime merge. So a change in the configuration could affect the claims of ALL users which could be considered as surprising.

---

> Or easier to just make it as template for every users... (or how useful is it for each user have its own claims map?)

The primary reason of user owning their claims mapping is that we support `kind: identity` claim mapping, which is a reference to one of the identity the user owns. This cannot be expressed in the claim mapping template in the configuration.

louischan-oursky:
Next time when we write a specification, if possible, we should list out all alternate designs and make a comparison between them. This would make the reader easier to evaluate the pros and cons of different designs.

docs/specs/user-model.md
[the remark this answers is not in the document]
chpapa: @louischan-oursky Any better suggestions that's not very complicated to develop in our MVP, but also not a textarea that is impossible for normal users?

Or we can have key textfields, and each value is textarea

docs/specs/user-model.md
chpapa: > The error message can explain why deletion is impossible and offer ways to resolve the error

Can you try to propose the most user friendlys flow and error message, when users delete their Google Login but got rejected? And how do we visualize the mapping of claims in Setting page for layman users?

> Not sure what you meant. Can you explain in more details?
I meant it is not a mapping that's generative / delcarative, but runs the mapping only once when a value change, effectively that will make it like if Google Login was removed, the email remains in the claims.

chpapa:
> Next time when we write a specification, if possible, we should list out all alternate designs and make a comparison between them. This would make the reader easier to evaluate the pros and cons of different designs.

Could be as a seperate document with link from the original document, otherwise it will be too long for reading...

docs/specs/user-model.md
chpapa: What about if we always have template claims mapping applied to all users, and each users have their own claims mapping which "overwrite" the default if conflicted?

I think copy / runtime merge might be too complicated? I think the change in the configuration that does not affect the claims of ALL users would be surprising in the real world, because most use cases would expect the whole auth system (or at least for each client-id) have only 1 claims mapping...

louischan-oursky:
@kiootic @chpapa I have updated the spec. Added standard attributes and simplified claims mapping.

docs/specs/user-model.md
```diff
@@ -1,71 +1,127 @@
...
+> Then we can support more Standard Claims, such as `picture`.
 
-The claims are used to detect duplicate identity. For example, an Email Login ID and the email claim of an OAuth Identity. This prevents duplicate user when the user forgets the original authentication method.
+The claim `email` is used to detect duplicate identity. For example, an Email Login ID and the email claim of an OAuth Identity. This prevents duplicate user when the user forgets the original authentication method.
 
 ### OAuth Identity
 
 OAuth identity is external identity from supported OAuth 2 IdPs. Only authorization code flow is supported. If the provider supports OIDC, OIDC is preferred over provider-specific OAuth 2 protocol.
 
+OAuth identity is updated every time authentication is performed.
```
kiootic, line 121: If an OAuth identity is updated to have the same email as an existing login ID identity in system, it would result in a conflict. This is related to the merge flow in #261.

I think we can clarify the for now, if updating an OAuth identity would result in a conflict, the update is skipped?

docs/specs/user-model.md
```diff
@@ -312,3 +338,207 @@ Once used, a recovery code is invalidated.
...
+- kind: "system"
+  name_pointer: "#/phone_number"
+- kind: "system"
+  name_pointer: "#/phone_number_verified"
+- kind: "system"
+  name_pointer: "#/preferred_username"
+
+- kind: "custom_attributes"
+  name_pointer: "#/zoneinfo"
+  value_pointer: "#/profile/preferred_timezone"
```
kiootic, line 450: I suppose it's forbidden to set mapping for an intermediate path of any value pointer?
i.e. If there is a mapping mapped to `#/profile/preferred_timezone`, developer cannot set mapping mapped to `#/profile` or `#`.

docs/specs/user-model.md
```diff
@@ -312,3 +338,207 @@ Once used, a recovery code is invalidated.
...
+          "value_pointer": { "type": "string" }
+        },
+        "required": ["kind", "name_pointer", "value_pointer"]
+      }
+    ]
+  }
+}
+```
+
+- `kind`: The kind of the mapping, either `identity` or `custom_attributes`.
```
kiootic, line 539: This point need update.

docs/specs/user-model.md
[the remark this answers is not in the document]
chpapa: @louischan-oursky can we resolve this problem in the specification before finalizing it?

docs/specs/user-model.md
```diff
@@ -1,71 +1,127 @@
...
+    properties:
+      email:
+        type: string
+        format: email
+    required: ["email"]
+```
+
+Any update on custom attributes must be valid against the given schema.
+
+> NOTE: In the future, we can add our own JSON schema keyword to annotate the access control of individual fields in the custom attributes.
```
chpapa, line 93: As a 2nd thought, I think we should include access control depends on Client ID on this spec.

Because I have a gut feeling that it might complicate the claims computation x custom attributes a lot (e.g. if a custom attribute is not accessible by Client A, but it is still in the claims, what would happen?) We can leave the implementation to later, but I hope our spec can cater it so we don't introduce a big backward incompatibility and redo everything in future...

docs/specs/user-model.md
```diff
@@ -312,3 +338,207 @@ Once used, a recovery code is invalidated.
...
+
+### Builtin claims mapping
+
+Builtin claims mapping is a list of mapping entries with `kind: "system"` and supported standard claims name.
+
+The full list is:
+
+```
+- kind: "system"
+  name_pointer: "#/email"
```
chpapa, line 361: Can this spec doc explain why it starts with `#/`?

chpapa (changes requested):
Some comments / clarification / request for further design 🙏

docs/specs/user-model.md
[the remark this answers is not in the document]
louischan-oursky: The custom attributes have a JSON schema defined the developer.
An example of similar thing is AWS Policy editor. AWS policy has schema as well. They have a wizard to generate policy as well as a plain old textarea for editing the policy JSON directly.

Since the provided JSON schema could be very complex and it is non-trivial to take a schema and generate form fields, I prefer we just stick with a textarea in our MVP. It ensures we provide the full capability of custom attributes editing experience.

On the other hand, how likely would a "normal user" edit the custom attributes? Custom attributes initially should be written by the developer via Admin API. The developer can store something tightly related to their business logic, which may require more application knowledge to make changes.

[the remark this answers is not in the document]
chpapa: > On the other hand, how likely would a "normal user" edit the custom attributes? 

My guess is very often, because:

1. Developer may use the custom attributes as some sort of customer profile, where business users will need to look at it or update it from management portal.
2. Following 1, it might even some time handy for developer to just enable changing it from our setting pages (this one might be a bit unclear how good / bad it is)

Hence I think without it might vastly limited the usage of custom attributes...

However I'm okay if we implement it in multiple phase, like phase 1 MVP only text area, followed by phase 2 make it view and changeable by business user in management portal; After that we can stop and wait for a few months with more real users, and see if they even want it in setting page.

docs/specs/user-model.md
kiootic: If we provide access control on custom attributes/user claims, I think we'll want to provide access control on APIs for OAuth clients as well.

Usually OAuth scopes are hierarchical, with possibly complex interpretation (e.g. `user.email`, `user.email.read`). We may need to take that into account when designing the access control mechanism.

louischan-oursky: I have added a document about access control.

docs/specs/access-control.md
```diff
@@ -0,0 +1,55 @@
...
+- OAuth Client
+- The User on web UI
+- The Caller of Admin API
+
+## Overview of Access Control between Information and Party
+
+|                       |Standard Attributes|Custom Attributes|Claims                               |
+|-----------------------|-------------------|-----------------|-------------------------------------|
+|OAuth Client           |No access          |No access        |Read access controlled by OAuth scope|
+|The User on web UI     |Read and Write     |No access (yet)  |No access                            |
```
chpapa, line 28: Would there be a situation like if user A logged in with a OAuth Client ID, which has no access to a "claimA" which is based on "attrA".

But the user on the Web UI have access to "attrA"?

Or on the other hand, would it be technically possible to identify if a user on web UI was logged in from which OAuth Client ID? And is it useful to give them different access based on which Client ID they logged in?

Or maybe too complicated...

louischan-oursky: > Would there be a situation like if user A logged in with a OAuth Client ID, which has no access to a "claimA" which is based on "attrA".
> 
> But the user on the Web UI have access to "attrA"?

In OAuth model, the user is authorizing access to their resources on the resource server to a third party (i.e. the Client), so the user should have access to their resources on the resource server. In our case, that means the user have access to "attrA" on the web UI.

However, if we allow custom attributes to be accessible by the user, the situation becomes more complicated. The user is no longer the sole owner of the resources on the resource server. The developer can define some custom attributes that cannot be read by the user, but at the same time, the developer can map them as claims.
During the OAuth authorization flow, only the user is involved in deciding which scope is granted to the Client.

For example, the developer defines a custom attribute "rbac" that is not readable by the user. At the same time, the developer maps custom attribute "rbac" to claim "app:rbac", with the restriction that scope "app:rbac" is required.
The user cannot read it but they are responsible for granting access to it during authorization flow, which is quite contradictory.

> Or on the other hand, would it be technically possible to identify if a user on web UI was logged in from which OAuth Client ID? And is it useful to give them different access based on which Client ID they logged in?

Actually the relationship is reverse. A IdP session may be associated with many refresh tokens. Strictly speaking, the web UI should not be aware of which client ID. It is because it is perfectly fine to login directly, without any Client. In that case, we show all information of the user. Technically speaking, on the web UI we are only seeing the IdP session cookie, which solely means the user is authenticated, not specific to any Client.

Imagine there are 2 apps using Google Login, you won't expect to see different information on myaccount.google.com after you login one of the app.

chpapa: > The user cannot read it but they are responsible for granting access to it during authorization flow, which is quite contradictory.

Hence I asked the question. I think in that case we might better define the permission as:

1. Claims via each Client ID (under OAuth)
2. Attributes via Setting Page
3. Attributes via GraphQL Endpoint (assume it is admin so all R/W access)

So in future, we might have more permissions such as, if we have user roles: 
* Attributes via User Roles on Management Portal
* Attributes via User Roles on Setting Page

louischan-oursky:
Superseded by https://github.com/authgear/authgear-server/pull/1498

PR #349, opened 2020-09-21
Expose claim verification status in GraphQL queries

### LINKED ISSUE #345: [Admin API] Return the verified claims in user node
Background: PR #289 make claims verifiable. In portal authentication screen connected identities, we want to toggle verified state, so we need a way to determine which claim is verified by querying user node in admin API in frontend portal.

- [ ] query `_auth_verified_claims` table with user ID
- [ ] add field in user node to return the verified claims

kiootic:
@chpapa Exposing the verification status of user claims in Admin API

fix #345


chpapa:
@kaojohnny @louischan-oursky umm, if I just look at the spec, it wasn't obvious what is `Claim.name` and `Claim.value`, can we add some samples in the schema file, assuming it is currently the only doc and doc would be generated from there?

chpapa:
@kiootic @louischan-oursky umm, if I just look at the spec, it wasn't obvious what is Claim.name and Claim.value, can we add some samples in the schema file, assuming it is currently the only doc and doc would be generated from there?

PR #357, opened 2020-09-22
Fix MFA setup

kiootic:


louischan-oursky:
Conflicted.

kiootic:
@louischan-oursky rebased.

Issue #375, opened 2020-09-23
[Portal] Handle adding password on adding username identity


axolotl0212:
If in primary authenticator
-> starting with oob_otp (no password needed if there is no username identity)
-> adding username identity
-> if there is no primary password authenticator, username identity cannot be used for login

- [x] create function, determine whether password is needed 
- [x] if password needed after adding username identity, show password field, provide password on adding identity
- [x] handle corresponding error

PR #376, opened 2020-09-23
[Portal] Integrate create delete identity mutation

### LINKED ISSUE #367: [Portal] Integrate create identity mutation
- [x] add create identity hook
- [x] call mutation in add username, email, phone page
- [x] call mutation with corresponding login ID, userID from url
- [x] on success, back to user detail screen
- [x] make sure connected identities tab is selected when go back to user detail screen
- [x] make sure updated data is shown

TODO: 
- how to handle the following case
   - If in primary authenticator
   -> starting with `oob_otp` + `password` (no password needed if there is no username identity)
   -> adding username identity
   -> if there is no existing password, username identity cannot be used for login

- mapping:
  - username: username
  - phone: phone
  - email: email

### LINKED ISSUE #368: [Portal] integrate delete identity mutation
- [x] create delete identity hook
- [x] call mutation on remove button click in user detail connected identities tab
- [x] handle error (show dialog? TBD)

axolotl0212:
Closes #367
Closes #368

TODO: handle show password field for add username (#375)

### DIFF
--- a/portal/src/graphql/adminapi/AddEmailScreen.tsx
+++ b/portal/src/graphql/adminapi/AddEmailScreen.tsx
@@ -73,7 +73,7 @@ const AddEmailScreen: React.FC = function AddEmailScreen() {
     createIdentity({ key: "email", value: email })
       .then((identity) => {
         if (identity != null) {
-          navigate("../");
+          navigate("../#connected-identities");
         } else {
           throw new Error();
         }
--- a/portal/src/graphql/adminapi/AddEmailScreen.tsx
+++ b/portal/src/graphql/adminapi/AddEmailScreen.tsx
@@ -30,7 +30,6 @@ const AddEmailScreen: React.FC = function AddEmailScreen() {
   } = useCreateLoginIDIdentityMutation(userID);
   const { renderToString } = useContext(Context);
 
-  const [violations, setViolations] = useState<Violation[]>([]);
   const [unhandledViolations, setUnhandledViolations] = useState<Violation[]>(
     []
   );
@@ -74,18 +73,15 @@ const AddEmailScreen: React.FC = function AddEmailScreen() {
       .then((identity) => {
         if (identity != null) {
           navigate("../#connected-identities");
-        } else {
-          throw new Error();
         }
       })
-      .catch((err) => {
+      .catch(() => {
         setDisableBlockNavigation(false);
-        const violations = parseError(err);
-        setViolations(violations);
       });
   }, [email, navigate, createIdentity]);
 
   const errorMessage = useMemo(() => {
+    const violations = parseError(createIdentityError);
     const emailFieldErrorMessages: string[] = [];
     const unknownViolations: Violation[] = [];
     for (const violation of violations) {
@@ -107,7 +103,7 @@ const AddEmailScreen: React.FC = function AddEmailScreen() {
     return {
       email: defaultFormatErrorMessageList(emailFieldErrorMessages),
     };
-  }, [violations, renderToString]);
+  }, [createIdentityError, renderToString]);
 
   return (
     <div className={styles.root}>
--- a/portal/src/graphql/adminapi/AddEmailScreen.tsx
+++ b/portal/src/graphql/adminapi/AddEmailScreen.tsx
@@ -1,4 +1,10 @@
-import React, { useCallback, useContext, useMemo, useState } from "react";
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from "react";
 import { useNavigate, useParams } from "react-router-dom";
 import { TextField, Toggle } from "@fluentui/react";
 import deepEqual from "deep-equal";
@@ -30,9 +36,7 @@ const AddEmailScreen: React.FC = function AddEmailScreen() {
   } = useCreateLoginIDIdentityMutation(userID);
   const { renderToString } = useContext(Context);
 
-  const [disableBlockNavigation, setDisableBlockNavigation] = useState<boolean>(
-    false
-  );
+  const [submittedForm, setSubmittedForm] = useState<boolean>(false);
 
   const navBreadcrumbItems = useMemo(() => {
     return [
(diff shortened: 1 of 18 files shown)


portal/src/graphql/adminapi/AddPhoneScreen.tsx
```diff
@@ -84,16 +107,61 @@ const AddPhoneForm: React.FC<AddPhoneFormProps> = function AddPhoneForm(
...
-    // TODO: mutation to be integrated
-  }, []);
+    setDisableBlockNavigation(true);
+    const combinedPhone = `+${countryCode}${phone}`;
+    createIdentity({ key: "phone", value: combinedPhone })
+      .then((identity) => {
+        if (identity != null) {
+          navigate("../#connected-identities");
+        } else {
+          throw new Error();
```
kiootic, line 117: It seems we have two kinds of error handling:
- call the mutation function, and use `catch` on the result promise
- check for the `error` returned from `useMutation` hook

Can you help checking whether the first case will actually trigger? i.e. GraphQL errors would throw a JS error from the mutation function promise. If not, we can simplify and ignore it, and use `error` from mutation hook exclusively.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
```diff
@@ -230,22 +296,79 @@ const UserDetailsConnectedIdentities: React.FC<UserDetailsConnectedIdentitiesPro
...
+
+  const onConfirmRemoveIdentity = useCallback(() => {
+    if (confirmationDialogData == null) {
+      return;
+    }
+    const { identityID } = confirmationDialogData;
+    deleteIdentity(identityID)
+      .then((success) => {
+        if (success) {
+          refetchUserDetail();
```
kiootic, line 321: Instead of refetching manually after a mutation, I prefer using `refetchQueries` of mutation options to automatically refetch queries to decouple dependencies between queries and mutations from presentational concerns.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
```diff
@@ -273,44 +398,86 @@ const UserDetailsConnectedIdentities: React.FC<UserDetailsConnectedIdentitiesPro
...
       directionalHintFixed: true,
-    }),
-    [renderToString, navigate]
-  );
+    };
+  }, [renderToString, navigate, availableLoginIdIdentities]);
 
   return (
     <div className={styles.root}>
+      {confirmationDialogData && (
```
kiootic, line 444: It seems the dismiss animation is not there. I suppose we should use `hidden` prop to close the dialog properly.

portal/src/graphql/adminapi/AddPhoneScreen.tsx
axolotl0212: > first case will actually trigger?

It will, so we can either use `.catch(() => {})` and use the error given be `useMutation` hook, or just use the error given by `catch`.

louischan-oursky: So both of them are the same error instance? I prefer we use the error returned from the hook.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
axolotl0212: Didn't realise that there is animation, will update this

axolotl0212:
updated

portal/src/graphql/adminapi/AddPhoneScreen.tsx
axolotl0212: > So both of them are the same error instance?

same instance, at least the data structure is exactly the same

axolotl0212:
updated to using error from useMutation hook

portal/src/graphql/adminapi/mutations/deleteIdentityMutation.ts
```diff
@@ -0,0 +1,34 @@
...
+`;
+
+export function useDeleteIdentityMutation(): {
+  deleteIdentity: (identityID: string) => Promise<boolean>;
+  loading: boolean;
+  error: unknown;
+} {
+  const [mutationFunction, { error, loading }] = useMutation<
+    DeleteIdentityMutation
+  >(deleteIdentityMutation, { refetchQueries: ["UserDetailsScreenQuery"] });
```
kiootic, line 20: I suppose we may want to extract the query to an individual file, to avoid referencing screen-specific queries in a generic mutation.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
```diff
@@ -230,22 +295,77 @@ const UserDetailsConnectedIdentities: React.FC<UserDetailsConnectedIdentitiesPro
...
+      setConfirmationDialogData({
+        identityID,
+        identityName,
+      });
+    },
+    [setConfirmationDialogData]
+  );
+
+  const onDismissConfirmationDialog = useCallback(() => {
+    setConfirmationDialogData(null);
```
kiootic, line 309: During the dismiss animation of confirmation dialog, the identity name (email) disappears and cause strange layout. We should use a separate flag to control visibility of dialog. For example:
- `isHidden`: initially true, set to false when presenting, set to true when dismissing.
- `dialogData` initially an empty value, set to correct value when presenting, do not change when dismissing.

kiootic (commented):
- I got this warning in console `Pivot property 'initialSelectedKey' was used but has been deprecated. Use 'defaultSelectedKey' instead.`

portal/src/graphql/adminapi/AddEmailScreen.tsx
```diff
@@ -44,20 +68,62 @@ const AddEmailScreen: React.FC = function AddEmailScreen() {
...
   const onAddClicked = useCallback(() => {
-    // TODO: mutation to be integrated
-  }, []);
+    setDisableBlockNavigation(true);
+    createIdentity({ key: "email", value: email })
+      .then((identity) => {
+        if (identity != null) {
+          navigate("../#connected-identities");
+        } else {
+          throw new Error();
```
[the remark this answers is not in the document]
axolotl0212: thanks, missed out add email and add password screen

portal/src/graphql/adminapi/UserDetailsScreen.tsx
```diff
@@ -20,20 +21,29 @@ import UserDetailsSession from "./UserDetailsSession";
...
   props: UserDetailsProps
 ) {
-  const { data, loading } = props;
+  const { data, loading, appConfig } = props;
+  const location = useLocation();
+  const hash = location.hash.slice(1);
   const { renderToString } = React.useContext(Context);
 
+  const availableLoginIdIdentities = useMemo(() => {
+    const rawLoginIdKeys = appConfig?.identity?.login_id?.keys ?? [];
```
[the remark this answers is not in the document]
axolotl0212: thanks, same logic as options in create user

axolotl0212:
@louischan-oursky updated

PR #379, opened 2020-09-24
Use correct Host when proxying Admin API

### LINKED ISSUE #374: Use correct Host when proxying Admin API
We have to provide the correct Host such that Admin API can resolve which app we are referring to.
We can use parse `http.public_origin` and use the host field.

louischan-oursky:
fix #374

### DIFF
--- a/pkg/portal/transport/admin_api_handler.go
+++ b/pkg/portal/transport/admin_api_handler.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/util/httproute"
+	"github.com/authgear/authgear-server/pkg/util/log"
 )
 
 func ConfigureAdminAPIRoute(route httproute.Route) httproute.Route {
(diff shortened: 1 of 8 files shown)


pkg/portal/transport/admin_api_handler.go
```diff
@@ -44,26 +54,45 @@ func (h *AdminAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	hostURL, err := url.Parse(cfg.AppConfig.HTTP.PublicOrigin)
+	if err != nil {
+		h.Logger.WithError(err).Debugf("failed to parse public origin: %v", cfg.AppConfig.HTTP.PublicOrigin)
+		http.Error(w, fmt.Errorf("failed to parse public origin: %v", cfg.AppConfig.HTTP.PublicOrigin).Error(), http.StatusInternalServerError)
+		return
+	}
+
 	authKey, ok := cfg.SecretConfig.LookupData(config.AdminAPIAuthKeyKey).(*config.AdminAPIAuthKey)
 	if !ok {
+		h.Logger.WithError(err).Debugf("failed to look up admin API auth key")
```
kiootic, line 71: I don't think value of `err` is relevant here?

pkg/portal/transport/admin_api_handler.go
```diff
@@ -44,26 +54,45 @@ func (h *AdminAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		"endpoint": endpoint.String(),
+		"host":     hostURL.Host,
+	}).Debugf("proxy admin API")
+
 	proxy := httputil.ReverseProxy{
 		Director: func(req *http.Request) {
 			req.URL = endpoint
-			req.Host = endpoint.Host
+			// We have to set both to ensure Admin API server see the correct host.
+			req.Host = hostURL.Host
```
kiootic, line 94: I don't think this is correct? `AppConfig.HTTP.PublicOrigin` is supposed to reach the main server from Internet, while we want to reach the admin API server from Internet. I suppose this is a cluster specific logic to derive correct endpoint for admin API using app ID? For example: `ADMIN_API_ENDPOINT_TEMPLATE=admin.{{ .AppID}}.authgear.com`.

kiootic (approved):

PR #382, opened 2020-09-24
[Portal] Integrate create user mutation

### LINKED ISSUE #369: [Portal] Integrate create user mutation
- update add user page
  - [x] radio button choose only one login ID
  - [x] only show text field if corresponding type is chosen
  - [x] always show password field, disable if not needed
         - in primary authenticators:
         - `username` login id type need password authenticator
         - password -> all login id type need password
         - oob_otp -> only email and phone allowed, no password needed
         - oob_otp + password -> email and phone no password needed, username need password
         - password + oob_otp -> all login id type need password.   (password has higher priority -> need password)

- [x] create hook for mutation, password nullable
- [x] call mutation on submit form
  - [x] use login id fields for `Identity definition`
          - selected field -> key
          - text field value -> value
- [x] leave password field null if not required
- [x] handle error
- [x] on success, back to user screen 
- mapping: 
  - username: username
  - phone: phone
  - email: email

axolotl0212:
Closes #369

- update user screen
- add hook for mutation
- handle error

### DIFF
--- a/portal/src/graphql/adminapi/AddUserScreen.tsx
+++ b/portal/src/graphql/adminapi/AddUserScreen.tsx
@@ -3,7 +3,14 @@ import { useParams } from "react-router-dom";
 import zxcvbn from "zxcvbn";
 import deepEqual from "deep-equal";
 import { Context, FormattedMessage } from "@oursky/react-messageformat";
-import { Label, PrimaryButton, Text, TextField } from "@fluentui/react";
+import {
+  PrimaryButton,
+  Text,
+  TextField,
+  ChoiceGroup,
+  IChoiceGroupOption,
+  Label,
+} from "@fluentui/react";
 
 import { useAppConfigQuery } from "../portal/query/appConfigQuery";
 import NavBreadcrumb, { BreadcrumbItem } from "../../NavBreadcrumb";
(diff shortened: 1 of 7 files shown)


portal/src/graphql/adminapi/mutations/createUserMutation.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+    ) {
+      user {
+        id
+      }
+    }
+  }
+`;
+
+interface Identity {
```
kiootic, line 23: Let's name it `LoginIDIdentity` instead.

portal/src/graphql/adminapi/AddUserScreen.tsx
```diff
@@ -1,140 +1,329 @@
...
+        loginIdKeyOptions.add("email");
+        break;
+      case "phone":
+        loginIdKeyOptions.add("phone");
+        break;
+      default:
+        break;
+    }
+  }
+  return Array.from(loginIdKeyOptions);
```
kiootic, line 117: We should also check whether login ID identity is enabled (i.e. `login_id` is contained in `authentication.identities`).

axolotl0212: How should we handle the case where login ID identity is not enabled? create user requires loginIDIdentities, I think we need a simple empty state in case where user cannot be added? @kiootic

kiootic: I see there is an existing empty state when user cannot be added. If I understood correctly, returning an empty array from the function should trigger that state?

axolotl0212: Yes, that's correct. However it is a text only empty state which does not look good... I added `TODO` in the corresponding line. I have returned empty array as there is no possible login ID identity option if login ID identity itself is disabled.

louischan-oursky (commented):
Similar issue from https://github.com/authgear/authgear-server/pull/376 was also found in this PR.
mutationFunction error vs error returned from the hook

louischan-oursky:
conflicted.

axolotl0212:
> conflicted.

Will rebase and address comments in #376

axolotl0212:

PR #388, opened 2020-09-24
[Portal] Integrate delete authenticator mutation

### LINKED ISSUE #370: [Portal] Integrate delete authenticator mutation
- [x] create delete authenticator hook
- [x] call mutation on remove button click in user detail account security tab (only secondary authenticator)
- [x] handle error (show dialog)

axolotl0212:
Closes #370 

TODO: 
- handle error from backend
  - get unknown error from backend
  - can only add 1 secondary authenticator during setup, cannot test for error

get following error from backend
```json
{
	"data": null,
	"errors": [
		{
			"message": "reflect: call of reflect.Value.Type on zero Value",
			"locations": [
				{
					"line": 2,
					"column": 3
				}
			],
			"path": [
				"deleteAuthenticator"
			]
		}
	]
}
```

### DIFF
--- a/portal/src/graphql/adminapi/UserDetailsAccountSecurity.tsx
+++ b/portal/src/graphql/adminapi/UserDetailsAccountSecurity.tsx
@@ -1,8 +1,16 @@
-import React, { useMemo, useCallback, useContext } from "react";
+import React, {
+  useMemo,
+  useCallback,
+  useContext,
+  useState,
+  useEffect,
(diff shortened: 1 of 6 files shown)


axolotl0212:
@kiootic rebased

kiootic (commented):
I fixed the server bug and test, it seems the authenticator list is not updated after removing authenticator. I'll open an issue for that.

Issue #385, opened 2020-09-24
[Portal] Handle verified claims in connected identities and add identities


axolotl0212:
- [x] get verified claims from user node
- [x] if the claim is found in verified claims, treat as verified
- [x] reflect the verified state on UI

mutation:
- [x] in add identity, remove toggle button
- [x] integrate setVerifiedStatus mutation

PR #406, opened 2020-09-25
Display oauth identities in separate list, show icon according to provider type

### LINKED ISSUE #393: [Portal] Identity icon are incorrect
If I have a Google identity, it should be a google icon. Please refer to the UI for which icon to use. Right now it is an email icon.

axolotl0212:
Closes #393 

Screenshot 2020-09-25 at 11 31 08 PM

### DIFF
--- a/portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
+++ b/portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
@@ -26,6 +26,7 @@ import { useDeleteIdentityMutation } from "./mutations/deleteIdentityMutation";
 import { formatDatetime } from "../../util/formatDatetime";
 import { parseError } from "../../util/error";
 import { Violation } from "../../util/validation";
+import { OAuthSSOProviderType } from "../../types";
 import { destructiveTheme, verifyButtonTheme } from "../../theme";
 
 import styles from "./UserDetailsConnectedIdentities.module.scss";
@@ -34,11 +35,13 @@ interface IdentityClaim extends Record<string, unknown> {
   email?: string;
   phone_number?: string;
   preferred_username?: string;
+  "https://authgear.com/claims/oauth/provider_type"?: OAuthSSOProviderType;
+  "https://authgear.com/claims/login_id/type"?: LoginIDIdentityType;
 }
 
 interface Identity {
   id: string;
-  type: string;
+  type: "ANONYMOUS" | "LOGIN_ID" | "OAUTH";
   claims: IdentityClaim;
   createdAt: string;
   updatedAt: string;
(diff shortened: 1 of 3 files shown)


portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
```diff
@@ -49,39 +52,50 @@ interface UserDetailsConnectedIdentitiesProps {
...
 }
 
 interface UsernameIdentityListItem {
   id: string;
-  username: string;
+  username?: string;
   addedOn: string;
 }
 
 export interface IdentityLists {
```
kiootic, line 87: I suggest refactoring the list items into two interfaces:
```ts
interface LoginIDIdentityListItem {
  id: string;
  type: "login_id";
  key: "email" | "phone" | "username"
  value: string;
  verified: boolean;
  connectedOn: string;
}

interface OAuthIdentityListItem {
  id: string;
  type: "oauth";
  provider: string;
  name: string;
  verified: boolean;
  connectedOn: string;
}
```

Then, `IdentityListCell` can accept just a list item instead of a variety of fields.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
```diff
@@ -49,39 +52,50 @@ interface UserDetailsConnectedIdentitiesProps {
...
+  id: string;
+  email?: string;
+  verified: boolean;
+  addedOn: string;
+}
+
 interface PhoneIdentityListItem {
   id: string;
-  phone: string;
+  phone?: string;
```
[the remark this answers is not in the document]
axolotl0212: If we use the `http://.../login_id_type` determine the type, then the `claims.phone_number` has type `string | undefined`, cast it to string instead?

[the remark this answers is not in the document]
kiootic: Yes, the simplest way is to cast the phone number claim to string, since the server should ensure it is defined if login ID type is correct.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
axolotl0212: Nice idea, refactored to these two interfaces

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
```diff
@@ -49,39 +52,41 @@ interface UserDetailsConnectedIdentitiesProps {
   availableLoginIdIdentities: string[];
 }
 
-const identityTypes = ["email", "phone", "username"] as const;
-type IdentityType = typeof identityTypes[number];
+const loginIdIdentityTypes = ["email", "phone", "username"] as const;
+type LoginIDIdentityType = typeof loginIdIdentityTypes[number];
+type IdentityType = "login_id" | "oauth";
```
kiootic, line 57: This should be defined as `"ANONYMOUS" | "LOGIN_ID" | "OAUTH"`, and the `type` field in `Identity` type should use it.

The identity list should explicitly hide (i.e. ignore) anonymous identities.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
```diff
@@ -49,39 +52,41 @@ interface UserDetailsConnectedIdentitiesProps {
...
-  addedOn: string;
-}
-
-interface UsernameIdentityListItem {
-  id: string;
-  username: string;
+  type: "login_id";
+  key: "email" | "phone" | "username";
+  value: string;
+  verified?: boolean;
```
kiootic, line 73: Verified status should not be an optional boolean; instead, it should be a required nullable boolean. If it is null, the login ID identity cannot be verified.

Username login ID can never be verified. Phone & email can verified iff verification is enabled in configuration.

Let's fix that in #385.

portal/src/graphql/adminapi/UserDetailsConnectedIdentities.tsx
```diff
@@ -49,39 +52,41 @@ interface UserDetailsConnectedIdentitiesProps {
...
+  email: LoginIDIdentityListItem[];
+  phone: LoginIDIdentityListItem[];
+  username: LoginIDIdentityListItem[];
 }
 
 interface IdentityListCellProps {
   identityID: string;
   identityType: IdentityType;
-  identityName: string;
+  loginIdIdentityType?: LoginIDIdentityType;
```
kiootic, line 87: I'd like to combine the fields `loginIdIdentityType`, `oauthProviderType`, and `identityName`.

These fields are used to represent mutual exclusive structure: if `identityType` is OAuth, `oauthProviderType` presents and `loginIdIdentityType` absents, and same for `identityName`. We can use a union type `OAuthIdentityListItem | LoginIDIdentityListItem` to represent all necessary identity data, and extract view data through helper functions:
```
function getIcon(item: OAuthIdentityListItem | LoginIDIdentityListItem): React.ReactNode;
function getName(item: OAuthIdentityListItem | LoginIDIdentityListItem): string | undefined;
```

axolotl0212: I see, pushed 1 more commit

kiootic (approved):

PR #407, opened 2020-09-25
Warn user on change tab with dirty form

### LINKED ISSUE #333: Handle change tab (Pivot item) discards form state in the tab item
This is caused by pivot item is dismounted when user changes tab, causing state in the tab discarded

As we warn user when user tries to leave the page with dirty form, we should either warn users when user tries to switch tab with dirty form, or persist the form state

Possible solution includes:
- try to persist the state
  - perhaps move state to parent (component with `<Pivot>`) ?
  - local storage ?
- try to find callback on tab change
  - try to block tab switching and show dialog similar to navigation blocker and warn user

axolotl0212:
fix #333 

- use url fragment (hash) to control pivot
- change tab => navigation => can use NavigationBlockerDialog to block
- avoid unnecessary useRef, useEffect to handle warn user on change tab
- fix password screen not warning user when leaving page with dirty form

### DIFF
--- a/portal/src/hook/usePivot.ts
+++ b/portal/src/hook/usePivot.ts
@@ -1,0 +1,24 @@
+import { useCallback } from "react";
+import { useLocation, useNavigate } from "react-router-dom";
+import { IPivotItemProps } from "@fluentui/react";
+
+export function usePivot(): {
+  hash: string;
+  onLinkClick: (item?: { props: IPivotItemProps }) => void;
+} {
+  const navigate = useNavigate();
+  const location = useLocation();
+  const hash = location.hash.slice(1);
+
+  const onLinkClick = useCallback(
+    (item) => {
+      const itemKey = item?.props.itemKey;
+      if (typeof itemKey === "string") {
+        navigate(`./#${itemKey}`);
+      }
+    },
+    [navigate]
+  );
+
+  return { hash, onLinkClick };
+}
--- a/portal/src/hook/usePivot.ts
+++ b/portal/src/hook/usePivot.ts
@@ -3,7 +3,7 @@ import { useLocation, useNavigate } from "react-router-dom";
 import { IPivotItemProps } from "@fluentui/react";
 
 export function usePivot(): {
-  hash: string;
+  selectedKey: string;
   onLinkClick: (item?: { props: IPivotItemProps }) => void;
 } {
   const navigate = useNavigate();
@@ -20,5 +20,5 @@ export function usePivot(): {
     [navigate]
   );
 
-  return { hash, onLinkClick };
+  return { selectedKey: hash, onLinkClick };
 }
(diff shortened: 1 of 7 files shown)


portal/src/hook/usePivot.ts
```diff
@@ -0,0 +1,24 @@
...
+} {
+  const navigate = useNavigate();
+  const location = useLocation();
+  const hash = location.hash.slice(1);
+
+  const onLinkClick = useCallback(
+    (item) => {
+      const itemKey = item?.props.itemKey;
+      if (typeof itemKey === "string") {
+        navigate(`./#${itemKey}`);
```
kiootic, line 17: Please escape the fragment for safety.

kiootic (commented):
I tried checking/unchecking the `Activate` column in `Authenticator` tab, and it seems it doesn't count as modified. Is that intentional?

portal/src/hook/usePivot.ts
```diff
@@ -0,0 +1,24 @@
+import { useCallback } from "react";
+import { useLocation, useNavigate } from "react-router-dom";
+import { IPivotItemProps } from "@fluentui/react";
+
+export function usePivot(): {
+  hash: string;
```
kiootic, line 6: Please use a more specific name for this field (`selectedKey`?), since using URL fragment is an internal details of the hook.

A more descriptive name may be used for the hook too, e.g. `usePivotNavigation`.

axolotl0212:
@kiootic updated

portal/src/hook/usePivot.ts
```diff
@@ -14,11 +14,13 @@ export function usePivot(): {
     (item) => {
       const itemKey = item?.props.itemKey;
       if (typeof itemKey === "string") {
-        navigate(`./#${itemKey}`);
+        // NOTE: escape fragment for safety
+        // eslint-disable-next-line  no-useless-escape
+        navigate(`./\#${itemKey}`);
```
kiootic, line 19: I mean escape the fragment using percent encoding: https://stackoverflow.com/questions/13861273/how-should-fragment-part-of-a-url-be-escaped

portal/src/hook/usePivot.ts
axolotl0212: > I tried checking/unchecking the `Activate` column in `Authenticator` tab, and it seems it doesn't count as modified. Is that intentional?

This is not expected, after investigation this is cause by mutating the state in setState directly also mutate initial state as they shared same reference

portal/src/hook/usePivot.ts
axolotl0212: I see, I misunderstood your previous comment, updated to using RFC3986

axolotl0212:
When reading doc for RFC3986, found out that my implementation was wrong, fixed

portal/src/util/escapeURI.ts
```diff
@@ -0,0 +1,3 @@
+export function encodeURIComponentRFC3986(str: string): string {
```
kiootic, line 1: Just `encodeURIComponent` is enough practically.

axolotl0212: updated to using encodeURIComponent only @kiootic

kiootic (approved):

PR #419, opened 2020-09-29
Fix authenticator list display

### LINKED ISSUE #417: [Portal] Primary OOB authenticator not showing up correctly
The email/phone is missing in each authenticator row. It seems it's looking at incorrect authenticator claims.

kiootic:
fix #417 

@axolotl0212 Please note the correct authenticator claims to use in case other screens need to use it. Authenticator claims are different from identity claims.


axolotl0212 (approved):
Thanks, didn't realise there is a separate set of claim key for authenticators.

PR #420, opened 2020-09-29
Implement authentication login id phone number widget

### LINKED ISSUE #390: Implement loginID phone number widget
Design for phone number widget is updated

- [x] Country name translation -> i18n-iso-countries
- [x] Phone number country code to region code mapping
  - [x] script for generate hard coded map => google libphonenumber (getRegionCodesForCountryCode)
  - [x] use hard coded map for getting country code with calling code
- [x] Create list with selection and ordering in only selected item
- [x] selected all -> checked, partially selected -> indeterminate, none selected -> unchecked
- [x] Implement search box (local filtering, only filter unselected)
- [x] Implement search logic and filtering
- [x] Study mapping between config and UI state
- [x] Handle change config mutation

axolotl0212:
Closes #390

# Screenshots
Screenshot 2020-09-29 at 7 36 03 PM
Screenshot 2020-09-29 at 7 35 44 PM
Screenshot 2020-09-29 at 7 35 28 PM


portal/src/graphql/portal/AuthenticationLoginIDSettings.tsx
```diff
@@ -1,5 +1,6 @@
-import React, { useContext, useMemo, useState } from "react";
+import React, { useCallback, useContext, useMemo, useState } from "react";
 import produce from "immer";
+import { WritableDraft } from "immer/dist/internal";
```
kiootic, line 3: Avoid importing from internal files. I suppose for our purpose, `WritableDraft<T>` should be equivalent to `T`?

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
```diff
@@ -0,0 +1,490 @@
...
+}
+
+interface CountryCallingCodeListSelectAllProps extends ICheckboxProps {
+  isPartiallySelected: boolean;
+  isAllSelected: boolean;
+  selectAll: () => void;
+  unselectAll: () => void;
+}
+
+const HEADER_STYLE = {
```
kiootic, line 65: Use `checkboxVisibility` prop of `DetailsList` instead of custom styles.

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
```diff
@@ -0,0 +1,490 @@
...
+  check: {
+    width: "35px !important",
+    opacity: "1 !important",
+    paddingLeft: "15px !important",
+  },
+};
+
+const SEARCH_BOX_STYLE = {
```
kiootic, line 73: Why use custom styles instead of setting it in `searchBox` CSS class?

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
```diff
@@ -0,0 +1,490 @@
...
+function constructCallingCodeListData(
+  allCountryCallingCodes: string[],
+  getTelecomCountryName: (code: string) => string
+): CountryCallingCodeListData {
+  const callingCodeListData: CountryCallingCodeListData = {};
+  for (const callingCode of allCountryCallingCodes) {
+    const countryCodes =
+      countryCallingCodeMap[callingCode as keyof typeof countryCallingCodeMap];
+    const countryName = countryCodes
+      .map((code) => getTelecomCountryName(code))
```
kiootic, line 143: Let's just use the first country code, otherwise for some entries it gets too long.

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
```diff
@@ -0,0 +1,490 @@
...
+  let rawUnselectedCodeList: string[] = inputUnselectedCountryCallingCodes;
+  if (searchString.trim() !== "") {
+    const matchedCallingCodeItems = fuseSearcher.search(searchString);
+    const matchedUnselectedCodes = matchedCallingCodeItems
+      .filter((item) => !selectedCountryCallingCodeSet.has(item.callingCode))
+      .map((item) => item.callingCode);
+    rawUnselectedCodeList = matchedUnselectedCodes;
+  }
+
+  const unselectedCodeList = rawUnselectedCodeList.sort((code1, code2) => {
```
kiootic, line 175: FYI, a more idiomatic way to write comparer function: `(code1, code2) => Number(code1) - Number(code2)`.

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
```diff
@@ -0,0 +1,490 @@
...
+          })}
+        />
+      );
+    },
+    [selectedCountryCallingCodes]
+  );
+
+  return (
+    <section className={styles.root}>
+      <SearchBox
```
kiootic, line 470: For this kind of search box, it is more intuitive to search on text change.

kiootic (commented):
Can you make the code list scrollable? Currently the list is very long.

portal/src/graphql/portal/AuthenticationLoginIDSettings.tsx
axolotl0212: Using type without `WritableDraft` result in type error `not assignable to type Record<string, unknown>`, `index signature is missing in type ...` in one of the function mutating the fields

axolotl0212: Object with interface defined is different from general `Object`?

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
axolotl0212: I followed the example in Fluent UI doc, I think this can be achieved with CSS class

portal/src/graphql/portal/AuthenticationLoginIDSettings.tsx
kiootic: The function `setFieldIfListNonEmpty` has a signature too general, we can correct it to this:
```ts
function setFieldIfListNonEmpty<T, K extends keyof T>(
  map: T,
  field: K,
  list: T[K] extends unknown[] | undefined ? T[K] : never,
)
```

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
axolotl0212: We can use the first country code for now. But I think we need to come up with a better solution. For example, country code 1 is used by both US and Canada and other areas, only showing one could be confusing.

Also, country code 42 is for Czechoslovakia, I think it should be removed from backend + portal, cannot get country name as the country itself no longer exist.

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
axolotl0212: Thanks, this looks much cleaner

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
axolotl0212: Sure, the performance is OK with search on text change.

axolotl0212:
@kiootic rebased and updated

axolotl0212:
pushed 1 more commit to make list scrollable

portal/src/graphql/portal/AuthenticationLoginIDSettings.tsx
axolotl0212: Thanks, it works

axolotl0212:
updated type definition

portal/src/graphql/portal/AuthenticationCountryCallingCodeList.tsx
kiootic: Sure, please help delete it from `country_calling_code.go` as well.

axolotl0212: pushed 1 more commit to remove it

kiootic (commented):
When I scroll the list to the middle, it seems the items are not displayed until I resized the window.

axolotl0212:
> When I scroll the list to the middle, it seems the items are not displayed until I resized the window.

Found same problem, investigating

axolotl0212:
Fixed list not loading all items and backend test case

kiootic:
The issue you linked (microsoft/fluentui#10267) seems to be a user error. Instead, how about try to use `ScrollablePane` instead?

ref: https://developer.microsoft.com/en-us/fluentui/#/controls/web/scrollablepane

axolotl0212:
> The issue you linked ([microsoft/fluentui#10267](https://github.com/microsoft/fluentui/issues/10267)) seems to be a user error. Instead, how about try to use `ScrollablePane` instead?
> 
> ref: https://developer.microsoft.com/en-us/fluentui/#/controls/web/scrollablepane

The problem persist after using ScrollablePane, seems like making the list scrollable makes the list render item in chunk by default, but it fails to render new item when we scroll to end of page for some reason, the `onShouldVirtualize` props disable this behaviour

axolotl0212:
@kiootic updated

kiootic (approved):

PR #424, opened 2020-09-30
Layout OAuth client list screen

### LINKED ISSUE #423: [Portal] OAuth client list screen
- [x] study mapping from config
- layout List:
  - [x] Detail List with 
     - Name 
     - Creation date
     - Application (type, "android" | "ios" | "web-app")
     - Client ID
   - [x] client id button with edit button and copy button (click to copy to clipboard)
   - [x] Primary Button for add client button (Link to add client page)

axolotl0212:
Closes #423 

# Screenshots
Screenshot 2020-09-30 at 2 51 17 PM

Screenshot 2020-09-30 at 2 52 01 PM

### DIFF
--- a/portal/src/locale-data/en.json
+++ b/portal/src/locale-data/en.json
@@ -238,6 +238,7 @@
   "OAuthClientConfiguration.client-list.creation-date": "Creation Date",
   "OAuthClientConfiguration.client-list.application-type": "Application",
   "OAuthClientConfiguration.client-list.client-id": "Client ID",
+  "OAuthClientConfiguration.client-id-copied": "Client ID is copied to clipboard",
   "OAuthClientConfiguration.application-type.ios": "iOS",
   "OAuthClientConfiguration.application-type.android": "Android",
   "OAuthClientConfiguration.application-type.web-app": "Web Application",
--- a/portal/src/locale-data/en.json
+++ b/portal/src/locale-data/en.json
@@ -50,7 +50,7 @@
   "ScreenNav.anonymous-users": "Anonymous Users",
   "ScreenNav.single-sign-on": "Single-Sign On",
   "ScreenNav.passwords": "Passwords",
-  "ScreenNav.oauth-clients": "OAuth Clients",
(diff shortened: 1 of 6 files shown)


portal/src/locale-data/en.json
```diff
@@ -49,6 +49,7 @@
   "ScreenNav.anonymous-users": "Anonymous Users",
   "ScreenNav.single-sign-on": "Single-Sign On",
   "ScreenNav.passwords": "Passwords",
+  "ScreenNav.oauth-clients": "OAuth Clients",
```
kiootic, line 52: IIRC this should be changed to `Client Applications`.

kiootic (commented):
I suppose we should use `MessageBar` instead of a custom toast component.

axolotl0212:
> I suppose we should use `MessageBar` instead of a custom toast component.

Just show message bar on top for a few seconds?

portal/src/locale-data/en.json
axolotl0212: Also need to change title of the screen?

kiootic: Yes.

kiootic:
> Just show message bar on top for a few seconds?

Just providing a close button is enough.

axolotl0212:
@kiootic updated to only using message bar and the displayed name of OAuth clients page

portal/src/Notification.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { MessageBar } from "@fluentui/react";
+
+interface NotificationProps {
+  onDismiss: () => void;
+  visible: boolean;
+  msg: string;
+}
+
+export function useNotification(): {
+  showNotification: (msg: string, delay?: number) => void;
```
kiootic, line 11: `delay` parameter is no use? In fact, this hook & component seems rather trivial, so I would prefer to just include it in the parent screen.

axolotl0212: Moved to parent screen

kiootic (approved):

PR #434, opened 2020-10-05
Add oauth client screen

### LINKED ISSUE #426: [Portal] Add OAuth client screen
- [x] create modify oauth client form (reuse in edit screen)
     - application type => 3 button selection with icon and description => Choice group 
        - customize option => onRenderField
     - name => text field
     - access token lifetime, refresh token lifetime => number field => spin button, no decimal
     - redirect URIs => picker
- [x] handle state and onSave
- [x] validation
       - redirect uris necessary field, at least 1 item
- [x] client ID generated
- [x] dialog on create success, with button for copying new client ID
- [x] block navigation if form dirty

axolotl0212:
Closes #426

# Screenshot

Screenshot 2020-10-06 at 12 15 01 AM
Screenshot 2020-10-06 at 12 14 44 AM
Screenshot 2020-10-06 at 1 10 28 AM

### DIFF
--- a/portal/src/graphql/portal/CreateOAuthClientScreen.tsx
+++ b/portal/src/graphql/portal/CreateOAuthClientScreen.tsx
@@ -167,7 +167,6 @@ const CreateOAuthClientForm: React.FC<CreateOAuthClientFormProps> = function Cre
           onCreateClientSuccess();
         }
       })
-      // TODO: handle error
       .catch(() => {});
   }, [rawAppConfig, clientConfig, onCreateClientSuccess, updateAppConfig]);
 
@@ -186,6 +185,7 @@ const CreateOAuthClientForm: React.FC<CreateOAuthClientFormProps> = function Cre
         className={styles.modifyClientForm}
         clientConfig={clientConfig}
         onClientConfigChange={onClientConfigChange}
+        updateAppConfigError={updateAppConfigError}
       />
       <ButtonWithLoading
         onClick={onCreateClick}
--- a/portal/src/graphql/portal/CreateOAuthClientScreen.tsx
+++ b/portal/src/graphql/portal/CreateOAuthClientScreen.tsx
@@ -135,6 +135,8 @@ const CreateOAuthClientForm: React.FC<CreateOAuthClientFormProps> = function Cre
     initialState
   );
 
+  const [submittedForm, setSubmittedForm] = useState(false);
+
   const [
     createClientSuccessDialogVisible,
     setCreateClientSuccessDialogVisible,
@@ -148,6 +150,7 @@ const CreateOAuthClientForm: React.FC<CreateOAuthClientFormProps> = function Cre
   );
 
   const onCreateClientSuccess = useCallback(() => {
+    setSubmittedForm(true);
     setCreateClientSuccessDialogVisible(true);
   }, []);
 
@@ -176,7 +179,9 @@ const CreateOAuthClientForm: React.FC<CreateOAuthClientFormProps> = function Cre
 
   return (
     <form className={styles.form}>
-      <NavigationBlockerDialog blockNavigation={isFormModified} />
+      <NavigationBlockerDialog
+        blockNavigation={!submittedForm && isFormModified}
+      />
       <CreateClientSuccessDialog
         visible={createClientSuccessDialogVisible}
         clientId={clientConfig.client_id}
--- a/portal/src/graphql/portal/CreateOAuthClientScreen.tsx
+++ b/portal/src/graphql/portal/CreateOAuthClientScreen.tsx
@@ -1,27 +1,181 @@
-import React, { useMemo, useState } from "react";
-import { useParams } from "react-router-dom";
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
+import {
+  Callout,
+  Dialog,
+  DialogFooter,
+  DirectionalHint,
+  IconButton,
+  Label,
+  PrimaryButton,
+  Text,
+} from "@fluentui/react";
+import { useNavigate, useParams } from "react-router-dom";
 import produce from "immer";
 import { FormattedMessage } from "@oursky/react-messageformat";
 
 import ShowError from "../../ShowError";
 import ShowLoading from "../../ShowLoading";
+import ModifyOAuthClientForm from "./ModifyOAuthClientForm";
+import ButtonWithLoading from "../../ButtonWithLoading";
 import NavBreadcrumb, { BreadcrumbItem } from "../../NavBreadcrumb";
 import { useAppConfigQuery } from "./query/appConfigQuery";
-import { PortalAPIAppConfig } from "../../types";
+import { useUpdateAppConfigMutation } from "./mutations/updateAppConfigMutation";
+import { OAuthClientConfig, PortalAPIAppConfig } from "../../types";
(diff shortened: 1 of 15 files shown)


louischan-oursky (commented):
1. I created a client. I clicked done. It prompts unsaved changes warning :( 
2. Remove the created at column in the list view.
3. Add a button in the list view to remove a client.
4. It does not seem to be clear that "Name" and "Redirect URIs" are required, while other are optional. Do we have a design to indicate required and optional fields?
5. When I mouse hover the info icon, the cursor is text selection cursor. It should be `pointer` cursor.

axolotl0212:
> I created a client. I clicked done. It prompts unsaved changes warning :(

Thanks, forgotten to allow navigation on submit

> Remove the created at column in the list view.
> Add a button in the list view to remove a client.

Will handle list updates in separate PR

> It does not seem to be clear that "Name" and "Redirect URIs" are required, while other are optional. Do we have a design to indicate required and optional fields?

Not yet, but I think we can follow fluent UI doc https://developer.microsoft.com/en-us/fluentui#/controls/web/textfield , using the `required` props

Screenshot 2020-10-06 at 12 05 21 PM

> When I mouse hover the info icon, the cursor is text selection cursor. It should be pointer cursor. 

Didn't spot this, will add `cursor: pointer` to icon

axolotl0212:
@louischan-oursky rebased and updated add oauth client screen

portal/src/graphql/portal/CreateOAuthClientScreen.tsx
```diff
@@ -0,0 +1,250 @@
...
+  visible: boolean;
+  clientId: string;
+}
+const CreateClientSuccessDialog: React.FC<CreateClientSuccessDialogProps> = function CreateClientSuccessDialog(
+  props: CreateClientSuccessDialogProps
+) {
+  const { visible, clientId } = props;
+  const navigate = useNavigate();
+
+  const calloutIntervalRef = useRef<number>();
```
kiootic, line 52: Why use `useRef` instead of `useState`?

portal/src/graphql/portal/CreateOAuthClientScreen.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const [isCalloutVisible, setIsCalloutVisible] = useState(false);
+
+  useEffect(() => {
+    if (calloutIntervalRef.current != null) {
```
kiootic, line 56: This should be ran in effect cleanup function instead, otherwise as it is now, it will run just once on component mount.

portal/src/graphql/portal/ModifyOAuthClientForm.tsx
```diff
@@ -0,0 +1,216 @@
...
+  // Number("") = 0
+  const accessTokenLifetimeSec = Number(accessTokenLifetime);
+  const refreshTokenLifetimeSec = Number(refreshTokenLifetime);
+
+  return {
+    client_id: clientId,
+    redirect_uris: redirectUris,
+    post_logout_redirect_uris:
+      postLogoutRedirectUris.length > 0 ? postLogoutRedirectUris : undefined,
+    grant_types: ["authorization_code", "refresh_token"],
```
kiootic, line 37: Sorry I missed one grant type for anonymous user: `urn:authgear:params:oauth:grant-type:anonymous-request`. Let's add that as well. cc @louischan-oursky

portal/src/graphql/portal/ModifyOAuthClientForm.tsx
```diff
@@ -0,0 +1,216 @@
...
+  return (
+    <section className={cn(styles.root, className)}>
+      <TextField
+        className={styles.inputField}
+        label={renderToString("ModifyOAuthClientForm.name-label")}
+        value={clientName}
+        onChange={onClientNameChange}
+        required={true}
+      />
+      <TextField
```
kiootic, line 165: Can we use `type="number"` input field for lifetime inputs?

portal/src/util/random.ts
```diff
@@ -0,0 +1,10 @@
+export function genRandomHexadecimalString(length: number = 16): string {
+  let result = "";
+  if (length < 0) {
```
kiootic, line 3: This condition is redundant.

portal/src/graphql/portal/CreateOAuthClientScreen.tsx
axolotl0212: This is because `calloutIntervalRef` is only used for cleanup, UI does not depend on this variable

portal/src/graphql/portal/CreateOAuthClientScreen.tsx
axolotl0212: missed `return`, thanks

portal/src/graphql/portal/ModifyOAuthClientForm.tsx
axolotl0212: I considered using `type="number"` but there is a several problems.

As `type="number"` allows symbol like `.`, `-`, so I think intercepting the onChange function is simpler, avoiding the need to do extra validation to ensure number is positive whole number. 

Also, if I add `type="number"`, then onChange is fired only if `Number()` of value is not `NaN`, so user will be able to input something like `------` and the state will not change.

portal/src/util/random.ts
axolotl0212: so the user is responsible to input positive length? Let's remove the condition

portal/src/graphql/portal/CreateOAuthClientScreen.tsx
kiootic: As it is now, there could be multiple active timeout interval, causing glitchy UI when tapping copy button repeatedly. Also, I think dismissing the callout after a timeout is counted as a UI concern.

I'd prefer using the `isCalloutVisible` variable to drive `setTimeout` side effect and remove the `useRef`, so the timeout effect is contained entirely within a single useEffect call:
```ts
const [calloutActiveCount, setCalloutActiveCount] = useState(0);
useEffect(() => {
    if (calloutActiveCount === 0) {
        return;
    }

    setIsCalloutVisible(true);
    const handle = setTimeout(() => setIsCalloutVisible(false), 3000);
    return () => clearTimeout(handle);
}, [calloutActiveCount]);

const onCopyClick = useCallback(() => {
    copyToClipboard(clientId);
    setCalloutActiveCount(c => c + 1);
}, [clientId]);
```

Optionally, you can also set the `key` prop of `Callout` to be `calloutActiveCount` to replay the animation on tap.

portal/src/util/random.ts
kiootic: I mean if the length is negative, then loop would not be entered and the result is the same, so there's no need to check it.

portal/src/graphql/portal/ModifyOAuthClientForm.tsx
kiootic: Okay I see. I suppose we can keep using text type input for now.

portal/src/graphql/portal/CreateOAuthClientScreen.tsx
axolotl0212: I see, when repeatedly click copy, the callout will be set to invisible by first timeout, the code above fixes the problem

axolotl0212:
@kiootic updated

kiootic (approved):

PR #438, opened 2020-10-06
Edit OAuth client screen

### LINKED ISSUE #437: [Portal] Layout Edit OAuth Client Screen
- [x] Reuse modify OAuth client form
- [x] Add client ID field, read-only on top
- [x] pre-fill form with existing setting

axolotl0212:
base on #434

Closes #437

- reuse modify oauth client form

# Screenshot

Screenshot 2020-10-06 at 4 56 51 PM

### DIFF
--- a/portal/src/graphql/portal/EditOAuthClientScreen.tsx
+++ b/portal/src/graphql/portal/EditOAuthClientScreen.tsx
@@ -1,15 +1,9 @@
-import React, {
-  useCallback,
-  useContext,
-  useEffect,
-  useMemo,
-  useState,
-} from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
 import { useNavigate, useParams } from "react-router-dom";
 import deepEqual from "deep-equal";
 import produce from "immer";
-import { Text, TextField } from "@fluentui/react";
-import { Context, FormattedMessage } from "@oursky/react-messageformat";
+import { Label, Text } from "@fluentui/react";
+import { FormattedMessage } from "@oursky/react-messageformat";
 
 import NavBreadcrumb, { BreadcrumbItem } from "../../NavBreadcrumb";
 import ShowError from "../../ShowError";
@@ -33,7 +27,6 @@ const EditOAuthClientForm: React.FC<EditOAuthClientFormProps> = function EditOAu
   props: EditOAuthClientFormProps
 ) {
   const { clientConfig: clientConfigProps, rawAppConfig } = props;
-  const { renderToString } = useContext(Context);
   const { appID } = useParams();
   const navigate = useNavigate();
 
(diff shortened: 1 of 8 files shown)


portal/src/graphql/portal/EditOAuthClientScreen.tsx
```diff
@@ -0,0 +1,190 @@
...
+      .then((result) => {
+        if (result != null) {
+          setSubmittedForm(true);
+        }
+      })
+      .catch(() => {});
+  }, [clientConfig, updateAppConfig, rawAppConfig]);
+
+  useEffect(() => {
+    if (submittedForm) {
```
louischan-oursky, line 89: It seems that the only use of `submittedForm` is here. Why don't we just call `navigate` in line 82? Using an useEffect in this way sounds a bit indirect.

louischan-oursky (commented):
The client ID is not editable, which is correct. But it is not visually read-only. https://developer.microsoft.com/en-us/fluentui#/controls/web/textfield I see Fluent does have a disabled state. Can we use it?

axolotl0212:
> The client ID is not editable, which is correct. But it is not visually read-only. https://developer.microsoft.com/en-us/fluentui#/controls/web/textfield I see Fluent does have a disabled state. Can we use it?

Sure, can add disabled={true}. I used readOnly props simply because this is what we want to achieve. btw, using disabled the text will have less contrast.

portal/src/graphql/portal/EditOAuthClientScreen.tsx
axolotl0212: This is for making sure the state is set before calling navigate, as `setState` in functional component does not have callback @louischan-oursky

axolotl0212:
After using disabled props, it looks like this


Screenshot 2020-10-07 at 7 28 45 PM

louischan-oursky:
I just merged the add-`name` to client PR. Can you add support for it? Right now I cannot create new client :(

> After using disabled props, it looks like this

Maybe we should just render it with `Text`?

portal/src/graphql/portal/EditOAuthClientScreen.tsx
louischan-oursky: A second thought. I guess we should stay on the page even after saving.

axolotl0212:
> I just merged the add-`name` to client PR. Can you add support for it? Right now I cannot create new client :(

Sure, I just rebased, can quickly change it.

axolotl0212:
Render as `Text`

Screenshot 2020-10-07 at 7 43 04 PM

axolotl0212:
updated and added support for name field in config

louischan-oursky:
Right now when I click add client, the app will crash with calling `.trim()` on undefined. I guess this is related to our usage of `!`.

axolotl0212:
@louischan-oursky Fixed the problem, I forgotten to pass client name to the ModifyOAuthClientForm. Changed to `?? ""` to make it safer

PR #439, opened 2020-10-06
Handle verified claims in user detail screen

### LINKED ISSUE #385: [Portal] Handle verified claims in connected identities and add identities

axolotl0212:
Closes #385 

- get verified state from verified claims
- integrate set verified state mutation

### DIFF
--- a/portal/src/ButtonWithLoading.tsx
+++ b/portal/src/ButtonWithLoading.tsx
@@ -1,17 +1,23 @@
 import React, { useContext } from "react";
-import { IButtonProps, PrimaryButton, Spinner } from "@fluentui/react";
+import {
+  IButtonProps,
+  ISpinnerProps,
+  PrimaryButton,
+  Spinner,
+} from "@fluentui/react";
 import { Context, FormattedMessage } from "@oursky/react-messageformat";
 
 interface ButtonWithLoadingProps extends IButtonProps {
   loading: boolean;
   labelId: string;
   loadingLabelId?: string;
+  spinnerStyles?: ISpinnerProps["styles"];
 }
 
 const ButtonWithLoading: React.FC<ButtonWithLoadingProps> = function ButtonWithLoading(
   props: ButtonWithLoadingProps
 ) {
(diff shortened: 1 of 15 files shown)


kiootic (commented):
- I can trigger more than one simultaneous verify/unverify operations. It's better to disable all/relevant only buttons when verifying/unverifying.
- It seems the verify toggle in add email/phone screen is not yet removed, let's remove that as well.

axolotl0212:
> I can trigger more than one simultaneous verify/unverify operations. It's better to disable all/relevant only buttons when verifying/unverifying.

When I try to disable buttons when verifying, I ran into a problem. 

Currently the list items is rendered by passing a render function to `List` component in fluent UI. When I include the verifying state in render function, the list is not rerendered (still investigating). If I use key props to force the list to rerender, the loading state disappears as it is controlled by useState in the cell (unmount -> losing the loading state)

axolotl0212:
Will simply block pointer event to avoid simultaneous verify/unverify operations, cannot come up with a better solution

kiootic:
Some potential solutions:
- Add a `isVerifying` field in the list item model to propagate the state down to item cell component. (List should render the item cell iff the item changed)
- Use a React context to provide this flag (or other data/behavior) to descendant components.

Personally I prefer the 1st solution for simplicity.

axolotl0212:
> * Add a `isVerifying` field in the list item model to propagate the state down to item cell component. (List should render the item cell iff the item changed)
> * Use a React context to provide this flag (or other data/behavior) to descendant components.

I picked second one, it works perfectly fine. I actually tried the first one before asking, but the item is not updated when the state changes. I expect the render function gets re-evaluated and list will be updated but that's not the case.

louischan-oursky (approved):

kiootic (approved):

PR #443, opened 2020-10-07
Add OAuth client name for developer use

### LINKED ISSUE #440: Store OAuth client display name in config
This display name is for developer use only (the user-facing name should be localized in `translation.json`).

ref #426

kiootic:
fix #440 

@axolotl0212 Please note the new field in OAuth client config.


axolotl0212 (approved):
If name field is optional, then for in portal if name does not exist, we simply put a placeholder in the OAuth client list?

kiootic:
On a second thought, let's make it required to simplify logics. So developer is required to set three fields when creating a client: name, client ID (auto-generated in portal), redirect URIs.

axolotl0212 (approved):

PR #447, opened 2020-10-07
Add infrastructure for portal backend database mechanism

### LINKED ISSUE #445: Setup portal database infrastructure

kiootic:
fix #445


louischan-oursky (commented):
So the name of Authgear tables start with `_auth` and the name of portal tables start with another prefix? So we can have two migration tracking table in the same schema?

kiootic:
> So the name of Authgear tables start with _auth and the name of portal tables start with another prefix? So we can have two migration tracking table in the same schema?

Yes, that's the intention, though locally I used another schema for portal.

Do you prefer having a prefix for portal tables, or omit the prefix?

louischan-oursky:
> Do you prefer having a prefix for portal tables, or omit the prefix?

Use the prefix `_portal_`.

PR #450, opened 2020-10-07
Fix leaving min length field empty in password policy page unhandled error

axolotl0212:
# Error screenshot

Screenshot 2020-10-07 at 5 24 12 PM

## Cause 
parseInt("", 10) = NaN, NaN is not valid number in authgear config, leads to unhandled error

### DIFF
--- a/portal/src/graphql/portal/PasswordPolicySettings.tsx
+++ b/portal/src/graphql/portal/PasswordPolicySettings.tsx
@@ -17,6 +17,7 @@ import ToggleWithContent from "../../ToggleWithContent";
 import ButtonWithLoading from "../../ButtonWithLoading";
 import NavigationBlockerDialog from "../../NavigationBlockerDialog";
 import {
+  setNumericFieldIfChanged,
   setFieldIfChanged,
   setFieldIfListNonEmpty,
   isArrayEqualInOrder,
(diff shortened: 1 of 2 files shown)


portal/src/graphql/portal/PasswordPolicySettings.tsx
```diff
@@ -63,7 +63,7 @@ function constructStateFromAppConfig(
   const historySize = passwordPolicy?.history_size ?? 0;
 
   return {
-    minLength: passwordPolicy?.min_length ?? 0,
+    minLength: passwordPolicy?.min_length ?? NaN,
```
louischan-oursky, line 66: I usually store numeric form fields as string. And only convert them to number on submission. Would this approach be simpler? For this particular form, `minLength`, `historyDays` and `historySize` should be stored as string.

axolotl0212: > I usually store numeric form fields as string. And only convert them to number on submission. Would this approach be simpler?

I agree, this is what I usually do as well. I keep it as number as the least changes has to be made. Can change it to string. Using NaN as empty is not a good idea anyway

axolotl0212:
rebased and pushed 1 more commit to change to store numeric form fields as string

portal/src/util/misc.ts
```diff
@@ -14,6 +14,21 @@ export function setFieldIfChanged<
   }
 }
 
+export function setNumericFieldIfChanged<M, K extends keyof M>(
+  map: Partial<M>,
+  key: K,
+  initialValue: M[K] extends number ? M[K] : never,
+  value: M[K] extends number ? M[K] : never
```
louischan-oursky, line 21: I think `initialValue` and `value` should be string. This avoids the caller to use `Number()` to cast the string into number, which does not always succeed.

axolotl0212: If the input value is string, then type of `map[key]` will be incompatible with `number`. Fix this by casting to `any` ?

axolotl0212: After discussion, decided to use `as any` as typescript failed to infer the type from 

```typescript
initialValue: M[K] extends number ? string : never,
value: M[K] extends number ? string : never
```

axolotl0212:
@louischan-oursky updated

PR #451, opened 2020-10-08
Fix create OAuth client screen block navigation on new form

axolotl0212:
Just found this problem, due to initial state does not match with state returned by ModifyOAuthClientForm


louischan-oursky:
Another related problem: If I enter the edit client screen, initially the save button should be disabled.

axolotl0212:
@louischan-oursky This could be caused by hardcoding `grant_type` and `response_type`. If you manually set those field, it is different from the hardcoded one, ie

```yaml
      grant_types:
        - authorization_code
        - refresh_token
        - 'urn:authgear:params:oauth:grant-type:anonymous-request'
      response_types:
        - code
        - none
```

axolotl0212:
updated

PR #455, opened 2020-10-08
Add configuration to disable public signup

### LINKED ISSUE #250: Disable public signup
Some application may want to disable public sign up. So the user must be created via Admin API by an existing user who has access.

The portal is an example of such application. No one should be able to sign up as a new portal user.

---

We have several approaches to the issue.

## Introduce dedicated configuration for each scenario

Given the following configuration, 

```yaml
identity:
  login_id:
    keys:
    - type: email
  oauth:
    providers:
    - alias: google
       type: google
```

Both login and signup screen will show "Sign in with Google" and a text field for the email.

We may introduce dedicated configuration to control which is shown on which screen.

```yaml
identity:
  login_id:
    keys:
    - type: email
       authentication:
       - allowed_in: ["login", "signup"]
  oauth:
    providers:
    - alias: google
       type: google
       authentication:
       - allowed_in: ["login"]
```

This approach has an advantage of deprecating unwanted identity. For example, the application used to support Facebook and Google login but the developer wishes force new user to use Google only. With this configuration, the developer can mark Facebook login allowed in login screen only.

The downside of this approach is that we can no longer merge login, signup and promote screen into one screen, since different screen has different semantics and different identity options.

## Introduce a simple boolean flag to disable public signup

If the flag is true, signup screen will show something like "Please contact the administrator for access"

louischan-oursky:
fix #250

### DIFF
--- a/pkg/admin/loader/interaction.go
+++ b/pkg/admin/loader/interaction.go
@@ -78,6 +78,10 @@ type createUserInput struct {
 	password    string
 }
 
+func (i *createUserInput) BypassPublicSignupDisabled() bool {
+	return true
+}
+
 func (i *createUserInput) SkipVerification() bool {
 	return true
 }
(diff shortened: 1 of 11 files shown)


kiootic (commented):
I think we should hide the `Don't have an account yet? Create one!` link as well?

louischan-oursky:
Fixed.

kiootic (approved):

Issue #392, opened 2020-09-25
Support custom domain


louischan-oursky:
## Imagined Scenario

Authgear is deployed on `authgearapps.com`. John is a developer working for `company.com`. John wants to use Authgear for his company, deployed at `accounts.company.com`.

Things to be resolved:

- How to provide the TLS certificate of `accounts.company.com`?
- How should `www.company.com` authenticate the request?

kiootic:
> How to provide the TLS certificate of accounts.company.com?

I suppose Let's Encrypt with HTTP01 challenge should work when using CNAME record.

> How should www.company.com authenticate the request?

I suppose you mean how would the application backend validate the credentials (e.g. OAuth access token, cookies) in the request. I suppose we may expose the resolver endpoint at a random sub-path of the domain, in case we don't want public access to it? (e.g. `https://example-app.authgear.com/_resolve/4A102642-C16A-493C-A15C-6D239D40CC4F`)

kiootic:
- Remove `hosts.json`, replace with per-app ingresses
    - Use wildcard cert for SaaS deployment
    - Use per-app ingress? or wildcard ingress. (investigate cert-only Let's Encrypt cert-manager)
- verified domain state
    - add database to portal backend
    - two tables, one for pending (no unique), for verified domain (with unique constraint on eTLD + 1 & app)
    - need careful safety check for data consistency
    - provide TXT record & CNAME record
    - only one custom domain per app; only one authgear per eTLD + 1
- TLS cert
    - wildcard cert: cert-manager with Let's Encrypt DNS validation (cert-only, no ingress)
    - per-app custom cert: cert-manager with Let's Encrypt HTTP validation (automatic with ingress)
- Resolver endpoint (TBD)

PR #464, opened 2020-10-08
Create ingresses per app

kiootic:
ref #392


pkg/lib/config/configsource/kubernetes.go
```diff
@@ -190,19 +191,39 @@ func (k *Kubernetes) Close() error {
 }
 
 func (k *Kubernetes) AllAppIDs() ([]string, error) {
-	appIDs := k.appIDs.Load().([]string)
+	// FIXME(k8s): remove this after introducing proper authz
+	ingresses, err := k.Client.NetworkingV1beta1().Ingresses(k.Namespace).List(metav1.ListOptions{})
```
louischan-oursky: I guess we can pass a label selector to `List` to make the list result more relevant?

pkg/lib/config/configsource/kubernetes.go
```diff
@@ -190,19 +191,39 @@ func (k *Kubernetes) Close() error {
...
+	}
+	sort.Strings(appIDs)
+
 	return appIDs, nil
 }
 
 func (k *Kubernetes) ResolveAppID(r *http.Request) (string, error) {
 	host := httputil.GetHost(r, bool(k.TrustProxy))
-	hostMap := k.hostMap.Load().(map[string]string)
+	if h, _, err := net.SplitHostPort(host); err == nil {
```
louischan-oursky, line 224: Any particular reason why do we need to ignore the port?

pkg/portal/config/app.go
```diff
@@ -1,15 +1,76 @@
...
+	}
+}
+
 type AppKubernetesConfig struct {
-	NewResourcePrefix   string `envconfig:"NEW_RESOURCE_PREFIX" default:"app-"`
-	IngressTemplateFile string `envconfig:"INGRESS_TEMPLATE_FILE"`
+	NewResourcePrefix    string `envconfig:"NEW_RESOURCE_PREFIX" default:"app-"`
+	IngressTemplateFile  string `envconfig:"INGRESS_TEMPLATE_FILE"`
+	DefaultDomainTLSCert string `envconfig:"DEFAULT_DOMAIN_TLS_CERT" default:"none"`
+	CustomDomainTLSCert  string `envconfig:"CUSTOM_DOMAIN_TLS_CERT" default:"none"`
```
louischan-oursky, line 73: Instead of encoding the value as colon separated, can we use a struct?

```golang
type AppKubernetesConfig struct {
  DefaultDomainTLSCert TLSCertConfig string `envconfig:"DEFAULT_DOMAIN_TLS_CERT"`
}

type TLSCertConfig struct {
  Type string `envconfig:"TYPE" default:"none"`
  
  SecretName string `envconfig:"SECRET_NAME"`
  
  IssuerKind string `envconfig:"ISSUER_KIND"`
  IssuerName string `envconfig:"ISSUER_NAME"`
}
```

louischan-oursky:
Can we provide a default template for ingress for development?

pkg/lib/config/configsource/kubernetes.go
kiootic: Kubernetes Ingress does not support routing on request host port.

pkg/lib/config/configsource/kubernetes.go
kiootic: This is just a temporary fix for the lack of proper authz, and should be removed after a proper authz mechanism is in place, so I don't think it need to be very performant.

louischan-oursky: Fair enough.

kiootic:
@louischan-oursky updated.

PR #465, opened 2020-10-08
Block dismiss dialog while loading

### LINKED ISSUE #444: [Portal] Dialog should not be dismissed while API request is in progress
Many of the dialogs can be dismissed by tapping the background overlay. Usually when a API request is triggered from the dialog, the primary button is set to loading state and the dialog is kept open. However, user can still dismiss the dialog through tapping the background overlay, and the API request would still be performed in background. We should disable this way of dismissal while API request is in progress.

axolotl0212:
fix #444

### DIFF
--- a/portal/src/graphql/adminapi/UserDetailsAccountSecurity.tsx
+++ b/portal/src/graphql/adminapi/UserDetailsAccountSecurity.tsx
@@ -321,7 +321,7 @@ const RemoveConfirmationDialog: React.FC<RemoveConfirmationDialogProps> = functi
     deletingAuthenticator,
     authenticatorID,
     authenticatorName,
-    onDismiss,
+    onDismiss: onDismissProps,
   } = props;
 
   const { renderToString } = useContext(Context);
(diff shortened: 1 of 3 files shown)


portal/src/graphql/adminapi/UserDetailsAccountSecurity.tsx
```diff
@@ -350,15 +350,15 @@ const RemoveConfirmationDialog: React.FC<RemoveConfirmationDialogProps> = functi
     <Dialog
       hidden={!visible}
       dialogContentProps={removeConfirmDialogContentProps}
-      onDismiss={onDismiss}
+      onDismiss={deletingAuthenticator ? undefined : onDismiss}
```
louischan-oursky, line 353: I think we should also set `modalProps` to make the dialog blocking. https://developer.microsoft.com/en-us/fluentui#/controls/web/dialog

Instead of conditionally passing `onDismiss`, can we check `deletingAuthenticator` in `onDismiss` to determine whether the dialog can be dismissed

Can we make sure all usage of Dialog follow this new convention?

axolotl0212: > I think we should also set modalProps to make the dialog blocking.

Then clicking background cannot dismiss the dialog. Only block when loading?

> Instead of conditionally passing onDismiss, can we check deletingAuthenticator in onDismiss to determine whether the dialog can be dismissed

If we do so, we do not need to set modalProps?

louischan-oursky: > Then clicking background cannot dismiss the dialog. Only block when loading?

Yes, only block the dialog from dismissal while loading.

> If we do so, we do not need to set modalProps?

Technically you are correct, but let's do both to make our intention clear.

axolotl0212:

kiootic (approved):

louischan-oursky (approved):

PR #472, opened 2020-10-09
Disallow navigate to same page

### LINKED ISSUE #427: [Portal] Erroneous confirmation dialog on navigation to same page
When form is modified and the same tab is selected in the nav bar, a confirmation dialog would appear but 'Confirm' button is not working.

axolotl0212:
fix #427

### DIFF
--- a/portal/src/ScreenNav.tsx
+++ b/portal/src/ScreenNav.tsx
@@ -11,9 +11,13 @@ function getAppRouterPath(location: Location) {
 }
 
 function getPath(url: string) {
-  // remove leading trailing slash, remove fragment
-  const regexpMatch = /^\/?([^#]*[^/#])\/?(#.*)?$/.exec(url);
-  return regexpMatch ? regexpMatch[1] : null;
+  // remove fragment
+  const path = new URL("scheme:" + url).pathname;
+  // remove leading trailing slash
+  const pathWithoutLeadingTrailingSlash = path
+    .replace(/^\//, "")
+    .replace(/\/$/, "");
+  return pathWithoutLeadingTrailingSlash;
 }
 
 function isPathSame(url1: string, url2: string) {
--- a/portal/src/ScreenNav.tsx
+++ b/portal/src/ScreenNav.tsx
@@ -2,6 +2,25 @@ import React, { useMemo, useCallback, useContext } from "react";
 import { useNavigate, useLocation } from "react-router-dom";
 import { Context } from "@oursky/react-messageformat";
 import { Nav, INavLink, INavLinkGroup, INavProps } from "@fluentui/react";
+import { Location } from "history";
+
+function getAppRouterPath(location: Location) {
+  // app router -> /app/:appID/*
+  // discard first 3 segment (include leading slash)
+  return location.pathname.split("/").slice(3).join("/");
+}
+
+function getPath(url: string) {
+  // remove leading trailing slash, remove fragment
+  const regexpMatch = /^\/?([^#]*[^/#])\/?(#.*)?$/.exec(url);
+  return regexpMatch ? regexpMatch[1] : null;
+}
+
+function isPathSame(url1: string, url2: string) {
+  const path1 = getPath(url1);
+  const path2 = getPath(url2);
+  return path1 === path2;
+}
 
 const ScreenNav: React.FC = function ScreenNav() {
   const navigate = useNavigate();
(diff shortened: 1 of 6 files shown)


portal/src/ScreenNav.tsx
```diff
@@ -2,6 +2,25 @@ import React, { useMemo, useCallback, useContext } from "react";
...
+  // app router -> /app/:appID/*
+  // discard first 3 segment (include leading slash)
+  return location.pathname.split("/").slice(3).join("/");
+}
+
+function getPath(url: string) {
+  // remove leading trailing slash, remove fragment
+  const regexpMatch = /^\/?([^#]*[^/#])\/?(#.*)?$/.exec(url);
+  return regexpMatch ? regexpMatch[1] : null;
+}
```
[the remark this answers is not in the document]
louischan-oursky: We should use the URL constructor directly.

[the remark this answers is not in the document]
axolotl0212: Tried using URL constructor, but it does not solve the leading / trailing slash inconsistency, take the example

```javascript
new URL("about:"+"/abc/def#123").pathname = "/abc/def"
new URL("about:"+"/abc/def/#123").pathname = "/abc/def/"
new URL("about:"+"abc/def#123").pathname = "abc/def"
```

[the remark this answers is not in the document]
louischan-oursky: I mean you should use URL constructor to convert a string to an instance of URL, and then use the pathname property to get the path, after that, apply your remove-trailing-slash logic on pathname.

axolotl0212:

portal/src/ScreenNav.tsx
```diff
@@ -2,6 +2,27 @@ import React, { useMemo, useCallback, useContext } from "react";
...
+  // app router -> /app/:appID/*
+  // discard first 3 segment (include leading slash)
+  return location.pathname.split("/").slice(3).join("/");
+}
+
+function getPath(url: string) {
+  // remove fragment
+  const path = new URL("scheme:" + url).pathname;
+  // remove leading trailing slash
+  const regexpMatch = /^\/?(.*[^/])\/?$/.exec(path);
```
louischan-oursky, line 17: `/^\/?(.*[^/])\/?$/.exec("/")` is null.

I would write

```typescript
// "/" would become empty string
// Is this result we are expecting?
const pathWithoutLeadingTrailingSlash = path.replace(/^\//, "").replace(/\/$/, "")
```

axolotl0212: > "/" would become empty string
Is this result we are expecting?

I think we won't have "/" in navigation bar item, and removing slash is for consistency in comparison, so I think "/" becomes empty string is OK (it will work even if the path is "/"). 

> const pathWithoutLeadingTrailingSlash = path.replace(/^\//, "").replace(/\/$/, "")

This looks much nicer

axolotl0212:
updated

portal/src/hook/usePivot.ts
```diff
@@ -1,24 +1,28 @@
...
   const location = useLocation();
   const hash = location.hash.slice(1);
 
   const onLinkClick = useCallback(
-    (item) => {
+    (item?: { props: IPivotItemProps }) => {
       const itemKey = item?.props.itemKey;
       if (typeof itemKey === "string") {
-        navigate(`./#${encodeURIComponent(itemKey)}`);
+        if (itemKey !== pivotRef.current?.state.selectedKey) {
```
louischan-oursky, line 19: So the `selectedKey` from the argument is different from `pivotRef.current?.state.selectedKey`?  The `ref.state.selectedKey` seems very hacky to me. It looks like accessing the internal state of the Pivot.

axolotl0212: > So the selectedKey from the argument is different from pivotRef.current?.state.selectedKey?

Yes, when we click cancel in navigation blocker dialog, `selectedKey` state is updated (unexpected) even if hash (`selectedKey` from the argument) stays the same, and the `PivotItem` is unchanged as well (expected)

> It looks like accessing the internal state of the Pivot.

It is the internal state of Pivot. But I am not sure why the internal state does not match with the blue underline style and the rendered PivotItem

axolotl0212:
Re-written the first commit to solve the problem @louischan-oursky

nicolaslee2775 (approved):

PR #473, opened 2020-10-09
Fix button in loading state behaviour

### LINKED ISSUE #448: [Portal] Buttons in loading state can be interacted with
A button in loading state can still be tapped, triggering multiple same request.

axolotl0212:
fix #448

### DIFF
--- a/portal/src/ButtonWithLoading.tsx
+++ b/portal/src/ButtonWithLoading.tsx
@@ -4,6 +4,7 @@ import {
   ISpinnerProps,
   PrimaryButton,
   Spinner,
+  SpinnerSize,
 } from "@fluentui/react";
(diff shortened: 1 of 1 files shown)


kiootic (commented):
How about just setting `disabled={true}` when it is loading? In my opinion it is quite confusing to see a button that looks interactive but actually not.

axolotl0212:
> How about just setting `disabled={true}` when it is loading? In my opinion it is quite confusing to see a button that looks interactive but actually not.

Will look like this, I think we can set button to disabled

Screenshot 2020-10-09 at 6 19 00 PM

axolotl0212:
@kiootic  updated

louischan-oursky (approved):

Issue #482, opened 2020-10-12
[UI] Create user interface screen


nicolaslee2775:
- [x] Create UserInterfaceScreen
- [x] Add UI components
  - [x] Custom css: `ui.custom_css`
  - [ ] ~~Supported Languages: `?`~~
  - [ ] ~~Default country / area code: `ui.country_calling_code.default`~~
- [x] Block navigation if there are unsaved changes
- [x] Integrate data query
- [x] Integrate data mutation

PR #486, opened 2020-10-12
Implement user interface screen

nicolaslee2775:
connect #482 

The `Supported Languages` and `Default country / area code` fields are removed.

image


axolotl0212:
> The Supported Languages and Default country / area code fields are removed.

There is an issue to follow update portal (#390)

axolotl0212 (approved):

Issue #478, opened 2020-10-12
Improve settings page


kiootic:
- MFA
    - TOTP
        - List
        - Add
        - Remove
    - OOB-OTP
        - List
        - Add
        - Remove
    - Additional Password
        - Setup
        - Change
    - Recovery code
        - List
        - Regenerate
    - Device tokens
        - Revoke

PR #488, opened 2020-10-13
Implement MFA section details

kiootic:
ref #478


louischan-oursky:
LGTM. When are we going to address the FIXME of remove all recovery codes when the last secondary authenticator is about to be removed?

kiootic:
I'm working on secondary password authenticator UI, let me also fix that in another PR.

PR #491, opened 2020-10-13
Validate password in resetPassword mutation

nicolaslee2775:
connect #462

Frontend already handled empty password.

https://github.com/authgear/authgear-server/blob/2d15d634c5ace571628b38694c594e782206a262/portal/src/PasswordField.tsx#L89-L114

### DIFF
--- a/pkg/lib/config/authenticator.go
+++ b/pkg/lib/config/authenticator.go
@@ -32,13 +32,6 @@ type AuthenticatorPasswordConfig struct {
 	Policy *PasswordPolicyConfig `json:"policy,omitempty"`
 }
 
-func (c *AuthenticatorPasswordConfig) SetDefaults() {
-	if c.Policy == nil {
-		c.Policy = &PasswordPolicyConfig{}
-	}
-	c.Policy.SetDefaults()
-}
-
 var _ = Schema.Add("PasswordPolicyConfig", `
 {
 	"type": "object",
--- a/pkg/lib/config/authenticator.go
+++ b/pkg/lib/config/authenticator.go
@@ -32,12 +32,20 @@ type AuthenticatorPasswordConfig struct {
 	Policy *PasswordPolicyConfig `json:"policy,omitempty"`
 }
 
+func (c *AuthenticatorPasswordConfig) SetDefaults() {
+	if c.Policy != nil {
+		if c.Policy.MinLength == nil {
+			c.Policy.MinLength = newInt(1)
+		}
+	}
+}
+
 var _ = Schema.Add("PasswordPolicyConfig", `
 {
 	"type": "object",
 	"additionalProperties": false,
 	"properties": {
-		"min_length": { "type": "integer" },
+		"min_length": { "type": "integer", "minimum": 1 },
 		"uppercase_required": { "type": "boolean" },
 		"lowercase_required": { "type": "boolean" },
 		"digit_required": { "type": "boolean" },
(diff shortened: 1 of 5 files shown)


kiootic (commented):
I don't think frontend has handled the empty password case? It only checks for mismatched password and password policy.

louischan-oursky:
@kiootic Do you think we should handle empty password in `pkg/lib/authn/authenticator/password/checker.go#49` `checkPasswordLength`? So if minLength is less than or equal to 0, minLength should be by default 1. Or we change our configuration schema to set `minimum` of  `min_length` to 1, and if it is absent, it is 1 by default.

kiootic:
> Or we change our configuration schema to set minimum of min_length to 1, and if it is absent, it is 1 by default.

I agree this one seems better.

louischan-oursky:
In case Nicolas wants to help, let me give some guidelines on what we want to implement

- Drop your current changes
- In `pkg/lib/config/authenticator.go#40`, you should see the JSON schema of `min_length`. We would like to add `"minimum": 1` to the JSON schema.
- We want to add a default value to `min_length`. So we need to change the type from `int` to `*int`. There is an example in the same file. See `AuthenticatorTOTPConfig.Maximum`. After that we need to add a method `SetDefaults` to `AuthenticatorPasswordConfig` to populate default value for `min_length` in case it is absent. Beware of nil struct.

nicolaslee2775:
@louischan-oursky Dropped previous code and added minimum constraint to `min_length`.

But I found that password validation is only performed if password is not empty. So I can still set an empty password.

https://github.com/authgear/authgear-server/blob/50d096333ae87fb6719bb6742a283be433dcb710/pkg/lib/authn/authenticator/password/provider.go#L78-L100

kiootic:
Referring to the comments in the `New` method:
https://github.com/authgear/authgear-server/blob/50d096333ae87fb6719bb6742a283be433dcb710/pkg/lib/authn/authenticator/password/provider.go#L60-L62

A password authenticator is created with empty password (it cannot be used to authenticate) when adding a new login ID identity without existing password authenticator using old Auth API (ref: https://github.com/SkygearIO/skygear-server/pull/1383#discussion_r422919124). Since that behavior is no longer relevant, we can always hash the password and check password policy for empty password.

pkg/lib/config/authenticator.go
```diff
@@ -32,12 +32,20 @@ type AuthenticatorPasswordConfig struct {
 	Policy *PasswordPolicyConfig `json:"policy,omitempty"`
 }
 
+func (c *AuthenticatorPasswordConfig) SetDefaults() {
+	if c.Policy != nil {
+		if c.Policy.MinLength == nil {
+			c.Policy.MinLength = newInt(1)
+		}
+	}
```
louischan-oursky, line 40: @kiootic Do we still have the populate-struct logic when we parse the config? 

I think we should also set default value if Policy is nil. In that case, initialize an empty Policy first.

kiootic: The populate default logic process the object graph in post-order traversal (i.e. populate recursively for the fields, then running the custom `SetDefaults` function), so there's no problem.

louischan-oursky: So the SetDefaults method should be implemented on PasswordPolicyConfig?

kiootic: Yes, I suppose so.

nicolaslee2775:
@louischan-oursky @kiootic Updated.

PR #506, opened 2020-10-14
Layout verify domain screen

### LINKED ISSUE #495: [Portal] Layout verify domain screen
- [x] get domain details from domain id (CNAME, TXT)
- [x] handle empty state (domain not found)
- [x] create hook for verify domain mutation
- [x] detail list showing domain detail, copy button in value column (onRenderItemColumn)
- [x] verify button calls verify domain mutation on click, with loading state
- [x] handle error (error dialog)
- [x] navigate to DNS list screen with search param `?verify=success`
- [x] cancel button navigates to DNS list on click

axolotl0212:
close #495

# Screenshot

Screenshot 2020-10-14 at 6 12 22 PM

### DIFF
--- a/portal/src/AppRoot.tsx
+++ b/portal/src/AppRoot.tsx
@@ -22,6 +22,7 @@ import CreateOAuthClientScreen from "./graphql/portal/CreateOAuthClientScreen";
 import EditOAuthClientScreen from "./graphql/portal/EditOAuthClientScreen";
 import UserInterfaceScreen from "./graphql/portal/UserInterfaceScreen";
 import DNSConfigurationScreen from "./graphql/portal/DNSConfigurationScreen";
+import VerifyDomainScreen from "./graphql/portal/VerifyDomainScreen";
 import SettingsScreen from "./graphql/portal/SettingsScreen";
 
 const AppRoot: React.FC = function AppRoot() {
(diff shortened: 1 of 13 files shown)


axolotl0212:
rebased

kiootic (commented):
- Please help to localize the error messages, you can find a list of possible domain-related errors here: https://github.com/authgear/authgear-server/blob/master/pkg/portal/service/domain.go#L27-L40
- Please help to make it so that when a custom domain is deleted, the public origin is updated to use the non-custom domain endpoint.

axolotl0212:
> Please help to make it so that when a custom domain is deleted, the public origin is updated to use the non-custom domain endpoint.

Do you mean if custom domain chosen in public origin field is deleted, update to use non-custom domain endpoint?

kiootic:
> Do you mean if custom domain chosen in public origin field is deleted, update to use non-custom domain endpoint?

Yes, that's what I intended. 

> Also by non-custom domain endpoint you mean the initial endpoint (not customised) ? How to get the that endpoint?

The first domain with `isCustom === false` should be the non-custom domain. (i.e. the provided default domain)

axolotl0212:
> The first domain with isCustom === false should be the non-custom domain. (i.e. the provided default domain)

I observed that the domain list does not include default domain @kiootic

kiootic:
> I observed that the domain list does not include default domain

The first app (`http://accounts.portal.localhost:3000`) would not have it, since we did not update the database entries during the bootstrap process. If you create a new app in multi-tenant mode, you should see a default domain in the list.

axolotl0212:
> The first app (http://accounts.portal.localhost:3000) would not have it, since we did not update the database entries during the bootstrap process. If you create a new app in multi-tenant mode, you should see a default domain in the list.

I see. Tried to add another app and default domain is included in the list, thanks

kiootic (approved):

PR #507, opened 2020-10-14
Portal access control - Part 2

### LINKED ISSUE #433: Portal Access Control
- Add table `_portal_app_collaborator`
  - `user_id`: The user ID of the Portal's tenant.
  - `app_id`: The app ID that the user has access to.
  - `created_at`: Creation time
  - Note that we do not store the app ID of the Portal's tenant, as it is part of the runtime configuration.
  - Columns like `role` are omitted for now. It is because we do not need them at the moment. We consider all collaborator as owners now.
- Add, list, remove collaborator
- Use `_portal_app_collaborator` to enforce access control
- Add row to `_portal_app_collaborator` when a new app is created by an user
- Expose list as graphql query
- Expose remove as graphql mutation

---

- Support calling Portal's Authgear Admin API
- Make the encoding/decoding of GraphQL ID public.
  - This is because the portal is going to receive the user ID from HTTP header. And then it has to encode the user ID in GraphQL format, call Admin API to see if the user is the intended invitee.

---

- Add table `_portal_app_collaborator_invitation`
  - `invited_by`: The user ID who sent the invitation
  - `target`: The email
  - `created_at`: Creation time
  - `expire_at`: Expiration time. Expire in 3 days
  - `code`: A randomly generated code
- List non-expiring invitation
- Invite collaborator
  - Check if the invitee is collaborator (FIXME: It is impossible now because Admin API does not have getUserByClaim)
  - Check if the invitee is on the invitation list already
  - Add row to `_portal_app_collaborator_invitation`
  - Send invitation email
- Remove invitation
  - Remove invitation by ID
- Accept invitation
  - Accept non-expiring invitation by code
  - Check if the authenticated user matches the `target`. (Call Admin API to get a single user node)
  - Remove the invitation
  - Add row to `_portal_app_collaborator`

louischan-oursky:
fix #433 

There are two remaining FIXMEs

- Send the invitation email. I am going to reuse the template mechanism from pkg/lib. The problem is where to store the default template. Storing it in templates/ seems a bit inappropriate because it is a place for authgear.
- Able to call Admin API of `accounts`.

### DIFF
--- a/pkg/portal/graphql/app_mutation.go
+++ b/pkg/portal/graphql/app_mutation.go
@@ -64,6 +64,12 @@ var _ = registerMutationField(
 			appID := resolvedNodeID.ID
 
 			gqlCtx := GQLContext(p.Context)
+
+			err := gqlCtx.AuthzService.CheckAccessOfViewer(appID)
+			if err != nil {
+				return nil, err
+			}
+
 			lazy := gqlCtx.Apps.Get(appID)
 			return lazy.
 				Map(func(value interface{}) (interface{}, error) {
--- a/pkg/portal/graphql/app_mutation.go
+++ b/pkg/portal/graphql/app_mutation.go
@@ -65,11 +65,6 @@ var _ = registerMutationField(
 
 			gqlCtx := GQLContext(p.Context)
 
-			err := gqlCtx.AuthzService.CheckAccessOfViewer(appID)
-			if err != nil {
-				return nil, err
-			}
-
 			lazy := gqlCtx.Apps.Get(appID)
 			return lazy.
 				Map(func(value interface{}) (interface{}, error) {
--- a/pkg/portal/graphql/collaborator.go
+++ b/pkg/portal/graphql/collaborator.go
@@ -8,8 +8,36 @@ var collaborator = graphql.NewObject(graphql.ObjectConfig{
 	Name:        "Collaborator",
 	Description: "Collaborator of an app",
 	Fields: graphql.Fields{
-		"id":        &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
-		"userID":    &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
+		"id": &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
+
+		// AppID is intentionally excluded because
+		// if we wanted to reference back the app, we would add "app"
+		// But adding it would result in a circular schema.
+
+		// The value is a plain user ID.
+		// Is there any better way to handle this field?
+		"userID": &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
+
 		"createdAt": &graphql.Field{Type: graphql.NewNonNull(graphql.DateTime)},
+	},
+})
+
+var collaboratorInvitation = graphql.NewObject(graphql.ObjectConfig{
+	Name:        "CollaboratorInvitation",
+	Description: "Collaborator invitation of an app",
+	Fields: graphql.Fields{
+		"id": &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
+
+		// AppID is intentionally excluded because
+		// if we wanted to reference back the app, we would add "app"
+		// But adding it would result in a circular schema.
+
+		// The value is a plain user ID.
+		// Is there any better way to handle this field?
+		"invitedBy": &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
+
+		"inviteeEmail": &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
+		"createdAt":    &graphql.Field{Type: graphql.NewNonNull(graphql.DateTime)},
+		"expireAt":     &graphql.Field{Type: graphql.NewNonNull(graphql.DateTime)},
 	},
 })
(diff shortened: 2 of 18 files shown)


pkg/portal/graphql/collaborator.go
```diff
@@ -8,8 +8,36 @@ var collaborator = graphql.NewObject(graphql.ObjectConfig{
 	Name:        "Collaborator",
 	Description: "Collaborator of an app",
 	Fields: graphql.Fields{
-		"id":        &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
-		"userID":    &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
+		"id": &graphql.Field{Type: graphql.NewNonNull(graphql.String)},
+
+		// AppID is intentionally excluded because
```
kiootic, line 13: Circular reference between types in GraphQL schema should be okay. I think the problem is Go does not allow circular dependency between package variable initializer? Can we use `init()` function to initialize them instead?

ref: https://github.com/graphql-go/graphql/issues/164

pkg/portal/graphql/collaborator_mutation.go
```diff
@@ -49,3 +51,147 @@ var _ = registerMutationField(
...
+			Type:        graphql.NewNonNull(graphql.String),
+			Description: "Invitee email address.",
+		},
+	},
+})
+
+var createCollaboratorInvitationPayload = graphql.NewObject(graphql.ObjectConfig{
+	Name: "CreateCollaboratorInvitationPayload",
+	Fields: graphql.Fields{
+		"app": &graphql.Field{Type: graphql.NewNonNull(nodeApp)},
```
kiootic, line 167: I suppose we should include the full `CollaboratorInvitation` in the payload? Our convention is to include both the parent & created entity when creating, and only the parent entity when deleting.

pkg/portal/service/collaborator.go
```diff
@@ -138,6 +158,178 @@ func (s *CollaboratorService) DeleteCollaborator(c *model.Collaborator) error {
...
+			return nil, err
+		}
+		is = append(is, i)
+	}
+
+	if len(is) <= 0 {
+		return nil, ErrCollaboratorInvitationInvalidCode
+	}
+
+	if len(is) > 1 {
```
kiootic, line 282: This seems strange, since we have no unique constraint on code, so an unacceptable invitation may be created. We may want to add code to the uniqueness constraint as well?

pkg/portal/graphql/collaborator_mutation.go
```diff
@@ -110,6 +111,23 @@ var createCollaboratorInvitationInput = graphql.NewInputObject(graphql.InputObje
 	},
 })
 
+var createCollaboratorInvitationInputSchemaName = "CreateCollaboratorInvitationInputSchema"
+
+var createCollaboratorInvitationInputSchema = validation.NewMultipartSchema("").
```
kiootic, line 149: Actually I'm thinking to replace most of usage of multipart schema with simple schema, since they just have single schema. What do you think?

pkg/portal/graphql/app_mutation.go
```diff
@@ -64,6 +64,12 @@ var _ = registerMutationField(
 			appID := resolvedNodeID.ID
 
 			gqlCtx := GQLContext(p.Context)
+
+			err := gqlCtx.AuthzService.CheckAccessOfViewer(appID)
```
kiootic, line 68: I think this check should be done in `loader`/`service` layer (i.e. within `gqlCtx.Apps.Get` call), otherwise it's easy to miss the checks when authoring new query/mutation.

kiootic (commented):
Since the expired invitations cannot be retrieved from API (so collaborators cannot delete them), we should do housekeeping to delete them regularly. I suppose we are not going to implement that cron job for now.

pkg/portal/graphql/collaborator.go
louischan-oursky: It is not very useful to add `app`. Do we really need this?

pkg/portal/service/collaborator.go
louischan-oursky: OK.

pkg/portal/graphql/collaborator_mutation.go
louischan-oursky: We can keep the existing multipart schema API. The single schema API should be a shorthand over the current API.

pkg/portal/graphql/collaborator.go
kiootic: Not really... I just saw the comment and provide a possible workaround.

pkg/portal/graphql/app_mutation.go
louischan-oursky: Service is at the building build level so it may not have enough context to determine access control. Let me review whether we can do it at loader layer.

pkg/portal/graphql/collaborator_mutation.go
kiootic: When I was developing the validation schema abstraction, I was thinking of using multipart schema only for the user-facing schemas (i.e. app/secret configs), while the ad-hoc validation schemas would use simple schema for simplicity.

louischan-oursky: Can you open an issue to track this?

louischan-oursky:
Updated.

pkg/portal/graphql/collaborator_mutation.go
kiootic: ref #513

kiootic (approved):

PR #509, opened 2020-10-15
Implement settings creen portal admin tab

### LINKED ISSUE #498: [UI] Implement settings screen portal admins tab
- [x] Create `PortalAdminsSettings` component
- [x] Display admin user list
  - [x] email column
  - [x] status column (pending, accepted)
  - [x] action button column
- [x] Create remove admin user confirm dialog
- [x] Integrate admin user query
- [x] Integrate remove admin user mutation
- [ ] ~~Integrate invite admin user mutation~~
- [x] Link to add admin screen #499 
- [ ] ~~Block navigation if there are unsaved changes~~

nicolaslee2775:
fix #498 

### Settings screen portal admins tab
(admin email is displayed as `dummy@example.com` since collaborator info cannot be obtained in current stage)
image

### Add admin
image

### Remove admin
image

### Remove admin invitation
image


axolotl0212 (approved):

axolotl0212:
Seems like we fetch all collaborator in 1 query, I think we should remove pagination?

nicolaslee2775:
The collaborator invitation duplicate error is not yet handled. Should we display the error message under email field or display it via dialog?

image

axolotl0212:
> The collaborator invitation duplicate error is not yet handled. Should we display the error message under email field or display it via dialog?

I will vote for `error message under email field`

portal/src/graphql/portal/PortalAdminsSettings.tsx
```diff
@@ -58,100 +69,22 @@ const PortalAdminsSettings: React.FC<PortalAdminsSettingsProps> = function Porta
...
+      }
       const collaborator = collaborators.find(
         (collaborator) => collaborator.id === id
       );
       if (collaborator) {
         setRemovePortalAdminConfirmationDialogData({
           userID: id,
-          email: collaborator.email,
+          // TODO: obtain admin user email
+          email: "dummy@example.com",
```
louischan-oursky, line 97: Show the ID for now.

nicolaslee2775:
> > The collaborator invitation duplicate error is not yet handled. Should we display the error message under email field or display it via dialog?
> 
> I will vote for `error message under email field`

If so, we need to add `non validation error to json pointer mapping` rules to `useValidationError` hook. Like
```ts
interface Rule {
  errorMessageID: string;
  reason: APIError["reason"];
  kind?: string;
  cause?: string;
  jsonPointer: string;
}

type useValidationError = (error: unknown, rules?: Rule[]) => {
  unhandledCauses?: ValidationFailedErrorInfoCause[];
  otherError?: unknown;
  value: FormContextValue;
};
```

axolotl0212:
> If so, we need to add non validation error to json pointer mapping rules to useValidationError hook. 

I will just get the error message from `otherError`, then stick it to the form text field. Although this approach will hide the validation error, but I believe we won't have both generic + validation error in the same time. We can change the form text field to stack the error passed by props and validation error if we want to fix the problem?

nicolaslee2775:

##### Admin list
image
##### Remove admin user
image
##### Duplicated invitation error
image

louischan-oursky:
I added one more commit to handle a possible error.

Issue #412, opened 2020-09-28
Default template for Forgot Password Email & SMS


chpapa:
We need some good looking default template for Forgot Password Email & SMS (and other templates)

@keithtkl can you help do it when u got time, or get someone help like Elliot?

PR #512, opened 2020-10-15
Update forgot password template #412

elly0t:
only updated forgot password template


elly0t:
closing this as this was opened for easy communication between Louis and I, but now is no longer needed

PR #515, opened 2020-10-15
Resource FS

### LINKED ISSUE #511: Revamp resource resolution
To provide more robust customization of resources, we should revamp the existing system:
- Use virtual FS for every resources:
    - authgear.yaml
    - authgear.secrets.yaml
    - templates (templates that goes through the template engine)
        - authgear
          - web
            - login.html
            - logout.html
            - etc.
          - messages
            - forgot_password_email.html
            - verification_sms.txt
            - etc.
          - translation.json
        - portal
          - messages
            - invitation_email.html
            - etc.
    - static (static assets that mounted at the authgear server)
        - img
            - logo.png
            - etc.
        - authgear.css
        - authgear.js
    - data (additional data needed by logic)
        - reserved_name.txt
- 2 layer of FS:
    - global FS: bundled in image, can be mounted by cluster admin to customize branding
    - app FS: app-specific resources, customized by developer
- custom layer merge function for each file type
    - config: for authgear.yaml, same as static for now
    - secrets: for authgear.secrets.yaml, merge from each layer using custom logic
    - templates: for templates that goes through the template engine, the file from top-most layer is used
    - static: for static assets that is used as-is, the file from top-most layer is used
- remove new app environment variables from portal: customize the global FS in main authgear server instead
- R/W resources:
    - reading resources:
      - portal GraphQL API to read effective (i.e. merged from both FS) & raw (i.e. in app FS) resources
      - for secret file type: only raw can be read
    - writing resources:
      - portal GraphQL API to update resources in app FS
      - validate FS after merge
    - assumption: portal server & authgear server must share the same mounted resource FS

kiootic:
ref #511 

This PR updates the portal GraphQL API to use resource FS. Remaining additional works to do:
- disallow overriding secrets provided in base FS
- allow configuring a template resource FS for new apps
- move other resources into FS


pkg/util/template2/translation.go
```diff
@@ -0,0 +1,104 @@
...
+func (t *translationJSON) MatchResource(path string) bool {
+	return matchTemplatePath(path, translationJSONName)
+}
+
+func (t *translationJSON) Merge(layers []resource.LayerFile, args map[string]interface{}) (*resource.LayerFile, error) {
+	return mergeTranslations(layers, args)
+}
+
+func (t *translationJSON) Parse(data []byte) (interface{}, error) {
+	var translations map[string]template.Translation
```
louischan-oursky, line 30: This is a bit confusing. Before `Merge`, the data of LayerFile is `map[string]string`, after Merge is done and result in a single LayerFile, the data becomes `map[string]template.Translation`.

Maybe we should add a distinct type for MergeFile and rename `Parse(data []byte)` to `ParseMerged(resource.MergedFile)`?

pkg/util/resource/manager.go
```diff
@@ -0,0 +1,47 @@
...
+		if err != nil {
+			return nil, err
+		}
+		layers = append(layers, files...)
+	}
+	if len(layers) == 0 {
+		return nil, ErrResourceNotFound
+	}
+
+	merged, err := desc.Merge(layers, args)
```
louischan-oursky, line 32: When do we call Parse?

pkg/util/template2/translation.go
kiootic: Agreed, let me change that.

pkg/util/resource/manager.go
kiootic: Parse is used in resolver to parse templates & translation map.

cmd/portal/server/config.go
```diff
@@ -32,8 +32,8 @@ type Config struct {
 	// Mail configures email settings.
 	Mail portalconfig.MailConfig `envconfig:"MAIL"`
 
-	// DefaultTemplateDirectory sets the directory for default template files
-	DefaultTemplateDirectory portalconfig.DefaultTemplateDirectory `envconfig:"DEFAULT_TEMPLATE_DIRECTORY" default:"templates"`
+	// DefaultResourceDirectory sets the directory for default resource files
+	DefaultResourceDirectory portalconfig.DefaultResourceDirectory `envconfig:"DEFAULT_RESOURCE_DIRECTORY" default:"resources/portal"`
```
louischan-oursky, line 36: Portal may need to read `resources/authgear`.

portal/src/graphql/portal/mutations/updateAppAndSecretMutation.ts
```diff
@@ -8,29 +8,27 @@ import {
...
   PortalAPISecretConfig,
 } from "../../../types";
-import { UpdateAppAndSecretConfigMutation } from "./__generated__/UpdateAppAndSecretConfigMutation";
+import {
+  UpdateAppAndSecretConfigMutation,
+  UpdateAppAndSecretConfigMutationVariables,
+} from "./__generated__/UpdateAppAndSecretConfigMutation";
 
 // relative to project root
 const APP_CONFIG_PATH = "./authgear.yaml";
```
louischan-oursky, line 17: Should we also remove the `./` prefix?

PR #516, opened 2020-10-15
Portal access control - Part 3

louischan-oursky:
ref #433


.make-lint-expect
```diff
@@ -7,3 +7,8 @@ pkg/lib/oauth/handler/handler_authz.go cannot import github.com/authgear/authgea
 pkg/lib/oauth/handler/mock_test.go cannot import github.com/authgear/authgear-server/pkg/auth/webapp
 
 exit status 1
+pkg/portal/deps/deps_task.go cannot import github.com/authgear/authgear-server/pkg/worker/tasks
```
kiootic, line 10: How about just make a copy of `SendMessagesTask` in `portal/task/tasks` to avoid introducing new lint exceptions?

kiootic (approved):

PR #522, opened 2020-10-16
Add nodes to root query

### LINKED ISSUE #521: Add `nodes` to graphql root query
https://developer.github.com/v4/query/#nodes

louischan-oursky:
fix #521

### DIFF
--- a/pkg/admin/graphql/entity.go
+++ b/pkg/admin/graphql/entity.go
@@ -8,20 +8,21 @@ import (
 	"github.com/authgear/graphql-go-relay"
 	"github.com/graphql-go/graphql"
 
-	"github.com/authgear/authgear-server/pkg/api/apierrors"
 	"github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/util/graphqlutil"
 )
 
 var nodeDefs = relay.NewNodeDefinitions(relay.NodeDefinitionsConfig{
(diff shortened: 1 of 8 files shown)


pkg/admin/graphql/entity.go
```diff
@@ -8,20 +8,21 @@ import (
...
 	"github.com/graphql-go/graphql"
 
-	"github.com/authgear/authgear-server/pkg/api/apierrors"
 	"github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/util/graphqlutil"
 )
 
 var nodeDefs = relay.NewNodeDefinitions(relay.NodeDefinitionsConfig{
 	IDFetcher: func(id string, info graphql.ResolveInfo, ctx context.Context) (interface{}, error) {
+		// If the ID is invalid, we should return null instead of returning an error.
```
kiootic, line 17: Should we follow the advice in [GraphQL Relay spec](https://relay.dev/graphql/objectidentification.htm#sel-EALVNCAACCgBhuc) and return null for nodes with invalid ID?

louischan-oursky: I haven't read the part of the spec before so my change is just a coincidence. I just feel intuitive that if an invalid ID (or valid, but the underlying typeless ID is invalid) is given, we should return null

kiootic: Are we going to change our fork to produce null for node fetch errors, so it'll leave nulls in the list instead of failing the whole query?

kiootic (approved):

PR #523, opened 2020-10-16
Implement accept admin invitation screen

### LINKED ISSUE #518: [UI] Implement accept admin invitation screen
- [x] Create `AcceptPortalAdminInvitationScreen` component
  - [x] obtain invitation code from url
  - [x] display accept button
- [x] Handle redirection if not yet logged in
  - [x] redirect to login page
  - [x] back to invitation accept screen
- [x] Handle accept collaborator invitation mutation
- [x] Handle mutation error
- [x] Direct to app specific screen after successfully accept invitation

nicolaslee2775:
close #518 

##### Accept admin invitation screen
image
##### Invalid invitation code error
image

### DIFF
--- a/portal/src/AppRoot.tsx
+++ b/portal/src/AppRoot.tsx
@@ -25,7 +25,6 @@ import DNSConfigurationScreen from "./graphql/portal/DNSConfigurationScreen";
 import VerifyDomainScreen from "./graphql/portal/VerifyDomainScreen";
 import SettingsScreen from "./graphql/portal/SettingsScreen";
 import InviteAdminScreen from "./graphql/portal/InviteAdminScreen";
-import AcceptAdminInvitationScreen from "./graphql/portal/AcceptAdminInvitationScreen";
 
 const AppRoot: React.FC = function AppRoot() {
   const { appID } = useParams();
@@ -111,10 +110,6 @@ const AppRoot: React.FC = function AppRoot() {
           <Route
             path="/configuration/settings/invite-admin"
             element={<InviteAdminScreen />}
-          />
-          <Route
-            path="/collaborators/invitation"
-            element={<AcceptAdminInvitationScreen />}
           />
         </Routes>
       </ScreenLayout>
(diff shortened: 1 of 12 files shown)


axolotl0212:
#509 is merged, please rebase

portal/src/graphql/portal/AcceptAdminInvitationScreen.tsx
```diff
@@ -0,0 +1,54 @@
...
+        if (appID !== null) {
+          navigate("../../");
+        }
+      })
+      .catch(() => {});
+  }, [acceptCollaboratorInvitation, invitationCode, navigate]);
+
+  return (
+    <main className={cn(styles.root, { [styles.loading]: loading })}>
+      {error && <ShowError error={error} />}
```
axolotl0212, line 39: The error thrown by backend should contain a `reason` field? I think we should handle the error and show localized error message.

louischan-oursky:
This screen is for accepting invitation to an unknown app so we should not see the side bar. In `pkg/portal/endpoint/endpoint`, the endpoint is `collaborators/invitation` so the path served by the portal client should be `/collaborators/invitation`.

axolotl0212:
Actually I think this page should not be included in AppRouter (AppRoot.tsx) as this page is not `within` any app? The URL should look something like `localhost:8000/collaborators/invitation?code={base64 string}`. The appID should be encoded in the code so we should only include the appID into the url when user accepted the invitation and chosen the new app?

portal/src/AppRoot.tsx
```diff
@@ -34,7 +34,7 @@ const AppRoot: React.FC = function AppRoot() {
   }, [appID]);
   return (
     <ApolloProvider client={client}>
-      <ScreenLayout>
+      <ScreenLayout showScreenNav={true}>
```
louischan-oursky, line 36: I think we should not be touching AppRoot and ScreenLayout. Can we follow the structure of `AppsScreen`?

axolotl0212 (approved):

nicolaslee2775:

louischan-oursky (approved):
Thanks

PR #531, opened 2020-10-19
Integrate forgot password query and mutation

### LINKED ISSUE #528: Integrate forgot password template query and mutation
- [x] Integrate template query
- [x] Integrate template mutation

nicolaslee2775:
close #528

### DIFF
--- a/portal/src/graphql/portal/ForgotPasswordSettings.tsx
+++ b/portal/src/graphql/portal/ForgotPasswordSettings.tsx
@@ -1,17 +1,11 @@
-import React, {
-  useMemo,
-  useContext,
-  useState,
-  useCallback,
-  useEffect,
-} from "react";
+import React, { useMemo, useContext, useState, useCallback } from "react";
 import { FormattedMessage, Context } from "@oursky/react-messageformat";
 import { TextField, Label } from "@fluentui/react";
 import cn from "classnames";
 import deepEqual from "deep-equal";
 import produce from "immer";
 
-import { AppTemplatesUpdater } from "./mutations/updateAppTemplatesMutation";
+import { UpdateAppTemplatesData } from "./mutations/updateAppTemplatesMutation";
 import CodeEditor from "../../CodeEditor";
 import ButtonWithLoading from "../../ButtonWithLoading";
 import NavigationBlockerDialog from "../../NavigationBlockerDialog";
@@ -34,12 +28,13 @@ interface ForgotPasswordSettingsProps {
   effectiveAppConfig: PortalAPIAppConfig | null;
   rawAppConfig: PortalAPIAppConfig | null;
   templates: Record<ForgotPasswordMessageTemplateKeys, string>;
-  updateAppConfig: (
-    appConfig: PortalAPIAppConfig
+  updateAppConfigAndTemplates: (
+    appConfig: PortalAPIAppConfig,
+    updateTemplatesData: UpdateAppTemplatesData<
+      ForgotPasswordMessageTemplateKeys
+    >
   ) => Promise<PortalAPIApp | null>;
-  updateTemplates: AppTemplatesUpdater<ForgotPasswordMessageTemplateKeys>;
-  updatingAppConfig: boolean;
-  updatingTemplates: boolean;
+  updatingAppConfigAndTemplates: boolean;
 }
 
 interface ForgotPasswordSettingsState {
@@ -126,10 +121,8 @@ const ForgotPasswordSettings: React.FC<ForgotPasswordSettingsProps> = function F
     effectiveAppConfig,
     rawAppConfig,
     templates,
-    updateAppConfig,
-    updateTemplates,
-    updatingAppConfig,
-    updatingTemplates,
+    updateAppConfigAndTemplates,
+    updatingAppConfigAndTemplates,
   } = props;
 
   const { renderToString } = useContext(Context);
@@ -218,19 +211,13 @@ const ForgotPasswordSettings: React.FC<ForgotPasswordSettingsProps> = function F
         state
       );
 
-      updateAppConfig(newAppConfig).catch(() => {});
-      updateTemplates(updateTemplatesData).catch(() => {});
+      updateAppConfigAndTemplates(
+        newAppConfig,
+        updateTemplatesData
+      ).catch(() => {});
     },
-    [rawAppConfig, initialState, state, updateAppConfig, updateTemplates]
+    [rawAppConfig, initialState, state, updateAppConfigAndTemplates]
   );
-
-  // Set state to initialState after updating
-  useEffect(() => {
-    if (!updatingAppConfig || !updatingTemplates) {
-      setState(initialState);
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [updatingAppConfig, updatingTemplates]);
 
   return (
     <form className={cn(styles.root, className)} onSubmit={onFormSubmit}>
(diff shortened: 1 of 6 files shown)


portal/src/graphql/portal/ForgotPasswordSettings.tsx
```diff
@@ -31,28 +36,39 @@ interface ForgotPasswordSettingsState {
...
+  emailAndSmsTemplates: PortalAPIEmailAndSmsTemplates | null
 ): ForgotPasswordSettingsState {
   const forgot_password = appConfig?.forgot_password;
 
   return {
-    emailHtmlTemplate: "", // TODO: handle email template
-    emailPlainTextTemplate: "", // TODO: handle email template
-    smsTemplate: "", // TODO: handle sms template
+    emailHtmlTemplate: emailAndSmsTemplates?.emailHtml ?? "",
+    emailPlainTextTemplate: emailAndSmsTemplates?.emailHtml ?? "",
```
axolotl0212, line 47: `emailHtmlTemplate`, `emailPlainTextTemplate` are same, both equals `emailAndSmsTemplates?.emailHtml`?

axolotl0212: should be `emailAndSmsTemplates?.emailText` ?

portal/src/graphql/portal/ForgotPasswordSettings.tsx
```diff
@@ -5,23 +5,28 @@ import cn from "classnames";
...
 import ButtonWithLoading from "../../ButtonWithLoading";
 import NavigationBlockerDialog from "../../NavigationBlockerDialog";
 import { setFieldIfChanged, clearEmptyObject } from "../../util/misc";
-import { PortalAPIAppConfig, PortalAPIApp } from "../../types";
+import { PortalAPIAppConfig, PortalAPIEmailAndSmsTemplates } from "../../types";
 
 import styles from "./ForgotPasswordSettings.module.scss";
-import TodoButtonWrapper from "../../TodoButtonWrapper";
+
+export const FORGOT_PASSWORD_EMAIL_HTML_TEMPLATE_NAME =
```
axolotl0212, line 17: Seems like there is a 1-1 relation between template item type and template name? I suggest we group them into a map, eg.

```typescript
export const TEMPLATE_NAME: Record<keyof PortalAPIEmailAndSmsTemplates, string> = {
    "emailHtml": "forgot_password_email.html",
    "emailText": "forgot_password_email.txt",
    "smsText": "forgot_password_sms.txt"
}
```

portal/src/graphql/portal/PasswordsScreen.tsx
```diff
@@ -1,16 +1,21 @@
...
 import { useParams } from "react-router-dom";
 import { FormattedMessage, Context } from "@oursky/react-messageformat";
 import { Pivot, PivotItem, Text } from "@fluentui/react";
 import cn from "classnames";
 
 import PasswordPolicySettings from "./PasswordPolicySettings";
-import ForgotPasswordSettings from "./ForgotPasswordSettings";
-import { useAppConfigQuery } from "./query/appConfigQuery";
-import { useUpdateAppConfigMutation } from "./mutations/updateAppConfigMutation";
+import ForgotPasswordSettings, {
```
axolotl0212, line 8: I think we should define template name in this file and import in Setting component as Screen is parent?

portal/src/graphql/portal/ForgotPasswordSettings.tsx
```diff
@@ -61,13 +77,66 @@ function constructAppConfigFromState(
       screenState.resetCodeExpirySeconds
     );
 
-    // TODO: update email template
-    // TODO: update sms template
+    if (
```
axolotl0212, line 80: Template items comes from config, which might not be empty. If user cleared an existing template (changed to empty string), we should remove the corresponding element? In current implementation we only append?

As are doing exactly same thing for three templates, I think we can abstract this to a function? something like

```typescript
updateTemplateItems(screenState, initialScreenState, key, templateItems) {
  // early return if no changes 
  if (initialScreenState[key] === screenState[key]) {
    return
  }
  // append if content non-empty and corresponding template item not exist
  ...
  // remove if content empty and corresponding template item exist
  ...
}
```

portal/src/graphql/portal/PasswordsScreen.tsx
```diff
@@ -47,8 +70,14 @@ const PasswordsScreen: React.FC = function PasswordsScreen() {
...
   return (
-    <main className={cn(styles.root, { [styles.loading]: updatingAppConfig })}>
-      {updateAppConfigError && <ShowError error={updateAppConfigError} />}
+    <main
+      className={cn(styles.root, {
+        [styles.loading]: updatingAppAndEmailSmsTemplateConfig,
+      })}
+    >
+      {updateAppAndEmailSmsTemplateConfigError && (
+        <ShowError error={updateAppAndEmailSmsTemplateConfigError} />
```
axolotl0212, line 79: If there is `reason` in API error response, we can handle those error and show it in dialog? (using `ErrorDialog`)

portal/src/graphql/portal/ForgotPasswordSettings.tsx
nicolaslee2775: @louischan-oursky Should we remove template item or simply set content to empty string?

louischan-oursky: Remove. So that default value can be restored

nicolaslee2775:
@louischan-oursky Rebased. Removed template if template is empty.

nicolaslee2775:
`useAppTemplatesQuery` don't fetch data when switching page.

image
image

axolotl0212:
> `useAppTemplatesQuery` don't fetch data when switching page.

If the data is fetched once then it is cached, so not refetching when switching page is expected behaviour?

nicolaslee2775:
> > `useAppTemplatesQuery` don't fetch data when switching page.
> 
> If the data is fetched once then it is cached, so not refetching when switching page is expected behaviour?

`useAppTemplatesQuery` should refetch as resource paths are different.

axolotl0212:
> useAppTemplatesQuery should refetch as resource paths are different.

Argument is different so it should refetch, not sure why that's not the case.
`fetchPolicy: "network-only"` will make sure data is refetched on page change but I don't think this is a good fix

nicolaslee2775:
`keyArgs` of resources was set to `[]` so apollo returns cached value even paths are different.
https://github.com/authgear/authgear-server/blob/fd1a86f133de7c68710d7e6027efc7775abecabc/portal/src/graphql/portal/apollo.ts#L9-L14

nicolaslee2775:
@louischan-oursky @axolotl0212 Updated. Removed `keyArgs: []` from resources' type policy so apollo can cache data correctly.

portal/src/graphql/portal/ForgotPasswordSettings.tsx
```diff
@@ -218,6 +224,14 @@ const ForgotPasswordSettings: React.FC<ForgotPasswordSettingsProps> = function F
     [rawAppConfig, initialState, state, updateAppConfig, updateTemplates]
   );
 
+  // Set state to initialState after updating
+  useEffect(() => {
+    if (!updatingAppConfig || !updatingTemplates) {
+      setState(initialState);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [updatingAppConfig, updatingTemplates]);
```
louischan-oursky, line 233: Can we try to set `key` on `<ForgotPasswordSettings>`? The value of the key `version` should be a increasing integer. Every time a update is done, `version` should be incremented by one. By changing the key, the component will unmount and mount again.

nicolaslee2775:
`templates` is not updated after update app templates. So value is reset to previous value after saving.

image
image

nicolaslee2775:
Apollo cache resources by their paths
image

nicolaslee2775:
@louischan-oursky Fixed.

Issue #527, opened 2020-10-19
Make the definition of usage of dataloader more like canonical implementation


louischan-oursky:
- [ ] Change the interface of dataloader to match canonical one
- [ ] Add one dataloader per entity to graphql context
- [ ] Change the usage of dataloader to deal with loading entity by IDs only

PR #532, opened 2020-10-19
Portal access control - Part 4

### LINKED ISSUE #433: Portal Access Control

louischan-oursky:
fix #433


pkg/portal/service/authz.go
```diff
@@ -54,18 +54,21 @@ func (s *AuthzService) AddAuthorizedUser(appID string, userID string) error {
 	return s.Collaborators.CreateCollaborator(c)
 }
 
-func (s *AuthzService) CheckAccessOfViewer(appID string) error {
+func (s *AuthzService) CheckAccessOfViewer(appID string) (userID string, err error) {
```
kiootic, line 57: It seems the returned `userID` is not used everywhere, is there any reason to return it?

louischan-oursky: It may be used in https://github.com/authgear/authgear-server/issues/527

kiootic (approved):

PR #541, opened 2020-10-20
Fix DNS screen firing unexpected mutation

### LINKED ISSUE #538: Visit DNS page will fire unwanted mutation

axolotl0212:
fix #538 

Cause:

If domain in config is not included in verified domains (domains in DB with
isVerified = true), it will trigger useEffect on mount, firing mutation. If default domain
(non-custom verified domain) is not found, mutation fails as public
origin is required field

Fix:

Move update config to screen, check if domain in config is included in
verified domain only after delete domain mutation instead of useEffect.

Public origin section only observe public origin in config, will reset
dropdown state if config is changed (after delete domain)

### DIFF
--- a/portal/src/graphql/portal/DNSConfigurationScreen.tsx
+++ b/portal/src/graphql/portal/DNSConfigurationScreen.tsx
@@ -26,7 +26,6 @@ import {
   MessageBar,
   MessageBarType,
   Dropdown,
-  IDropdownOption,
 } from "@fluentui/react";
 
 import { useAppConfigQuery } from "./query/appConfigQuery";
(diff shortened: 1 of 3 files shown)


louischan-oursky:
@kiootic Should we handle the coordination of `public_origin` and custom domain deletion in the backend? This fix is very hacky to me.

kiootic:
> Should we handle the coordination of public_origin and custom domain deletion in the backend? This fix is very hacky to me.

That sound good to me. So we need to do this when deleting a custom domain:
1. Get a list of verified domains
2. Get the raw app config
3. If the public origin in raw app config is not present in the verified domain list:
4. Set public origin to the verified non-custom domain if present (it doesn't in case of missing database entries during bootstrapping)

louischan-oursky:
Or in step 4, set to the concatenation of `"https://"`, the app ID and `APP_HOST_SUFFIX`?

axolotl0212:
> @kiootic Should we handle the coordination of `public_origin` and custom domain deletion in the backend? This fix is very hacky to me.

I guess backend should handle the changes, as deleting a domain might invalidate the config. If user delete the domain by directly calling the graphQL API, the config would be invalid?

Actually the fix moved the `setPublicOrigin` part from the public origin section to `then(...)` statement after delete mutation which is doing something similar. Removing the `then(...)` and handle this in backend sounds better. 

The public origin section can observe the effective app config (implementation in current fix)

louischan-oursky:
@axolotl0212 Can you help remove the hack in the frontend? We are going to implement the fix in the backend. The frontend should ensure it fetches the latest app after a domain is removed.

axolotl0212:
> set to the concatenation of `"https://"`, the app ID and `APP_HOST_SUFFIX`

This seems better as it always exists?

axolotl0212:
> @axolotl0212 Can you help remove the hack in the frontend? We are going to implement the fix in the backend. The frontend should ensure it fetches the latest app after a domain is removed.

Sure, will add another commit to remove the set public origin part after delete domain

axolotl0212:
@louischan-oursky removed the `set public origin after delete` part, added `effectiveAppConfig` and `rawAppConfig` to delete domain mutation

axolotl0212:
updated typegen and rebased

axolotl0212:
rebased

PR #543, opened 2020-10-20
Fix the path of translation.json in new app

louischan-oursky:


louischan-oursky:
@kiootic This fix is quite important. Could you have a look on this first?

kiootic (approved):

PR #544, opened 2020-10-20
Do not treat the project root as local fs config directory by default

louischan-oursky:


pkg/lib/config/configsource/config.go
```diff
@@ -24,5 +24,5 @@ type Config struct {
 	// Watch indicates whether the configuration source would watch for changes and reload automatically
 	Watch bool `envconfig:"WATCH" default:"true"`
 	// Directory sets the path to app configuration directory file for local FS sources
-	Directory string `envconfig:"DIRECTORY" default:"."`
+	Directory string `envconfig:"DIRECTORY" default:"./var"`
```
kiootic, line 27: I suppose changing this default value would break Docker deployments? Should we instead just update the `.env` examples?

louischan-oursky: Updated

kiootic (approved):

PR #545, opened 2020-10-20
Revamp dataloader

### LINKED ISSUE #527: Make the definition of usage of dataloader more like canonical implementation

louischan-oursky:
fix #527


pkg/portal/graphql/app_mutation.go
```diff
@@ -69,30 +71,35 @@ var _ = registerMutationField(
...
-							Data: data,
-						})
-					}
-
-					return map[string]interface{}{
-						"app": gqlCtx.Apps.UpdateResources(app, resourceUpdates),
-					}, nil
-				}).
-				Value, nil
+			app, err := gqlCtx.AppService.Get(appID)
```
kiootic, line 74: If I remember correctly we're enforcing access control in loader layer, it seems authz check would be skipped if we call services directly?

pkg/portal/graphql/app_mutation.go
```diff
@@ -69,30 +71,35 @@ var _ = registerMutationField(
...
+					Data: data,
+				})
+			}
+
+			err = gqlCtx.AppService.UpdateResources(app, resourceUpdates)
+			if err != nil {
+				return nil, err
+			}
+
+			// App is not primed here intentionally.
```
kiootic, line 99: Any reason behind this?

louischan-oursky: I meant the app from line 74 was not primed because this mutation mutates app. So it is safer to not prime it.

pkg/portal/graphql/app_mutation.go
louischan-oursky: Maybe we should enforce access control in resolver? It is the place where the context is clear enough.

kiootic: That sounds good to me.

louischan-oursky: Updated.

kiootic (approved):

PR #546, opened 2020-10-20
Follow up on resource FS

### LINKED ISSUE #511: Revamp resource resolution

kiootic:
fix #511

TODO:
- portal css resource theming


kiootic:
@louischan-oursky let's review it first, will make portal themable in another PR.

cmd/authgear/Dockerfile
```diff
@@ -43,10 +43,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 RUN update-ca-certificates
 COPY ./migrations ./migrations
 COPY --from=stage1 /src/authgear /usr/local/bin/
-COPY --from=stage2 /usr/src/app/dist/ ./dist/
 COPY ./resources/authgear/ ./resources/authgear/
+COPY --from=stage2 /usr/src/app/dist/* ./resources/authgear/static/
```
louischan-oursky, line 47: Can we avoid using * in COPY?

pkg/lib/web/image.go
```diff
@@ -0,0 +1,155 @@
...
+func (a imageAsset) Parse(merged *resource.MergedFile) (interface{}, error) {
+	mimeType := http.DetectContentType(merged.Data)
+	ext, ok := imageExtensions[mimeType]
+	if !ok {
+		return nil, fmt.Errorf("invalid image format: %s", mimeType)
+	}
+
+	var path string
+	if langTag, ok := merged.Args[argResolvedLanguageTag]; ok {
+		path = fmt.Sprintf("%s_%s%s", a.Name, langTag, ext)
```
louischan-oursky, line 146: Any reason why we do not follow the convention of template `/__default__/logo.png`?

kiootic: `__default__` tag is just an alias to the fallback language configured in configuration. Since we already have the real resolved language tag when deriving URL to the image, there is no need to have this tag alias.

I'd prefer using `app_logo_en.png` instead of `en/app_logo.png` since the path looks better to me. I can change it back to use `<language tag>/<name>` if you found it more consistent.

louischan-oursky: Let's be more consistent.

PR #558, opened 2020-10-22
Update redirect URI in README and k8s config

### LINKED ISSUE #557: Update redirect URI in readme and k8s secret
commit 1caba981bf updated the OAuth redirect URI, but the redirect in 
- README.md (single tenant mode)
- k8s-manifest.yaml (multi-tenant mode)
is not updated

also there is a typo in muti-tenant mode section in readme.

axolotl0212:
fix #557


louischan-oursky:
Thanks!

PR #577, opened 2020-10-23
Return entity reference only in mutating service facade functions

### LINKED ISSUE #568: Add time for username is always at 1/1/1, 7:36 AM
When try to add a username under "Add Identity" dropdown, will see the add time is 1/1/1 7:36AM. But the date time will change to a correct one after page reload

Screenshot 2020-10-23 at 14 17 20

kiootic:
fix #568

### DIFF
--- a/pkg/admin/facade/identity.go
+++ b/pkg/admin/facade/identity.go
@@ -53,7 +53,7 @@ func (f *IdentityFacade) Remove(identityInfo *identity.Info) error {
 	return nil
 }
 
-func (f *IdentityFacade) Create(userID string, identityDef model.IdentityDef, password string) (*identity.Info, error) {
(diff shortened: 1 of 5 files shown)


louischan-oursky (commented):
I am curious of why it did not work as expected. Any details?

kiootic:
We pass the new `*identity.Info` to service through facade. The service translates the `*identity.Info` to the concrete type of each identity type (e.g. `*loginid.Identiy`), the specific identity type provider then populate the timestamps before updating database. Therefore, the updated timestamp did not propagated back to the passed in `*identity.Info`.

louischan-oursky:
Can we instead ensure the returned Info has latest information?

kiootic:
I'd prefer the updated approach, so that all resolved entity data comes from loaders instead of from service directly.

PR #579, opened 2020-10-23
Include verification code in email

### LINKED ISSUE #570: Missing code on the verification email
There is only the verify link on the email, but in some flow the app owner will prefer to have the user to input the code themselves, and therefore we should put the code in the email.

kiootic:
fix #570

### DIFF
--- a/resources/authgear/templates/__default__/messages/verification_email.mjml
+++ b/resources/authgear/templates/__default__/messages/verification_email.mjml
@@ -19,6 +19,7 @@
         This email is sent to verify {{ .Email }} on {{ template "app.name" }}. To do so, please click the button below.
       </mj-text>
       <mj-button href="{{ .URL }}" width="300px" padding="24px 0">Verify Email</mj-button>
+      <mj-text font-size="14px" color="#555555" font-weight="light" align="left">Alternatively, use this code in the verification page: {{ .Code }}</mj-text>
       <mj-text font-size="14px" color="#555555" font-weight="light" align="left">If you didn't sign in or sign up, please ignore this email.</mj-text>
     </mj-column>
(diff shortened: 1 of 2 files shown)


resources/authgear/templates/__default__/messages/verification_email.mjml
```diff
@@ -19,6 +19,7 @@
         This email is sent to verify {{ .Email }} on {{ template "app.name" }}. To do so, please click the button below.
       </mj-text>
       <mj-button href="{{ .URL }}" width="300px" padding="24px 0">Verify Email</mj-button>
+      <mj-text font-size="14px" color="#555555" font-weight="light" align="left">Alternatively, use this code in the verification page: {{ .Code }}</mj-text>
```
louischan-oursky, line 22: `Alternatively use this code in the mobile app: {{ .Code }}`.

Can we add this paragraph to the plaintext email template as well?

louischan-oursky: Can we follow the design of one time password email so that the code is more visually important?

kiootic: This message is already in the plaintext email template, and I copyed it from there.

kiootic: > Can we follow the design of one time password email so that the code is more visually important?

Not sure if the UX is better this way, since there would be two prominent elements in the email: the big button and big text. It may be confusing to user on whether use the code or click the button.

louischan-oursky: OK. Let's keep what we have now.

PR #581, opened 2020-10-23
Fix development setup

kiootic:


axolotl0212:
@kiootic got this error when creating app after updating environment variable
```json
{
	"data": null,
	"errors": [
		{
			"message": "invalid resource 'authgear.secrets.yaml': cannot override secret 'oidc' defined in base config",
			"locations": [
				{
					"line": 2,
					"column": 3
				}
			],
			"path": [
				"createApp"
			]
		}
	]
}
```

Updated some config as well?

kiootic:
I suppose the secrets in `authgear.yaml` for local mode is in conflict of the template secrets for k8s mode. Therefore we  should not symlink the k8s secret config to local secret config. Let me change that.

axolotl0212:
hack/k8s-resources/authgear.secrets.yaml missing first line `secret:`

axolotl0212 (approved):
It works

PR #584, opened 2020-10-23
Preserve state in entrypoint

### LINKED ISSUE #578: OldStateID is actually required in most interaction entrypoint
Otherwise the first error cannot be shown.

louischan-oursky:
fix #578


kiootic (commented):
If I understand correctly, there are two changes here:
- `GetIntent` -> `GetState`: ensure error of the failed operation is shown.
- Set `OldStateID` in forget password flow entry point: preserve OAuth parameters after the flow.

Am I correct about these changes?

louischan-oursky:
Yes.

If GetData does not require `graph`, I converted them to use GetState.

Otherwise set OldStateID to the original state ID if present.

The third fix is avoid a nil crash about accessing `state.Error`.

kiootic (approved):

PR #600, opened 2020-10-27
Create new url list input component

### LINKED ISSUE #535: UX problems of long URL in Allowed Origins
allowed-origins-ux

As shown in the screenshot above, the allowed origins URL would be cut off in the current input control.

This is problematic if we can't display the whole URL. (There are also security implication)

We should either: 

- Make it display the full URL no matter what, or
- Use a list view to display all added URLs

axolotl0212:
fix #535 

# Screenshot

Screenshot 2020-10-27 at 3 39 07 PM
Screenshot 2020-10-27 at 3 13 12 PM

### DIFF
--- a/portal/src/locale-data/en.json
+++ b/portal/src/locale-data/en.json
@@ -303,6 +303,7 @@
   "ModifyOAuthClientForm.redirect-uris-tooltip-message": "Users will be redirected to this path after they have authenticated.",
   "ModifyOAuthClientForm.post-logout-redirect-uris-label": "Post Logout Redirect URIs",
   "ModifyOAuthClientForm.post-logout-redirect-uris-tooltip-message": "Users will be redirected to this path after they have logged out.",
+  "ModifyOAuthClientForm.add-uri": "Add URI",
 
   "DNSConfigurationScreen.title": "DNS",
   "DNSConfigurationScreen.desc": "You can add a custom domain so that users will login via, for example, “identity.yourdomain.com”. In this case, a session cookie of domain “yourdomain.com” is written so that your websites can share the session cookies under “*.yourdomain.com” for better customers experiences.",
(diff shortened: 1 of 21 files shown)


louischan-oursky:
1. I can save an empty string in allowed origins. Can we update the schema to prevent this? Maybe we also need to prevent it in the frontend.

axolotl0212:
> 1. I can save an empty string in allowed origins. Can we update the schema to prevent this? Maybe we also need to prevent it in the frontend.

Actually can we validate the format of the origin in backend in backend? I think uri has such validation but not sure whether there is format check for origin (maybe custom `pattern` validation?). Empty string should not pass validation (can use `minLength: 1`). If we add validation in backend we can let backend handle the checking? @louischan-oursky

axolotl0212:

louischan-oursky:
Actually we can add a custom format `http_origin` to validate the format of an origin. An origin must have scheme (http or https), host, optional port, and nothing else (i.e. no path, query nor fragment). See `pkg/util/validation/formats.go` for our custom formats.

axolotl0212:
Removed unnecessary `omitempty`

axolotl0212:
> Actually we can add a custom format http_origin

Will add this custom validation

axolotl0212:
@louischan-oursky currently the allowed origin has no scheme, need another validator?

axolotl0212:
@louischan-oursky added custom validator with test case

portal/src/locale-data/en.json
```diff
@@ -55,6 +55,8 @@
   "validation-error.minItems": "Expect at least {expected} items",
   "validation-error.minimum": "Expect number larger than or equal to {minimum}",
   "validation-error.maximum": "Expect number smaller than or equal to {maximum}",
+  "validation-error.minLength": "Expect string with length greater than or equal to {expected} characters",
```
louischan-oursky, line 58: Use plural rule here.

axolotl0212:
@louischan-oursky pushed 1 more commit to use plural rules

PR #605, opened 2020-10-27
Fix state comparison in authenticator screen

### LINKED ISSUE #565: Authenticator page form dirty after edit
Cause: Multiple UI state produces same config, as config only remembers an ordered list of active authenticator. The UI state could be different with state generated from the new config after save

axolotl0212:
fix #565


kiootic (approved):
This logic should be applicable globally (form is not modified if and only if the result config is same). We may want to extract an helper function to perform this check, something like:
```ts
function isModified<State>(initialConfig: AppConfig, currentState: State, configMapper: (initial: AppConfig, current: State) => AppConfig): boolean;
```

Let's keep that in mind when authoring new screens.

PR #608, opened 2020-10-27
Implement basic rate limiting

### LINKED ISSUE #592: Introduce rate limiter

kiootic:
fix #592


pkg/lib/ratelimit/bucket.go
```diff
@@ -0,0 +1,9 @@
+package ratelimit
```
louischan-oursky, line 1: Should this package live in `pkg/util` instead?

resources/authgear/templates/__default__/translation.json
```diff
@@ -30,6 +30,7 @@
   "error-remove-last-secondary-authenticator": "Cannot remove. Multi-factor authentication is required.",
   "error-new-password-typo": "Typo in your re-typed password",
   "error-web-ui-invalid-state": "The state associated with this URL is invalid. Maybe you have idled too long or switched your browser?",
+  "error-rate-limited": "Please wait a bit before retrying",
```
louischan-oursky, line 33: `Please wait for a moment before retrying.`

pkg/lib/ratelimit/bucket.go
kiootic: The package depends on Redis, which is in `pkg/lib`.

PR #609, opened 2020-10-27
Return to previous URL when forgot password ends

### LINKED ISSUE #559: Remove "Sign in" button in forgot password success page
~~If we allow that button to be there, `prompt=login` is not respected. If the user does have a valid session, they will be redirected to the settings page. This behavior is very confusing.~~

**Edited**

But if we remove the button, in a native application, if the user click authorize but they triggered forgot password, they have to close the in-app browser once and click authorize again. This is very bad UX. Can we make the sign in button preserve the Oauth params?

louischan-oursky:
fix #559

### DIFF
--- a/pkg/auth/handler/webapp/forgot_password.go
+++ b/pkg/auth/handler/webapp/forgot_password.go
@@ -71,11 +71,12 @@ type ForgotPasswordHandler struct {
 }
 
 func (h *ForgotPasswordHandler) MakeIntent(r *http.Request) *webapp.Intent {
+	redirectURI := r.URL.Query().Get("redirect_uri")
 	return &webapp.Intent{
 		RedirectURI: "/forgot_password/success",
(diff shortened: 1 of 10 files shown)


pkg/auth/handler/webapp/forgot_password.go
```diff
@@ -71,11 +71,12 @@ type ForgotPasswordHandler struct {
 }
 
 func (h *ForgotPasswordHandler) MakeIntent(r *http.Request) *webapp.Intent {
+	redirectURI := r.URL.Query().Get("redirect_uri")
```
kiootic, line 74: We should validate whether it's the same origin before using it.

louischan-oursky: It is actually done at the use-site. Only the Path and Query is used. Other url components are ignored.

kiootic (approved):

PR #618, opened 2020-10-28
Update dev setup instruction

kiootic:
@elly0t Updated README according to yesterday's setup troubleshooting session. Please let me know if I miss anything.

### DIFF
--- a/portal/README.md
+++ b/portal/README.md
@@ -2,7 +2,7 @@
 
 # Prerequisite
 
-Follow [authgear setup guide](../README.md) to start Authgear
+Follow [authgear setup guide](../README.md) to start Authgear.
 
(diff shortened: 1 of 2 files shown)


portal/README.md
```diff
@@ -22,6 +22,7 @@ This container listen to port 8000. For redirection configuration, please refer
 ## Setup environment variable
 
 We need to set up the environment variables:
+- `AUTHGEAR_APP_ID=accounts`
```
louischan-oursky, line 25: It is better to just tell the reader to copy either `.env.example` or `.env.example.k8s`

kiootic: Sure, let me delete the whole list of environment variables in portal README.

PR #619, opened 2020-10-28
Handle missing OAuth provider client secret

### LINKED ISSUE #548: Validate oauth provider config when save

nicolaslee2775:
fix #548 

image

### DIFF
--- a/pkg/lib/config/secret.go
+++ b/pkg/lib/config/secret.go
@@ -126,7 +126,12 @@ func (c *SecretConfig) Validate(appConfig *AppConfig) error {
 					}
 				}
 				if !found {
-					ctx.EmitErrorMessage(fmt.Sprintf("OAuth client credentials for '%s' is required", p.Alias))
+					ctx.Child("secrets", "sso.oauth.client", p.Alias).EmitError(
+						"required",
+						map[string]interface{}{
+							"missing": "client_secret",
+						},
+					)
 				}
 			}
 		}
--- a/pkg/lib/config/secret.go
+++ b/pkg/lib/config/secret.go
@@ -86,26 +86,33 @@ func (c *SecretConfig) Overlay(layers ...*SecretConfig) *SecretConfig {
 	return merged
 }
 
-func (c *SecretConfig) Lookup(key SecretKey) (*SecretItem, bool) {
-	for _, item := range c.Secrets {
+func (c *SecretConfig) Lookup(key SecretKey) (*SecretItem, bool, int) {
+	for index, item := range c.Secrets {
 		if item.Key == key {
-			return &item, true
+			return &item, true, index
 		}
 	}
-	return nil, false
+	return nil, false, -1
 }
 
 func (c *SecretConfig) LookupData(key SecretKey) SecretItemData {
-	if item, ok := c.Lookup(key); ok {
+	if item, ok, _ := c.Lookup(key); ok {
 		return item.Data
 	}
 	return nil
 }
 
+func (c *SecretConfig) LookupDataWithIndex(key SecretKey) (SecretItemData, int) {
+	if item, ok, index := c.Lookup(key); ok {
+		return item.Data, index
+	}
+	return nil, -1
+}
+
 func (c *SecretConfig) Validate(appConfig *AppConfig) error {
 	ctx := &validation.Context{}
 	require := func(key SecretKey, item string) {
-		if _, ok := c.Lookup(key); !ok {
+		if _, ok, _ := c.Lookup(key); !ok {
 			ctx.EmitErrorMessage(fmt.Sprintf("%s (secret '%s') is required", item, key))
 		}
 	}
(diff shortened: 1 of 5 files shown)


portal/src/graphql/portal/SingleSignOnConfigurationScreen.tsx
```diff
@@ -428,10 +428,15 @@ const SingleSignOnConfigurationWidgetWrapper: React.FC<WidgetWrapperProps> = fun
       : "";
   }, [providerIndex]);
 
+  const clientSecretJsonPointer = useMemo(() => {
+    return `/secrets/sso.oauth.client/${providerType}`;
```
axolotl0212: In backend this is `/secrets/sso.oauth.client/{alias}`, `alias` might not be the same as `providerType`

pkg/lib/config/secret.go
```diff
@@ -116,22 +123,30 @@ func (c *SecretConfig) Validate(appConfig *AppConfig) error {
...
-						},
-					)
+				if matchedItem == nil {
+					ctx.EmitErrorMessage(fmt.Sprintf("OAuth client credentials for '%s' is required", p.Alias))
+				} else {
+					if matchedItem.ClientSecret == "" {
+						ctx.Child("secrets", fmt.Sprintf("%d", secretIndex), "data", "items", fmt.Sprintf("%d", matchedItemIndex)).EmitError(
+							"required",
+							map[string]interface{}{
+								"missing": "client_secret",
```
axolotl0212, line 146: missing is string array in other "required" error, with actual and expected attribute. I suggest keeping the error shape consistent such that the type definition in frontend is valid

portal/src/graphql/portal/SingleSignOnConfigurationScreen.tsx
```diff
@@ -124,6 +124,11 @@ function getOrCreateSecret(secretConfigState: PortalAPISecretConfig) {
       data: { items: [] },
     };
     secretConfigState.secrets.push(secret);
+    // Sync secret item order with server side.
+    // Otherwise the error json pointer would not match.
+    secretConfigState.secrets = secretConfigState.secrets.sort((a, b) =>
```
axolotl0212, line 129: Why do we need this sorting? the json pointer should refer to the new secret config sent from portal, so we just need to find the index with secretConfigState and it should always sync with server side no matter what order `secretConfigState.secrets` is in?

nicolaslee2775: In server side, the secret items are sorted by their key.

pkg/lib/config/secret.go
```diff
@@ -86,26 +86,33 @@ func (c *SecretConfig) Overlay(layers ...*SecretConfig) *SecretConfig {
 	return merged
 }
 
-func (c *SecretConfig) Lookup(key SecretKey) (*SecretItem, bool) {
-	for _, item := range c.Secrets {
+func (c *SecretConfig) Lookup(key SecretKey) (*SecretItem, bool, int) {
```
louischan-oursky, line 89: If there is a bool in the return value to represent ok or not, the convention is to make it the last element in the return value tuple. So the return type should be `(int, *SecretItem, bool)`. Index is the first item so that it is consistent with range loop.

pkg/lib/config/secret.go
```diff
@@ -86,26 +86,33 @@ func (c *SecretConfig) Overlay(layers ...*SecretConfig) *SecretConfig {
...
 
 func (c *SecretConfig) LookupData(key SecretKey) SecretItemData {
-	if item, ok := c.Lookup(key); ok {
+	if item, ok, _ := c.Lookup(key); ok {
 		return item.Data
 	}
 	return nil
 }
 
+func (c *SecretConfig) LookupDataWithIndex(key SecretKey) (SecretItemData, int) {
```
louischan-oursky, line 105: It is better to add a last bool in the return value.

pkg/lib/config/secret.go
nicolaslee2775: Fixed

pkg/lib/config/secret.go
nicolaslee2775: Fixed

pkg/lib/config/secret.go
nicolaslee2775: Fixed

nicolaslee2775:

axolotl0212 (approved):

nicolaslee2775:
@louischan-oursky Changed to use regexp for client secret json pointer and removed secret item sorting.

louischan-oursky:
Screen Shot 2020-10-30 at 10 12 15 AM

Still got this result.

nicolaslee2775:
@louischan-oursky Fixed.
image

nicolaslee2775:
@louischan-oursky Handled single cause with multiple missing fields.
image

nicolaslee2775:
Some error message would be displayed twice
image

nicolaslee2775:
@louischan-oursky Fixed.
image

axolotl0212:
@nicolaslee2775 Just curious, as validation error cause has location (json pointer), and we expect each pointer corresponds to 1 single field, why do we need to match 1 cause with multiple fields?

nicolaslee2775:
@axolotl0212 Sometime cause contains multiple missing fields.
image

axolotl0212:
I see, `required` error with multiple missing fields while the json pointer refers to parent...

Issue #549, opened 2020-10-21
Reset using ModifiedIndicator does not reset Apollo error state


axolotl0212:
Currently, resetting state using `ModifiedIndicator` does not reset error state stored in Apollo client.

- [x] Investigate how to reset error state in Apollo client, is it necessary to `useState` and `useEffect` in each query hook to achieve reset error state
- [x] Add resetErrorState to each mutation hook, consider wrapping useMutation to provide this function?
- [x] For each resetForm function, reset error state using resteErrorState provided by query hook

axolotl0212:
fixed via pr #628

PR #628, opened 2020-10-28
Reset error state on switching tab and reset form state

### LINKED ISSUE #576: Error bar sometimes have no content
1. SSO -> enable any one of them and press save without further configuration
Screenshot 2020-10-23 at 14 51 17

2. Try to save an incorrect configuration on Authentication > Authenticator
Screenshot 2020-10-23 at 14 52 18

axolotl0212:
fix #576 #549

### DIFF
--- a/portal/src/graphql/adminapi/AddEmailScreen.tsx
+++ b/portal/src/graphql/adminapi/AddEmailScreen.tsx
@@ -22,11 +22,18 @@ import { useGenericError } from "../../error/useGenericError";
 
 import styles from "./AddEmailScreen.module.scss";
 
+interface AddEmailFormProps {
+  resetForm: () => void;
+}
+
 interface AddEmailFormData {
   email: string;
 }
 
-const AddEmailScreen: React.FC = function AddEmailScreen() {
+const AddEmailForm: React.FC<AddEmailFormProps> = function AddEmailForm(
+  props: AddEmailFormProps
+) {
+  const { resetForm } = props;
   const { userID } = useParams();
   const navigate = useNavigate();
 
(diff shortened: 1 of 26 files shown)


portal/src/hook/usePivot.ts
```diff
@@ -7,7 +7,8 @@ function isHashValid(validItemKeys: string[], hash: string): boolean {
 }
 
 export function usePivotNavigation(
-  validItemKeys: string[]
+  validItemKeys: string[],
+  switchTabEffect?: () => void
```
louischan-oursky, line 11: It is better to call this `onSwitchTab`

portal/src/graphql/adminapi/AddEmailScreen.tsx
```diff
@@ -63,8 +64,9 @@ const AddEmailScreen: React.FC = function AddEmailScreen() {
   }, [initialFormData, formData]);
 
   const resetForm = useCallback(() => {
+    resetCreateIdentityError();
     setFormData(initialFormData);
-  }, [initialFormData]);
+  }, [initialFormData, resetCreateIdentityError]);
```
louischan-oursky, line 69: Can we use `key` to force re-mount of the component to implement form reset?

nicolaslee2775: Currently, only `PasswordlessAuthenticatorScreen` and `PasswordsScreen` use key to force re-mount components. Maybe create new issue for updating all other screens?

louischan-oursky: I prefer using key to re-mount the form to reset the form. It is simpler and less error-prone. Would a new abstraction (I don't know, maybe a hook) be helpful here? Maybe you guys can discuss and come up with a solution.

axolotl0212: For some of the screen we can use key to re-mount and reset the form, but for some other screen, the mutation hook is called on screen level, so remounting the form does not reset the error state (for example screen with tab). Then we will need to change the component structure (say introduce another wrapper element, call mutation in its child and remount its child on reset).

nicolaslee2775: Use can keep `resetCreateIdentityError()` and change `setFormData(initialFormData)` to `setRemountIdentifier((prev) => prev + 1)`.

axolotl0212:
@louischan-oursky @nicolaslee2775 updated and rebased. For some screen we will use resetError as @nicolaslee2775 suggested, and use key to remount form for all screen to reset form state

PR #660, opened 2020-10-29
Fix back button no response in page with tab

axolotl0212:
fix https://github.com/authgear/authgear-deploy/issues/28

Cause:
This is caused by redirection if the fragment (hash) is not valid (each tab has a key, selected tab specified by fragment), as this affects the history, for example:
`previous page` -> `/app/:appID/configuration/authentication` -> `/app/:appID/configuration/authentication#login-id`

So back button has no effect, including the confirm button in navigation blocker dialog.

Fix:

Use `replaceState` to avoid pushing to history stack

### DIFF
--- a/portal/src/hook/usePivot.ts
+++ b/portal/src/hook/usePivot.ts
@@ -22,7 +22,8 @@ export function usePivotNavigation(
 
   useEffect(() => {
     if (!isHashValid(validItemKeys, hash)) {
-      navigate(`#${initialSelectedKey}`);
+      // NOTE: avoid adding extra entry to history stack
+      navigate(`#${initialSelectedKey}`, { replace: true });
     }
   }, [validItemKeys, hash, initialSelectedKey, navigate]);
 


portal/src/hook/usePivot.ts
```diff
@@ -22,7 +22,8 @@ export function usePivotNavigation(
 
   useEffect(() => {
     if (!isHashValid(validItemKeys, hash)) {
-      navigate(`#${initialSelectedKey}`);
+      // NOTE: avoid adding extra entry to history stack
+      window.history.replaceState(undefined, "", `#${initialSelectedKey}`);
```
louischan-oursky, line 26: This is incorrect. If we want to replace, `navigate` accepts options in the second argument. See the type definition for more details.

louischan-oursky (changes requested):

portal/src/hook/usePivot.ts
axolotl0212: @louischan-oursky fixed with `{replace: true}`

PR #672, opened 2020-10-29
Add error handling for invalid app ID

axolotl0212:
fix https://github.com/authgear/authgear-deploy/issues/32

Motivation:

- Avoid error like unhandled forbidden error
- App ID which has valid base64 encoding but not valid UTF8 encoding for Postgres

Screenshot 2020-10-29 at 4 18 31 PM

- App ID with incorrect encoding leads to empty response without error

Screenshot 2020-10-29 at 4 17 55 PM

### DIFF
--- a/pkg/portal/graphql/app.go
+++ b/pkg/portal/graphql/app.go
@@ -147,9 +147,10 @@ var nodeApp = node(
 	&model.App{},
 	func(ctx context.Context, id string) (interface{}, error) {
 		gqlCtx := GQLContext(ctx)
+		// return nil without error for both inaccessible / not found apps
 		_, err := gqlCtx.AuthzService.CheckAccessOfViewer(id)
 		if err != nil {
-			return nil, err
+			return nil, nil
 		}
 		return gqlCtx.Apps.Load(id).Value, nil
 	},
--- a/portal/src/AppRoot.tsx
+++ b/portal/src/AppRoot.tsx
@@ -1,8 +1,11 @@
 import React, { useMemo } from "react";
 import { Routes, Route, useParams, Navigate } from "react-router-dom";
 import { ApolloProvider } from "@apollo/client";
+
 import { makeClient } from "./graphql/adminapi/apollo";
+import { useAppConfigQuery } from "./graphql/portal/query/appConfigQuery";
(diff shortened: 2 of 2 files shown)


pkg/portal/graphql/app.go
```diff
@@ -147,11 +148,18 @@ var nodeApp = node(
 	&model.App{},
 	func(ctx context.Context, id string) (interface{}, error) {
 		gqlCtx := GQLContext(ctx)
+		// check for existence of app
```
kiootic, line 151: I don't think this change is needed? The data loader should translate the errors to null data, so client-side only need to check whether the node is null.

axolotl0212: Non-existing app will trigger `Forbidden` error, so this is used to distinguish `not exist` from `no access`

axolotl0212: It also prevent the error in first screenshot

kiootic: We should be allow client-side to distinguish between a non-existant app and an inaccessible app to prevent information leakage. Instead, treat both cases as not found.

For the pg error, I think we can use `utf8.ValidString` to validate all decode IDs.

axolotl0212: `utf8.ValidString` cannot block `0x00` (when cast string seems like `0x00` is omitted), I changed to throw forbidden if app is not on list

axolotl0212:
Rebased and updated @kiootic

pkg/portal/graphql/app.go
louischan-oursky: We should keep simply returning null for non-existant app and inaccessible app.

axolotl0212: > simply returning null for non-existant app and inaccessible app

no error for inaccessible app?

kiootic: Yes, there should be no errors returned from API for inaccessible/not found apps.

axolotl0212:
Updated @louischan-oursky @kiootic

louischan-oursky (approved):

kiootic (approved):

PR #678, opened 2020-10-29
Fix phone number validation

axolotl0212:
fix https://github.com/authgear/authgear-deploy/issues/31

### DIFF
--- a/pkg/util/phone/phone.go
+++ b/pkg/util/phone/phone.go
@@ -3,6 +3,7 @@ package phone
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -11,6 +12,9 @@ import (
 
 // ErrNotInE164Format means the given phone number is not in E.164 format.
 var ErrNotInE164Format = errors.New("not in E.164 format")
+
+// ErrPhoneNumberInvalid means phone number doesn't pass validation
+var ErrPhoneNumberInvalid = errors.New("invalid phone number")
 
 // EnsureE164 ensures the given phone is in E.164 format.
 func EnsureE164(phone string) error {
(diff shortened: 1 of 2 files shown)


pkg/util/phone/phone.go
```diff
@@ -41,6 +51,29 @@ func Parse(nationalNumber string, callingCode string) (e164 string, err error) {
 	return
 }
 
+func ParseWithValidation(nationalNumber string, callingCode string) (e164 string, err error) {
```
louischan-oursky, line 54: I think we do not need both Parse and ParseWithValidation. Make Parse behave like ParseWithValidation. Parse was always meant to return validated phone number in E.164 format.

louischan-oursky: Can you also add test cases to verify if the sibling function EnsureE164 accepts invalid input like `+85222` or `+85222a`? Please help fix it if it is the case.

louischan-oursky: BTW, the two commits should be a single commit. It is more logical to have the new test cases committed along with the changed behavior.

axolotl0212: I expect +85222 passes EnsureE164 as it can be parsed, and formatted correctly, +85222a should fail the test, as even it can be parsed, formatting from uint64 national number gives number, hence not equal to original. You want EnsureE164 to do phone number validation as well? This means phone number like +85212345678 will fail the check as well, as +852 number do not begins with 1 for example.

louischan-oursky: Yes. For our use case, we never accept invalid phone number. We use phone numbers to deliver SMS message so they must be valid.

axolotl0212: OK, I will call `isValidNumber` in `EnsureE164`

axolotl0212:
@louischan-oursky rebased and updated, but we might need to update the error message.

Screenshot 2020-10-30 at 11 47 14 AM

PR #685, opened 2020-10-30
Handle error in the first step in the interaction

### LINKED ISSUE #684: Investigate why sometimes back button would result in invalid state

louischan-oursky:
fix #684


kiootic (commented):
Not sure if the fix is secure enough: suppose an attacker obtains a URL with correct SID, the attacker may be able to extract sensitive informations just from the rendered page (e.g. phone number, email, CSRF token), which enable further attacks.

I think we'll need a more robust way to handle error presentation. Is this fix urgent?

louischan-oursky:
Not very urgent though.

PR #687, opened 2020-10-30
Add checking for add identity form

axolotl0212:
ref #680 

# screenshot

Screenshot 2020-10-30 at 4 46 19 PM

### DIFF
--- a/portal/src/graphql/adminapi/AddEmailScreen.module.scss
+++ b/portal/src/graphql/adminapi/AddEmailScreen.module.scss
@@ -3,16 +3,21 @@
   min-height: 100%;
 }
 
-.breadcrumb {
-  margin: 0 25px;
-  padding-top: 20px;
+.wrapper {
+  padding: 10px 20px 40px;
+}
+
+.helpText {
+  display: block;
+  margin: 10px 20px;
+  font-size: 15px;
 }
 
 .content {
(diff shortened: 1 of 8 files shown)


louischan-oursky:
Conflicted. BTW, does this PR fix the problem described in #680 ?

axolotl0212:
> does this PR fix the problem described in #680

@kiootic mentioned that if `login_id` is not included in identities, then user cannot create identity, so this PR add checking for login_id identity, show message to explain the situation if `login_id` is not included (for user going into page by URL). The add identity button is already disabled before.

Also, this case (`login_id` not included in identities) is not handled by backend (unhandled error is thrown), may be we need to handle the error in backend as well? As we cannot stop user from calling the API directly (not using portal)

louischan-oursky:
We definitely need to handle password is required when new identity is added. So this PR does not fix #680 actually. Let me change the description.

axolotl0212:
@louischan-oursky rebased

kiootic (approved):

axolotl0212:
> We definitely need to handle password is required when new identity is added. So this PR does not fix #680 actually. Let me change the description.

@louischan-oursky opened PR https://github.com/authgear/authgear-server/pull/697 to fix this

PR #691, opened 2020-10-30
Fix SSO screen modified indicator not cleared after save

### LINKED ISSUE #663: [SSO] Unsaved change bar still show after clicking the save button
Step to reproduce:
1. Create a new app
2. In SSO, set credentials for any one of the items in the list
3. Click save button
4. Will still see the unsaved change bar shown

axolotl0212:
fix #663 

Cause:

In effective app config, an extra field "claims" is added to oauth provider in
/identity/oauth/providers. We use the app config directly as our screen
state.

When new provider is added, only the fields appear on portal sso screen is added to state,
so getting initial state after save from effective app config will get extra field.

Fix:

use raw app config for initial screen state.


louischan-oursky:
Conflicted.

axolotl0212:
@louischan-oursky rebased

PR #697, opened 2020-11-02
Show password field for all add identity form if needed

### LINKED ISSUE #680: [Portal] Add login ID page does not prompt for password for user with OAuth identity only

axolotl0212:
require #687 
fix #680


axolotl0212:
rebased

PR #700, opened 2020-11-02
Fix stale host map when ingress is deleted

### LINKED ISSUE #674: Expire all relevant k8s config source cache data
We currently only expire `appMap` in k8s config source, we should also expire entries in `hostMap`, etc.

louischan-oursky:
fix #674

### DIFF
--- a/pkg/lib/config/configsource/kubernetes.go
+++ b/pkg/lib/config/configsource/kubernetes.go
@@ -146,6 +146,7 @@ func (k *Kubernetes) invalidateHostMap(ingress *networkingv1beta1.Ingress) {
 			k.Logger.WithField("host", host).Info("host invalidated")
 			k.hostMap.Delete(host)
 		}
+		k.ingressMap.Delete(ingress.UID)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/lib/config/configsource/kubernetes.go
```diff
@@ -155,7 +156,13 @@ func (k *Kubernetes) invalidateHostMap(ingress *networkingv1beta1.Ingress) {
 }
 
 func (k *Kubernetes) updateHostMap(appID string, ingress *networkingv1beta1.Ingress) {
-	for _, host := range extractIngressHosts(ingress) {
+	hosts := extractIngressHosts(ingress)
+
+	k.ingressMap.Store(ingress.UID, &ingressSnapshot{
```
kiootic, line 161: We should delete the previous hosts, for example, when ingress hosts are updated from `["test1.localhost"]` to `["test2.localhost"]`, `test1.localhost` should be deleted.

louischan-oursky: fixed.

kiootic (approved):

PR #701, opened 2020-11-02
Fix missing validation for login ID form

axolotl0212:
fix https://github.com/authgear/authgear-deploy/issues/40

Add new rule to `enterLoginIDSchema` which requires `x_national_number` if
`login_id_key` is phone

Screenshot 2020-11-02 at 4 31 44 PM


axolotl0212:
rebased

PR #703, opened 2020-11-02
Only show reuse password policy when creating primary password

axolotl0212:
fix https://github.com/authgear/authgear-deploy/issues/42


resources/authgear/templates/__default__/web/__password_policy.html
```diff
@@ -1,4 +1,5 @@
 {{ define "__password_policy.html" }}
+{{ $isNew := .IsNew }}
```
louischan-oursky, line 2: This declaration is unnecessary. If we want to reference the fields of the root data object, we can use `{{ $.IsNew }}`.

axolotl0212: Removed the declaration

PR #709, opened 2020-11-03
Fix horizontal scrollbar appears in country code list on windows browser

### LINKED ISSUE #646: [Portal, Authentication] Unnecessary scroll bar is shown in phone number search result
## Environment:
- Platform: Web Portal
- Browser: Chrome on Win 10

## Description:
An unnecessary horizontal scroll is found in the phone number search result area.

## Steps to reproduce:
1. Log in to the Authgear portal
2. Select any app entity
3. Go to the 'Authentication' page and enable the phone number authenticator
4. Scroll to the bottom of the country code listings and see the unnecessary horizontal scroll 

## Expected result:
Should remove the unnecessary scroll bar

## Screenshot:
image

nicolaslee2775:
fix #646

It seem to be fluent ui bug and not yet fixed (https://github.com/microsoft/fluentui/issues/3608).

The unnecessary horizontal scrollbar appears in DetailsList as it has `overflow-x: auto` style. So I simply added `overflow-x: hidden` to hide horizontal scrollbar. (I have tried setting `overflow-x: visible` to DetailsList, but the horizontal scrollbar would appears in scrollable pane.)

##### Before
authgear_portal_646-1
##### After
authgear_portal_646-2

### DIFF
--- a/portal/src/graphql/portal/AuthenticationCountryCallingCodeList.module.scss
+++ b/portal/src/graphql/portal/AuthenticationCountryCallingCodeList.module.scss
@@ -25,3 +25,8 @@
 .lastPinnedCallingCode {
   border-bottom: 1px solid #c4c4c4;
 }
+
+// Fix horizontal scrollbar appears on windows browser
+.detailsList {
+  overflow-x: hidden;
+}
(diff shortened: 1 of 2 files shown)


axolotl0212 (approved):
This fix works, ~but did you check whether the list is wider than the widget for some reason (say maxWidth of column is not restricted)?~ Seems like the width of each row is computed by the fluentUI and it is always longer than the list itself, as the y scroll bar width is not taken into account

PR #711, opened 2020-11-04
Add description to change login ID page

nicolaslee2775:
fix authgear/authgear-deploy#48

image
image
image

### DIFF
--- a/pkg/lib/authn/identity/service/service.go
+++ b/pkg/lib/authn/identity/service/service.go
@@ -395,26 +395,6 @@ func (s *Service) CheckDuplicated(is *identity.Info) (dupeIdentity *identity.Inf
 	return
 }
 
-func (s *Service) GetLoginIDDisplayID(userID string, loginIDKey string) (displayID string, err error) {
-	var loginIDs []*loginid.Identity
-
-	if userID != "" {
-		loginIDs, err = s.LoginID.List(userID)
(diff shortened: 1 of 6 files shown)


axolotl0212 (approved):

pkg/lib/authn/identity/service/service.go
```diff
@@ -395,6 +395,26 @@ func (s *Service) CheckDuplicated(is *identity.Info) (dupeIdentity *identity.Inf
...
+
+	for _, iden := range loginIDs {
+		if iden.LoginIDKey == loginIDKey {
+			displayID = loginIDToIdentityInfo(iden).DisplayID()
+		}
+	}
+
+	return
+}
+
```
louischan-oursky, line 417: Actually no need to add this function. We should reuse the existing `func (s *Service) Get(userID string, typ authn.IdentityType, id string) (*identity.Info, error)`.

userID: You got it already.
typ: authn.IdentityTypeLoginID
id: the value of r.Form.Get("x_identity_id")

Of course we need to handle the identity not found error.

nicolaslee2775:
@louischan-oursky Updated.

louischan-oursky:
It crashes when I try to add a phone number (I did not have one previously)

nicolaslee2775:
@louischan-oursky Fixed.

PR #715, opened 2020-11-09
Introduce Turbolinks to improve navigation behavior

### LINKED ISSUE #665: Improve Auth UI navigation handling
We may want to use other battle-tested libraries to handle browser history & navigation in Auth UI.

ref: https://github.com/turbolinks/turbolinks

### LINKED ISSUE #668: Press back button on Connected Identities page cannot back to main setting page
Step to reproduce:
1. Signup an account with the application have SSO configured
2. Go to Connected Identities
3. Try connecting to a SSO that will result login failed
4. Press back button on the Connected Identities page will back to the SSO page instead of back to the previous Authgear setting page

kiootic:
fix #665 
fix #668


authui/src/authgear.ts
```diff
@@ -8,6 +8,15 @@ window.api.onLoad(() => {
   document.body.classList.add("js");
 });
 
+// Handle history tracking.
+
+let inHistorySettings = false;
+window.api.onLoad(() => {
+  if (window.location.pathname === "/settings") {
```
louischan-oursky, line 15: I just realized `/change_password` is not under `/settings`.

louischan-oursky (commented):
Now we have to build the static asset to make the server functional. Do you have any suggested workflow? Can we document it in README or even include it in Makefile?

authui/src/authgear.ts
kiootic: I suppose we can move it under settings, so it'll be `/settings/change_password` and `/settings/change_secondary_password`.

I just remembered `/enter_login_id`, its name looks like part of interaction flow, yet it's part of settings page. Should we change its name too?

kiootic:
> Now we have to build the static asset to make the server functional. Do you have any suggested workflow? Can we document it in README or even include it in Makefile?

I suppose we can invoke `$(MAKE) static` in `make vendor` script to make setup more convenient.

authui/src/authgear.ts
louischan-oursky: > I suppose we can move it under settings, so it'll be /settings/change_password and /settings/change_secondary_password.

Agree.

> I just remembered /enter_login_id, its name looks like part of interaction flow, yet it's part of settings page. Should we change its name too?

Agree. Let's change from `/enter_login_id` to `/settings/enter_login_id`?

Issue #150, opened 2020-08-10
Delete User


chpapa:
* Admin API
    * Anything should do to help developers delete data on their own system after Delete User?
* Portal UI
* Check anything else for GDPR compliance?

Issue #148, opened 2020-08-10
Disable User


chpapa:
* Admin API
* UI Portal
* Effect: Prevent login, show error code with disabled reason

PR #716, opened 2020-11-10
Add spec for deleting/disabling users

kiootic:
ref #148 
ref #150

### DIFF
--- a/docs/specs/delete-user.md
+++ b/docs/specs/delete-user.md
@@ -21,3 +21,5 @@ until its natural expiry.
 - Web-hooks
 - Reserve deleted login ID
 - Self-service deletion
+- Soft-delete
+- Delayed deletion (i.e. soft-delete for N days, then hard-delete)
--- a/docs/specs/delete-user.md
+++ b/docs/specs/delete-user.md
@@ -1,0 +1,23 @@
+# Delete User
+
+  * [Deleting user](#deleting-user)
+  * [Cached data](#cached-data)
+  * [Future works](#future-works)
+
+## Deleting user
+
+A user can be deleted by admins. A deleted user would have all its associated
+data (exposed from API) erased from database (hard-delete), including
+identities, authenticators, password data, sessions, etc. 
+
+## Cached data
+
+Some internal data may be present in cache (Redis), such as OAuth states,
+MFA device tokens, rate limit counter. These data would remain in the cache
+until its natural expiry.
+
+## Future works
+
+- Web-hooks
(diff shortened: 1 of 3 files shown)


docs/specs/disable-user.md
```diff
@@ -0,0 +1,37 @@
...
+
+Admin may optionally provide a reason when disabling user. This reason would be
+shown when user attempted to login.
+
+## Disabled user session
+
+When a user is disabled, the user can no longer create new sessions. However,
+there may be existing logged in sessions.
+
+The authentication tokens (i.e. sessions, access token, refresh tokens, etc.)
```
louischan-oursky, line 21: Do we update all token at once or do we check user disabled status every time we resolve a token?

docs/specs/delete-user.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Deleting user
+
+A user can be deleted by admins. A deleted user would have all its associated
+data erased from database (hard-delete), including identities, authenticators,
+password data, etc. 
+
+## Cached data
+
+Some data may be present in caches (Redis), such as sessions, rate limit
```
louischan-oursky, line 15: We should also remove sessions as part of the deletion process, as we treat sessions as entity as well.

docs/specs/disable-user.md
kiootic: I intended to check disabled status for every resolver request for accuracy. Consumer may cache the resolver result/use planned JWT access token for performance.

docs/specs/delete-user.md
```diff
@@ -0,0 +1,23 @@
...
+
+Some internal data may be present in cache (Redis), such as OAuth states,
+MFA device tokens, rate limit counter. These data would remain in the cache
+until its natural expiry.
+
+## Future works
+
+- Web-hooks
+- Reserve deleted login ID
+- Self-service deletion
```
chpapa, line 23: A common feature needed is "Soft delete for N days and delete", I think it sort of like "Disable for N days and then delete" in our case. Maybe put it in future work so it is considered? (Although it should be relatively trivial)

chpapa: BTW, same applies to self-service deleteion too, for a lot of consumer apps the Business would prefer to let the user delete but have a 30 days period to revert its decision. In those case we need to display a better default copywriting of reason for disable users for this case.

chpapa (approved):
Doesn't see any problem besides want to add a note on future work.

PR #721, opened 2020-11-11
Multilingual template support

### LINKED ISSUE #718: Multilingual messages configuration on portal
Wireframe: [Multilingual templates.pdf](https://github.com/authgear/authgear-server/files/5521367/Multilingual.templates.pdf)

axolotl0212:
require #720 
closes #718 

- [x] Move template to separate screen
- [x] Handle fetch template with different localization
- [x] Add dropdown for choosing language from a list of supported language (default = effective default language)
- [x] Add dropdown for choosing default language (default english)
- [x] Add action button and dialog for managing template locales
- [x] Handle parse supported locales from configured paths
- [x] Handle delete all template with certain locale
- [x] Modify backend to accept path with matching format with supported locale


portal/src/util/stringTemplate.ts
```diff
@@ -0,0 +1,75 @@
...
+export function renderTemplateString(
+  values: Record<string, string>,
+  template: string
+): string {
+  return Object.entries(values).reduce((result, [key, value]) => {
+    return applyArgumentOnTemplateString(key, result, value);
+  }, template);
+}
+
+export function parseTemplateString(
```
kiootic, line 24: This function seems rather complicated for what we need.
I suggest try to hook into template string mechanism for this. Something like:
```ts
function template<ArgKey extends string>(parts: readonly string[], ...args: readonly ArgKey[]) {
    // should receive { parts: ["templates/", "/messages/test.html"], args: ["locale"] }
}
const localeTemplate = template`templates/${'locale'}/messages/test.html`;
```

portal/src/graphql/portal/TemplateLocaleManagement.tsx
```diff
@@ -32,9 +34,12 @@ interface TemplateLocaleManagementProps {
   defaultTemplateLocale: TemplateLocale;
   onTemplateLocaleSelected: TemplateLocaleUpdater;
   onDefaultTemplateLocaleSelected: TemplateLocaleUpdater;
+  pendingTemplateLocales: TemplateLocale[];
```
kiootic, line 38: I suppose `pendingTemplateLocales` is used to indicate new template locales? The interaction between locale delete & add seems rather complex.

I would suggest having a single source of truth for the templates & locales, and have just a `templateUpdates: Record<string, string | null>` in the state:
- deleting a locale means setting `templateUpdates["template/path"] = null` for all templates of the locale
- adding a locale means setting `templateUpdates["template/path"] = ""` for all known templates of the locale
- the templates are merged rom the configured templates and the `templateUpdates` state. (null content means deleted)
- the locale list would be derived (useMemo) from the template paths.
- empty content would be translated to deletion on submission.
- dismissing the locale management dialog would not make request to server.

axolotl0212: > I suppose pendingTemplateLocales is used to indicate new template locales?

That's correct

> dismissing the locale management dialog would not make request to server.

Any thought @louischan-oursky ? If it is OK to only send request on save (the apply button in dialog only changes screen state), then having `templateUpdates: Record<string, string | null>` in state is cleaner and simpler

portal/src/util/stringTemplate.ts
axolotl0212: @kiootic I understand this can be used for rendering the string from template, but not sure how to use this for parsing string with template

kiootic: IIRC `parts.length` is always `args.length + 1`, so:
- for each argument `arg` with index `i` in args:
    - check if the input starts with `parts[i]`: if so remove it; otherwise reject the input.
    - find next occurrence of `parts[i+1]` in input: if exists, assign the substring between the two parts to `arg`; otherwise reject the input.

axolotl0212: I see, using this to replace first part of `parseTemplateString`. So we are using this object
```
{ parts: ["templates/", "/messages/test.html"], args: ["locale"] }
```
as template? @kiootic

kiootic: I'm thinking of returning the parser and renderer as part of tagged template function, something like:
```ts
interface ResourcePath<Arg> {
    parse(path: string): Record<Arg, string> | null;
    render(args: Record<Arg, string>): string;
}
function resourcePath<Arg extends string>(parts: readonly string[], ...args: readonly Arg[]): ResourcePath<Arg> {
    // ...
}

const localizedEmailMessageHTML = resourcePath`templates/${"locale"}/messages/${"type"}_email.html`;
localizedMessageHTML.render({ locale: "en", type: "forgot_password" });
```

axolotl0212: I see, will implement this

louischan-oursky (commented):
When I tried to remove the default template language, the error message says "Cannot remove last template language".

Some improvements we can have here

1. Disable the delete button for default template language with a hovering tooltip explaining that default template language cannot be removed.
2. Update the error message to be "Cannot remove the default template language"

axolotl0212:
> Disable the delete button for default template language with a hovering tooltip explaining that default template language cannot be removed

Disallow removing default language? I only blocked removing last configured language

louischan-oursky:
> Disallow removing default language? I only blocked removing last configured language

Removing the default language is not a very sensible action. We should forbid that.

axolotl0212:
> Removing the default language is not a very sensible action. We should forbid that.

OK

portal/src/graphql/portal/TemplateLocaleManagement.tsx
axolotl0212: @kiootic Actually shall we fetch all template in all configured locale? I think it is too complicated to fetch templates on locale is selected if we are using `templateUpdates: Record<string, string | null>`?

kiootic: I agree. I would prefer that once the screen is loaded, no further requests are made until save/discard changes.

axolotl0212: Actually I run into another problem, we do not know the configured locale before fetching template @kiootic Split it into 2 query?

kiootic: I see, let's just use two queries for simplicity for now, one to fetch all resource paths, another to fetch relevant resources.

axolotl0212:
@kiootic @louischan-oursky updated

louischan-oursky (commented):
1. Previously, if the template text field is empty, we can click save button. The behavior is to delete the template. Now when the template text field is empty, save button is disabled.
2. In Add/Remove language dialog, the add-language behavior is fine. Click apply does not trigger API call. However, it seems unintuitive to me that if I remove a language and then click apply, the language is not removed yet. I have to click the save button below to take effect.
3. If I am viewing the template of language X and I remove language X, we should change the selected language to the next one.

axolotl0212:
1. This is because saving empty string has no effect, so I omitted empty template
2. I think this is what @kiootic suggested?
3. OK

louischan-oursky:
> In Add/Remove language dialog, the add-language behavior is fine. Click apply does not trigger API call. However, it seems unintuitive to me that if I remove a language and then click apply, the language is not removed yet. I have to click the save button below to take effect.

Let's say I am viewing English. I open the dialog and remove Chinese (HK). Now I have to click save in the context of English to apply my change to Chinese (HK). Does this sound intuitive?

axolotl0212:
Keep the delete locale on apply behaviour?

louischan-oursky:
> This is because saving empty string has no effect, so I omitted empty template

Actually it did have effect previously. We can delete the template by saving an empty string. Before this PR, we make user of this behavior to revert changes to the default template value.

axolotl0212:
If we count empty string as changes, user add new language then save, if there is empty string it is counted as modified. eg
template1: "some string"
template2: ""

click save -> form still modified

axolotl0212:
@louischan-oursky updated to only save forms on screen, select next item if selected item is deleted, block navigation on locale change if form dirty by using query param

louischan-oursky:
Thank you!

PR #731, opened 2020-11-13
Multilingual template support part 1

axolotl0212:
- Move template to separate screen
- Handle fetch template with different localization
- Add dropdown for choosing language from a list of supported language (default = effective default language)
- Add dropdown for choosing default language (default english)
- Add action button and dialog for managing template locales

Screenshot 2020-11-13 at 12 49 49 AM


portal/src/templates.ts
```diff
@@ -1,39 +1,85 @@
...
+  | SetupPrimaryOOBMessageTemplateKeys
+  | AuthenticatePrimaryOOBMessageTemplateKeys
+  | ForgotPasswordMessageTemplateKeys;
+
+function applyArgumentOnTemplateString(
+  argument: string,
+  template: string,
+  value: string
+): string {
+  return template.replace(new RegExp(`{{[ ]*${argument}[ ]*}}`, "g"), value);
```
kiootic, line 56: This substitution looks a bit weird to me. How about making the template types themselves typed? Something like:
```ts
type TemplateType = (locale: string) => string;

const TemplateForgotPasswordEmailHTML: TemplateType = (locale) => `templates/${locale}/messages/forgot_password_email.html`;
```

pkg/portal/model/system_config.go
```diff
@@ -1,9 +1,10 @@
...
 type SystemConfig struct {
-	AuthgearClientID string      `json:"authgearClientID"`
-	AuthgearEndpoint string      `json:"authgearEndpoint"`
-	AppHostSuffix    string      `json:"appHostSuffix"`
-	Themes           interface{} `json:"themes,omitempty"`
-	Translations     interface{} `json:"translations,omitempty"`
+	AuthgearClientID        string      `json:"authgearClientID"`
+	AuthgearEndpoint        string      `json:"authgearEndpoint"`
+	AppHostSuffix           string      `json:"appHostSuffix"`
+	PossibleTemplateLocales []string    `json:"possibleTemplateLocales"`
```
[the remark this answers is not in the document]
axolotl0212: Actually I was going to named it supportedTemplateLocales, however frontend has another app-specific supportedTemplateLocales so I needed another name. Maybe `possibleTemplateLocales` is not a good name, will come up with another name

portal/src/templates.ts
axolotl0212: Actually the template is also used for parsing path. Writing template as `(locale: string) => string` will work for rendering the path. Not sure if we should make this symmetric

pkg/portal/model/system_config.go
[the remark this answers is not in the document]
kiootic: Actually the we can localize things other than templates (e.g. static assets, translations). I think we can name the the list of locales to `supportedResourceLocales`, and app-specific list of locales to `configuredResourceLocales`.

portal/src/templates.ts
kiootic: I suppose front-end would need to parse the path to extract list of configured locales? Then, after adding the purposed field to backend, front-end should not need to parse the path.

We can keep it for now, and refactor it after introducing that field.

axolotl0212: > I suppose front-end would need to parse the path to extract list of configured locales? 

Yes

> Then, after adding the purposed field to backend, front-end should not need to parse the path.

If backend provide the locales, then the whole template thing can be removed and the path can simply be functions `(locale: string) => string`

axolotl0212:
@kiootic updated

louischan-oursky (commented):
1. Can we make the 2 lists in "Add or remove languages" dialog scroll independently? This can also make the dialog always smaller and visible fully in the viewport.

axolotl0212:
> 1. Can we make the 2 lists in "Add or remove languages" dialog scroll independently? This can also make the dialog always smaller and visible fully in the viewport.

Sure

kiootic (approved):

PR #745, opened 2020-11-18
Add ToC & privacy policy footer in signup page

### LINKED ISSUE #626: Add T&C and Privacy Policy to the auth page
Application will use the auth page provided by Authgear as the signup / login page. We should allow user to put Terms & Condition and Privacy Policy to the page.

Suggest to have admin to supply the link to T&C and Privacy Policy on portal, and we display them on the auth page so that user can open the corresponding pages

kiootic:
fix #626

### DIFF
--- a/resources/authgear/templates/en/web/signup.html
+++ b/resources/authgear/templates/en/web/signup.html
@@ -131,6 +131,9 @@
 					{{ template "next-button-label" }}
 				</button>
 			</form>
+			{{ if and ($.Translations.HasKey "terms-of-service-link") ($.Translations.HasKey "privacy-policy-link") }}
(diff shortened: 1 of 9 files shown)


resources/authgear/templates/en/web/signup.html
```diff
@@ -131,6 +131,9 @@
 					{{ template "next-button-label" }}
 				</button>
 			</form>
+			{{ if and ($.Translations.HasKey "terms-of-service-link") ($.Translations.HasKey "privacy-policy-link") }}
```
louischan-oursky, line 134: We should also put this in the promote page.

PR #750, opened 2020-11-19
Support using JWT as access token

### LINKED ISSUE #318: Support issuing access token in JWT
The draft spec is here https://tools.ietf.org/html/draft-ietf-oauth-access-token-jwt-08

kiootic:
fix #318

### DIFF
--- a/pkg/lib/config/oauth.go
+++ b/pkg/lib/config/oauth.go
@@ -41,7 +41,7 @@ var _ = Schema.Add("OAuthClientConfig", `
 		"post_logout_redirect_uris": { "type": "array", "items": { "type": "string", "format": "uri" } },
 		"access_token_lifetime_seconds": { "$ref": "#/$defs/DurationSeconds", "minimum": 300 },
 		"refresh_token_lifetime_seconds": { "$ref": "#/$defs/DurationSeconds" },
-		"use_jwt_access_token": { "type": "boolean" }
+		"issue_jwt_access_token": { "type": "boolean" }
 	},
 	"required": ["name", "client_id", "redirect_uris"]
 }
@@ -57,7 +57,7 @@ type OAuthClientConfig struct {
 	PostLogoutRedirectURIs []string        `json:"post_logout_redirect_uris"`
 	AccessTokenLifetime    DurationSeconds `json:"access_token_lifetime_seconds"`
 	RefreshTokenLifetime   DurationSeconds `json:"refresh_token_lifetime_seconds"`
-	UseJWTAccessToken      bool            `json:"use_jwt_access_token"`
+	IssueJWTAccessToken    bool            `json:"issue_jwt_access_token"`
 }
 
 func (c *OAuthClientConfig) SetDefaults() {
--- a/pkg/lib/config/oauth.go
+++ b/pkg/lib/config/oauth.go
@@ -14,10 +14,10 @@ type OAuthConfig struct {
 	Clients []OAuthClientConfig `json:"clients,omitempty"`
 }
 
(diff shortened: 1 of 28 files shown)


pkg/lib/config/secret.go
```diff
@@ -174,7 +174,7 @@ const (
 	SMTPServerCredentialsKey  SecretKey = "mail.smtp"
 	TwilioCredentialsKey      SecretKey = "sms.twilio"
 	NexmoCredentialsKey       SecretKey = "sms.nexmo"
-	OIDCKeyMaterialsKey       SecretKey = "oidc"
+	OAuthKeyMaterialsKey      SecretKey = "oauth"
```
louischan-oursky, line 177: This would break existing apps. Can we avoid renaming?

pkg/lib/config/oauth.go
```diff
@@ -40,7 +40,8 @@ var _ = Schema.Add("OAuthClientConfig", `
 		"response_types": { "type": "array", "items": { "type": "string" } },
 		"post_logout_redirect_uris": { "type": "array", "items": { "type": "string", "format": "uri" } },
 		"access_token_lifetime_seconds": { "$ref": "#/$defs/DurationSeconds", "minimum": 300 },
-		"refresh_token_lifetime_seconds": { "$ref": "#/$defs/DurationSeconds" }
+		"refresh_token_lifetime_seconds": { "$ref": "#/$defs/DurationSeconds" },
+		"use_jwt_access_token": { "type": "boolean" }
```
louischan-oursky, line 44: `issue_jwt_access_token` would be a more descriptive name.

pkg/lib/config/secret.go
kiootic: Sure we can keep the name, however we will be stuck with an inaccurate name thereafter.

louischan-oursky:
Could you help resolve the conflicts?

pkg/lib/oauth/token_encoding.go
```diff
@@ -3,19 +3,102 @@ package oauth
...
+		return token, nil
+	}
+
+	claims, err := e.UserClaims.LoadUserClaims(userID)
+	if err != nil {
+		return "", err
+	}
+
+	// iss & sub claims are in user claims.
+	_ = claims.Set(jwt.AudienceKey, e.BaseURL.BaseURL().String())
```
louischan-oursky, line 46: In ID token, we put client ID as `aud`. Should we do the same?

kiootic: The audience of access token is the resource server (i.e. ourselves). Clients are not supposed to interpret access tokens, therefore it's inappropriate to include them in audiences.

louischan-oursky: But I think one of the motivation to have introduce JWT access token is to allow the developer's backend server to see the claims. So are they also audience?

kiootic: In that case, the backend server is a resource server, so it should be included in audiences. Referring this scenario:
- Developer has authorization server (i.e. authgear)
- Developer has a backend API server (i.e. resource server)
- Developer has a iOS app (i.e. OAuth client)
- Developer has an Android app (i.e. OAuth client)

It can be observed that resource servers are not related to OAuth clients. Instead, according to the [draft spec](https://tools.ietf.org/html/draft-ietf-oauth-access-token-jwt-10#section-3), the authorization server should know a list of resource server indicators. However, I'm hesitant to add such a list to config at the moment, since it's unclear whether it would be useful.

PR #751, opened 2020-11-19
Fix not bypassing all IP rate limit for admin APIs

### LINKED ISSUE #747: Should not have rate limit for create user

kiootic:
fix #747

### DIFF
--- a/pkg/lib/interaction/nodes/do_create_user.go
+++ b/pkg/lib/interaction/nodes/do_create_user.go
@@ -30,10 +30,18 @@ func (e *EdgeDoCreateUser) Instantiate(ctx *interaction.Context, graph *interact
 		return nil, ErrNoPublicSignup
 	}
 
-	ip := httputil.GetIP(ctx.Request, bool(ctx.TrustProxy))
-	err := ctx.RateLimiter.TakeToken(interaction.SignupRateLimitBucket(ip))
-	if err != nil {
-		return nil, err
+	bypassRateLimit := false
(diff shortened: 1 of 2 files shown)


pkg/lib/interaction/nodes/do_create_user.go
```diff
@@ -30,10 +30,18 @@ func (e *EdgeDoCreateUser) Instantiate(ctx *interaction.Context, graph *interact
...
-		return nil, err
+	bypassRateLimit := false
+	var bypassInput interface{ BypassInteractionIPRateLimit() bool }
+	if interaction.Input(rawInput, &bypassInput) {
+		bypassRateLimit = bypassInput.BypassInteractionIPRateLimit()
+	}
+
+	if !bypassRateLimit {
+		ip := httputil.GetIP(ctx.Request, bool(ctx.TrustProxy))
+		err := ctx.RateLimiter.TakeToken(interaction.SignupRateLimitBucket(ip))
```
louischan-oursky, line 41: Would it be better if we have a RateLimiter implementation specific to interaction that also take input as argument? Then we could prevent this kind of programming error in the future.

kiootic: The input indicates skipping IP-based rate limit, since it is common for admin API requests come from same IP (i.e. backend server). However there is no function to know what kind of rate limit is the bucket, so the rate limiter proxy for interaction cannot ignore specific kind of rate limit based on input. Do we want to bypass every rate limit for admin APIs?

PR #753, opened 2020-11-19
Enhance security of HTTP headers

### LINKED ISSUE #752: Improve CSP headers
- Use strict CSP
- Main server
  - generate nonce: provide to template
- Portal server
  - setup nonce generation in web server
  - static asset: template nonce in index.html
- Optional config
  - frame ancestor? main server derived from OAuth client redirect URIs
  - report_uri?

ref: https://csp.withgoogle.com/docs/strict-csp.html

kiootic:
fix #752


pkg/lib/web/sec_headers_middleware.go
```diff
@@ -0,0 +1,44 @@
...
+var DefaultStrictCSPDirectives = []string{
+	"default-src 'self'",
+	"object-src 'none'",
+	"base-uri 'none'",
+	"block-all-mixed-content",
+}
+
+// SecHeadersMiddleware sends default security headers according to configuration.
+type SecHeadersMiddleware struct {
+	FrameAncestors []string
```
louischan-oursky, line 18: How do we populate this field?

kiootic: Previously, we derive it using OAuth client redirect URIs to support silent token refresh, however that mechanism is most likely non-functional in modern browsers due to strict cookies (and also we're not officially supporting it). Therefore I omitted it for now.

louischan-oursky: Can you put this as comment in the code?

PR #766, opened 2020-11-26
Add web-hook config to portal

### LINKED ISSUE #744: [Portal] Hook configuration screen

kiootic:
fix #744


louischan-oursky (commented):
I think we can hide the timeout configuration for now. Or we can keep them if we add a description for them. I guess no developer can set them correctly.

Issue #780, opened 2020-11-30
[Portal, Create App] Let user input and display the App Name instead of App ID


joyzoursky:
## Background:
- Device / Browser (OS version): Safari on Mac OS 
- Build: staging server 23-11-2020

## Description:
In the Authgear Portal. The apps are not showing the App Name but the App ID.

## Steps to reproduce:
1. Login to the Auth Gear Portal https://ourskyqa.authgear-staging.com.
2. Go through All App.  
4. Check All APP Name. The App Name not exist.
3. Check All app tool tip. The tool tip is mix with App Url and the App ID.

## Expected result:
Compare to the Design specification. 
Each app should show the App Name rather than the App ID.

## Screenshot:
![Screenshot_2020-11-23_at_3.09.20_PM](https://gitlab.com/skytester/authgear-server/uploads/08b8cde565e2de496c51b0c34e67e95f/Screenshot_2020-11-23_at_3.09.20_PM.png)![Screenshot_2020-11-23_at_2.58.13_PM](https://gitlab.com/skytester/authgear-server/uploads/f21f45f1898f29b236ef2a7d12b15376/Screenshot_2020-11-23_at_2.58.13_PM.png)

PR #785, opened 2020-12-01
Display decoded app id in app list and header

carmenlau:
refs #780 

In the issue, the design shows the app id and name, but we don't have the concept of app name. So we showed the encoded app id and the origin. 

I updated to show the decoded app id and kept displaying the app origin, it doesn't look nice actually. Maybe we could consider to show the app id only in the app list?

Screenshot 2020-12-01 at 5 59 15 PM


louischan-oursky:
This looks way better than before. I am merging this.

louischan-oursky:
BTW, we have a convention to use `fix #issue-number` in the PR description if the PR fixes the issue trivially.

Issue #748, opened 2020-11-19
Put reset and save button to command bar


keithtkl:

kiootic:
Some pages have pivots (i.e. tab pages). The current behavior is treating each tab page as separate form context, requiring user to save/discard changes before saving tab. After switching to use command bar for save & reset, it would become strange that a lower component (pivot tabs) controls the behavior of upper component (command bar).

Also, the current placement of command bar in 'Portal Admin' tab of 'Settings' page is strange.

Instead, we should use [hierarchical navigation](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#hierarchical-navigation) pattern of Fluent UI design. I suppose the equivalent component on web platform would be nested `Nav` links.

PR #788, opened 2020-12-01
Update resource config pages UI design

kiootic:
ref #748


louischan-oursky (commented):
The only concern I have is that the developer can no longer "reset to default value" by saving an empty template. Is there any way to allow them to reset the template to default value?

kiootic:
I suppose you encountered 'Some required resources do not have localizations in default language' error?

'Default value' for templates are only available for `en` locale templates. Maybe we can make that a special case and allow saving empty templates (i.e. delete app specific template value) for `en` locale.

kiootic:
@louischan-oursky Updated. Now only forbid removing a required resource of default locale which is not the built-in locale (en).

louischan-oursky:
Two bugs

1. Cannot save images
2. Just by changing selected locale, the save button will become enabled.

kiootic:
> Just by changing selected locale, the save button will become enabled.

I can't reproduce this bug, but I've fixed the 'cannot save images' bug, which is actually 'cannot display saved images'.

louischan-oursky:
1. Remove all languages and all templates so that the app does not have any resources
2. Add Chinese Hong Kong, and then save
3. Edit the English version of translation.json, and then save
4. Change selected to Chinese Hong Kong, the save button becomes enabled.

PR #801, opened 2020-12-04
Unify one-time code / verification code

### LINKED ISSUE #798: Unify one-time code / verification code
- Remove `authenticator.oob_otp.sms.code_digits`
- Remove `authenticator.oob_otp.email.code_digits`
- Use 6 digit numeric code for all one-time code / verification code (non-configurable for now)
- Store the code with the verification store (i.e. on Redis)

cc @carmenlau

carmenlau:
fix #798

### DIFF
--- a/pkg/lib/authn/authenticator/oob/provider.go
+++ b/pkg/lib/authn/authenticator/oob/provider.go
@@ -4,18 +4,32 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"time"
 
 	"github.com/authgear/authgear-server/pkg/lib/authn"
 	"github.com/authgear/authgear-server/pkg/lib/authn/otp"
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/util/clock"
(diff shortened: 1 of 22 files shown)


pkg/lib/authn/authenticator/oob/provider.go
```diff
@@ -61,34 +74,44 @@ func (p *Provider) Create(a *Authenticator) error {
...
-	ok := otp.ValidateTOTP(secret, code, p.Clock.NowUTC(), p.getOTPOpts(channel))
-	if !ok {
-		return errors.New("invalid OOB code")
+func (p *Provider) CreateCode(authenticatorID string) (*Code, error) {
+	code := otp.FormatNumeric.Generate()
+	codeModel := &Code{
+		AuthenticatorID: authenticatorID,
+		Code:            code,
+		// TODO(oob): Expiry should be configurable
+		ExpireAt: p.Clock.NowUTC().Add(time.Duration(3600) * time.Second),
```
carmenlau, line 87: Do you think we need config for the oob code expiry? Or too avoid too many config from now, so set a fixed value is fine?

louischan-oursky (approved):

Issue #796, opened 2020-12-02
Introduce concept of 'first-party' OAuth clients


kiootic:
- Update specs to describe the concept, and its interaction with features:
    - settings page/api
    - oauth protocol (consent, scopes, etc.)
    - security implications

PR #810, opened 2020-12-09
Add spec for first-party clients

kiootic:
ref #796


kiootic:
@louischan-oursky had raised a concern on potential confusion of two concepts:
- 'first-party client' & 'third-party client' by OAuth framework; mobile apps are always third-party client.
- 'first-party client' & 'third-party client' by developers' intuition; self-developed mobile apps are considered first-party client.

We may want to use 'first-party application' & 'third-party application' instead for the second case. Any thoughts?

Issue #763, opened 2020-11-25
Allow some interaction steps to be performed by another user agent


louischan-oursky:
## Motivation

When the user is signing up in a mobile application with email address and password, they will receive a verification email. In the verification email, there is a link. It is expected that most of the users will click on the link. At the moment, clicking the link will result in error.

## Approach 1 - Allow some specific steps to be performed by another user agent

We would like to allow some specific steps to be performed by another user agent. Example of the steps are verification, in which the verification code is associated with the interaction. Therefore, by the verification code only, it is possible to perform this interaction step. Since the another user agent does not have the interaction cookie, after performing the step, we will display a page telling the user that they should return to the original application to resume the interaction.

Since some of the interaction steps can now be performed by another user agent, we need to have a persistence websocket connection per interaction to keep the interaction in sync. There is new websocket endpoint which accepts connection by interaction cookie. If the request has a valid interaction cookie, connection is established. The client will then listen for the event of interaction step being performed by another user agent. The client will also ping the server periodically to see if the interaction is stale. If the interaction is stale, the server will return a new URL, in which the client will redirect to it.

## Implementation of Approach 1

- Ensure interaction cookie is only written once when the interaction starts.
- Add a static page to tell the user to return to the original application.
- Make verification special case. Allow it to be performed without interaction cookie.
- Redirect to the static page after interaction step is performed.
- Figure out how to handle the external interaction step is the last step of the interaction.
- Introduce the websocket endpoint at server side.
- Implement the webcocket protocol between the server and the client.

## The reason why approach 1 cannot solve our problem

In an edge case where the interaction step to be performed by another user agent is the last step of the interaction, the interaction is destroyed and the original user agent cannot resume the interaction.

## Approach 2 - Allow another user agent to contribute to input to the original agent

The interaction mechanism is a finite state machine, where it constantly renders a page that collects user input and steps forward. If a specific step is opened by another user agent, we want to model this as an alternative way to collect user input. Specifically, the another user agent will show page to tell the user to click the continue button. Clicking the button will submit the form to the server. The server will store the form data in the web session. The page will try to use the form data in the web session as pre-filled form values. There also exists a persistent websocket connection for listening the change in the form data in the web session.

[...]

PR #815, opened 2020-12-10
Handle verification email being opened in another user agent

louischan-oursky:
This is the first part of #763. The user has to refresh the page manually to have the code filled in in the original user agent. The next PR will implement a websocket protocol between the server and the client to avoid manual refreshing.

### DIFF
--- a/pkg/auth/handler/webapp/verify_identity.go
+++ b/pkg/auth/handler/webapp/verify_identity.go
@@ -1,7 +1,6 @@
 package webapp
 
 import (
-	"errors"
 	"net/http"
 
 	"github.com/authgear/authgear-server/pkg/auth/handler/webapp/viewmodels"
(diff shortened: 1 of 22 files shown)


resources/authgear/templates/en/translation.json
```diff
@@ -246,5 +249,8 @@
   "show-password-label": "Show password",
   "hide-password-label": "Hide password",
 
+  "return-page-title": "Return to where you were",
```
kiootic, line 252: I think the wording of this page can use some amendment, but that's good enough for now.

pkg/auth/handler/webapp/verify_identity.go
```diff
@@ -116,34 +142,105 @@ func (h *VerifyIdentityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
...
 				return err
 			}
 
-			h.Renderer.RenderHTML(w, r, TemplateWebVerifyIdentityHTML, data)
+			var action string
+			if session == nil {
+				// The web session is invalid.
+				// Here we assume the verification is requested by the user.
+				// This is because if verification is NOT requested by the user, the interaction can never continue.
+				action = VerifyIdentityActionResume
```
kiootic, line 189: This logic seems a bit strange? I think this case is when user perform verification as part of authentication, and use the link in email after the interaction expired/completed/invalidated. In that case, would returning a `WebUIInvalidSession` error and render error page more appropriate?

pkg/auth/handler/webapp/verify_identity.go
```diff
@@ -166,36 +263,50 @@ func (h *VerifyIdentityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
...
+			return err
 		}
 
-		var result *webapp.Result
-		if session != nil {
-			result, err = ctrl.InteractionPost(inputFn)
-		} else {
-			result, err = ctrl.EntryPointPost(opts, intent, inputFn)
+		step := session.CurrentStep()
+		if step.FormData == nil {
```
kiootic, line 296: Can we assume by convention `FormData` is never nil, since it seems we're switching to use `NewSessionStep` instead of constructing it directly?

pkg/auth/webapp/session_step.go
```diff
@@ -86,8 +94,9 @@ func (k SessionStepKind) MatchPath(path string) bool {
 }
 
 type SessionStep struct {
-	Kind    SessionStepKind `json:"kind"`
-	GraphID string          `json:"graph_id"`
+	Kind     SessionStepKind        `json:"kind"`
+	GraphID  string                 `json:"graph_id"`
+	FormData map[string]interface{} `json:"form_data"`
```
kiootic, line 99: Can we add a comment on its intended usage?

pkg/auth/handler/webapp/verify_identity.go
louischan-oursky: This can also be the case where the user requested verification in the settings and wait until the original interaction become expired. If we do not assume VerifyIdentityActionResume here, IntentVerifyIdentityResume is never reachable.

kiootic: I think the root cause is that we can't distinguish whether the verification code comes from authentication flow (non-resumable if invalid) or settings page (resumable if invalid). Maybe we can add a flag to verification code to indicate this? Otherwise in my mentioned case, user may see confusing error when trying to click next?

louischan-oursky:
Updated.

kiootic (approved):

PR #816, opened 2020-12-11
Add session token spec

kiootic:
requires #810 

TODO: make SDK use web view for settings page

### DIFF
--- a/docs/specs/oidc.md
+++ b/docs/specs/oidc.md
@@ -431,25 +431,26 @@ Developers should note the security implications for first-party clients:
   (TODO: on-behalf-of flow https://tools.ietf.org/html/rfc7523)
   (TODO: authenticate clients using client secret)
   
-### Session Token
+### App Session Token
 
 For mobile first-party clients, developer may want to 'transfer' the
 user session from the native app (obtained through OAuth protocol) to
-web UI. In this case, developer may use access token to exchange for a
-session token, which can be used to copy the user session to web user agent.
+web UI. In this case, developer may use refresh token to exchange for a
+one-time-use app session token, which can be used to open authenticated pages
+using the refresh token.
 
 First, the native app should perform authentication through standard
 OAuth flow to obtain an access token and refresh token.
 
-When the native app wants to copy the user session to web user agent,
+When the native app wants to copy the user session from app to web user agent,
 the native app may use the refresh token in the session token endpoint to
-obtain a session token:
+obtain a one-time-use app session token:
 ```
 POST /oauth2/session-token HTTP/1.1
 Host: accounts.example.com
 Content-Type: application/json
 
-{"refresh_token":"<refresh token>","lifetime":3600}
+{"refresh_token":"<refresh token>"}
 
 ---
 HTTP/1.1 200 OK
@@ -458,22 +459,29 @@ Content-Type: application/json
 {"result":{"session_token":"<session token>"}}
 ```
 
-A session token would be returned, and the native app may inject it into the
-session token cookie in web view/header, and open the desired page:
+A one-time-use app session token would be returned, and the native app may
+then use it in OAuth authorization flow to open authenticated page in web user
+agent:
 ```
-GET /settings HTTP/1.1
+GET /oauth2/authorize?client_id=client_id&prompt=none&response_type=none
+    &login_hint=https%3A%2F%2Fauthgear.com%2Flogin_hint%3Ftype%3Dsession_token%26session_token%3D<session token>
+    &redirect_uri=<redirect URI> HTTP/1.1
 Host: accounts.example.com
-Cookie: session_token=<session token>
+
+---
+HTTP/1.1 302 Found
+Set-Cookie: <session cookie>
+Location: <redirect URI>
 ```
 
-When the session token is requested:
+When the app session token is requested:
 - The OAuth client associated with the access token must be a first-party
   client.
-- The requested session token lifetime would be capped by the refresh token
-  lifetime.
 
-When the session token value in cookie is validated:
-- Session token cookie takes priority over IdP session cookie, even if the
-  session token is invalid.
-- If the session token is invalid/expired, it'll be treated same as invalid
-  session.
+When the app session token is consumed:
+- If the app session token is invalid, normal OAuth authorization flow would be
+  performed instead.
+- The session cookie would contain a token referencing the refresh token,
+  instead of IdP sessions. Therefore, the lifetime of session cookie is bound
+  to refresh token instead of IdP session.
+ 
--- a/docs/specs/oidc.md
+++ b/docs/specs/oidc.md
@@ -436,22 +436,20 @@ Developers should note the security implications for first-party clients:
 For mobile first-party clients, developer may want to 'transfer' the
 user session from the native app (obtained through OAuth protocol) to
 web UI. In this case, developer may use access token to exchange for a
-one-time-use session token, which can be used to copy the user session
-to web user agent.
+session token, which can be used to copy the user session to web user agent.
 
 First, the native app should perform authentication through standard
-OAuth flow to obtain an access token. However, refresh token must not
-be requested (i.e. not including `offline_access` in requested scopes),
-so that the access token would be bound to the user session instead of
-a refresh token.
+OAuth flow to obtain an access token and refresh token.
 
 When the native app wants to copy the user session to web user agent,
-the native app may use the access token in the session token endpoint to
+the native app may use the refresh token in the session token endpoint to
 obtain a session token:
 ```
 POST /oauth2/session-token HTTP/1.1
 Host: accounts.example.com
-Authorization: Bearer <access token>
+Content-Type: application/json
+
+{"refresh_token":"<refresh token>","lifetime":3600}
 
 ---
 HTTP/1.1 200 OK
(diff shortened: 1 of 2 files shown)


docs/specs/oidc.md
```diff
@@ -405,3 +406,83 @@ Silent Authentication
...
+---
+HTTP/1.1 302 Found
+Set-Cookie: <session cookie>
+Location: <redirect URI>
+```
+
+When the session token is requested:
+- The OAuth client associated with the access token must be a first-party
+  client.
+- The access token must be bound to an IdP session (i.e. no refresh token)
```
chpapa, line 481: May I know why? (Or should the spec explain why?)

docs/specs/oidc.md
```diff
@@ -405,3 +406,83 @@ Silent Authentication
...
+
+When the session token is consumed:
+- If the session token is invalid, normal OAuth authorization flow would be
```
chpapa, line 484: Any scenario which a client with valid access token might pass in an invalid session token? (So it should be a re-authentication?), or it is an error case?

docs/specs/oidc.md
```diff
@@ -405,3 +406,83 @@ Silent Authentication
...
+  performed instead.
+- The session cookie associated with the IdP session is re-generated, to
```
chpapa, line 486: Is the session token use case solely for mobile apps to transfer its access token from native app to cookies in browser?

Asking coz I'm thinking if there would be cases where we invalidate the session cookie might have some strange outcome (like mobile devices with multiple browser sessions etc...)

chpapa (commented):
Took a quick look, don't really understand enough so I asked some clarification questions, thanks!

docs/specs/oidc.md
kiootic: By convention of OAuth, refresh tokens are independent of IdP sessions; they can be created/invalidated without affecting each other. Therefore they do not reference each other.

Session token represents an IdP session, therefore it need an IdP session.

docs/specs/oidc.md
kiootic: The session token is one-time-use and limited lifetime, so it can be already consumed, expired, or the associated session is invalidated/revoked/logout. Assume normal situation, these cases should not occur when consuming the token, since requesting the token would fail already.

docs/specs/oidc.md
kiootic: We have an implicit invariant, that an IdP session may be held by at most one user agent, since the carrier of IdP session (cookie) is not transferrable between user agents.

To enforce this invariant, we re-generates the session cookie to ensure no other user agents may held this IdP session.

chpapa: @kiootic that's interesting. It sounds like make sense (or at least no harm); Any reason/beneftis why we the rule of one user agent?

docs/specs/oidc.md
```diff
@@ -429,4 +429,59 @@ Developers should note the security implications for first-party clients:
...
```
louischan-oursky, line 486: What if the lifetime of the refresh token is longer than that of the IdP session? So it could happen that the IdP session has ended. In that case do we create a brand new IdP session?

docs/specs/oidc.md
kiootic: The i.e. part refers to 'bound to an IdP session': Access token is either bound to an IdP session (when refresh token is not requested) or a refresh token (when refresh token is requested).

docs/specs/oidc.md
kiootic: One of the reasons, as stated above, is that the carrier of IdP session (cookie) is not transferrable between user agents.

Also, having a one-to-one relationship between IdP session and user agent is easier to reason about the security properties.

docs/specs/oidc.md
```diff
@@ -429,4 +429,59 @@ Developers should note the security implications for first-party clients:
...
+OAuth flow to obtain an access token. However, refresh token must not
+be requested (i.e. not including `offline_access` in requested scopes),
+so that the access token would be bound to the user session instead of
+a refresh token.
+
+When the native app wants to copy the user session to web user agent,
+the native app may use the access token in the session token endpoint to
+obtain a session token:
+```
+POST /oauth2/session-token HTTP/1.1
```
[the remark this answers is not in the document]
kiootic: By OAuth spec, `/oauth2/token` is used to obtain access token/refresh token, therefore it's inappropriate to reuse it for other purpose.

kiootic:
@chpapa @louischan-oursky
Updated the session token mechanism to disregard IdP session, since common native app flow uses refresh token.

docs/specs/oidc.md
```diff
@@ -460,29 +458,22 @@ Content-Type: application/json
...
-A one-time-use session token would be returned, and the native app may
-then use it in OAuth authorization flow to copy user session to web user
-agent:
+A session token would be returned, and the native app may inject it into the
+session token cookie in web view/header, and open the desired page:
 ```
-GET /oauth2/authorize?client_id=client_id&prompt=none&response_type=none
-    &login_hint=https%3A%2F%2Fauthgear.com%2Flogin_hint%3Ftype%3Dsession_token%26session_token%3D<session token>
-    &redirect_uri=<redirect URI> HTTP/1.1
+GET /settings HTTP/1.1
```
louischan-oursky, line 464: Let's keep the original approach by using `/oauth2/authorize` to cause the server to write a valid cookie. The developer can also use `redirect_uri` to open a specific settings page.

louischan-oursky (approved):

carmenlau (approved):

carmenlau:
We have discussed the flow in the production meeting and the update looks good. See if you have further comment on the details? @chpapa :) If no, I am going to merge it!

louischan-oursky:
BTW, I have merged the parent PR.

docs/specs/oidc.md
```diff
@@ -72,6 +72,7 @@ Supported [standard client metadata](https://openid.net/specs/openid-connect-reg
 - `client_id`: OIDC client ID.
 - `access_token_lifetime`: Access token lifetime in seconds, default to 1800.
 - `refresh_token_lifetime`: Refresh token lifetime in seconds, default to max(access_token_lifetime, 86400). It must be greater than or equal to `access_token_lifetime`.
+- `is_first_party`: Indicate whether the client is a [first-party client](#first-party-clients), default to false.
```
chpapa, line 75: Is it related with the "first party client" vs "first party application" wordings? Or update the naming later?

docs/specs/oidc.md
```diff
@@ -405,3 +406,82 @@ Silent Authentication
...
+
+OAuth protocol allows user to delegate access to clients, so that clients can
+act on behalf of the user.
+
+Usually, OAuth clients are third-party clients, with limited trust:
+- User need to give consent to a client before clients is allowed access.
+- Privileged user operations (e.g. change password) is usually not exposed
+  through APIs accessible through OAuth protocol.
+
+However, developers may want to give first-party clients full trust:
```
chpapa, line 420: Similarly, for the clients vs applications wordings, shall we change these sentences (and other related similarly) to:

"However, developers may want to allow clients full trust as a first-party application:"

something like that.

docs/specs/oidc.md
```diff
@@ -405,3 +406,82 @@ Silent Authentication
...
+- First-party clients do not need user consent in authorization process.
+- First-party clients have access to privileged user operations through a
+  special OAuth scope value (`https://authgear.com/scopes/full-access`).
+
+To designate an OAuth client as first-party client, developer may set
+`is_first_party` attribute to `true` in the corresponding client metadata.
+
+Developers should note the security implications for first-party clients:
+- Access tokens for first-party clients should not be passed to third-party
```
chpapa, line 429: Similarly: "Access tokens for first-party applications should not be passed to third-party applications with limited trust."

docs/specs/oidc.md
```diff
@@ -405,3 +406,82 @@ Silent Authentication
...
+using the refresh token.
+
+First, the native app should perform authentication through standard
+OAuth flow to obtain an access token and refresh token.
+
+When the native app wants to copy the user session from app to web user agent,
+the native app may use the refresh token in the session token endpoint to
+obtain a one-time-use app session token:
+```
+POST /oauth2/session-token HTTP/1.1
```
chpapa, line 449: `/oauth2/one-time-use-app-session-token`, or
`/oauth2/app-session-token`?

docs/specs/oidc.md
```diff
@@ -405,3 +406,82 @@ Silent Authentication
...
+To designate an OAuth client as first-party client, developer may set
+`is_first_party` attribute to `true` in the corresponding client metadata.
+
+Developers should note the security implications for first-party clients:
+- Access tokens for first-party clients should not be passed to third-party
+  clients with limited trust.
+  (TODO: on-behalf-of flow https://tools.ietf.org/html/rfc7523)
+  (TODO: authenticate clients using client secret)
+  
+### App Session Token
```
chpapa, line 434: The rest of this page seems call it "One-time-use App Session Token", or we just call it "App Session Token"?

chpapa (approved):
The flow are good.

Only two comments on wordings:

* Is it "One time use App session token" or "App session token"?
* Is it "First/Third part Clients" or "Applications"?

docs/specs/oidc.md
kiootic: I've opened #823 to track updates on usage of terms, since most likely we want to review/update the whole spec in one go.

docs/specs/oidc.md
kiootic: 'One-time-use' is a description adjective for app session token: app session tokens are one-time-use.

PR #817, opened 2020-12-11
Allow configuring OAuth client as first-party client

kiootic:
ref #796 

TODO: update SDK to stop requesting full access scope for third-party clients.


louischan-oursky (commented):
The behavior change in the 2nd commit should be specified in the spec. Can we update the spec to reflect this new behavior?

kiootic:
Updated #810 to classify anonymous user flows as privileged operations.

PR #821, opened 2020-12-14
Use websocket to listen interaction change

### LINKED ISSUE #763: Allow some interaction steps to be performed by another user agent

louischan-oursky:
fix #763

### DIFF
--- a/authui/src/authgear.ts
+++ b/authui/src/authgear.ts
@@ -309,6 +309,10 @@ window.api.onLoad(() => {
     }
   }
 
-  connect();
+  // Avoid websocket connection unless it is necessary.
+  if (window.location.pathname.startsWith("/verify_identity")) {
+    connect();
+  }
+
   return dispose;
 });
--- a/authui/src/authgear.ts
+++ b/authui/src/authgear.ts
@@ -152,6 +152,17 @@ function back(e: Event) {
   window.history.back();
 }
 
+function refreshPage() {
+  let url = window.location.pathname;
+  if (window.location.search !== "") {
+    url += "?" + window.location.search;
+  }
+  if (window.location.hash !== "") {
+    url += "#" + window.location.hash;
+  }
+  Turbolinks.visit(url, { action: "replace" });
+}
+
 window.api.onLoad(() => {
   const elems = document.querySelectorAll(".back-btn");
   for (let i = 0; i < elems.length; i++) {
@@ -252,4 +263,52 @@ window.api.onLoad(() => {
       cancelAnimationFrame(animHandle);
     }
   };
+});
+
+window.api.onLoad(() => {
+  const scheme = window.location.protocol === "https:" ? "wss:" : "ws:";
+  const host = window.location.host;
+  const url = `${scheme}//${host}/ws`;
+
+  let ws: WebSocket | null = null;
+
+  function dispose() {
+    ws?.close();
+    ws = null;
+  }
+
+  function connect() {
+    ws = new WebSocket(url);
+
+    ws.onopen = function(e) {
+      console.log("ws onopen", e);
+    };
+
+    ws.onclose = function(e) {
+      console.log("ws onclose", e);
+      // Close code 1000 means we do not need to reconnect.
+      if (e.code === 1000) {
+        return;
+      }
+
+      dispose();
+      connect();
+    };
+
+    ws.onerror = function(e) {
+      console.error("ws onerror", e);
+    };
+
+    ws.onmessage = function(e) {
+      console.log("ws onmessage", e);
+      const message = JSON.parse(e.data);
+      switch (message.kind) {
+      case "refresh":
+        refreshPage();
+      }
+    }
+  }
+
+  connect();
+  return dispose;
 });
--- a/pkg/lib/config/redis.go
+++ b/pkg/lib/config/redis.go
@@ -21,8 +21,10 @@ type RedisConfig struct {
 }
 
 func (c *RedisConfig) SetDefaults() {
+	// Now we use redis pubsub, we need to have much greater number of connections.
+	// https://redis.io/topics/clients#maximum-number-of-clients
 	if c.MaxOpenConnection == nil {
-		c.MaxOpenConnection = newInt(2)
+		c.MaxOpenConnection = newInt(10000)
 	}
 	if c.MaxIdleConnection == nil {
 		c.MaxIdleConnection = newInt(2)
--- a/pkg/util/pubsub/http_handler.go
+++ b/pkg/util/pubsub/http_handler.go
@@ -22,7 +22,6 @@ type WebsocketOutgoingMessage struct {
 
 type Delegate interface {
 	Accept(r *http.Request) (channelName string, err error)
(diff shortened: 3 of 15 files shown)


pkg/util/pubsub/http_handler.go
```diff
@@ -0,0 +1,160 @@
...
+				if err != nil {
+					errChan <- err
+					return
+				}
+
+				if outgoing != nil {
+					writeCtx, cancel := context.WithTimeout(rootCtx, WebsocketWriteTimeout)
+					defer cancel()
+
+					err = wsConn.Write(writeCtx, outgoing.MessageType, outgoing.Data)
```
kiootic, line 148: `wsConn.Write` may be called concurrently in different gorountine. Although it seems the library allows it (by using mutex), I think it would be a trap if we were to switch to other libraries/implementation (e.g. gorilla), which does not allow concurrent  read/write.

pkg/util/pubsub/http_handler.go
```diff
@@ -0,0 +1,160 @@
...
+// HTTPHandler receives incoming websocket messages and delegates them to the delegate.
+// For each websocket connection, a redis pubsub connection is established.
+// Every message from the redis pubsub connection is forwarded to the websocket connection verbatim.
+type HTTPHandler struct {
+	RedisPool     RedisPool
+	Delegate      Delegate
+	LoggerFactory *log.Factory
+}
+
+func (h *HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
kiootic, line 43: The data flow of this handler seems a bit unnatural:
- Redis routine: Redis read -> WebSocket write
- WebSocket routine: WebSocket read -> delegate (?) -> WebSocket write (?)

It would be unclear to API consumer on what delegate must perform in order to complete the data flow. If I understand correctly:
- it publish the incoming message to Redis
- return some message (?) to be written to WebSocket.

I'd suggest to either abstract away Redis subscription in delegate, or handle Redis publish directly in the handler.

authui/src/authgear.ts
```diff
@@ -253,3 +264,51 @@ window.api.onLoad(() => {
     }
   };
 });
+
+window.api.onLoad(() => {
```
kiootic, line 268: `api.onLoad` handler would be called repeatedly whenever navigation happens, so WebSocket would be connected & disconnected repeatedly. I think we should subscribe to window's load event instead?

pkg/lib/config/redis.go
```diff
@@ -21,8 +21,10 @@ type RedisConfig struct {
 }
 
 func (c *RedisConfig) SetDefaults() {
+	// Now we use redis pubsub, we need to have much greater number of connections.
+	// https://redis.io/topics/clients#maximum-number-of-clients
 	if c.MaxOpenConnection == nil {
-		c.MaxOpenConnection = newInt(2)
+		c.MaxOpenConnection = newInt(10000)
```
kiootic, line 27: I think we should use a hub connection for handling redis subscription, by dynamically issuing `subscribe`/`unsubscribe` commands.

Let's open an issue to improve this later.

authui/src/authgear.ts
```diff
@@ -309,6 +309,10 @@ window.api.onLoad(() => {
     }
   }
 
-  connect();
+  // Avoid websocket connection unless it is necessary.
+  if (window.location.pathname.startsWith("/verify_identity")) {
```
kiootic, line 313: I remembered we discussed on this and decided to always keep a WebSocket connection for simplicity, is there any problem with that?

louischan-oursky: We can remove this once we have implemented dynamic `subscribe` / `unsubscribe`.

pkg/util/pubsub/http_handler.go
louischan-oursky: Actually we do not expect the websocket client to send us any message at the moment. For simplicity, maybe we can remove `OnWebsocketMessage` from `Delegate`. This also makes sure `wsConn.Write` is only called within the websocket goroutine.

pkg/lib/config/redis.go
louischan-oursky: https://github.com/authgear/authgear-server/issues/822

kiootic (approved):

PR #834, opened 2020-12-18
Improve pubsub implementation

### LINKED ISSUE #822: Improve websocket implementation
- Try to decouple the number of websocket clients with the number of required redis connections. Ideally we should have very small number of redis connections to handle many websocket clients.
- Always establish websocket connections, instead of only establish in some selected paths.
- Do not use `window.api.onLoad` to manage websocket lifecycle.

louischan-oursky:
fix #822 

NOTE

I did not complete the 3rd point in that issue. It is because when we do not have cookie, we try connect once and give up. So when navigation changes, we want to try again. The current behavior does this automatically.

### DIFF
--- a/pkg/lib/infra/redis/hub.go
+++ b/pkg/lib/infra/redis/hub.go
@@ -4,16 +4,11 @@ import (
 	"context"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/go-redis/redis/v8"
 
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/util/log"
-)
-
-const (
-	RedisReadTimeout = 1 * time.Second
 )
 
 type SubscriberID struct {
@@ -29,14 +24,32 @@ func (i *SubscriberID) Next() int64 {
 	}
 }
 
-type ChannelKey struct {
-	ConnKey     string
-	ChannelName string
+type Subscriber struct {
+	ChannelName  string
+	SubscriberID int64
+	// The drop pattern
+	// See https://www.ardanlabs.com/blog/2017/10/the-behavior-of-channels.html
+	MessageChannel chan *redis.Message
 }
 
-type Subscriber struct {
+// ControlMessageJoin is essentially Subscriber.
+type ControlMessageJoin = Subscriber
+
+// ControlMessageLeave is basically Subscriber without MessageChannel.
+type ControlMessageLeave struct {
+	ChannelName  string
 	SubscriberID int64
-	Chan         chan *redis.Message
+}
+
+// PubSub is an actor wrapping redis.PubSub
+type PubSub struct {
+	PubSub     *redis.PubSub
+	Subscriber map[string][]Subscriber
+	// Mailbox handles 3 messages.
+	// 1. When the mailbox is closed, that means the actor should die.
+	// 2. ControlMessageJoin
+	// 3. ControlMessageLeave
+	Mailbox chan interface{}
 }
 
 // Hub aims to multiplex multiple subscription to Redis PubSub channel over a single connection.
@@ -45,9 +58,8 @@ type Hub struct {
 	SubscriberID *SubscriberID
 	Logger       *log.Logger
 
-	mutex      sync.RWMutex
-	pubsub     map[string]*redis.PubSub
-	subscriber map[ChannelKey][]Subscriber
+	mutex  sync.RWMutex
+	pubsub map[string]*PubSub
 }
 
 func NewHub(pool *Pool, lf *log.Factory) *Hub {
@@ -55,9 +67,7 @@ func NewHub(pool *Pool, lf *log.Factory) *Hub {
 		Pool:         pool,
 		SubscriberID: &SubscriberID{},
 		Logger:       lf.New("redis-hub"),
-
-		subscriber: make(map[ChannelKey][]Subscriber),
-		pubsub:     make(map[string]*redis.PubSub),
+		pubsub:       make(map[string]*PubSub),
 	}
 	return h
 }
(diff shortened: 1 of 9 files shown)


pkg/lib/infra/redis/hub.go
```diff
@@ -0,0 +1,244 @@
...
+					ChannelName: n.Channel,
+				}
+
+				h.mutex.RLock()
+				subs := make([]Subscriber, len(h.subscriber[key]))
+				copy(subs, h.subscriber[key])
+				h.mutex.RUnlock()
+
+				for _, s := range subs {
+					s.Chan <- n
```
kiootic, line 149: I think this may panic (i.e. sending to closed channel), since `cleanupOne` may close a channel in `subs` concurrently?

pkg/lib/infra/redis/hub.go
```diff
@@ -0,0 +1,244 @@
...
+	h.pubsub[connKey] = pubsub
+
+	go func() {
+		for {
+			iface, err := pubsub.ReceiveTimeout(ctx, RedisReadTimeout)
+			if os.IsTimeout(err) {
+				continue
+			}
+			if err != nil {
+				h.cleanupAll(cfg, credentials)
```
kiootic, line 126: This goroutine would exit on error when receiving. I suppose a context cancellation error is the normal exit condition; can we log other errors (i.e. abnormal exit)?

pkg/util/pubsub/http_handler.go
```diff
@@ -72,47 +70,33 @@ func (h *HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-				if os.IsTimeout(err) {
-					continue
+			case n, ok := <-msgChan:
+				if !ok {
+					return
 				}
+
+				writeCtx, cancel := context.WithTimeout(rootCtx, WebsocketWriteTimeout)
+				defer cancel()
+				err := wsConn.Write(writeCtx, websocket.MessageText, []byte(n.Payload))
```
kiootic, line 95: I think think may cause unresponsive pubsub mechanism when client is malicious:
Since `msgChan` (created in `Hub.Subscribe`) is un-buffered, the spawned message dispatch goroutine would block if the previous message is still sending (blocked at `wsConn.Write`).

pkg/lib/infra/redis/hub.go
louischan-oursky: Any idea on how to avoid that? Wrap the statement `s.Chan <- n` in another goroutine with recover() igoring the panic?

pkg/util/pubsub/http_handler.go
louischan-oursky: We have a reverse proxy in front of us (traefik), so I guess the write should finish pretty fast?

pkg/lib/infra/redis/hub.go
kiootic: We should close the channel when there is no more message to send to the channel (i.e. should close by sender). Therefore, I think we should close the channel in this goroutine instead.

Another thing I just noticed (by reading the doc of goredis), is that `pubsub.Channel()` is the preferred API. I think we can have a control channel (send from cleanup), and select on the control channel and message channel in this goroutine?

pkg/util/pubsub/http_handler.go
kiootic: Even the buffers of reverse proxy can be filled up if client is really slow (intentionally), and I don't think the proxies would drop the connection once it gets full? With some amount of (malicious) clients, this case may be hit frequently.

pkg/lib/infra/redis/hub.go
```diff
@@ -29,14 +24,32 @@ func (i *SubscriberID) Next() int64 {
...
+	ChannelName  string
+	SubscriberID int64
+	// The drop pattern
+	// See https://www.ardanlabs.com/blog/2017/10/the-behavior-of-channels.html
+	MessageChannel chan *redis.Message
 }
 
-type Subscriber struct {
+// ControlMessageJoin is essentially Subscriber.
+type ControlMessageJoin = Subscriber
```
kiootic, line 36: This is a type alias instead of type definition, which should be seldom used. Can we just do `type ControlMessageJoin Subscriber` instead, and cast `ControlMessageJoin` to `Subscriber` in relevant places?

pkg/lib/infra/redis/hub.go
```diff
@@ -0,0 +1,228 @@
...
+	channelName string,
+) (chan *redis.Message, func()) {
+	pubsub := h.getPubSub(cfg, credentials)
+
+	id := h.SubscriberID.Next()
+	// The drop pattern
+	// See https://www.ardanlabs.com/blog/2017/10/the-behavior-of-channels.html
+	c := make(chan *redis.Message, 10)
+
+	pubsub.Mailbox <- ControlMessageJoin{
```
kiootic, line 87: A bit of nitpick: this may have same 'send to closed channel' problem when subscription & pubsub actor mailbox close is performed concurrently. I suppose this case would not happen as it is now, since we never close the pubsub actor actively (is it? what about abnormal closure of redis conn?).

Conceptually, we have three types of actors here:
- `Hub`: one per Authgear server instance, supervise a set of `PubSub`,
- `PubSub`: one per Redis connection, supervise a set of `Subscriber`,
- `Subscriber`: one per WebSocket connection.

To be more robust, we may send the `Subscriber` actor reference through `Hub`, which dispatch it to the correct `PubSub` instance. The death of `PubSub` actors would be performed by the supervisor (`Hub`) after being notified by the child `PubSub` actor when connection failed.

pkg/lib/infra/redis/hub.go
louischan-oursky: I changed it to type definition and ignored the lint error.

pkg/lib/infra/redis/hub.go
louischan-oursky: I converted Hub to an actor.

kiootic (approved):

PR #842, opened 2020-12-22
Improve verification UX

### LINKED ISSUE #836: Improve verification UX
- Make "Continue Verification" page look more like the real one
- Auto submit after prefill

louischan-oursky:
fix #836

### DIFF
--- a/authui/src/authgear.ts
+++ b/authui/src/authgear.ts
@@ -265,6 +265,14 @@ window.api.onLoad(() => {
   };
 });
 
+// Handle auto form submission
+window.api.onLoad(() => {
+  const e = document.querySelector('[data-auto-submit="true"]');
+  if (e instanceof HTMLElement) {
+    e.click();
+  }
+});
+
 window.api.onLoad(() => {
   const scheme = window.location.protocol === "https:" ? "wss:" : "ws:";
   const host = window.location.host;
--- a/authui/src/authgear.ts
+++ b/authui/src/authgear.ts
@@ -269,6 +269,7 @@ window.api.onLoad(() => {
 window.api.onLoad(() => {
   const e = document.querySelector('[data-auto-submit="true"]');
   if (e instanceof HTMLElement) {
+    e.removeAttribute("data-auto-submit");
     e.click();
   }
 });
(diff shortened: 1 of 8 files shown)


authui/src/authgear.ts
```diff
@@ -265,6 +265,14 @@ window.api.onLoad(() => {
   };
 });
 
+// Handle auto form submission
+window.api.onLoad(() => {
+  const e = document.querySelector('[data-auto-submit="true"]');
```
kiootic, line 270: When I press back button after it auto-submitted, it shows `There is a problem with the network.` error since it attempts to auto-submit again. I suppose we should auto-submit at most once per page?

louischan-oursky: I made the auto-submit runs at most once

kiootic (approved):

resources/authgear/templates/en/web/verify_identity.html
```diff
@@ -6,23 +6,6 @@
...
-
-<input type="hidden" name="x_code" value="{{ $.VerificationCode }}">
-<button class="btn primary-btn align-self-flex-end" type="submit" name="x_action" value="{{ $.Action }}">{{ template "next-button-label" }}</button>
-</form>
-
-{{ else }}
-
 {{ if $.IdentityDisplayID }}
 {{ template "__nav_bar.html" }}
 {{ else }}
```
carmenlau, line 11: Better not to show the back button when `update_session_step`?

resources/authgear/templates/en/web/verify_identity.html
```diff
@@ -48,49 +31,46 @@ <h1 class="title primary-txt">{{ template "verify-user-page-title--email" }}</h1
...
+	class="input text-input primary-txt"
+	type="text"
+	inputmode="numeric"
+	pattern="[0-9]*"
+	autocomplete="one-time-code"
+	autocapitalize="none"
+	name="x_code"
+	placeholder="{{ template "verify-user-placeholder" }}"
+	value="{{ $.VerificationCode }}"
+	{{ if eq $.Action "update_session_step" }}readonly{{ end }}
```
carmenlau, line 44: Use `disabled` or update css makes readonly to make it looks different from normal input?

louischan-oursky: We cannot use `disabled` because `disabled` input are NOT included in the form submission. I'll make `readonly` look the same as `disabled` instead.

carmenlau (approved):

Issue #773, opened 2020-11-30
WeChat login


louischan-oursky:
[Integration Documentation](https://developers.weixin.qq.com/doc/offiaccount/OA_Web_Apps/Wechat_webpage_authorization.html)

---

Implementation Notes

- Authorization endpoint: `https://open.weixin.qq.com/connect/oauth2/authorize?appid=APPID&redirect_uri=REDIRECT_URI&response_type=code&scope=SCOPE&state=STATE#wechat_redirect`
- Token endpoint: `https://api.weixin.qq.com/sns/oauth2/access_token?appid=APPID&secret=SECRET&code=CODE&grant_type=authorization_code`
- UserInfo endpoint: `https://api.weixin.qq.com/sns/userinfo`, needs scope `snsapi_userinfo`

---

[Apply for sandbox account](https://mp.weixin.qq.com/debug/cgi-bin/sandbox?t=sandbox/login)

---

[Useful article](https://blog.51cto.com/zero01/2117892)

---

The flow is very different from normal OAuth 2.0 providers (e.g. Facebook). The authorization URL we generate cannot be opened by a normal browser. We must generate a QR code image of it and display it to the user. Then we tell the user to scan the QR code with their Wechat app. By scanning the QR code, the authorization URL is visited by the embedded Wechat browser. If the authorization URL is visited by other browser, an error will be shown saying that the URL must be opened by the embedded Wechat browser. The user consent the authorization in the Wechat app and then being redirected to redirect URI. Since there are now two user agents, we have to think of a way to instruct the user to come back to the original user agent to continue the flow. The original user agent is showing the QR code while the redirect URI is visited by the embedded Wechat browser. We need to be able to support interaction in more than one user agent.

---

Known Caveats

- We must use publicly accessible domain name for the redirect URI.
- ngrok.com is blocked by Wechat.
- The wechat authorization URL is not a deep link of the Wechat app. So on mobile device, we cannot show the authorization URL as link and expect the user to click on it to redirect to the Wechat app.

---

Mobile Flow

1. The user clicks the Login with Wechat button.
1. We show a page with the authorization URL encoded as QR code image. Tell the user to scan the QR code with their Wechat app.
1. The user scans the QR code and authorize in the Wechat app.
1. Wechat redirects the user to the redirect URI we gave them.
1. The handler of the redirect URI steps forward the graph.

---

Web Flow

---

iOS SDK setup

1. In Build Phases, Link Binary with Libraries, add libc++.tbd
[...]

PR #843, opened 2020-12-23
Publish WeChatLoginStart event for SDK integration

carmenlau:
refs #773 

Updates:
- Publish `WeChatLoginStart` evnt in wechat auth endpoint
- Change wechat endpoints to `/sso/wechat/auth/:alias` and `/sso/wechat/callback`
- Support `x_ws_channel_id` parameter in authorize endpoint, allow sdk specify websocket channel id
- Support `x_sdk` in authorize endpoint
- Updated `/sso/wechat/callback` to support api call from sdk

### DIFF
--- a/pkg/auth/handler/webapp/wechat_auth.go
+++ b/pkg/auth/handler/webapp/wechat_auth.go
@@ -37,6 +37,7 @@ type WechatAuthHandler struct {
 	BaseViewModel     *viewmodels.BaseViewModeler
 	Renderer          Renderer
 	CSRFCookie        webapp.CSRFCookieDef
+	Publisher         *Publisher
 }
 
 func (h *WechatAuthHandler) GetData(r *http.Request, w http.ResponseWriter, session *webapp.Session, graph *interaction.Graph) (map[string]interface{}, error) {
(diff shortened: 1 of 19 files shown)


carmenlau:
@louischan-oursky I added commit fd4a9a7 to address the problem that the fresh event is fired before the authui connected to the web socket. You can check the detail in the commit message. I added one new `NavigationAction` `no_action` in `webapp.Result`, as I wish to stay in the same page after the post request and no navigation is needed. See if it is appropriate?

pkg/auth/handler/webapp/wechat_auth.go
```diff
@@ -143,6 +129,33 @@ func (h *WechatAuthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		err = h.Publisher.Publish(session, msg)
+		if err != nil {
+			return err
+		}
+
+		// stay in the same page, no navigation action
+		result := &webapp.Result{
+			NavigationAction: "no_action",
```
louischan-oursky, line 154: I think we should also set webapp.Result.RedirectURI to current URI, in case turbolinks is not being used. We can test this by disabling JavaScript in the browser.

louischan-oursky:
conflicted.

pkg/auth/handler/webapp/wechat_auth.go
carmenlau: The commit has been removed and handled in #857

carmenlau:
Rebased

PR #845, opened 2020-12-23
Portal style tweak

### LINKED ISSUE #844: Portal UI improvement
- Remove all nav icons. Some of the nested nav items are missing icon, causing them to look indented one level up more than its parent

Screen Shot 2020-12-23 at 2 59 31 PM

- Double check if margin/padding is consistent across different blocks of text.

Screen Shot 2020-12-23 at 2 58 45 PM

- The ordering UX in Login ID is too subtle

Screen Shot 2020-12-23 at 3 02 01 PM

louischan-oursky:
fix #844

### DIFF
--- a/portal/src/graphql/portal/CustomDomainListScreen.module.scss
+++ b/portal/src/graphql/portal/CustomDomainListScreen.module.scss
@@ -1,13 +1,13 @@
-.root {
+.content {
   margin: 0 32px;
-  position: relative;
 }
 
 .verifySuccessMessageBar {
-  position: absolute;
-  left: 0;
-  top: 0;
-  width: 100%;
+  visibility: hidden;
+
+  &.visible {
+    visibility: visible;
+  }
 }
 
 .description {
--- a/portal/src/graphql/portal/CustomDomainListScreen.module.scss
+++ b/portal/src/graphql/portal/CustomDomainListScreen.module.scss
@@ -1,5 +1,5 @@
 .root {
-  padding-top: 40px;
+  margin: 0 32px;
   position: relative;
 }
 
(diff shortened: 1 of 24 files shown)


portal/src/graphql/portal/UserInterfaceScreen.tsx
```diff
@@ -58,9 +58,9 @@ const UserInterfaceContent: React.FC<UserInterfaceContentProps> = function UserI
   return (
     <div className={styles.root}>
       <NavBreadcrumb items={navBreadcrumbItems} />
-      <Label>
+      <Text className={styles.title}>
```
kiootic, line 61: This looks strange, since the field labels on other screens are bolded.

kiootic (commented):
It seems DNS screens are missed?

portal/src/graphql/portal/UserInterfaceScreen.tsx
louischan-oursky: Changed back to Label.

louischan-oursky:
updated.

portal/src/graphql/portal/CustomDomainListScreen.module.scss
```diff
@@ -1,5 +1,5 @@
 .root {
-  padding-top: 40px;
+  margin: 0 32px;
```
kiootic, line 2: The top padding is intentional, otherwise the verify success message bar would overlap with the title. You can see it by going to `configuration/dns/custom-domains?verify=success`.

louischan-oursky: I updated the layout to use `visibility` instead.

kiootic: I think the message bar should be full-width, instead of having spacing on two horizontal side?

louischan-oursky: Fixed.

kiootic (approved):

PR #854, opened 2021-01-04
Replace app banner with favicon

### LINKED ISSUE #851: Remove app banner and add app favicon

louischan-oursky:
fix #851


carmenlau (changes requested):
The changes looks great! But do we have a larger default banner image? Now it looks a bit blurry.

louischan-oursky:
See if it looks better now.

Issue #291, opened 2020-09-11
Web UI theming


kiootic:
- [Update web UI to match latest design](https://github.com/authgear/authgear-server/issues/864)
- [Make web UI theme-aware](https://github.com/authgear/authgear-server/issues/864)
- [Store theme in app FS CSS file with special comment blocks surrounding the theme CSS declarations](https://github.com/authgear/authgear-server/pull/908)
- Add preview on theme change

---

#### Original description

Since we're using Fluent UI design in settings page, which is a plain web app (not SPA), we may want to look into using [Fabric Core](https://developer.microsoft.com/en-us/fluentui#/get-started#fabric-core), the official CSS-only framework for Fluent UI design.

louischan-oursky:
Fabric Core is basically this CSS file https://static2.sharepointonline.com/files/fabric/office-ui-fabric-core/11.0.0/css/fabric.css

It basically includes the following things

1. [A Grid system](https://developer.microsoft.com/en-us/fluentui#/styles/web/layout#grid). But the doc suggests that you can use CSS grid instead. We are using CSS grid already
2. A number of class names that form a theme. This may be not be applicable to us, since we are still not sure about how are we going to do theming.
3. A number of class names for using fluent UI icons. But we are actually using font-awesome for most icons.
4. A number of class names for controlling font size. I guess font size should be part of the theme?
5. Some utility class names, such as clearfix.

So adding Fabric Core may not be very useful to us. Thoughts? @carmenlau @kiootic

kiootic:
It seems the component styles are moved to `office-ui-fabric-js` (which is deprecated in favor of the new react-based library) at v3.0.0. I suppose we can hand-roll the component styles ourselves using SCSS?

louischan-oursky:
How [Fabric UI Theme Designer] works?

The code responsible for generating a theme has the following parts

1. https://github.com/microsoft/fluentui/blob/master/packages/react-internal/src/components/ThemeGenerator/ThemeRulesStandard.ts
2. https://github.com/microsoft/fluentui/blob/master/packages/react-internal/src/components/ThemeGenerator/ThemeGenerator.ts
3. https://github.com/microsoft/fluentui/blob/master/packages/react-internal/src/utilities/color/shades.ts
4. https://github.com/microsoft/fluentui/blob/master/packages/theme/src/utilities/makeSemanticColors.ts

[...]

PR #856, opened 2021-01-04
Make login and signup page more obvious

### LINKED ISSUE #556: Make default sign in page and sign up page more distinglishable
Current design's sign in and sign up page is almost indistinguishable. Which is sort of a usability nightmare.

louischan-oursky:
fix #556

### DIFF
--- a/authui/src/authgear.css
+++ b/authui/src/authgear.css
@@ -66,12 +66,6 @@ a {
   cursor: pointer;
 }
 
-/* a:link is more specific than .primary-btn so we need to add :not here */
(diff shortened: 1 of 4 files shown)


carmenlau:
@louischan-oursky The login and signup tab looks great. But from the design files, we should have more things that need to be updated? e.g. login with login id come first then social login. Add separation line for the word "or"...etc. Will we handle it in this issue? Let me reopen it first.

louischan-oursky:
Let's handle all these in https://github.com/authgear/authgear-server/issues/291

PR #858, opened 2021-01-05
Send refresh event again after connected to ws

### LINKED ISSUE #857: WebSocket improvement
There is a case that the session "refresh" event will be published before the authui connected to the websocket and authui may missed the event. After discussion, to handle this, authui will provide a timestamp parameter when connect to websocket, if server found that the session is updated after the provided timestamp, server will publish a "refresh" event again.

carmenlau:
fix #857

### DIFF
--- a/pkg/auth/webapp/session.go
+++ b/pkg/auth/webapp/session.go
@@ -25,6 +25,7 @@ type SessionOptions struct {
 	UILocales       string
 	Prompt          string
 	Extra           map[string]interface{}
+	UpdatedAt       time.Time
 }
 
 func NewSessionOptionsFromSession(s *Session) SessionOptions {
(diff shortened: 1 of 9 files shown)


pkg/auth/webapp/session.go
```diff
@@ -58,6 +59,9 @@ type Session struct {
 	// UILocales are the locale to be used to render UI, passed in from OAuth
 	// flow or query parameter.
 	UILocales string `json:"ui_locales,omitempty"`
+
+	// UpdatedAt indicate the session last updated time
+	UpdatedAt time.Time `json:"updated_at,omitempty"`
```
louischan-oursky, line 64: I think we need to initialize it when Session is being created.

carmenlau: Updated!

Issue #864, opened 2021-01-06
Theme-aware web UI


louischan-oursky:
- [x] Tidy up class names and styles
- [ ] Apply hard-coded theme to every component

PR #868, opened 2021-01-08
Revamp web UI

louischan-oursky:
ref #864

### DIFF
--- a/authui/src/authgear.css
+++ b/authui/src/authgear.css
@@ -77,8 +77,96 @@ h1, h2, h3, h4, h5, h6 {
   font-weight: 600;
 }
 
+@media (min-width: 320px) {
+  .content {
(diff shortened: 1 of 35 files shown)


carmenlau (changes requested):
The PR overall looks good, some minor comments.

1. Seems the some of the font size are wrong? e.g. SSO button font is a bit larger, toc-pp-footer font are larger than the design too...etc
1. The wechat auth page is removed accidentally?
1. The error widget from `__errors.html` added unexpected margin to the layout even there is no error.

louischan-oursky:
1. See the comment on `authui/src/authgear.css:3`. We had an issue in the past that focusable element should have a font-size of >= 16px.
2. Sorry for that
3. Will fix that

PR #883, opened 2021-01-14
Apply default light and dark theme to web UI

### LINKED ISSUE #864: Theme-aware web UI

louischan-oursky:
fix #864

### DIFF
--- a/authui/src/authgear.css
+++ b/authui/src/authgear.css
@@ -171,9 +171,40 @@
   --color-password-strength-meter-4: #baca58;
   --color-password-strength-meter-5: var(--color-good);
 }
-
 @media (prefers-color-scheme: dark) {
+  :root {
+    --color-primary-unshaded: #317BF4;
+    --color-primary-shaded-1: #f6faff;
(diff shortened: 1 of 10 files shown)


carmenlau:
💯 The theme looks great!! But I still have some comments on some style not related to theming. Let me merge this PR first, and handle it in another PR? @louischan-oursky 

1. From the design, the input field should have small left right padding.
2. The social login button left right padding should be smaller
3. Maybe we can apply color to social login button icon in dark mode, except the google one we cannot make multiple color.
4. The secondary btn should not have left right padding? We can check the enter password page.
5. Wrong padding in enter_login_id page

louischan-oursky:
1. Both Ben and Elliot complained that the input field should not have padding so that all text are aligned along the same vertical line. Therefore the padding deviates from the design.
2. Will fix it.
3. I will try to apply color to see if mono color looks great. Will share a screenshot for this.
4. Will fix it.
5. Will fix it.

PR #889, opened 2021-01-15
Tweak web UI style

### LINKED ISSUE #887: Auth ui style refinement
1. ~From the design, the input field should have small left right padding.~
1. The social login button left right padding should be smaller
1. Maybe we can apply color to social login button icon in dark mode, except the google one we cannot make multiple color.
1. The secondary btn should not have left right padding? We can check the enter password page.
1. Wrong padding in enter_login_id page

louischan-oursky:
fix #887

Screen Shot 2021-01-15 at 2 57 22 PM
Screen Shot 2021-01-15 at 2 57 02 PM

### DIFF
--- a/authui/src/authgear.css
+++ b/authui/src/authgear.css
@@ -746,20 +746,19 @@ h1, h2, h3, h4, h5, h6 {
   padding: 0;
 }
 
-.btn {
-  height: 36px;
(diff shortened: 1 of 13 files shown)


carmenlau:
💯

Issue #926, opened 2021-01-27
Unable to go back to app homepage through back button and app logo


mr9dollars:
## Description:
After going to the login page, can't go back to the main page by pressing "back" button or by clicking the application logo.

## Steps to reproduce:
1. Go to Sign up / Sign in Page
2. Try to press the Project House icon or Back button
3. See will stay on Sign up / Sign in page

## TODO
- [ ] Allow configuration on Portal UI for adding home/back to app button
- [ ] Allow redirection link for the home and back to app button

## Screenshot:

https://user-images.githubusercontent.com/69806274/105948974-e5bae300-60a6-11eb-9466-70b90248c33b.mov

mr9dollars:
1. Cannot go back to the website with back button.
2. Cannot click logo to back to home page

@carmenlau 

Not sure about whether the first issue is a bug from us.
The second issue should be an enhancement that we allow a redirect URL for clicking the logo?

chpapa:
@mr9dollars I think you should update this issues, maybe have a parameter options to add a "back" / "home" button in the login / signup page, and setting page.

I think it requires some design update as well.

louischan-oursky:
In https://github.com/authgear/authgear-server/pull/957 I make the logo an anchor if home URI is specified. Is that good enough for this issue? Adding a back button in login/signup/settings page seems too weird to me.

mr9dollars:
![image](https://user-images.githubusercontent.com/16920209/106733531-f6da9580-664c-11eb-95d4-e2eee23a8b42.png)

Do you think we can make a back to app/home button here when user finds a way to quit the login/signup flow? Logo as button might be too subtle.

louischan-oursky:
Let's come up with a design first and then discuss it in the product meeting.

PR #957, opened 2021-02-03
Add home_uri

### LINKED ISSUE #917: Set default redirect url
For some apps, it would be better for branding and avoid confusing users, to have a default redirect uri

So when users visit the [tenant].authgearapps.com, after login instead of goes to setting page, it goes to the "default app" and continue the usage (or they can build a page to tell users what to do next)

louischan-oursky:
fix #917 
ref #926


carmenlau:
We need to update portal to allow setting `home_uri`, any idea where should we put in? I think developer will want to set in when setting the auth ui banner, but put this under Appearance is strange. 🤔

louischan-oursky:
A new subtab in settings? But I cannot think of any good name :(

PR #964, opened 2021-02-04
[Breaking] Separate oob_otp config to oob_otp_email and oob_otp_sms

### LINKED ISSUE #904: Separate secondary authenticator from `OOB` to `OOB_email` and `OOB_sms`
The use case is a user will sign up email + password, but 2FA with SMS.

Currently, if someone enable both email and phone as login ID, it will have an unwanted side-effect of login is possible via both Phone and Email, not exactly fit the use case above.

The changes requires:

1. Refactor `OOB` to `OOB_email` and `OOB_sms`
2. Update of Setting page:
    - make sure it works as intended when login ID doesn't have `phone` and setup MFA with `OOB_sms`
    - Separate "Verification code via SMS / Email" at `/settings/mfa` to two items.
    - Remove "Setup [email] 2-step verification instead" at all setup page.
3. Update of AuthUI signup / login screen to work in new config
4. Update Portal so that authenticator screen include the new options, and on-boarding screen contain the new options too

Extended goal:
* Add a new option under `OOB_SMS` and `OOB_Email`, to specify if the SMS/Email for OOB channels have to be unique or not.

carmenlau:
fix #904


louischan-oursky (commented):
Looks good to me so far.

louischan-oursky (approved):

PR #995, opened 2021-02-22
Validate values of translation.json

### LINKED ISSUE #985: Add validation on the value of translation.json
Right now it is possible to save invalid syntax of messageformat string.

louischan-oursky:
fix #985

### DIFF
--- a/pkg/lib/config/configsource/resources.go
+++ b/pkg/lib/config/configsource/resources.go
@@ -79,6 +79,8 @@ func (d AuthgearYAMLDescriptor) ViewResources(resources []resource.ResourceFile,
 		return app()
 	case resource.EffectiveResourceView:
 		return effective()
+	case resource.ValidateResourceView:
+		return effective()
 	default:
 		return nil, fmt.Errorf("unsupported view: %T", rawView)
 	}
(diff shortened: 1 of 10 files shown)


carmenlau (commented):
Need updating to add new view, so only run validation when updating the resources, instead every `viewEffectiveResource`

carmenlau:
@louischan-oursky I checked the PR again and found actually the new approach doesn't help on the case that I mentioned (don't break the existing app if translation is invalid). Since in the original implementation, the render functions (`renderHTML` and `renderPlainText`) called `FormatTemplateParseTree` for all translation items already. That means if the translation is invalid, the pages are broken already. If this is the case, maybe we can keep your original updates (update `viewEffectiveResource`) to keep it simple? What do you think and sorry for that.

louischan-oursky:
I think we can keep this implementation. In the future we could optimize `renderHTML` and `renderPlainText` such that translation item are parsed as few as needed.

louischan-oursky:
But if we want to do that, we need to borrow the idea from Helm. They have [include](https://helm.sh/docs/howto/charts_tips_and_tricks/) which is similar to the builtin `template`, which means we can delay the parsing of translation item until we really need it.

PR #1005, opened 2021-02-24
Move public origin configuration to custom domain page in portal

### LINKED ISSUE #945: Custom public domain feature UX improvement
Solution after discussion:

1. Only 1 domain will be supported. Which is the public origin.
2. To avoid confusion, update auth ui to remove public origin configuration page. Instead, user can choose one active domain in custom domain page, which will be set into the public origin in app config.

=====

Original problem:

Tried adding a custom domain to Project House's production project, verified the domain, and set the new domain as the new public domain.

Saw 2 problems after doing that:
1. When I click the login/sign up button at kurious.live, I still see the domain name = https://kurious.authgearapps.com/login, not auth.kurious.live
2. The https://kurious.authgearapps.com/login?ui_locales=zh-Hant and https://auth.kurious.live/login page layout is broken, seems like the CSS is not loaded at all.

image
image

carmenlau:
fix #945


louischan-oursky (commented):
When I press `active`, the state has already changed. I suggest we delay the state change until the user has press the confirm button in the dialog.

carmenlau:
Updated, see if aa69018 addressed the problem that you mentioned? Instead of delay the state change, I updated the ui to show the saved config only.

PR #1010, opened 2021-02-25
Update error bar style

### LINKED ISSUE #939: Update error display UI
From https://github.com/authgear/authgear-server/issues/920,

We had a discussion on how to display error message properly. 

1. Putting generic error message in AuthUI as snack bar error. Allow putting URL in error msg

image

2. Put input related error back into textfield error
image

carmenlau:
fix #939

### DIFF
--- a/resources/authgear/static/authgear.css
+++ b/resources/authgear/static/authgear.css
@@ -104,9 +104,9 @@ h1, h2, h3, h4, h5, h6 {
 @media (min-width: 1025px) {
   .page {
     display: flex;
-    flex-direction: row;
-    justify-content: center;
-    align-items: flex-start;
+    flex-direction: column;
+    justify-content: start;
+    align-items: center;
   }
   .content {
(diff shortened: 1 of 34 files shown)


resources/authgear/static/authgear.css
```diff
@@ -313,9 +313,55 @@ h1, h2, h3, h4, h5, h6 {
...
 .errors {
-  list-style-type: none;
+  display: flex;
+  flex-direction: row;
+  background-color: var(--color-error-shaded-2);
+}
+
+.errors.display-none {
+  display: none;
+}
```
louischan-oursky, line 329: This is unnecessary. We have `.display-none` already.

carmenlau: Since `.errors` has `display: flex`, which override `display:none` of `.display-none`, `.errors` and `.display-none` are in the same element. That's the reason why I added this.

Issue #953, opened 2021-02-02
Email verification code expired UX enhancement


mr9dollars:
- If the code is expired, use the resend button to request a new email
- In the error message, "Resend a new one" should be a link to resend the code
- Then add a button to back to the app login page.
(Pending design)

Input the expired verification code:
![image](https://user-images.githubusercontent.com/16920209/106579467-b52dea00-657b-11eb-9d57-2446d8108351.png)
- Allow resending immediately

PR #1011, opened 2021-02-25
Update verification code resend button state to reflect real cooldown time

carmenlau:
refs #953

WIP: update oob_otp resend button in the same approach


carmenlau:
Will update to use checking rate limit approach

PR #1013, opened 2021-02-25
Explicit supported languages

### LINKED ISSUE #1004: Add supported languages in configuration
Currently, supported languages are determined implicitly by the templates the app has. But once we add more and more builtin supported languages, this approach will not work if the app want to support a subset of the builtin supported languages. Therefore we need a way for the app to declare its supported languages.

---

- Add `supported_languages` to the config.
- The default value of `supported_languages` is a singleton array of `fallback_language`.
- Maintain the invariant that `supported_languages` must contain `fallback_language`.
- Update resource resolution mechanism to respect `supported_languages`.
- Update resource resolution mechanism to consider effective supported languages.
- Update portal to read/write `supported_languages` directly.

louischan-oursky:
fix #1004


pkg/util/intlresource/intlresource.go
```diff
@@ -0,0 +1,47 @@
...
+		// Ignore resources in unsupported languages.
+		_, supported := supportedSet[langTag]
+		if !supported {
+			continue
+		}
+		err := add(langTag, resrc)
+		if err != nil {
+			return err
+		}
+		added = true
```
carmenlau, line 37: Would like to confirm the behavior of the `Prepare` function
1. is to ensure `intl.DefaultLanguage` is added even it is not listed in `supportedLanguageTags`?
OR
2. is to ensure there is at least one language added ? 

Now the function is doing 2.

Actucally I found a problem in `translationJSON` Descriptor that. If developer removed `en` language in supported language, e.g. the app only support `zh-TW`. If there is new items which is missing in `zh-TW/translation.json`, the app will crash due to missing translation item. Not sure always adding `en` can help solve the problem? Since we defined the `en`.

But if user provide `en/translation.json` that cannot have all items, will the same problem occur again?

The problem seems a bit complicated, see if we need to discuss for that?

carmenlau (changes requested):

pkg/util/intlresource/intlresource.go
carmenlau: > But if user provide en/translation.json that cannot have all items, will the same problem occur again?

Double check and found, missing items in user defined `en/translation.json` is not a problem.

PR #1018, opened 2021-03-01
Fix missing public origin in ui when portal custom domain page is just loaded

carmenlau:
refs #945


portal/src/graphql/portal/CustomDomainListScreen.tsx
```diff
@@ -523,7 +523,8 @@ const CustomDomainListContent: React.FC<CustomDomainListContentProps> = function
       savedPublicOriginRef.current = state.publicOrigin;
     }
   }, [isDirty, state]);
-  const prevSavedPublicOrigin = savedPublicOriginRef.current;
+  const prevSavedPublicOrigin =
+    savedPublicOriginRef.current || state.publicOrigin;
```
louischan-oursky, line 527: I suggest we initialize the ref with `state.publicOrigin`. `state.publicOrigin` should be non-empty at the 1st time this render function runs.

PR #1020, opened 2021-03-01
Separate appearance and localization

### LINKED ISSUE #1016: Separate Appearance and Localization

louischan-oursky:
depends on #1013 
fix #1016


louischan-oursky:
@carmenlau Fixed the forbidden error. The cause is a bit complicated. In `useResourceForm`, I forgot to skip update if there is no resource changes. If `updateAppTemplates` is called without any updates, `$path` will be empty. The behavior of `$path` is to fetch all resources, including `authgear.secrets.yaml`. But getting the field `effectiveValue` of `authgear.secrets.yaml` is forbidden.

PR #1024, opened 2021-03-02
Revise fallback language resource resolution

### LINKED ISSUE #1023: Revise fallback language resource resolution mechanism
For fallback language resource, we will always use the BaseFS of DefaultLanguage as the first layer of the effective resource view.

louischan-oursky:
fix #1023

### DIFF
--- a/pkg/util/intlresource/match_test.go
+++ b/pkg/util/intlresource/match_test.go
@@ -4,8 +4,6 @@ import (
 	"testing"
 
 	. "github.com/smartystreets/goconvey/convey"
-
-	"github.com/authgear/authgear-server/pkg/util/intl"
 )
 
 type testItem string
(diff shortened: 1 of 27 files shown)


pkg/util/intlresource/match_test.go
```diff
@@ -35,12 +35,5 @@ func TestMatch(t *testing.T) {
 		})
 		So(err, ShouldBeNil)
 		So(matched, ShouldEqual, testItem("ja"))
-
-		// Match intl.DefaultLanguageTag if possible
-		matched, err = Match([]string{"zh"}, "ja", []LanguageItem{
-			testItem(intl.DefaultLanguage),
-		})
-		So(err, ShouldBeNil)
-		So(matched, ShouldEqual, testItem(intl.DefaultLanguage))
```
carmenlau, line 44: We should remove `testItem(intl.DefaultLanguage)` in the other test case in the file to avoid confusion in future.

carmenlau (commented):
The resolution looks good!! Only one minor comment.

louischan-oursky:
Updated.

PR #1028, opened 2021-03-03
Rate limit issue

### LINKED ISSUE #1019: Rate limit issue
1. Signup rate limit is counted in incorrect timing. Now it is counted after user entered the login id, we should count it when signup success.
2. Rename `reset-password-generate-code` to `reset-password-send-code` and update size to once per min
3. Remove unnecessary rate limit `verification-generate-code`, it will be bounded by verification email/sms rate limit

carmenlau:
depends on #1027 

fix #1019 

- Fix incorrect timing for signup rate limit
- Remove `verification-generate-code` rate limit which is bounded by `verification-send-code` rate limit
- Rename and update `reset-password-send-code` rate limit

### DIFF
--- a/pkg/lib/interaction/nodes/do_create_user.go
+++ b/pkg/lib/interaction/nodes/do_create_user.go
@@ -3,6 +3,7 @@ package nodes
 import (
 	"github.com/authgear/authgear-server/pkg/api/apierrors"
 	"github.com/authgear/authgear-server/pkg/lib/interaction"
+	"github.com/authgear/authgear-server/pkg/lib/ratelimit"
 	"github.com/authgear/authgear-server/pkg/util/httputil"
 	"github.com/authgear/authgear-server/pkg/util/uuid"
 )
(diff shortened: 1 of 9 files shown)


pkg/lib/interaction/nodes/do_create_user.go
```diff
@@ -64,11 +59,37 @@ func (n *NodeDoCreateUser) GetEffects() ([]interaction.Effect, error) {
 			return err
 		}),
 		interaction.EffectOnCommit(func(ctx *interaction.Context, graph *interaction.Graph, nodeIndex int) error {
+			ip := httputil.GetIP(ctx.Request, bool(ctx.TrustProxy))
+			// check the rate limit only and don't take the token yet
+			if !n.BypassRateLimit {
+				pass, _, err := ctx.RateLimiter.CheckToken(interaction.SignupRateLimitBucket(ip))
```
louischan-oursky, line 65: Any reason why don't we just call TakeToken?

carmenlau: Since the real effects are run in `ctx.Users.AfterCreate`.
1. We want to ensure there is token before running the effects, so we check token here.
2. We only want to take the token if the effects are success. If effects failed (e.g. webhook disallow the operation), we don't take the token, so user can retry immediately. 

If the implementation looks good, let me add more comment in the code to explain.

Issue #943, opened 2021-02-01
Biometric authentication for mobile apps


louischan-oursky:
- [x] Add biometric identity data model (https://github.com/authgear/authgear-server/issues/1074)
- [x] Skip MFA for OAuth identity and biometric identity. (https://github.com/authgear/authgear-server/issues/1074)
- [x] Enforce the invariant that `A user must have at least one Login ID identity or OAuth identity`. (https://github.com/authgear/authgear-server/issues/1074)
- [x] `/oauth2/challenge` (https://github.com/authgear/authgear-server/issues/1079)
- [x] `/oauth2/token` (https://github.com/authgear/authgear-server/issues/1079)
  - [x] `action = setup`: require `jwt` and `device_info`
  - [x] `action = authenticate`
- [x] Show biometric identity in settings page (https://github.com/authgear/authgear-server/issues/1086)
- [x] Make sure anonymous user cannot setup biometric
- [x] Show device info (https://github.com/authgear/authgear-server/issues/1094)
- [x] Show biometric identity in portal (https://github.com/authgear/authgear-server/issues/1099)
- [x] SDK remove dependency on react-native-async-storage (https://github.com/authgear/authgear-sdk-js/issues/68)
- [x] SDK biometric APIs (https://github.com/authgear/authgear-sdk-ios/issues/49, https://github.com/authgear/authgear-sdk-android/issues/61, https://github.com/authgear/authgear-sdk-js/issues/66)

PR #1043, opened 2021-03-09
Add biometric authentication draft

louischan-oursky:
ref #943 

For all readers, they should read in this order:

1. user-model.md
2. sdk.md

For implementers, they should read in this order:

1. user-model.md
2. sdk.md
3. api.md
4. oidc.md

### DIFF
--- a/docs/specs/api.md
+++ b/docs/specs/api.md
@@ -2,28 +2,6 @@
 
 Here are the endpoints that are not part of OIDC, but essential to provide some features of Authgear.
 
-## /api/identity/biometric/remove
-
-This endpoint is for removing Biometric identity.
-The user still can remove biometric identity in the settings page.
-
-### Request Schema
-
-```json
-{
-  "type": "object",
-  "additionalProperties": false,
-  "properties": {
-    "kid": { "type": "string" }
-  },
-  "required": ["kid"]
-}
-```
-
-### Response Schema
-
-No response is necessary.
-
 ## /oauth2/challenge
 
 This endpoint is for requesting a one-time use, short-lived challenge.
--- a/docs/specs/api.md
+++ b/docs/specs/api.md
@@ -1,0 +1,56 @@
+# API
+
+Here are the endpoints that are not part of OIDC, but essential to provide some features of Authgear.
+
+## /api/identity/biometric/remove
+
+This endpoint is for removing Biometric identity.
+The user still can remove biometric identity in the settings page.
+
+### Request Schema
+
+```json
+{
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "kid": { "type": "string" }
+  },
+  "required": ["kid"]
+}
+```
+
+### Response Schema
+
+No response is necessary.
+
+## /oauth2/challenge
+
+This endpoint is for requesting a one-time use, short-lived challenge.
+The challenge is used in anonymous user and biometric authentication.
+
+## Request Schema
+
+```json
+{
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "purpose": { "type": "string" }
+  },
+  "required": ["purpose"]
+}
+```
+
+### Response Schema
+
+```json
+{
+  "type": "object",
+  "properties": {
+    "token": { "type": "string" },
+    "expire_at": { "type": "string" }
+  },
+  "required": ["token", "expire_at"]
+}
+```
--- a/docs/specs/biometric.md
+++ b/docs/specs/biometric.md
@@ -89,5 +89,5 @@ By default, biometric identity is hidden.
 ```yaml
 identity:
   biometric:
-    is_hidden: true
+    list_enabled: false
 ```
--- a/docs/specs/biometric.md
+++ b/docs/specs/biometric.md
@@ -57,28 +57,10 @@ This means enabling biometric authentication binds biometric authentication to t
 If the application is not a single-user application like the HSBC app,
 then we must ensure when User B logs in, User B can never access the keypair of User A.
 
-The SDK stores the following things:
-
-1. The keypair is stored in the device keychain per container, requiring biometric authentication to access.
-2. The last logged in user ID (LastUserID) is stored in the device keychain per container, NOT requiring biometric authentication to access.
-3. The user ID of last biometric setup (BiometricUserID) is stored in the device keychain per container, NOT requiring biometric authentication to access.
-
-Under normal condition, when biometric authentication is enabled, the user should log in with it.
-For some reasons like the user is wearing a mask or gloves, biometric authentication is impossible,
-the application should perform the typical web-based authentication flow to authenticate the user.
-It is possible that the user could log into a different account, which can lead to the situation that
-User B is logged in and the keypair of User A is still in the keychain.
-
-The HSBC app avoid this situation by making itself a single-user application.
-However, it is impossible for the SDK to enforce this.
-
-Instead, the SDK has the following behavior.
-If the BiometricUserID is not null, the SDK includes the BiometricUserID as `login_hint` in the authorization URL.
-The server will complain if the user tries to sign in a different user.
-If the developer really wants the user to sign in different accounts,
-they should prompt the user to disable biometric authentication first.
-
-Disabling biometric authentication deletes the keypair, which requires biometric authentication.
+The SDK stores the keypair in the keychain of the device per container, requiring biometric authentication to access.
+Under normal condition, when biometric authentication is enabled, the developer should let the user to use biometric authentication.
+Every time non-biometric authentication is used, the previous keypair will be deleted from the keychain.
+The user has to set up biometric authentication again.
 
 ## Configuration
 
--- a/docs/specs/biometric.md
+++ b/docs/specs/biometric.md
@@ -33,7 +33,11 @@ Next time the app is launched, the user can just use biometric authentication to
 
 It makes sense for a banking application to keep session lifetime short.
 
-## Design
+Once a user has logged in the HSBC app, the app remembers the user forever.
+It is impossible to switch user unless the app is uninstalled and installed again.
+The uninstallation wipes the keychain.
+
+## Design Overview
 
 Storing the refresh token implies long session lifetime.
 Storing the password poses a hard restriction on the authenticator the user can use.
@@ -43,3 +47,35 @@ we must store something unique.
 Borrowing the design of anonymous user, a keypair is stored in the keychain.
 Biometric authentication is required to retrieve the private key.
 The keypair is used to both identify and authenticate the user.
+
+## Security Concerns
+
+The common user experience of biometric authentication is by scanning fingerprint or face to
+log in an account which was signed in previously with non-biometric means.
+
+This means enabling biometric authentication binds biometric authentication to the application authentication.
+If the application is not a single-user application like the HSBC app,
+then we must ensure when User B logs in, User B can never access the keypair of User A.
+
+The SDK stores the following things:
+
+1. The keypair is stored in the device keychain per container, requiring biometric authentication to access.
+2. The last logged in user ID (LastUserID) is stored in the device keychain per container, NOT requiring biometric authentication to access.
+3. The user ID of last biometric setup (BiometricUserID) is stored in the device keychain per container, NOT requiring biometric authentication to access.
+
+Under normal condition, when biometric authentication is enabled, the user should log in with it.
+For some reasons like the user is wearing a mask or gloves, biometric authentication is impossible,
+the application should perform the typical web-based authentication flow to authenticate the user.
+It is possible that the user could log into a different account, which can lead to the situation that
+User B is logged in and the keypair of User A is still in the keychain.
+
+The HSBC app avoid this situation by making itself a single-user application.
+However, it is impossible for the SDK to enforce this.
+
+Instead, the SDK has the following behavior.
+If the BiometricUserID is not null, the SDK includes the BiometricUserID as `login_hint` in the authorization URL.
+The server will complain if the user tries to sign in a different user.
+If the developer really wants the user to sign in different accounts,
+they should prompt the user to disable biometric authentication first.
+
+Disabling biometric authentication deletes the keypair, which requires biometric authentication.
--- a/docs/specs/biometric.md
+++ b/docs/specs/biometric.md
@@ -79,3 +79,15 @@ If the developer really wants the user to sign in different accounts,
 they should prompt the user to disable biometric authentication first.
 
 Disabling biometric authentication deletes the keypair, which requires biometric authentication.
+
+## Configuration
+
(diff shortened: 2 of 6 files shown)


carmenlau:
The draft looks good to me so far

louischan-oursky:
@chpapa When you have time, can you have a look at this?

docs/specs/sdk.md
```diff
@@ -0,0 +1,140 @@
...
+ * 3. Call /oauth2/token to request refresh token and access token.
+ * 4. If the identity is not found, do the following steps.
+ *   4.1. Remove the kid from the list of keys so that listBiometricKeys no longer includes that kid.
+ */
+function authenticateBiometric(kid: string): Promise<AuthorizeResult>;
+```
+
+### Open questions
+
+#### Should we allow only 1 private key per container name?
```
chpapa, line 84: I think it is okay to allow only 1 private key per container, so we won't risk the problem of User A key is accessible to the app even when User B is using

chpapa (changes requested):
Basically just some comment, no big problem.

docs/specs/api.md
```diff
@@ -0,0 +1,56 @@
+# API
+
+Here are the endpoints that are not part of OIDC, but essential to provide some features of Authgear.
+
+## /api/identity/biometric/remove
+
+This endpoint is for removing Biometric identity.
+The user still can remove biometric identity in the settings page.
```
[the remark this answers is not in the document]
louischan-oursky: > Is it because all of them implement "Biometric" login like storing refresh/access token locally, and it is actually not a login, but a single session?

For HSBC, I guess they use biometric authentication to retrieve the mobile security key (a password) in the device keychain. So it should still be a new login every time biometric authentication is used.

> Or we just store the biometrics login data at server and ignore it...?

Do you mean we keep the current design and just do not show the biometric identity in the settings page?

> Or make it optional, so advanced users can also disable biometric login per device?

Not sure what you meant.

docs/specs/sdk.md
chpapa: @louischan-oursky if we support only 1 private key per container, Seems we don't need the API of `listBiometricKeys()` and `BiometricKey:kid`?

docs/specs/sdk.md
```diff
@@ -0,0 +1,148 @@
...
+1. It is possible to show a screen to let the user to log in either User A or User B with biometric only. However, in normal condition, the app should remember the last login user via non-biometric means. Biometric screen is only shown when there is such user. It is the application logic to make this happen.
+
+### Intended usage
+
+```typescript
+// Authenticate the user via non-biometric means.
+await authgear.authorize({ ... });
+
+// Store the last login user via non-biometric means.
+myapp.setUserID(userID);
```
chpapa, line 118: Why the user need to `setUserID` manually? `authgear.authorize()` success should already have the `userID`?

docs/specs/sdk.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+// When the session has expired.
```
chpapa, line 121: If we support one BioMetricKey per container only, I think the intended usage could be simplified to something like these?

```
if (authgear.hasBiometricKeys()) {
  // Show a screen to ask if they want to auth with biometric
 try {
   await authgear.authenticateBiometric();
 } catch (e) {
   // Check if identity error, fallback to non-biometric login
 }
}

File: docs/specs/sdk.md [hunk 5]
```diff
@@ -0,0 +1,148 @@
...
+1. User A logs out.
+1. User B logs in and sets up biometric. 2 keys is known to the SDK.
+1. User B logs out.
+1. It is possible to show a screen to let the user to log in either User A or User B with biometric only. However, in normal condition, the app should remember the last login user via non-biometric means. Biometric screen is only shown when there is such user. It is the application logic to make this happen.
+
+### Intended usage
+
+```typescript
+// Authenticate the user via non-biometric means.
+await authgear.authorize({ ... });
```
chpapa, line 115: After that, need to call `addBiometric()` to enable biometric?

docs/specs/api.md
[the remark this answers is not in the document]
chpapa: I meant, if we keep the current design, make it optional in setting page to show or hide the biometric identities.

It seems uncommon, but in our design it seems necessary 

(e.g. A potential security breach is someone stole the accounts, then create a Biometric login by API, but just store the key pair themselves, in this case the hacker have a way to always come back to the account without the user knowing it. Show list it in session page seems useful.)

chpapa (changes requested):
Questions on simplifying the API with only 1 key-pair per container only.

docs/specs/sdk.md
louischan-oursky: Adjusted the API to adapt 1 keypair per container.

docs/specs/sdk.md
louischan-oursky: Adjusted the API to store the last user ID.

docs/specs/sdk.md
louischan-oursky: Yes!

docs/specs/sdk.md
louischan-oursky: Yes. I have updated the intended usage to reflect this.

docs/specs/api.md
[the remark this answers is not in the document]
louischan-oursky: I prefer we just show the biometric identities along with other identities (login IDs, OAuth), until we learn a strong reason from the developer not to do so.

louischan-oursky:
@chpapa I have updated the spec to adopt 1 private key per container. See if it looks good to you?

docs/specs/api.md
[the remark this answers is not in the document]
chpapa: @louischan-oursky UX-wise, I don't even think it should be shown along with other identities. It should be shown in the sessions part?

I prefer to make it hidden by default, and let admin to enable it, just try to follows a "normal" behaviour before we know we should do differently otherwise.

chpapa, line 8: Can we put a note in the spec on where the Biometric identities should be shown and by default off?

docs/specs/api.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Request Schema
+
+```json
+{
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "kid": { "type": "string" }
```
chpapa, line 17: kid is still needed here?

docs/specs/sdk.md
```diff
@@ -0,0 +1,180 @@
...
+## Biometric authentication
+
+Biometric authentication is provided by the following API
+
+```typescript
+/**
+ * This function can be called at anytime.
+ * This function reports whether biometric authentication is available.
+ */
+function isBiometricAvailable(): Promise<boolean>;
```
chpapa, line 60: What does this meant? Is "available" meant the biometric setting is enabled or not in the server side? (If yes, maybe better specified here)

chpapa (changes requested):
Further review

docs/specs/api.md
louischan-oursky: Yes, in order to support the case that multiple devices have biometric authentication set up.

docs/specs/sdk.md
louischan-oursky: This is a feature checking on the device. For iOS devices, this function will probably return true. But for Android devices, this function reports whether the device support biometric. Let me clarify.

louischan-oursky: Our SDK does NOT support checking whether a particular feature is enabled on the server.

First, the Authgear server itself does not have an endpoint to get the configuration. Only the portal server can do this. The SDK communicates with the Authgear server only, not the portal server.

Second, the developer should have known which feature they have enabled, and only use the subset of API of SDK to complete their job.

chpapa: Ic ok. I was wondering too, and feel strange why we have such a api call.

docs/specs/sdk.md
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * This function can be called when the user has logged in, or at anytime if an option is specified.
+ * This function WILL trigger biometric authentication if necessary.
+ *
+ * The following steps are performed:
+ *
+ * 1. Call /api/identity/biometric/remove
+ * 2. If the error is identity not found, the error is always ignored.
+ * 3. If `force` is specified, the error is ignored, otherwise the error is thrown.
+ * 4. The key pair is deleted from the keychain.
```
carmenlau, line 102: Would like to know, do we know if step 4 requires face/touch id? If it requires face/touch id. In one scenario, developer use this api, the app asks for face id in the UI but user cancel it. I think the user may assume nothing should be deleted?

1. Need check if delete require face/touch id, if it is not needed. I think this flow is fine.
2. If it requires face/touch id, maybe better change the order of flow to delete the key pair first and the kid should be stored in somewhere else so we can perform `/api/identity/biometric/remove` in next step? But this is a bit awkward that if key pair is removed successfully but remove identity is failed. What should be returned in `isBiometricEnabled`, seems false? But the identity will be there forever and no one can access it.

Thought or any other idea?

docs/specs/sdk.md
```diff
@@ -0,0 +1,180 @@
...
+ */
+function isBiometricAvailable(): Promise<boolean>;
+
+/**
+ * This function can be called at anytime.
+ * This function returns the user ID of last logged in user.
+ * The data is stored with kSecAttrAccessibleWhenUnlockedThisDeviceOnly,
+ * therefore the data is not included in backup, nor transfered to new device.
+ */
+function getLastUserID(): Promise<string | null>;
```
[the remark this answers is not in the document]
louischan-oursky: if `getLastUserID` returns something and `isBiometricEnabled()` returns false, the the developer knows biometric was not setup and they should show a normal login screen. So it is quite useful to the developer.

louischan-oursky:
Some updates:

1. Remove the API to remove biometric identity. It is unavoidable to have dangling biometric identity (e.g. the app was uninstalled). So let's do not bother to remove it. This makes thing much simpler.
2. Rename isBiometricAvailable to isBiometricSupported.
3. Specify where the biometric identity is shown. Note that instead of show them in the session list, a new dedicated section in the settings page is introduced.

@chpapa @carmenlau

docs/specs/biometric.md
```diff
@@ -79,3 +79,15 @@ If the developer really wants the user to sign in different accounts,
...
+## Configuration
+
+Biometric identity is not shown along with Login ID identity and OAuth identity.
+Biometric identity is shown separately in the settings page.
+By default, biometric identity is hidden.
+
+```yaml
+identity:
+  biometric:
+    is_hidden: true
```
carmenlau, line 92: We had `list_enabled` in recovery code, do you think this name is suitable here too? Looks more consistent.

louischan-oursky: I was aware of this too. I was not sure whether we would have something else to enable in the future. But you seem agree to use `list_enabled` for consistency. Let me update to use `list_enabled`.

docs/specs/biometric.md
```diff
@@ -0,0 +1,93 @@
...
+For some reasons like the user is wearing a mask or gloves, biometric authentication is impossible,
+the application should perform the typical web-based authentication flow to authenticate the user.
+It is possible that the user could log into a different account, which can lead to the situation that
+User B is logged in and the keypair of User A is still in the keychain.
+
+The HSBC app avoid this situation by making itself a single-user application.
+However, it is impossible for the SDK to enforce this.
+
+Instead, the SDK has the following behavior.
+If the BiometricUserID is not null, the SDK includes the BiometricUserID as `login_hint` in the authorization URL.
```
chpapa, line 76: I'm skeptical if it is really necessary, and may cause more confusion.

Imagine there could be multiple ways the UX of what you describe can be designed:

1. Biometrics login if enabled, if failed, switch back to page page where you can choose to try Biometrics login again, or forgot your account (in this way, no need the behavior you described)
2. Biometrics login if enabled, if failed, display the normal web login page. However, there are no way the developer can allow the user to disable the biometrics here again? Since it is already redirected to our login page; In this page, if user want to login to another account, what can he do? There are no obvious way to do it, a deadlock and UX problem.

Can we instead, just make our SDK's `authorize({ ... })` always clear the biometrics key if there is one?

chpapa (changes requested):
Sorry another question when I read this @louischan-oursky

docs/specs/biometric.md
louischan-oursky: > Isn't that no matter what, if isBiometricEnabled() returns false, the developer knows biometric was not setup? How is it related if getLastUserID() return something or not?

It isBiometricEnabled() returns true, then getLastUserID() must return something.
If getLastUserID() returns nothing, then isBiometricEnabled() must return false.

> How should that be implemented? The developer should store the biometrics's user's Login ID or any meta data they want? is it possible?

This is only possible if we include all identities data in the user info endpoint. This is NOT yet done.

> Biometrics login if enabled, if failed, switch back to page page where you can choose to try Biometrics login again, or forgot your account (in this way, no need the behavior you described)

Failed for what reason? It is a failure due to covered face, then a retry on biometric authentication should work. Otherwise fallback to webview login limited to the last user.

> Biometrics login if enabled, if failed, display the normal web login page. However, there are no way the developer can allow the user to disable the biometrics here again? Since it is already redirected to our login page; In this page, if user want to login to another account, what can he do? There are no obvious way to do it, a deadlock and UX problem.

Xero shows "Sign in", which translates to web login limited to the last user. "Forget [email login ID]" translated to calling `disableBiometric()` followed by `authorize()`.

docs/specs/sdk.md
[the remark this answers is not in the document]
chpapa: Trying to put the relevant discussion under each comments.

> It isBiometricEnabled() returns true, then getLastUserID() must return something.
If getLastUserID() returns nothing, then isBiometricEnabled() must return false.

@louischan-oursky So in this case, why do we need to expose this API? Isn't that developers always only need to call `isBiometricEnabled` to determine if a login screen should be shown, or they should try to authenticate?

> This is only possible if we include all identities data in the user info endpoint. This is NOT yet done.

So for now developer should store the login ID on their own, if they want to display that when Biometric Login failed right?

docs/specs/biometric.md
chpapa: > Xero shows "Sign in", which translates to web login limited to the last user. "Forget [email login ID]" translated to calling disableBiometric() followed by authorize().

Actually this is not true, the "Sign in" actually called `authorizeBiometric()` again in Xero's case.

So go back to our original discussion, what do you think if we just by default make `authorize({ ... })` always clear the biometrics key if there is one?

carmenlau:
@louischan-oursky I think we could update the spec to match current implementation (e.g. remove `getLastUserID`... etc) and then merge it.

carmenlau (approved):

PR #1044, opened 2021-03-10
Support idle timeout for refresh token

### LINKED ISSUE #1000: Support idle timeout for refresh token

louischan-oursky:
fix #1000

### DIFF
--- a/pkg/lib/config/oauth.go
+++ b/pkg/lib/config/oauth.go
@@ -72,18 +72,18 @@ func (c *OAuthClientConfig) SetDefaults() {
 	}
 
 	if c.RefreshTokenLifetime == 0 {
-		if c.AccessTokenLifetime > DefaultSessionLifetime {
+		if c.AccessTokenLifetime > DefaultRefreshTokenLifetime {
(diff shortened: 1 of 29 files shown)


pkg/lib/config/oauth.go
```diff
@@ -49,28 +51,39 @@ var _ = Schema.Add("OAuthClientConfig", `
...
 			c.RefreshTokenLifetime = DefaultSessionLifetime
 		}
 	}
+
+	if c.RefreshTokenIdleTimeoutEnabled == nil {
+		b := DefaultSessionIdleTimeoutEnabled
+		c.RefreshTokenIdleTimeoutEnabled = &b
+	}
+	if c.RefreshTokenIdleTimeout == 0 {
+		c.RefreshTokenIdleTimeout = DefaultSessionIdleTimeout
```
carmenlau, line 87: I think better to use another constant, they can be the same value. Or rename the constant, so when we change the value we will know that it is affecting 2 config.

carmenlau (commented):
Please update portal to allow update the new idle setting

louischan-oursky:
Updated.

carmenlau (approved):

PR #1046, opened 2021-03-10
Onboarding in portal

### LINKED ISSUE #950: Onboarding

carmenlau:
fix #950 

Not able to test create app locally, will check it after deploying to staging.

Now the onboarding flow will be triggered this way:
1. After user authorize, portal will redirect user to "/onboarding-redirect". `OnboardingRedirect` will check and further redirect user to `/apps/create` if user doesn't have any apps.
2. After creating app in `/apps/create`, user will be redirected to `/app/:appID/onboarding` to configure the app.
3. After configuring the app, user will be redirected to `/app/:appID/done`. Completion page will show different screen (login / sso screenshot) based on user selected identities.

We can visit `/app/:appID/onboarding` to try the configure flow locally.


portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
```diff
@@ -0,0 +1,300 @@
...
+    if (btnItem.loginIDType) {
+      return state.pendingForm.loginIDKeys.has(btnItem.loginIDType);
+    } else if (btnItem.identityType) {
+      return state.pendingForm.identities.has(btnItem.identityType);
+    }
+    console.error(
+      "IdentitiesButton should have either identityType or loginIDType"
+    );
+    return false;
+  }, [state, btnItem]);
```
louischan-oursky, line 172: It is more natural to define this as `const checked = useMemo(...)`.

portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
```diff
@@ -0,0 +1,300 @@
...
+        identityType: "oauth",
+      },
+      {
+        labelId: "Onboarding.identities.anonymous-user",
+        iconName: "FollowUser",
+        identityType: "anonymous",
+      },
+    ],
+    []
+  );
```
louischan-oursky, line 291: This array is static. So it can be defined as a module level constant, saving us from using `useMemo`.

portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
```diff
@@ -378,6 +383,110 @@ const SecondaryAuthenticationModeContent: React.FC<SecondaryAuthenticationModeCo
...
+        labelId: "Onboarding.secondary-authenticators.oob-email",
+        authenticatorType: "oob_otp_email",
+      },
+      {
+        labelId: "Onboarding.secondary-authenticators.additional-password",
+        authenticatorType: "password",
+      },
+    ],
+    []
+  );
```
louischan-oursky, line 559: This can be made as a module level constant.

portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
```diff
@@ -487,6 +500,110 @@ const SecondaryAuthenticatorsContent: React.FC<SecondaryAuthenticatorsContentPro
...
+
+  const getSelectedKey = useCallback(() => {
+    if (state.pendingForm.verificationClaims[claimName]?.enabled) {
+      if (state.pendingForm.verificationClaims[claimName]?.required) {
+        return "required";
+      }
+      return "optional";
+    }
+    return "disabled";
+  }, [state, claimName]);
```
louischan-oursky, line 623: Change this to `const selectedKey = useMemo(...)`.

portal/src/hook/useAppConfigForm.ts
```diff
@@ -97,7 +99,10 @@ export function useAppConfigForm<State>(
     setIsUpdating(true);
     setUpdateError(null);
     updateConfig(newConfig)
-      .then(() => setCurrentState(null))
+      .then(() => {
+        setCurrentState(null);
+        setIsSubmitted(true);
```
louischan-oursky, line 104: When should be reset it to false? When the form is reset?

portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+interface FormState {
+  pendingForm: PendingFormState;
+}
+
+function constructFormState(_config: PortalAPIAppConfig): FormState {
+  return {
+    pendingForm: {
+      identities: new Set<IdentityType>(),
+      loginIDKeys: new Set<LoginIDKeyType>(),
```
louischan-oursky, line 65: By default, we should enable Login ID identity and the Email Login ID.

portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
```diff
@@ -254,6 +321,7 @@ const OnboardingConfigAppScreenForm: React.FC<OnboardingConfigAppScreenFormProps
         <FormattedMessage id="Onboarding.desc" />
       </Text>
       <IdentitiesListContent form={form} />
+      <PrimaryAuthenticatorsContent form={form} />
```
louischan-oursky: 1. Change primary authenticator to OTP
2. Add username
3. See OTP is disabled and STILL selected.

portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
carmenlau: Found that we cannot simply change the initial state to default value. If we change the initial state to default value, the AppConfigFormModel will treat the form isDirty as false and not allow saving.  Let me think how to handle it..

portal/src/graphql/portal/OnboardingFormContainer.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { useCallback } from "react";
```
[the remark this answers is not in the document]
carmenlau: Since the `FormContainer` and `OnboardingFormContainer` have different layout, `FormContainer` have the save and reset bottom on the top bar, and `OnboardingFormContainer` need to have the button at the bottom. My first thought is I don't want to make `FormContainer` supports multiple layout as we may have more layouts in future and it will make `FormContainer` difficult to maintain. Thought?

portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
louischan-oursky: And we should disable save button if nothing has been selected for the first question.

louischan-oursky:
We should sort

- Login ID keys
- Secondary authenticators

portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
carmenlau: Updated in 90d379c

carmenlau:
Updated!

portal/src/hook/useAppConfigForm.ts
```diff
@@ -45,6 +48,7 @@ export function useAppConfigForm<State>(
   const [isUpdating, setIsUpdating] = useState(false);
   const [updateError, setUpdateError] = useState<unknown>(null);
   const [isSubmitted, setIsSubmitted] = useState(false);
+  const [canSave, setCanSave] = useState<boolean | undefined>(initialCanSave);
```
louischan-oursky, line 51: Can we receive `canSave` from argument directly? I mean something like

```
interface Options {
  validate?: (state: State) => APIError | null;
  canSave?: boolean;
}

function useAppConfigForm(..., options?: Options) {
  const canSave = options?.canSave;
  const validate = options?.validate;
}
```

Then we do not need to expose `setCanSave`, and do not need to use `useEffect` to sync `canSave` in the hook caller.

carmenlau: Sure, let me revise :)

carmenlau: Oh, I found we cannot update this way. In most of the case, the canSave state is determined by the current form state, and the form state is created by  `useAppConfigForm`. So it is difficult for us to provide canSave in `useAppConfigForm` argument. Seems having `setCanSave` is more flexible. Thought?

louischan-oursky: OK. let's keep what we have now.

PR #1053, opened 2021-03-11
Set more cache friendly headers in HTTP response

### LINKED ISSUE #1052: Resources served via HTTP should be cacheable
Right now all responses have the `Cache-Control: no-store` header

louischan-oursky:
fix #1052

### DIFF
--- a/pkg/auth/routes.go
+++ b/pkg/auth/routes.go
@@ -32,24 +32,31 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource, st
 		p.Middleware(newPanicLogMiddleware),
 		p.Middleware(newSessionMiddleware),
 		p.Middleware(newSecHeadersMiddleware),
-		httproute.MiddlewareFunc(httputil.NoCache),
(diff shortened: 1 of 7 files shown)


pkg/auth/routes.go
```diff
@@ -32,23 +32,31 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource, st
 		p.Middleware(newPanicLogMiddleware),
 		p.Middleware(newSessionMiddleware),
 		p.Middleware(newSecHeadersMiddleware),
-		httproute.MiddlewareFunc(httputil.NoCache),
+	)
+
+	staticChain := httproute.Chain(
+		rootChain,
+		p.Middleware(newCORSMiddleware),
+		httproute.MiddlewareFunc(httputil.ETag),
```
carmenlau, line 40: Seems better to separate chain for static and oauth config? Although we can use same middleware now, we may want to have different caching strategy in future. The name `staticChain` makes me think we only use it in static asset.

carmenlau (commented):
Very minor comment, overall looks good.:)

Issue #906, opened 2021-01-22
Domain blacklist and whitelist for Email Login ID signup


chpapa:
Allow users to configure, either:

- a blacklist of domains that is disallowed to signup
- a whitelist of domains that is only allowed for signup

This issue should consist of both:

- server daemon, config file
- portal

Some notes for implementation:

- the blacklist/whitelist should affect only signup, so users created from portal or via admin API are not affected
- it would be nice, on portal end, we have a default free email provider list, that users could have a checkbox to include those in the blacklist. (It should be possible to google for such a list)

mr9dollars:
1. Do we want support regex for black/whitelist?
2. If regex, what should be the behaviour if the domain falls in both blacklist and whitelist?

chpapa:
1. I think no need regex? (Or we can add it if we do have a use cases)
2. I think the blacklist and whitelist should be mutually exclusive (so you either configure to use the whitelist, or use the blacklist), that's why "either".

carmenlau:
Added a simple spec for the feature https://github.com/authgear/authgear-server/pull/1055

1. We will not support regex, only exact match the domain
2. blacklist and whitelist is mutually exclusive

PR #1055, opened 2021-03-12
Add domain blacklist and whitelist for email login id spec

carmenlau:
ref #906 

Maybe we can use this list as the default blacklist: https://gist.github.com/tbrianjones/5992856

### DIFF
--- a/docs/specs/user-model.md
+++ b/docs/specs/user-model.md
@@ -173,7 +173,8 @@ Login ID type determines the validation, normalization and unique key generation
 - Disallow confusing homoglyphs
 - Validate against PRECIS IdentifierClass profile
 - Disallow builtin reserved usernames (Configurable, default ON)
-- Disallow developer-provided reserved usernames (Configurable, default empty list)
+- Disallow username contains developer-provided keywords (Configurable, default OFF)
+  - Developer will need to provide their exclude keywords in txt file, separated by newline.
 - Check ASCII Only (`a-zA-Z0-9_-.`) (Configurable, default ON)
 
 ###### Normalization of Username Login ID
--- a/docs/specs/user-model.md
+++ b/docs/specs/user-model.md
@@ -148,6 +148,12 @@ Login ID type determines the validation, normalization and unique key generation
 
 - [RFC5322](https://tools.ietf.org/html/rfc5322#section-3.4.1)
 - Disallow `+` sign in the local part (Configurable, default OFF)
+- Domain blacklist / whitelist
+  - Block blacklisted domains (Configurable, default OFF, can be ON only if *Allow whitelisted domains* is OFF)
+  - Include default blacklisted domains (Configurable, default OFF, can be ON only if *Block blacklisted domains* is ON)
+  - Allow whitelisted domains only (Configurable, default OFF, can be ON only if *Block blacklisted domains* is OFF)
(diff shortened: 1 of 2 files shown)


docs/specs/user-model.md
```diff
@@ -148,6 +148,12 @@ Login ID type determines the validation, normalization and unique key generation
 
 - [RFC5322](https://tools.ietf.org/html/rfc5322#section-3.4.1)
 - Disallow `+` sign in the local part (Configurable, default OFF)
+- Domain blacklist / whitelist
+  - Block blacklisted domains (Configurable, default OFF, can be ON only if *Allow whitelisted domains* is OFF)
+  - Include default blacklisted domains (Configurable, default OFF, can be ON only if *Block blacklisted domains* is ON)
```
chpapa, line 153: 1. I don't think it should be called "default backlisted domains", it is unclear for user "why there is a default?" "what does default meant?"; Instead it should be specific, it is an option to "block email addresses from free email provider domains"
2. I'm not sure if we should have an option here? Maybe it should just be an option for Portal?

docs/specs/user-model.md
```diff
@@ -148,6 +148,12 @@ Login ID type determines the validation, normalization and unique key generation
...
+  - Allow whitelisted domains only (Configurable, default OFF, can be ON only if *Block blacklisted domains* is OFF)
+  - Domain blacklist / whitelist only affect user signup, users created from portal or via admin API are not affected
+  - Developer will need to provide their blacklist / whitelist in txt file, separated by newline. A default blacklist will be included when *Include default blacklisted domains* is ON
```
chpapa, line 156: Seems inconsistent with keywords in username? Shall we stick with one?

pkg/lib/config/testdata/default_config.yaml
```diff
@@ -270,6 +270,9 @@ identity:
         case_sensitive: false
         block_plus_sign: false
         ignore_dot_sign: false
+        domain_blacklist_enabled: false
+        domain_whitelist_enabled: false
+        include_default_blacklisted_domains: false
```
chpapa, line 275: as mentioned above, "default blacklisted domains" is not clear.

chpapa (changes requested):

docs/specs/user-model.md
carmenlau: 1. Sure, let me change it!
2. Actually in the implementation, we will not copy the list during portal save config. Instead, we will load the list dynamically from the code, so we can update the list.

pkg/lib/config/testdata/default_config.yaml
carmenlau: Let me change it to `block_free_email_provider_domains: false`

docs/specs/user-model.md
carmenlau: Just afraid that the list will be too long to save in config, maybe we can:

1. If we think the developer provided list will not be too long, we can move the blacklist / whitelist config back to the yaml config file. So it will be consistent to username's excluded keywords.

or

2. Move username's excluded keywords to txt file. But this will be another breaking change..>"<

Thought?

carmenlau:
Updated @louischan-oursky @chpapa

chpapa (approved):

louischan-oursky (approved):

PR #1088, opened 2021-03-18
Support email blocklist allowlist

### LINKED ISSUE #906: Domain blacklist and whitelist for Email Login ID signup

carmenlau:
require #1083 

fix #906


pkg/admin/facade/interaction.go
```diff
@@ -43,6 +43,13 @@ func (adminAPIOp) BypassPublicSignupDisabled() bool {
 	return true
 }
 
+// BypassLoginIDEmailByPassBlocklistAllowlist indicates Admin API operations
+// bypass email domains blocklist allowlist checking; Admin API can create
+// email login id that doesn't have blocklist allowlist restriction
+func (adminAPIOp) BypassLoginIDEmailByPassBlocklistAllowlist() bool {
```
louischan-oursky, line 49: It seems that the `ByPass` in the middle is extra. Would `BypassLoginIDEmailBlocklistAllowlist` a better name?

portal/src/graphql/portal/LoginIDConfigurationScreen.tsx
```diff
@@ -450,6 +513,41 @@ const AuthenticationLoginIDSettingsContent: React.FC<AuthenticationLoginIDSettin
           "LoginIDConfigurationScreen.email.blockPlusTooltipMessage"
         )}
       />
+      <CheckboxWithContent
```
louischan-oursky, line 604: I have a suggestion here. Instead of disabling the checkboxes, can we smartly change the checked status of the checkboxes based on what have been clicked?

For example, if allowlist is checked, then blocklist is disabled and block free provider is disabled.

I think this is more user-friendly.

portal/src/graphql/portal/LoginIDConfigurationScreen.tsx
```diff
@@ -547,6 +648,19 @@ const AuthenticationLoginIDSettingsContent: React.FC<AuthenticationLoginIDSettin
...
+              "LoginIDConfigurationScreen.email.domainAllowlist"
+            ),
+            placeholder: "example.com",
+          }}
+          className={styles.widgetInputField}
+          disabled={!state.email.domain_allowlist_enabled}
+          selectedItems={domainAllowlist}
+          onChange={onDomainAllowlistChange}
+          onResolveSuggestions={onDomainAllowlistSuggestions}
+        />
```
louischan-oursky, line 663: It would be great if we can warn the developer that allowlist is enabled but the allowlist is empty. If the allowlist is enabled and is empty, signup is not possible. Maybe we can display the warning as error under the control?

portal/src/graphql/portal/LoginIDConfigurationScreen.tsx
```diff
@@ -525,6 +613,19 @@ const AuthenticationLoginIDSettingsContent: React.FC<AuthenticationLoginIDSettin
...
+              "LoginIDConfigurationScreen.email.domainBlocklist"
+            ),
+            placeholder: "example.com",
+          }}
+          className={styles.widgetInputField}
+          disabled={!state.email.domain_blocklist_enabled}
+          selectedItems={domainBlocklist}
+          onChange={onDomainBlocklistChange}
+          onResolveSuggestions={onDomainBlocklistSuggestions}
+        />
```
louischan-oursky, line 628: TagPicker has a UX problem. If I enter some text and then clear the focus, the text is not "added" to the picker. I think the behavior should be add-item-on-blur. What do you think?

portal/src/graphql/portal/LoginIDConfigurationScreen.tsx
louischan-oursky: Discussed with Carmen and we decided to keep the current implementation.

louischan-oursky:
Can I close https://github.com/authgear/authgear-server/pull/1083 and just merge this PR when it is ready?

portal/src/graphql/portal/LoginIDConfigurationScreen.tsx
carmenlau: Added tooltip to labels

portal/src/graphql/portal/LoginIDConfigurationScreen.tsx
carmenlau: I tried use the method we discussed to fix the problem, but still have another problem. When I added the input value to list, I cannot clear the text in input as the input element is controlled by fluent ui. :(

carmenlau: Updated, thanks @louischan-oursky !!

Issue #1041, opened 2021-03-09
Update webhook events


carmenlau:
1. Include user preferred language in webhook context for developer to customise message (title and reason)
1. Update webhook response to add title and drop data. e.g. 
        ```
        {
        "is_allowed": false,
        "title": "Signup is not allowed",
        "reason": "Any resaon"
        }
        ```
1. If reason is not provided in hook, error page will show the default reason with basic context. e.g. "Signup is blocked by webhook".
1. Rename / drop some existing webhooks
     - Rename "before_user_create, after_user_create" to "before_signup_completion, after_signup_completion"
     - (TBC)

carmenlau:
@louischan-oursky @chpapa @mr9dollars 

Here is a summary for us to review the webhooks,

The webhooks that are using by some existing projects:

- after_identity_create
- after_identity_delete
- after_user_promote
- after_user_create

The webhooks that we implemented in code:

- before_user_create / after_user_create
- before_identity_create / after_identity_create
- before_identity_delete / after_identity_delete
- before_identity_update / after_identity_update
- before_session_create / after_session_create
- before_session_delete / after_session_delete
- before_user_promote / after_user_promote
- user_sync

My thought:

1. In product meeting, Ben mentioned that wants to change `(before|after)_user_create` to `(before|after)_signup_completion`, but after checking, I found that the hook will be called in all user creation cases (signup by user / create user by admin). Seems `user_create` is more accurate.

1. After reading the detail, I think the current hooks definition are quite well actually. Instead of separating different scenarios by the hook type, developer can use the webhook payload to identify what is happening. For example, hook `(before|after)_session_create` will be triggered when creating session, there is `reason` in the payload for developer to know it is `signup` or `login`. If developer wants to block both `signup` and `login`, they only need to implement 1 hook. If they want to block login only, it is clear for them to check the payload reason. The current hooks definition are easy to imagine when the hooks will be triggered, but the developer may need to know what is `session` and `identity`. So I prefer to keep the existing hooks. What do you think?

1. See if you guys have better idea to rename / change the hooks? Some implementation detail: To handle the existing project, we can make a authgear version that support both old and new hook type. We have a `type` in the hook payload, we may need further ask if anyone of them are using that attribute and see how to handle it gracefully. If no one is using that attribute, we can write migration script to migrate all app configs to new hooks. After all apps config have been migrated, then we can update the code to drop the old hooks.

louischan-oursky:
My suggestions

[...]

PR #1091, opened 2021-03-22
[Don't merge] Webhook update proposal

carmenlau:
ref #1041 

This PR is for discussion only. `webhook_update.md` will be removed and combined to `webhook.md`.

For reading:
https://github.com/carmenlau/authgear-server/blob/1040-webhooks-spec/docs/specs/webhook_update.md

Discussion questions:
https://github.com/carmenlau/authgear-server/blob/1040-webhooks-spec/docs/specs/webhook_update.md#discussion-questions

### DIFF
--- a/docs/specs/webhook_update.md
+++ b/docs/specs/webhook_update.md
@@ -1,0 +1,87 @@
+## Update webhooks
+
+### Concepts
+
+Instead of having before or after hooks, introduce new concept: blocking and non-blocking hooks.
+
+Actually blocking hooks are the before books and non-blocking is after. But they don't need to be exists in pair anymore.
+
+Blocking hooks accept below response body, operation will be aborted when `is_allowed` is `false`. `title` and `reason` will be shown in auth ui.
+
+```json5
+{
+  "is_allowed": false,
+  "title": "any title",
+  "reason": "any string",
+}
+```
+
+Non-blocking hooks are the same as the after event hooks, response body will be ignored.
+
+
+### Naming convention
+
+For blocking hooks, we create them by different scenarios, so no specific naming convention.
+
+For non-blocking hooks, the main purposes are for data synchronization / app server notification, we will follow the pattern of `<entity>.<action>.<reason>`.
+
+
+### Proposed hooks
+
+#### Blocking hooks
+
+- pre_registration
+- admin_create_user
+
+#### Non-blocking hooks
+
+- user.created.user_signup
+- user.created.admin_api_create
+- identity.created.user_create
+- identity.created.admin_api_create
+- identity.deleted.user_delete
+- identity.deleted.admin_api_delete
+- identity.updated.user_update
+- session.created.user_signup
+- session.created.user_login
+- session.deleted.user_revoke
+- session.deleted.user_logout
+- session.deleted.admin_api_revoke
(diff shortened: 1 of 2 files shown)


docs/specs/webhook_update.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+Please check the below list for webhook shape
+- [Webhook shape](./webhook.md#webhook-event-shape)
+- [Payload of events](./webhook.md#blocking-webhook-event-list)
+
+
+### Discussion questions
+
+1. Seems we cannot solve developers may missed some of the events after we add new events problem, unless we support subscribing event type `user.created.*`. But we cannot do it in my blocking hooks definition, so eventually they will miss. I found the existing services don't have this problem, since they didn't put the reason in the event name, even in the request payload. Developer has no way to know the reason why the event is triggered. e.g.
```
carmenlau, line 80: > Louis: Do we follow Stripe and send all events to a single endpoint? If yes, then the developer can actually match the event type with a prefix like user.created. provided that they do not care about the reason.

I think the major problems is blocking hooks, stripe doesn't have blocking events. If we allow user subscribe all blocking hooks in one endpoint, the default implementation of his webhook will be `is_allowed` = true? There are some problems that I can think of:

1. There are some performance tradeoff to have many unnecessary for **blocking hooks**, do we think it is acceptable?
2. We may need to have naming convention so developer can match prefix, but the naming need to be flexible enough. e.g. Ben mentioned example `after_signup_filled_email`.

If we want to go for this direction, we can think about how to solve point 2.

docs/specs/webhook_update.md
```diff
@@ -0,0 +1,87 @@
...
+  "id": "0E1E9537-DF4F-4AF6-8B48-3DB4574D4F24",
+  "seq": 435,
+  "type": "after_user_create",
+  "payload": { /* ... */ },
+  "context": { /* ... */ }
+}
+```
+
+1. `type` will be the new event type
+1. Add `locale` to context
```
carmenlau, line 69: Louis's comment:

Let's include more information here

`preferred_languages`: A list of BCP47 tags we inferred from the request
`supported_languages`: The list of supported languages in the configuration.
`fallback_language`: The fallback language in the configuration.
`resolved_language`: The language we resolved(?) Actually it is quite hard to derive this. If the request is made to the web UI. We actually resolved various resources individually. So there could be multiple resolved languages.

The main concern is that if we just pass `preferred_languages`, `supported_languages` and `fallback_language`, the developer could resolve something different from us.

docs/specs/webhook_update.md
louischan-oursky: Actually my comment is mainly about the non-blocking events.

For blocking events, I think the following properties are important to the developer.

- If the developer has defined a blocking hook for a particular event, then it is very likely that they want to know about the event and take actions. For example, if the developer has defined a blocking hook to inspect the email, it is very likely that they want to know about ALL scenarios. If our design could make them miss the event, it would be very surprising.
- If the developer has defined a blocking hook for a particular event, then the developer must explicitly tell us to allow the operation. Otherwise, the operation is allowed by default.

carmenlau:
Spec update will be handled in #1092

PR #1096, opened 2021-03-24
Fix unexpected error when aborting signup in before_session_create hook

carmenlau:
ref #1041 

This PR is fixing an existing bug in hook provider.

### Steps to reproduce
- Register `before_session_create` hook only and return `is_allowed=false`
- Signup  a user

### About the fix
In hook provider, DispatchEvent will be called multiple times. If the first DispatchEvent
is success, dbHook is registered already even the latter events have error. If there is
error, all the later operations should be skipped, so we need check and skip them.
We cannot see this problem when aborting signup in before_user_create, as
before_user_create is the first event in signup, so dbHook is not be registered in this case 
and no further DispatchEvent will be called.

### DIFF
--- a/pkg/lib/hook/provider.go
+++ b/pkg/lib/hook/provider.go
@@ -51,10 +51,22 @@ type Provider struct {
 
 	persistentEventPayloads []event.Payload `wire:"-"`
 	dbHooked                bool            `wire:"-"`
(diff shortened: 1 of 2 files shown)


carmenlau:
Updated!

PR #1097, opened 2021-03-25
Revamp webhook

carmenlau:
ref #1041 

- [x] New blocking and non-blocking events
- [x] Drop all old events, config
- [x] Update test case to use new events
- [x] Add language to webhook context
- [x] Display webhook error title and reason
- [x] Update portal


pkg/lib/hook/provider.go
```diff
@@ -89,6 +93,27 @@ func (provider *Provider) DispatchEvent(payload event.Payload) (err error) {
 		provider.persistentEventPayloads = append(provider.persistentEventPayloads, payload)
 		err = nil
 
+	case event.BlockingPayload:
+		if provider.Deliverer.WillDeliverBlockingEvent(typedPayload.BlockingEventType()) {
+			seq, err = provider.Store.NextSequenceNumber()
+			if err != nil {
+				err = errorutil.HandledWithMessage(err, "failed to dispatch event")
```
louischan-oursky, line 100: HandledWithMessage is for hiding error. If we want to add context to the error, we should write `err = fmt.Errorf("failed to dispatch event: %w", err)`

pkg/lib/interaction/nodes/do_create_identity.go
```diff
@@ -17,13 +18,21 @@ type EdgeDoCreateIdentity struct {
 }
 
 func (e *EdgeDoCreateIdentity) Instantiate(ctx *interaction.Context, graph *interaction.Graph, rawInput interface{}) (interaction.Node, error) {
+	isAdminAPI := false
+	var adminInput interface{ IsAdminAPI() bool }
+	if interaction.Input(rawInput, &adminInput) {
+		isAdminAPI = adminInput.IsAdminAPI()
+	}
```
louischan-oursky, line 25: This 5 lines may worth extracting to a helper function.

portal/src/graphql/portal/WebhookConfigurationScreen.tsx
```diff
@@ -38,13 +43,49 @@ interface FormState {
...
+  // always adding non_blocking_handler in index 0
+  // to fix the json pointer for rendering error
+  if (currentState.non_blocking_handler_endpoint !== "") {
+    nonBlockingConfig.unshift({
+      events: ["*"],
+      url: currentState.non_blocking_handler_endpoint,
+    });
+  }
+
+  return nonBlockingConfig.length > 0 ? nonBlockingConfig : undefined;
```
louischan-oursky, line 80: I think we should allow the developer to add multiple URLs, so the UI should look like the blocking ones, without event type dropdown.

pkg/lib/config/hook.go
```diff
@@ -7,7 +7,6 @@ var _ = Schema.Add("HookConfig", `
 	"properties": {
 		"sync_hook_timeout_seconds": { "$ref": "#/$defs/DurationSeconds" },
 		"sync_hook_total_timeout_seconds": { "$ref": "#/$defs/DurationSeconds" },
-		"handlers": { "type": "array", "items": { "$ref": "#/$defs/HookHandlerConfig" } },
```
louischan-oursky: If we remove the config, then do we need to provide migration for existing production apps? Otherwise once we upgrade the server image tag, the existing app will have config error.

carmenlau: I thought we will update the config for the specific project manually, or you think using script is better? I can add the script

louischan-oursky: Either way has downtime. We just have 1 app. So manually is fine.

portal/src/graphql/portal/WebhookConfigurationScreen.tsx
carmenlau: Updated!

carmenlau:
Some comments note:
- [ ] In portal, add placeholder to url 
- [ ] Add doc link to portal webhook description
- [ ] Rename `context.resolved_language` to `context.language`

carmenlau:
Updated.

PR #1101, opened 2021-03-30
Biometric portal support

### LINKED ISSUE #1099: Biometric portal support
- Allow removing biometric identity
- Update config enable biometric

carmenlau:
fix #1099


portal/src/util/deviceinfo.ts
```diff
@@ -0,0 +1,149 @@
...
+  "iPad11,3": "iPad Air (3rd generation)",
+  "iPad11,4": "iPad Air (3rd generation)",
+  "AppleTV2,1": "Apple TV", // Apple TV (2nd Generation)
+  "AppleTV3,1": "Apple TV", // Apple TV (3rd Generation)
+  "AppleTV3,2": "Apple TV", // Apple TV (3rd Generation - Rev A)
+  "AppleTV5,3": "Apple TV", // Apple TV (4th Generation)
+  "AppleTV6,2": "Apple TV 4K", // Apple TV 4K
+};
+
+export function formatDeviceInfo(deviceInfo: any): string | undefined {
```
louischan-oursky, line 111: I think it is better to use the formatted device info from the server. I suggest we add `IdentityClaimBiometricFormattedDeviceInfo` whose value is a string. Then we can avoid duplicate the device info formatting logic.

carmenlau: Since the formatted device info can be derived from device info anytime, so I didn't store `IdentityClaimBiometricFormattedDeviceInfo` in db. Updated, see if it looks good?

pkg/lib/authn/identity/service/conversion.go
```diff
@@ -123,9 +123,10 @@ func anonymousFromIdentityInfo(i *identity.Info) *anonymous.Identity {
 
 func biometricToIdentityInfo(b *biometric.Identity) *identity.Info {
 	claims := map[string]interface{}{
-		identity.IdentityClaimBiometricKeyID:      b.KeyID,
-		identity.IdentityClaimBiometricKey:        string(b.Key),
-		identity.IdentityClaimBiometricDeviceInfo: b.DeviceInfo,
+		identity.IdentityClaimBiometricKeyID:               b.KeyID,
+		identity.IdentityClaimBiometricKey:                 string(b.Key),
+		identity.IdentityClaimBiometricDeviceInfo:          b.DeviceInfo,
+		identity.IdentityClaimBiometricFormattedDeviceInfo: b.FormattedDeviceInfo,
```
louischan-oursky, line 129: My original thought was to call `deviceinfo.Format(b.DeviceInfo)` here. So we do not need to add a new field to `biometric.Identity`. On the other hand, we do not need to do additional things in biometricFromIdentityInfo because the new field does not exist.

carmenlau: Oh, I got your point. Let me update it again.

carmenlau: Updated!

PR #1114, opened 2021-04-07
Update portal deps

louischan-oursky:
If we update `@monaco-editor/react` to 4, our current usage will make it hang.

If we update `immer` to 9, there are many type errors.


carmenlau:
Please rebase!

louischan-oursky:
Rebased.

PR #1123, opened 2021-04-08
Support AD FS

### LINKED ISSUE #831: Research AD FS

louischan-oursky:
fix #831


pkg/lib/authn/sso/adfs.go
```diff
@@ -0,0 +1,128 @@
...
+	if err == nil {
+		// upn looks like an email address.
+		normalizer := f.LoginIDNormalizerFactory.NormalizerWithLoginIDType(config.LoginIDKeyTypeEmail)
+		email, err = normalizer.Normalize(upn)
+		if err != nil {
+			return
+		}
+	} else {
+		// upn does NOT look like an email address.
+		preferredUsername = upn
```
carmenlau, line 110: Or we should always fill `upn` in `preferredUsername`? So adfs always have `preferredUsername` consistently. I am open to this

pkg/lib/authn/sso/userinfo.go
```diff
@@ -61,6 +63,16 @@ func DecodeAzureADv2(userInfo map[string]interface{}) *ProviderUserInfo {
...
 }
 
+func DecodeADFS(userInfo map[string]interface{}) *ProviderUserInfo {
+	id, _ := userInfo["sub"].(string)
+	email, _ := userInfo["email"].(string)
+
+	return &ProviderUserInfo{
+		ID:    id,
+		Email: email,
+	}
```
carmenlau, line 73: I found that not all our oauth provider is calling `DecodeUserInfo`? `DecodeUserInfo` is useful only if we call it in `<provider>.go`? If this is the case, maybe we should not add this, since this is not the way to convert `userInfo` to `ProviderUserInfo` for adfs, it is confusing.

pkg/lib/authn/sso/adfs.go
louischan-oursky: Let me change to always fill in upn.

louischan-oursky:
updated.

Issue #1129, opened 2021-04-09
Direct logout


louischan-oursky:
- For web cookie use case, introduce a new grant type for the token endpoint to echo the ID token for the current cookie.
- For web refresh token use case, use `grant_type=refresh_token` (i.e. refresh access token) to obtain ID token.
- Add `sid` claim to ID token. https://openid.net/specs/openid-connect-frontchannel-1_0.html#OPLogout
- Make `end_session_endpoint` accept `id_token_hint`. If `id_token_hint` is present and is valid, then logout directly. Otherwise fallback to existing confirmation flow. https://openid.net/specs/openid-connect-rpinitiated-1_0.html

> When an id_token_hint parameter is present, the OP MUST validate that it was the issuer of the ID Token.

> The OP SHOULD accept ID Tokens when the RP identified by the ID Token's aud claim and/or sid claim has a current session or had a recent session at the OP, even when the exp time has passed.

> If the ID Token's sid claim does not correspond to the RP's current session or a recent session at the OP, the OP SHOULD treat the logout request as suspect, and MAY decline to act upon it.

---

- Update portal frontend to use the SDK version that includes this change.

PR #1134, opened 2021-04-13
Direct logout

louischan-oursky:
ref #1129

### DIFF
--- a/pkg/lib/oauth/oidc/handler/handler_end_session.go
+++ b/pkg/lib/oauth/oidc/handler/handler_end_session.go
@@ -16,13 +16,41 @@ type WebAppURLsProvider interface {
 	SettingsURL() *url.URL
 }
 
+type IDTokenIssuer interface {
+	ValidateIDToken(idToken string) error
+}
+
+type LogoutSessionManager interface {
+	Logout(session.Session, http.ResponseWriter) error
+}
+
 type EndSessionHandler struct {
-	Config    *config.OAuthConfig
-	Endpoints oidc.EndpointsProvider
-	URLs      WebAppURLsProvider
+	Config         *config.OAuthConfig
+	Endpoints      oidc.EndpointsProvider
+	URLs           WebAppURLsProvider
+	IDTokens       IDTokenIssuer
+	SessionManager LogoutSessionManager
 }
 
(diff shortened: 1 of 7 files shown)


pkg/lib/oauth/oidc/handler/handler_end_session.go
```diff
@@ -16,13 +16,41 @@ type WebAppURLsProvider interface {
...
+	Endpoints      oidc.EndpointsProvider
+	URLs           WebAppURLsProvider
+	IDTokens       IDTokenIssuer
+	SessionManager LogoutSessionManager
 }
 
 func (h *EndSessionHandler) Handle(s session.Session, req protocol.EndSessionRequest, r *http.Request, rw http.ResponseWriter) error {
+	idTokenHint := req.IDTokenHint()
+	if s != nil && idTokenHint != "" {
+		err := h.IDTokens.ValidateIDToken(idTokenHint)
```
carmenlau, line 38: Seems we only validate the id token here, but we have not compared it with the current session?
If this is the case, there will be a problem that people can use another peoples'  id token to logout current user.

pkg/lib/oauth/oidc/id_token.go
```diff
@@ -48,11 +94,16 @@ func (ti *IDTokenIssuer) IssueIDToken(client *config.OAuthClientConfig, s sessio
 	now := ti.Clock.NowUTC()
 
 	_ = claims.Set(jwt.AudienceKey, client.ClientID)
+
+	_ = claims.Set("sid", EncodeSID(s))
```
carmenlau, line 167: Seems `sid` is custom claim. Will we don't some prefix to it, do we have similar case in the existing code base?

https://openid.net/specs/openid-connect-core-1_0.html#IDToken

louischan-oursky: No. It is defined here. https://openid.net/specs/openid-connect-frontchannel-1_0.html#OPLogout

pkg/lib/oauth/oidc/handler/handler_end_session.go
louischan-oursky: This is quite tricky. Can we have a chat on this.

Issue #880, opened 2021-01-13
Allow MFA to be skipped if a particular OAuth provider is used as primary authenticator


louischan-oursky:
In practice, the developer configure a OAuth provider which has already enforced all users to enable MFA. So it is a bit weird if the user has to, once again, perform MFA during authentication with Authgear. However, we still want users who are using other primary authenticators (e.g. email login ID with primary password) to go through MFA.

chpapa:
I think a config for each external login:

- Require MFA (default)
- Require MFA only if not performed (only if the provider support [Authentication Method Reference Values](https://tools.ietf.org/html/rfc8176#section-2))
- Skip MFA

mr9dollars:
@louischan-oursky Should also able to skip MFA when using OAuth to sign up.

louischan-oursky:
- If the user has only OAuth identities, then the user cannot add MFA. The user does not need to setup MFA in any mode. The user does not need to perform MFA during login.
- If the user does not need MFA, then they cannot see MFA in settings page.

chpapa:
@louischan-oursky A few questions / thoughts:

1. Does most common OAuth provider we support now provide any information if the user already have MFA on their side? I remember there are something like that in OAuth spec but not sure if it is implemented.

2. Is it a configuration now to disable / enable MFA for OAuth provider, or this issue avoid all? Just wanna know, not sure if such a configuration make sense.

louischan-oursky:
1. No, none of them provide such information.
2. No configuration at the moment. All OAuth provider will skip MFA.

PR #1135, opened 2021-04-13
Skip MFA for OAuth during signup AND login

louischan-oursky:
ref #880


pkg/lib/interaction/nodes/create_authenticator_begin.go
```diff
@@ -58,9 +58,7 @@ func (n *NodeCreateAuthenticatorBegin) Prepare(ctx *interaction.Context, graph *
 		return err
 	}
 
-	if n.Stage == interaction.AuthenticationStagePrimary {
```
carmenlau, line 61: Would like to know why don't we need this checking any more?
Would creating mfa in setting page go through this node, in that case we may not have identity? Or setting is different?

louischan-oursky: You're correct! Identity may be nil if authenticator is being added in settings page. However, it means that we have to hide MFA section in the settings page if the user has only one OAuth identity. This is related to the REDACTED project's requirement though.

PR #1137, opened 2021-04-15
DB implementation refactoring

### LINKED ISSUE #1136: DB implementation refactoring
Preparation work for #1127 

- Update SQLBuilder to remove namespace, refactor to use same SQLBuilder for portal and auth
- To have TenantSQLExecutor and GlobalSQLExecutor, which is alias of original SQLExecutor. The handle of 2 SQLExecutor are different to fulfil tenant and global case. e.g. TenantSQLExecutor should connect to tenant db from app config, and GlobalSQLExecutor should connect to db from environment variable. All those logic should be in `./pkg/lib/...`
- Update use same GlobalSQLExecutor for portal and TenantSQLExecutor for auth, portal should not have own code to connect db
- Move portal `DatabaseConfig` to `EnvironmentConfig`, `EnvironmentConfig` is shared between auth and portal. So auth can have its `DatabaseConfig` in #1127

carmenlau:
fix #1136


louischan-oursky (commented):
1. `GlobalSQLExecutor` should be put in `pkg/lib/infra/db/global`. The lint may complain that the struct should not start with `Global` because the package is `global`.
2. `TenantSQLExecutor` should be put in `pkg/lib/infra/db/tenant`.
3. The import rename should be `globaldb "pkg/lib/infra/db/global"` and `tenantdb "pkg/lib/infra/db/tenant"` because it is more explicit.

carmenlau:
@louischan-oursky Updated!

Issue #1127, opened 2021-04-08
Support db as config source


louischan-oursky:
- Whether it is possible to read/write to the filesystem, with changes reflected in the configmap.
- Whether another pod with the same configmap mount can see changes immediately.

louischan-oursky:
Introduce database configsource

- How to handle `/`, `/_api/admin`, `/_resolver` routing.
- How to listen for changes.
- One Ingress and one Certificate for default domain (wildcard)
- Create one Ingress and one Certificate for each custom domain

louischan-oursky:
Some PoC on listen

```golang
package main

import (
	"fmt"
	"time"

	"github.com/lib/pq"
)

// CREATE TABLE portal_app (
// 	app_id TEXT NOT NULL,
// 	config JSONB NOT NULL
// );
//
// CREATE FUNCTION notify_me() RETURNS TRIGGER AS $$
// DECLARE
//   record RECORD;
// BEGIN
//   IF (TG_OP = 'DELETE') THEN
//     record := OLD;
//   ELSE
//     record := NEW;
//   END IF;
//   PERFORM pg_notify('notify_me', record.app_id);
//   RETURN NULL;
// END;
// $$ LANGUAGE plpgsql;
//
// CREATE TRIGGER notify_me
// AFTER INSERT OR UPDATE OR DELETE
// ON portal_app
// FOR ROW
// EXECUTE FUNCTION notify_me();

func main() {
	listener := pq.NewListener(
		"postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable",
		10*time.Second,
		1*time.Minute,
		nil,
	)
	err := listener.Listen("notify_me")
	if err != nil {
		panic(err)
	}
	for {
		select {
		case e := <-listener.Notify:
			fmt.Printf("app_id: %v\n", e.Extra)
		case <-time.After(time.Minute):
			listener.Ping()
		}
	}
}

```

carmenlau:
### Todo for supporting new config source
- New configuresource implementation for db
- New command: go run cmd/authger.go bootstrap --authgear-yaml=./file --authgear-secrets-yaml=./file
- New command: go run cmd/authger.go breaking_change migrate_k8s_to_db
- New command: go run cmd/authger.go config migrate <migration_name> --database-url=
- Rename: go run cmd/authger.go database migrate up --database-url=
- Rename: go run cmd/authger.go config init

### Deployment

### Cleanup

PR #1140, opened 2021-04-16
Support database config source

carmenlau:
ref #1127 

Initial draft for review only, not fully working

Done
- Basic ResolveAppID from db
- ResolveContext from db
- UpdateResources from db
- CreateResources from db
- Support cache
- Update portal nginx config to have correct host header when calling resolve endpoint

Todo
- TBD: Add `APP_HOST_SUFFIX` config and use it in ResolveAppID
- Command to setup first `accounts` app

### DIFF
--- a/cmd/authgear/server/server.go
+++ b/cmd/authgear/server/server.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"context"
 	golog "log"
 
 	"github.com/authgear/authgear-server/pkg/admin"
@@ -54,7 +55,7 @@ func (c *Controller) Start() {
 		c.logger.Warn("development mode is ON - do not use in production")
 	}
 
-	configSrcController := newConfigSourceController(p)
+	configSrcController := newConfigSourceController(p, context.Background())
 	err = configSrcController.Open()
 	if err != nil {
 		c.logger.WithError(err).Fatal("cannot open configuration")
--- a/nginx.conf
+++ b/nginx.conf
@@ -67,6 +67,7 @@ http {
             proxy_pass http://host.docker.internal:3001/resolve;
             proxy_pass_request_body off;
             proxy_set_header Host $http_host;
+            proxy_set_header X-Forwarded-Host "accounts.localhost";
             proxy_set_header Content-Length "";
         }
     }
@@ -109,6 +110,7 @@ http {
             proxy_pass http://host.docker.internal:3001/resolve;
             proxy_pass_request_body off;
             proxy_set_header Host $http_host;
+            proxy_set_header X-Forwarded-Host "accounts.localhost";
             proxy_set_header Content-Length "";
         }
     }
--- a/pkg/lib/config/configsource/database.go
+++ b/pkg/lib/config/configsource/database.go
@@ -1,0 +1,86 @@
+package configsource
+
+import (
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/spf13/afero"
+
+	"github.com/authgear/authgear-server/pkg/lib/config"
+	globaldb "github.com/authgear/authgear-server/pkg/lib/infra/db/global"
+	"github.com/authgear/authgear-server/pkg/util/httputil"
+	"github.com/authgear/authgear-server/pkg/util/log"
+	"github.com/authgear/authgear-server/pkg/util/resource"
+)
+
+type DatabaseSource struct {
+	ID        string
+	Data      map[string][]byte
+	CreatedAt time.Time
+	UpdatedAt time.Time
+}
+
+type DatabaseLogger struct{ *log.Logger }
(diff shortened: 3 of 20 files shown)


pkg/lib/config/configsource/database.go
```diff
@@ -52,7 +52,20 @@ func (d *Database) ResolveAppID(r *http.Request) (string, error) {
 		host = h
 	}
 
-	return "", nil
+	var appID string
+	err := d.Database.WithTx(func() error {
+		aid, err := d.Store.GetAppIDByDomain(host)
```
louischan-oursky, line 64: We should be able to resolve default domain even if they are not in the `domain` table.

pkg/lib/config/configsource/store.go
```diff
@@ -35,3 +74,85 @@ func (s *Store) GetAppIDByDomain(domain string) (string, error) {
...
+			return nil, ErrAppNotFound
+		}
+		return nil, err
+	}
+
+	return dbs, nil
+}
+
+func (s *Store) CreateDatabaseSource(dbs *DatabaseSource) error {
+	data, err := json.Marshal(dbs.Data)
```
louischan-oursky, line 99: How does `data` look like in JSON? Why am I asking this is because data is of type `map[string][]byte`. JSON does not have native byte array type.

pkg/lib/config/configsource/database.go
carmenlau: @louischan-oursky I found that we stored both default and custom domain in the `domain` table, so would suggest to resolve the app id in the same way. What do you think?

pkg/lib/config/configsource/store.go
carmenlau: The data is the same as format as the secret data in k8s. JSON key is the filename and the value is base64 encoded file content.

cmd/authgear/server/server.go
```diff
@@ -54,7 +55,7 @@ func (c *Controller) Start() {
 		c.logger.Warn("development mode is ON - do not use in production")
 	}
 
-	configSrcController := newConfigSourceController(p)
+	configSrcController := newConfigSourceController(p, context.Background())
```
carmenlau, line 58: @louischan-oursky Would like to ask, I am not sure if it is appreciate to use the background context here? The context is for the `SQLExecutor` of config source store. Any idea?

pkg/lib/config/configsource/database.go
louischan-oursky: If we store the default domain in the table, then we cannot simply change a env var to change every default domain of all apps.

pkg/lib/config/configsource/database.go
```diff
@@ -34,22 +40,54 @@ func NewDatabaseLogger(lf *log.Factory) DatabaseLogger {
...
+		switch channel {
+		case PGChannelConfigSourceChange:
+			d.invalidateApp(extra)
+		case PGChannelDomainChange:
+			d.invalidateHost(extra)
+		default:
+			panic(fmt.Sprintf("db_config: unknown notification channel: %s", channel))
+		}
+	}, func(e error) {
+		panic(fmt.Sprintf("db_config: error on listening pgsql: %s", e))
```
louischan-oursky, line 82: If we panic here, this goroutine will die. However, this config source will no longer listen anything :(

nginx.conf
```diff
@@ -67,6 +67,7 @@ http {
             proxy_pass http://host.docker.internal:3001/resolve;
             proxy_pass_request_body off;
             proxy_set_header Host $http_host;
+            proxy_set_header X-Forwarded-Host "accounts.localhost";
```
louischan-oursky, line 70: Any particular reason?

cmd/authgear/server/server.go
louischan-oursky: The main usage of the context is to provide cancellation. In case it is a HTTP request, the context is the request context, which is cancelled if the connection closes. If we use background context here, then the SQLExecutor may hang indefinitely because the context never cancel itself.

louischan-oursky: But I guess it is fine to use background context here.

pkg/lib/config/configsource/database.go
carmenlau: Since the domain is written in `authgear.yaml` `public_origin` field, even we support resolve default domain from env var, we cannot change it. After discussion, we will resolve both default and custom domain from db for consistent.

pkg/lib/config/configsource/database.go
carmenlau: Tested the listener has mechanism to reconnect, so we don't need to call `panic` here to break the program. Changed to log the error instead.

PR #1154, opened 2021-04-25
Add migrate-k8s-to-db command

### LINKED ISSUE #1153: Add migrate config source from k8s to database command

louischan-oursky:
fix #1153

### DIFF
--- a/cmd/portal/internal.go
+++ b/cmd/portal/internal.go
@@ -10,6 +10,8 @@ import (
 
 var defaultAuthgearDomain string
 var customAuthgearDomain string
+var kubeConfigPath string
+var namespace string
 
 var cmdInternal = &cobra.Command{
 	Use:   "internal [setup-portal]",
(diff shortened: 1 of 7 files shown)


cmd/portal/internal.go
```diff
@@ -51,4 +80,13 @@ func init() {
...
 	_ = cmdInternalSetupPortal.MarkFlagRequired("default-authgear-domain")
 	_ = cmdInternalSetupPortal.MarkFlagRequired("custom-authgear-domain")
+
+	cmdInternalBreakingChangeMigrateK8SToDB.Flags().StringVar(&DatabaseURL, "database-url", "", "Database URL")
+	cmdInternalBreakingChangeMigrateK8SToDB.Flags().StringVar(&DatabaseSchema, "database-schema", "", "Database schema name")
+	cmdInternalBreakingChangeMigrateK8SToDB.Flags().StringVar(&kubeConfigPath, "kubeconfig", "", "Path to kubeconfig")
+	cmdInternalBreakingChangeMigrateK8SToDB.Flags().StringVar(&namespace, "namespace", "", "Namespace")
+
+	// FIXME: Respect KUBECONFIG environment variable.
+	_ = cmdInternalBreakingChangeMigrateK8SToDB.MarkFlagRequired("kubeconfig")
```
carmenlau, line 90: I think `kubeconfig` should not be required? As `MakeKubeConfig` accept empty `kubeConfigPath` and home kube config (e.g. `~/.kube/config`) will be returned.

carmenlau (commented):
Looks good, and I tried run in my local db + staging k8s, it works great~!

cmd/portal/internal.go
louischan-oursky: But it still do not respect the `KUBECONFIG` environment variable. I googled and found out that people typically use cobra with viper to accept both environment variable and command line argument. I think we should update our code to that. So I added a FIXME here.

PR #1157, opened 2021-04-26
Integrate viper

louischan-oursky:


cmd/portal/migrate_resources.go
```diff
@@ -12,7 +13,7 @@ var cmdInternalBreakingChangeMigrateResources = &cobra.Command{
 }
 
 func init() {
-	cmdInternalBreakingChangeMigrateResources.PersistentFlags().StringVar(&DatabaseURL, "database-url", "", "Database URL")
-	cmdInternalBreakingChangeMigrateResources.PersistentFlags().StringVar(&DatabaseSchema, "database-schema", "", "Database schema name")
+	ArgDatabaseURL.Bind(cmdInternalBreakingChangeMigrateResources.Flags(), viper.GetViper())
+	ArgDatabaseSchema.Bind(cmdInternalBreakingChangeMigrateResources.Flags(), viper.GetViper())
```
carmenlau, line 17: We should use `PersistentFlags` instead of `Flags` in  `cmdInternalBreakingChangeMigrateResources`,
since those flags are global flags for all migrate resources sub-commands.

carmenlau (commented):
The other parts look great!

Issue #1156, opened 2021-04-26
Human readable error messages


chpapa:
These common error messages should be updated:

* Invalid email address => Please sign up with a valid email address
* This identity has been claimed by another user. => [Email/Phone/Username] already exists, maybe you want to login?
* Invalid credentials => Incorrect password

chpapa:
Just realize some common error messages is not very user friendly, maybe @louischan-oursky / @carmenlau can update the default quickly?

PR #1160, opened 2021-04-28
Improve error message

louischan-oursky:
connect #1156


pkg/lib/interaction/graph.go
```diff
@@ -243,19 +245,25 @@ func (g *Graph) GetACR(amrValues []string) string {
 	return ""
 }
 
+func (g *Graph) FillDetails(err error) error {
+	return errorutil.WithDetails(err, errorutil.Details{
+		"IntentKind": apierrors.APIErrorDetail.Value(IntentKind(g.Intent)),
```
carmenlau, line 250: I found some of the validation is done in handler, so `IntentKind` will be missing for the error in that case. e.g. For signup with phone number, we parse the phone number by function `FormToLoginID` in handler.

pkg/lib/interaction/nodes/create_authenticator_oob_setup.go
```diff
@@ -55,6 +57,26 @@ func (e *EdgeCreateAuthenticatorOOBSetup) Instantiate(ctx *interaction.Context,
...
+		if err != nil {
+			validationCtx.EmitError("format", map[string]interface{}{"format": "phone"})
+		}
+	}
+
+	err := validationCtx.Error("invalid target")
+	if err != nil {
+		return nil, err
+	}
+
```
carmenlau, line 79: The original logic in handler not only validate the phone number, it also do parse. That means the phone number we store in DB is a tidied up format (e.g. no space and hyphen..etc). Maybe we should
1. check if phone login id parsed inside interaction?
2. ensure oob also parse the phone

louischan-oursky: I double checked and the current behavior is that the raw input from user must be in E.164 format, otherwise a validation error will be thrown.

PR #1164, opened 2021-04-28
Include OIDC state parameter in user.pre_create hook

### LINKED ISSUE #1141: Allow OIDC state parameter visible and utilised in the user.pre_create webhook
To support projects that require passing in parameter before user created.

carmenlau:
fix #1141 

- Added new interface `IntentWithWebhookState` to allow passing state to the webhook, see if this approach looks good?

### DIFF
--- a/docs/specs/webhook.md
+++ b/docs/specs/webhook.md
@@ -185,15 +185,20 @@ Operation can be aborted by providing specific response in your webhook, details
 
 ```json5
 {
+  "context": {
+    ....
+    "oauth": {
+      "state": "the-value-of-state-if-provided"
+    }
+  },
   "payload": {
     "user": { /* ... */ },
-    "identities": [ { /* ... */ } ],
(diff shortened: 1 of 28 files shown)


docs/specs/webhook.md
```diff
@@ -187,11 +187,14 @@ Operation can be aborted by providing specific response in your webhook, details
...
   "payload": {
     "user": { /* ... */ },
-    "identities": [ { /* ... */ } ]
+    "identities": [ { /* ... */ } ],
+    "state": ""
   }
 }
 ```
 
+- `state`: OIDC state if the signup is triggered through authorize endpoint.
```
louischan-oursky, line 196: I think this should be put in context. 

```yaml
{
  "context": {
    "oauth": {
      "state": "the-value-of-state-if-provided"
    }
  }
}
```

carmenlau:
Updated!

Issue #1132, opened 2021-04-09
[Portal] User search and sort


mr9dollars:
**User search and sort**
- [x] User Table
- [x] Allow to search user through email, username, user id or phone number
- [x] Allow to sort user through last login, signed up

louischan-oursky:
# Choice of search engine

Elasticsearch

# Deployment of Elasticsearch

- Local: Official docker image `docker.elastic.co/elasticsearch/elasticsearch:7.12.0` with [single node cluster](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#docker-cli-run-dev-mode)
- Staging: [Official helm chart](https://github.com/elastic/helm-charts/tree/master/elasticsearch)
- Production: [Official helm chart](https://github.com/elastic/helm-charts/tree/master/elasticsearch)

# Mapping

```json
{
  "properties": {
    "app_id": { "type": "keyword" },
    "id": { "type": "keyword" },
    "created_at": { "type": "date" },
    "updated_at": { "type": "date" },
    "last_login_at": { "type": "date" },
    "is_anonymous": { "type": "boolean" },
    "is_verified": { "type": "boolean" },
    "is_disabled": { "type": "boolean" },
    "email": { "type": "keyword" },
    "preferred_username": { "type": "keyword" },
    "phone_number": { "type": "keyword" }
  }
}
```

# Operations

## `authgear search reindex <app-id>`

Re-index all data for the given app.

## `authgear search reindex --all`

Re-index all data for all apps.
Mainly for initial deployment, or we have breaking changes on the mapping.

## Re-index a user

- When a user is created.
- When a user is disabled.
- When a user is enabled.
- When a user is promoted.
- When an identity of a user is added.
- When an identity of a user is updated.
- When an identity of a user is deleted.
- When an identity of a user is verified.

# Search

```json5
{
  "query": {
    "bool": {
      "filter": [
        { "term": { "app_id": "APP_ID" } }, // 1
      ],
      "should": [
        { "term": { "email": "KEYWORD" } }, // 2
        { "term": { "preferred_username": "KEYWORD" } }, // 3
        { "term": { "phone_number": "KEYWORD" } }, // 4
        {
          "regexp": { // 5
            "email": {
              "value": "@KEYWORD$",
              "case_insensitive": true
            }
          }
        },
        {
          "regexp": { // 6
            "phone_number": {
              "value": "KEYWORD$"
            }
          }
        }
      ]
    }
  },
[...]

PR #1173, opened 2021-05-06
Add search box to users screen

louischan-oursky:
ref #1132

### DIFF
--- a/portal/src/graphql/adminapi/UsersScreen.tsx
+++ b/portal/src/graphql/adminapi/UsersScreen.tsx
@@ -1,11 +1,4 @@
-import React, {
-  useCallback,
-  useContext,
-  useEffect,
-  useMemo,
-  useRef,
-  useState,
-} from "react";
+import React, { useCallback, useContext, useMemo, useState } from "react";
 import { useNavigate } from "react-router-dom";
 import { ICommandBarItemProps, SearchBox } from "@fluentui/react";
 import { Context, FormattedMessage } from "@oursky/react-messageformat";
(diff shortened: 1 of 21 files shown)


portal/src/graphql/adminapi/UsersScreen.tsx
```diff
@@ -83,33 +167,125 @@ const UsersScreen: React.FC = function UsersScreen() {
...
+        cursor,
+      },
+    });
+  }, [execute, sortBy, sortDirection, searchKeyword, cursor]);
+
+  // Initial execute
+  useEffect(() => {
+    search();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
carmenlau, line 225: I found that the user list page would flicker when I land on the user list page, is it the problem that you mentioned? I checked a bit and here are some of the observations. 

1. We request the graphql server twice when we land on the user page.
2. `useEffect` will be called after the first render. We have 3 `useEffect` to control calling `search`, so `search` was called 3 times when we landed on the page.
3. From point 2, graphql server should be called 3 times but not twice? I found that since the first two `search` call the timing is too close, we had called  `execute` twice but it only fire one graphql request (apollo handled it?). And the third `search` was called after 500ms, so it fire another graphql request. So the page flickers.

Maybe we should update to add some logic to avoid the second and third `useEffect` call `search` in the first render?

ref: https://reactjs.org/docs/hooks-effect.html
> Does useEffect run after every render? Yes! By default, it runs both after the first render and after every update.

louischan-oursky:
I removed all useEffect in UsersScreen and it looks much better now.

Issue #1128, opened 2021-04-09
How to show IdP session and offline grant


louischan-oursky:
> Special case on how to show sessions. When an IdP session is associated with exact one offline grant, we show the offline grant only. Of course, this implies we have to revoke them at the same time. When an IdP session is associated with more than 1 offline grant, we show them separately.

chpapa:
We need to consider the following scenario, and show the session list (or device list) with the following results:

1. Desktop Web
= 1 new Session
= 1 device

2. Mobile App login
= 1 new Session 
= 1 device

3. Mobile Safari Login > Mobile App Login
= 2 new Sessions (or 
= 1 device

4. (Case 1 - Desktop Web) + SPA login with offline grant
= 1 new Session
= 1 device

louischan-oursky:
# Sessions in online services

## Google

On myaccount.google.com, sessions are grouped by devices.

Google has its own way to somehow identify a device.

### Grouping browser sessions into mobile session is likely based on heuristic

At the beginning of the experiment, MacBookPro15,2 (my macbook) and rfc1123 (my iPhone 12 mini) are my identified devices.

I signed in with a private tab in the mobile Safari on rfc1123.
No new device was identified.

I signed out rfc1123.
The remaining device is MacBookPro15,2.

I borrowed my friend's iPhone and then signed in with a private tab in the mobile Safari.
The rfc1123 device was resurrected.
Their iPhone is neither named rfc1123 nor the model is iPhone 12 mini.

So device identification is not precise and is likely based on heuristic.

### HTTP User-Agent header and navigator affects how a device is identified.

I installed User-Agent Switcher on Firefox on my macbook.
I switched the User-Agent to Windows/Edge and I signed in.
A new device Windows was identified.

## Facebook

On facebook.com, sessions are not grouped.

Here is my session list.

> Mac, Hong Kong
> Chrome, Active Now
> 
> Mac, Hong Kong
> Firefox, just now
> 
> iPhone 12 mini, Hong Kong
> Facebook app, about an hour ago
> 
> iPhone 12 mini, Hong Kong
> Messenger for iOS, 2021-03-27

As you can see, sessions are not grouped.

# Insights from online services

Device identification is hard and the goal of Authgear is not about device tracking.
Therefore I suggest we do not group sessions by devices.

If our session list looks like that of Facebook, most user should understand it.

The real challenge for ourselves is to group IdP sessions and offline grants so that they appear logically to the user.

# Session properties that can be used for identification

[...]

PR #1205, opened 2021-05-21
Group session by group

louischan-oursky:
ref #1128


louischan-oursky:
1. Handle offline grant on web
2. Show "Active Now" in green instead of last access at for the current session.

carmenlau (commented):
Two comments:
- Need to handle web app offline grant which doesn't have device info
- Show active label to the current session

louischan-oursky:
Updated.

carmenlau (approved):

PR #1220, opened 2021-05-28
Improve support for prompt

### LINKED ISSUE #1207: Improve support for prompt
- It should be a space delimited value, so it should be parsed as a list of string.
- What to do if an unsupported value is given? Ignore
- `none` cannot co-exist with any other values. Return error in this error.
- Forward supported prompt values to underlying OIDC providers.
- https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/operations/ad-fs-prompt-login
- https://developers.google.com/identity/protocols/oauth2/openid-connect#authenticationuriparameters
- https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow

TODO:
- [X] Change prompt to list and support passing prompt to OAuth providers
- [ ] Cleanup logic translating old prompt parameter to prompt list in web session

carmenlau:
ref #1207

### DIFF
--- a/pkg/auth/webapp/session_store.go
+++ b/pkg/auth/webapp/session_store.go
@@ -77,6 +77,14 @@ func (s *SessionStoreRedis) Get(id string) (session *Session, err error) {
 			return err
 		}
 		err = json.Unmarshal(data, &session)
+		// FIXME(webapp): remove LegacyPrompt in the next deployment
+		// translate old LegacyPrompt to Prompt list in web session
+		// since websession only have 20 mins lifetime
+		// the translation logic can be removed in the next deployment
+		if session.LegacyPrompt != "" && len(session.Prompt) == 0 {
+			session.Prompt = []string{session.LegacyPrompt}
+		}
+		// translation logic end
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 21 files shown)


pkg/lib/authn/sso/azureadv2.go
```diff
@@ -149,6 +150,24 @@ func (f *Azureadv2Impl) OpenIDConnectGetAuthInfo(r OAuthAuthorizationResponse, p
 	return
 }
 
+func (f *Azureadv2Impl) GetPrompt(prompt []string) []string {
+	// Azureadv2 only support single value for prompt
+	// the first supporting value in the list will be used
```
carmenlau, line 155: Azureadv2 only supports single value for prompt, and I will pick the first supporting value from the list. What do you think?

louischan-oursky: This is the best we can do. Maybe we should document the order MAY matter in the SDK documentation.

pkg/auth/webapp/session_store.go
```diff
@@ -77,6 +77,13 @@ func (s *SessionStoreRedis) Get(id string) (session *Session, err error) {
 			return err
 		}
 		err = json.Unmarshal(data, &session)
+		// translate old LegacyPrompt to Prompt list in web session
```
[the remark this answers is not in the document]
carmenlau: Updated!

Issue #1214, opened 2021-05-26
Audit log implementation


louischan-oursky:
- Refactor webhook into event; Allow multiple event sinks; Make webhook a sink.
- Refactor database to support more than one tenant database.
- Introduce audit database connection in authgear.secrets.yaml
- Introduce new migration for audit database
- Introduce new command for running audit database pg_partman maintenance procedure
- Add audit specific event types
- Add audit log event sink
- Add auditLogs to Admin API GraphQL API
- Show audit log table in the portal

PR #1226, opened 2021-06-01
Add audit event sink

louischan-oursky:
ref #1214


pkg/lib/audit/sink.go
```diff
@@ -0,0 +1,61 @@
...
+		}).Debug("skip persisting event")
+		return
+	}
+
+	s.Logger.WithFields(map[string]interface{}{
+		"event": e,
+	}).Debug("persisting event")
+
+	err = s.Database.WithTx(func() error {
+		return s.Store.PersistEvent(e)
```
carmenlau, line 32: I think we need to update the sinks' implementation to skip some events? e.g. hook sink should not dispatch `user.failed_authentication` and `user.signed_out`.  And audit sink should not store `user.pre_create`?

pkg/lib/event/service.go
```diff
@@ -131,12 +136,19 @@ func (s *Service) makeContext(payload event.Payload) event.Context {
 		triggeredBy = event.TriggeredByTypeAdminAPI
 	}
 
+	ipAddress := httputil.GetIP(s.Request, bool(s.TrustProxy))
+	userAgent := s.Request.UserAgent()
+	clientID := clientid.GetClientID(s.Context)
```
carmenlau, line 141: If the event is triggered by the portal, the IP address will be the cluster IP of the portal pod? Does it matter if we expose it in the event or we should hide it?

louischan-oursky:
I just realized I need to filter out the new event type for the hook event sink. Let me add a commit for that.

pkg/lib/interaction/nodes/authn_end.go
```diff
@@ -30,12 +31,28 @@ type EdgeAuthenticationEnd struct {
...
+		if err != nil {
+			return nil, err
+		}
+		err = ctx.Events.DispatchEvent(&nonblocking.UserFailedAuthenticationEventPayload{
+			User: *user,
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
```
carmenlau, line 53: Thinking should we also dispatch `UserFailedAuthenticationEventPayload` if the user failed to select identity? e.g. user input invalid email. It should be useful for audit purpose, e.g. one IP keep trying to burst force the email address.
I tried auth0, they had an audit event for that case, the event context will have the `username` that the user is trying and an empty user id.

What do you think?

pkg/lib/event/service.go
louischan-oursky: Let me omit the IP address if the event is triggered by admin API.

pkg/lib/interaction/nodes/authn_end.go
louischan-oursky: But in that case, we cannot associate the event with any user. The event can only be viewed in the audit log screen. Should we add a new event type for that? So the admin can select this particular event type to view all malicious brute force attack.

carmenlau: Sure, any ideas for the names? We need to have a name that is easy to identify 🤔 For auth0, they are `Failed Login (invalid email/username)` and `Failed Login (wrong password)`

pkg/lib/event/service.go
louischan-oursky: But on the other hand, we want to log the IP address calling the admin API if the caller is not ourselves. I guess we should not hide it at the moment. What do you think?

PR #1230, opened 2021-06-03
Add audit logs to admin graphql endpoint

louischan-oursky:
ref #1214

### DIFF
--- a/pkg/lib/audit/log.go
+++ b/pkg/lib/audit/log.go
@@ -1,7 +1,10 @@
 package audit
 
 import (
+	"encoding/json"
 	"time"
+
+	"github.com/authgear/authgear-server/pkg/api/event"
 )
 
 type Log struct {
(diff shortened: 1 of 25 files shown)


pkg/lib/audit/log.go
```diff
@@ -0,0 +1,48 @@
...
+	"encoding/json"
+	"time"
+
+	"github.com/authgear/authgear-server/pkg/api/event"
+)
+
+type Log struct {
+	ID           string                 `json:"id"`
+	CreatedAt    time.Time              `json:"createdAt"`
+	ActivityType string                 `json:"activityType"`
```
carmenlau, line 13: I think we use underscore rather than camelcase for JSON key? Or remove the annotation first if we don't need it in this stage?

louischan-oursky: Add a commit to explain the reason for using camel case in JSON key.

carmenlau (approved):

carmenlau:
Will be merged in #1234

Issue #1200, opened 2021-05-17
Auth Login UI adjustment


mr9dollars:
- [x] Delete session confirmation dialog
- [x] Fix Font Sizes
- [x] Update icons of whole site
- [x] Add label login to continue to {client_name}
- [x] Setting page back button style
- [x] Setting page back to home page link style
- [x] Portal update UI settings preview and theme preset colors
- [x] Templates to handle: 
    - [x] change_password.html
    - [x] change_secondary_password.html
    - [x] create_password.html
    - [x] enter_login_id.html
    - [x] enter_oob_otp.html
    - [x] enter_password.html
    - [x] enter_recovery_code.html
    - [x] enter_totp.html
    - [x] fatal_error.html
    - [x] forgot_password.html
    - [x] forgot_password_success.html
    - [x] login.html
    - [x] logout.html
    - [x] promote.html
    - [x] reset_password.html
    - [x] reset_password_success.html
    - [x] return.html
    - [x] settings.html
    - [x] settings_biometric.html
    - [x] settings_identity.html
    - [x] settings_mfa.html
    - [x] settings_oob_otp.html
    - [x] settings_recovery_code.html
    - [x] settings_sessions.html
    - [x] settings_totp.html
    - [x] setup_oob_otp.html
    - [x] setup_recovery_code.html
    - [x] setup_totp.html
    - [x] signup.html
    - [x] user_disabled.html
    - [x] verify_identity.html
    - [x] verify_identity_success.html
    - [x] wechat_auth.html

PR #1233, opened 2021-06-07
Add confirmation dialog for revoking sessions

carmenlau:
refs #1200


louischan-oursky:
Looks very nice. I just found a very minor room for improvement. When the confirm button in the modal is pressed, the modal is closed immediately without transition :(

PR #1234, opened 2021-06-07
Show audit log in portal

louischan-oursky:
ref #1214


carmenlau (commented):
The PR overall looks great and I love the way `useTranactionalState` handles the date selection state!

Only 1 comment that we may need to handle when the audit log is not enabled?

louischan-oursky:
Added 2 commits.

carmenlau (commented):
@louischan-oursky I think we need to handle when developer doesn't provide `audit.db` in `authgear.secrets.yaml`. Currently, admin api `GraphQLHandler` called `h.AuditDatabase.ReadOnly` and seems we haven't handled the db is not configured case.

louischan-oursky:
Fixed.

PR #1238, opened 2021-06-09
Update auth ui icons

carmenlau:
ref #1200 

Update using tabler-icons in the auth ui

FIXME:
- oauth icons in account settings and resend verification icon is not confirmed yet


pkg/lib/web/static_assets.go
```diff
@@ -31,6 +32,38 @@ var AuthgearCSS = resource.RegisterResource(CSSDescriptor{
...
 
+var NormalizeCSS = resource.RegisterResource(CSSDescriptor{
+	Path: StaticAssetResourcePrefix + "normalize.min.css",
+})
+
+var NormalizeCSSMap = resource.RegisterResource(resource.SimpleDescriptor{
+	Path: StaticAssetResourcePrefix + "normalize.min.css.map",
+})
+
+var IconsCSS = resource.RegisterResource(CSSDescriptor{
```
carmenlau, line 43: I found I cannot simply change `CSSDescriptor` to `resource.SimpleDescriptor`, as their `ViewResources` functions are returning different object. `CSSDescriptor` returns struct `web.StaticAsset` which has the path and data. But `resource.SimpleDescriptor` returns data directly. 

When I called `$.StaticAssetURL` in the template, the function expect the descriptor return `web.StaticAsset` but `resource.SimpleDescriptor` doesn't.

Should we create another `SimpleDescriptor` under `/pkg/lib/web/` that will return `web.StaticAsset`, but then we will need to copy the logic from `pkg/util/resource`. Or just keep using the `CSSDescriptor` to avoid too many descriptors implementation?

louischan-oursky: Just keep using CSSDescriptor is fine.

carmenlau: Then the PR is good to go:)

pkg/lib/web/static_assets.go
```diff
@@ -31,6 +32,30 @@ var AuthgearCSS = resource.RegisterResource(CSSDescriptor{
 	Path: StaticAssetResourcePrefix + "authgear.css",
 })
 
+var IconsCSS = resource.RegisterResource(CSSDescriptor{
```
louischan-oursky, line 43: Can we add a comment here to document the version of tabler-icon we have embedded? It is easier than to `git blame` and then see the version in the commit message.

pkg/lib/web/static_assets.go
```diff
@@ -32,6 +32,14 @@ var AuthgearCSS = resource.RegisterResource(CSSDescriptor{
 	Path: StaticAssetResourcePrefix + "authgear.css",
 })
 
+var NormalizeCSS = resource.RegisterResource(CSSDescriptor{
```
louischan-oursky, line 35: Document the version of the normalize.css we have embedded.

resources/authgear/static/authgear-light-theme.css
```diff
@@ -158,7 +158,7 @@
 
   --color-recovery-code-list-background: #eaeaea;
 
-  --color-settings-security-item-active-background: rgba(0, 0, 0, 0.03);
+  --color-action-btn-active-background: rgba(0, 0, 0, 0.03);
```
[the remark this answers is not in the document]
carmenlau: We don't have the exact same color, I find a similar one and updated!

PR #1246, opened 2021-06-11
Specify how to do reauthentication via OIDC authentication request

### LINKED ISSUE #1243: Research account selection and reauthentication
- Research how `prompt`, `max_age `, `id_token_hint`, `acr_values` work together to give different use cases.
- Research how other identity provider do reauthentication.

louischan-oursky:
fix #1243

### DIFF
--- a/docs/specs/oidc.md
+++ b/docs/specs/oidc.md
@@ -157,7 +157,7 @@ Refresh token is not used.
 ### max_age
 
 - `0`: It is equivalent to prompt=login.
-- Any positive integer: It is equivalent to prompt=login if the elapsed time of the last authentication is greater than n. Otherwise, it has no effect.
+- `n`, where n is Any positive integer: It is equivalent to prompt=login if the elapsed time in seconds of the last authentication is greater than `n`. Otherwise, it has no effect.
 
 ### id_token_hint
 
@@ -323,7 +323,7 @@ If no authentication method is included in `amr` claim, `amr` claim would be omi
 
 ### `auth_time`
 
-The time when the authentication was performed.
+The time in Unix epoch when the authentication was performed.
 This claim is only present if the ID token is obtained via the authorization code flow or the implicit flow.
 
 This is the authenticated at of the IdP session.
--- a/docs/specs/oidc.md
+++ b/docs/specs/oidc.md
@@ -1,57 +1,65 @@
-# OIDC
-
-Authgear acts as OpenID Provider (OP).
-
+- [OIDC](#oidc)
   * [OAuth 2 and OIDC Conformance](#oauth-2-and-oidc-conformance)
   * [Client Metadata](#client-metadata)
-    * [Standard Client Metadata](#standard-client-metadata)
-    * [Custom Client Metadata](#custom-client-metadata)
-      * [Generic RP Client Metadata example](#generic-rp-client-metadata-example)
-      * [Native application Client Metadata example](#native-application-client-metadata-example)
-      * [Web application sharing the same root domain Client Metadata example](#web-application-sharing-the-same-root-domain-client-metadata-example)
-      * [Silent Authentication Client Metadata example](#silent-authentication-client-metadata-example)
+    + [Standard Client Metadata](#standard-client-metadata)
+    + [Custom Client Metadata](#custom-client-metadata)
+      - [Generic RP Client Metadata example](#generic-rp-client-metadata-example)
+      - [Native application Client Metadata example](#native-application-client-metadata-example)
+      - [Web application sharing the same root domain Client Metadata example](#web-application-sharing-the-same-root-domain-client-metadata-example)
+      - [Silent Authentication Client Metadata example](#silent-authentication-client-metadata-example)
   * [Authentication Request](#authentication-request)
-    * [scope](#scope)
-    * [response_type](#response_type)
-    * [display](#display)
-    * [prompt](#prompt)
-    * [max_age](#max_age)
-    * [id_token_hint](#id_token_hint)
-    * [login_hint](#login_hint)
-    * [acr_values](#acr_values)
-    * [code_challenge_method](#code_challenge_method)
+    + [scope](#scope)
+    + [response_type](#response_type)
+    + [prompt](#prompt)
+    + [max_age](#max_age)
+    + [id_token_hint](#id_token_hint)
+    + [login_hint](#login_hint)
+    + [acr_values](#acr_values)
+    + [code_challenge_method](#code_challenge_method)
   * [Token Request](#token-request)
-    * [grant_type](#grant_type)
-    * [jwt](#jwt)
+    + [grant_type](#grant_type)
+    + [id_token_hint](#id_token_hint-1)
+    + [jwt](#jwt)
   * [Token Response](#token-response)
-    * [token_type](#token_type)
-    * [refresh_token](#refresh_token)
-    * [scope](#scope-1)
+    + [token_type](#token_type)
+    + [refresh_token](#refresh_token)
+    + [scope](#scope-1)
   * [The metadata endpoint](#the-metadata-endpoint)
-    * [authorization_endpoint](#authorization_endpoint)
-    * [token_endpoint](#token_endpoint)
-    * [userinfo_endpoint](#userinfo_endpoint)
-    * [revocation_endpoint](#revocation_endpoint)
-    * [jwks_uri](#jwks_uri)
-    * [scopes_supported](#scopes_supported)
-    * [response_types_supported](#response_types_supported)
-    * [grant_types_supported](#grant_types_supported)
-    * [subject_types_supported](#subject_types_supported)
-    * [id_token_signing_alg_values_supported](#id_token_signing_alg_values_supported)
-    * [claims_supported](#claims_supported)
-    * [code_challenge_methods_supported](#code_challenge_methods_supported)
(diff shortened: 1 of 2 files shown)


docs/specs/oidc.md
```diff
@@ -184,7 +190,13 @@ Unknown parameters are ignored, and invalid parameters are rejected. However, if
...
 ### acr_values
 
-Unsupported yet.
+`acr_values` determines how the authentication is performed when both a valid session and `id_token_hint` are present.
+
+The value is an URL `https://authgear.com/acr` with the following query parameters.
+
+- `primary_authenticator=password`: Indicate that the primary authenticator was password. Other values are possible.
+- `secondary_authenticator=totp`: Indicate that the secondary authenticator was TOTP. Other values are possible.
+- `reauthenticate=primary_authenticator`: Require that one of primary authenticators of the users to be used to reauthenticate the user.
```
carmenlau, line 199: Can we list out the values that we support for re-authentication? Or we will only support `primary_authenticator` in this stage?

docs/specs/oidc.md
```diff
@@ -1,57 +1,65 @@
...
-    * [login_hint](#login_hint)
-    * [acr_values](#acr_values)
-    * [code_challenge_method](#code_challenge_method)
+    + [scope](#scope)
+    + [response_type](#response-type)
+    + [prompt](#prompt)
+    + [max_age](#max-age)
+    + [id_token_hint](#id-token-hint)
+    + [login_hint](#login-hint)
+    + [acr_values](#acr-values)
```
carmenlau, line 17: Seems the anchors are wrong after updating?

docs/specs/oidc.md
```diff
@@ -489,4 +520,38 @@ When the app session token is consumed:
...
+If the user continues with another account, the user is authenticated fully.
+
+### The user has signed in. I want to reauthenticate the user before they can perform sensitive operation.
+
+The authentication request is `response_type=id_token&prompt=login&max_age=0&id_token_hint=ID_TOKEN&acr_values=https://authgear.com/acr?reauthenticate=primary`.
+
+The user will NOT see select account screen because `id_token_hint` is given.
+
+If the user has valid session, the user is reauthenticated with one of their primary authenticator.
+It is an error if the user does not have any primary authenticator.
```
carmenlau, line 549: Could we have some examples for this case? e.g. reauthenticate a user who only connected to OAuth?

docs/specs/oidc.md
```diff
@@ -136,27 +144,25 @@ Refresh token is not used.
...
 
-- `login`
-- `none`
+> In the future, we may support select_account and content.
 
 ### max_age
 
-Unsupported.
+- `0`: It is equivalent to prompt=login.
+- Any positive integer: Reauthenticate the end-user if the elapsed time of the last authentication is greater than n.
```
carmenlau, line 161: What will happen if the auth-time is still within the max_age?

docs/specs/oidc.md
louischan-oursky: I used https://ecotrust-canada.github.io/markdown-toc/ to generate the TOC, but it is buggy as you see. Now I use https://github.com/jonschlinkert/markdown-toc

chpapa:
@louischan-oursky as spoken, while I haven't review the spec in details, please try to see if we can enhance it with:

If the logged in user is using an authenticator that's not supported to do re-auth (such as 3rd parties SSO), the developer:
1. should be able to detect it and prevent going to the re-auth page, without showing an error message to the user, or
2. even if it have to go to the re-auth page, it won't show error message to the end user, but the developer can know that the re-auth was not supported so they could decide if they still want to continue with the "sensitive operations" at all.

docs/specs/oidc.md
```diff
@@ -136,27 +144,26 @@ Refresh token is not used.
...
 
-- `login`
-- `none`
+> In the future, we may support select_account and content.
 
 ### max_age
 
-Unsupported.
+- `0`: It is equivalent to prompt=login.
+- Any positive integer: It is equivalent to prompt=login if the elapsed time of the last authentication is greater than n. Otherwise, it has no effect.
```
chpapa, line 160: What unit is n? Hours minutes seconds?

docs/specs/oidc.md
```diff
@@ -300,24 +317,25 @@ To indicate the authenticator used in authentication, `amr` claim is used in OID
 - If password authenticator is used: `pwd` is included.
 - If any OTP (TOTP/OOB-OTP) is used: `otp` is included.
 - If WebAuthn is used: `hwk` is included.
+- If Biometric is used: `x_biometric` is included.
```
chpapa, line 320: I think we should also include as much detailed information in the amr, if it is a list allow multiple values? Eg if it is iOS FaceID, than ['face', 'x_biometric']. If not than just the x_biometric is fine.

docs/specs/oidc.md
```diff
@@ -300,24 +317,25 @@ To indicate the authenticator used in authentication, `amr` claim is used in OID
...
 
-If no authentication method is to be included in `amr` claim, `amr` claim would be omitted from the ID token.
+If no authentication method is included in `amr` claim, `amr` claim would be omitted from the ID token.
 
-### `acr`
+### `auth_time`
 
-If any secondary authenticator is performed, `acr` claim would be included in ID token with value `http://schemas.openid.net/pape/policies/2007/06/multi-factor`.
+The time when the authentication was performed.
```
chpapa, line 326: In what unit?

docs/specs/oidc.md
```diff
@@ -300,24 +317,25 @@ To indicate the authenticator used in authentication, `amr` claim is used in OID
...
+This claim is only present if the ID token is obtained via the authorization code flow or the implicit flow.
 
-To perform step-up authentication, developer can pass a `acr_values` of  `http://schemas.openid.net/pape/policies/2007/06/multi-factor` to the authorize endpoint.
+This is the authenticated at of the IdP session.
```
chpapa, line 329: Question: do our current design allow step-up authentication in parallel with re-auth?

docs/specs/sdk.md
```diff
@@ -23,6 +23,116 @@ It is more preferable to have ephemeral refresh token that are only stored in me
...
+    // You must send the ID token to your server.
+    // Your server must verify the signature and check the claims inside.
+    // Your server should allow user to skip reauthentication if it is impossible.
+    await performSensitiveOperation(authgear.getIDTokenHint());
+    return;
+  }
+
+  const options: AuthorizeOptions = {
+    redirectURI: "myapp://host/path",
+    prompt: "login",
```
chpapa, line 123: I don't think prompt and maxAge is needed for reauthenticate?

chpapa (commented):
I have finally read it once (on mobile sorry lol)

Leave some minor comments and questions.

Btw this PR seems mixed some formatting / minor wording changes? Would appreciate if those could be breakdown to another PR next time @louischan-oursky 

As we already covered in last product meeting, I think a design principle problem we have is, in client side SDK, we want to make it simple and usable even without prior knowledge of OIDC or reading our spec, so asking users to "put different combination of authorizeOptions" to achieve reauthentication is a bad design, and hence we should have the new function "reauthenticate". I still see the same authorizeOptions as a parameter for reauthenticate, actually that seems a problem?

docs/specs/oidc.md
louischan-oursky: [In seconds](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest). Let me be more specific here.

docs/specs/oidc.md
louischan-oursky: Agree. Opened this issue to follow up https://github.com/authgear/authgear-server/issues/1295

docs/specs/oidc.md
louischan-oursky: [Unix epoch](https://openid.net/specs/openid-connect-core-1_0.html#IDToken). Let me be more specific here.

docs/specs/oidc.md
louischan-oursky: We now have `auth_time` and `amr` in the ID token. From these two claims, we can know **when** was the last authentication done (from `auth_time`) and **how** was it done (from `amr`). In the future, we can add our own parameters to the authorization endpoint to specify what kind of authentication the developer wants. When we add such option, the default is re-auth with any authenticator (i.e. the current behavior)

docs/specs/sdk.md
louischan-oursky: maxAge may be needed sometimes. When the app has a series of sensitive operation to perform, it should be acceptable that the user has reauthenticated themselves recently enough, say 5 minutes. So in this case, the developer should specify a maxAge of `5 * 60`, to provide a better UX. The user does not need to go through reauthenticate themselves repeatedly.

docs/specs/oidc.md
chpapa: Got it

docs/specs/sdk.md
chpapa: @louischan-oursky thanks got it.

What about prompt? There are no way prompt is changable in reauthenticate?

louischan-oursky: Yes, you're correct. I have removed `prompt` from ReauthenticateOptions.

louischan-oursky:
Is this PR good to merge? @carmenlau

PR #1280, opened 2021-06-18
Reauthentication part 1

louischan-oursky:


carmenlau (commented):
Looks good to me so far:)

carmenlau (commented):
Need update the token endpoint to obtain the updated id token by using session (idp session or offline grant session).

louischan-oursky:
@carmenlau Updated.

PR #1292, opened 2021-06-24
Support app plan part 1

carmenlau:


carmenlau:
For the models in the portal pkg, we only have service but not store. I created a store for plan model, so that i can reuse it in commands. I added a `lib` folder in `pkg/portal`, see if it looks good for you?

pkg/lib/config/feature.go
```diff
@@ -88,3 +88,7 @@ func ParseFeatureConfig(inputYAML []byte) (*FeatureConfig, error) {
 
 	return &config, nil
 }
+
+func NewEffectiveDefaultFeatureConfig() *FeatureConfig {
+	return &FeatureConfig{}
```
louischan-oursky, line 93: I think we should populate the struct recursively so that we will not encounter nil pointer exception when we access the fields. I remember we have done the same to authgear.yaml.
That should be done for both `NewEffectiveDefaultFeatureConfig` and `ParseFeatureConfig`

pkg/portal/graphql/app.go
```diff
@@ -90,6 +90,12 @@ var nodeApp = node(
 					return p.Source.(*model.App).Context.Config.FeatureConfig, nil
 				},
 			},
+			"planName": &graphql.Field{
+				Type: graphql.NewNonNull(AppConfig),
```
louischan-oursky, line 94: The type should be string.

portal/src/graphql/portal/schema.graphql
```diff
@@ -30,6 +30,9 @@ type App implements Node {
   """The ID of an object"""
   id: ID!
 
+  """"""
+  planName: AppConfig!
```
louischan-oursky, line 34: The type should be string. After we correct the type in the code, re-generate the schema should correct this as well.

pkg/portal/model/plan.go
```diff
@@ -0,0 +1,11 @@
+package model
+
+import (
+	"github.com/authgear/authgear-server/pkg/lib/config"
+)
+
+type Plan struct {
+	ID            string
+	Name          string
+	FeatureConfig *config.FeatureConfig
```
louischan-oursky, line 10: This struct here is left as raw. That is, accessing the fields may trigger nil pointer exception. Do we need to use another name to remind us this?

louischan-oursky (commented):
> For the models in the portal pkg, we only have service but not store. I created a store for plan model, so that i can reuse it in commands. I added a lib folder in pkg/portal, see if it looks good for you?

I prefer we organize the code by feature. For example `pkg/portal/plan/store.go`, `pkg/portal/plan/service.go`, instead of `pkg/portal/lib/plan/store.go`, `pkg/portal/service/plan.go`. Putting related things together (in the case, plan-related) in the same Go package should give better encapsulation.

pkg/portal/model/plan.go
carmenlau: How about changing the pointer to struct?

pkg/lib/config/feature.go
carmenlau: I did it in the later commit, by using the `setFieldDefaults` function like `authgear.yaml`.

pkg/portal/model/plan.go
louischan-oursky: Using struct would be more problematic if we accidentally serialize it with `encoding/json`. Maybe rename it to `RawFeatureConfig`?

carmenlau:
Updated, also move plan service to `pkg/portal/lib/plan`

PR #1293, opened 2021-06-24
Reauthentication part 2

louischan-oursky:


pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -109,19 +109,38 @@ func (h *AuthorizationHandler) doHandle(
...
-		// after authentication
-		r2 := protocol.AuthorizationRequest{}
-		for k, v := range r {
-			r2[k] = v
+	// Handle max_age and prompt=login
+	prompt := r.Prompt()
+	if maxAge, ok := r.MaxAge(); ok {
+		impliesPromptLogin := false
+		// When there is no session, the presence of max_age implies prompt=login.
+		if s == nil {
```
carmenlau, line 127: We should check `s == nil || s.SessionType() != session.TypeIdentityProvider` here. Although in the normal case authz endpoint should not be called with access token, we should only accept idp session here. We have the same checking when we handle prompt=none.

carmenlau (commented):
Looks great to me so far, when I read the PR let me think of one thing!

For handling re-auth + the idp session is missing case, should we also hide the signup tab from the auth ui and redirect to /login only?

pkg/auth/webapp/service2.go
```diff
@@ -54,7 +53,7 @@ type Service2 struct {
 }
 
 func (s *Service2) CreateSession(session *Session, redirectURI string) (*Result, error) {
-	if err := s.Sessions.Create(session); err != nil {
+	if err := s.Sessions.Save(session); err != nil {
```
carmenlau, line 56: I am ok with the change, but i am curious to know why do we need it?

resources/authgear/templates/en/web/login.html
```diff
@@ -14,12 +14,12 @@
 	{{ end }}
 	<div class="pane flex flex-col">
 	<div class="flex flex-col">
-		{{ if not $.PublicSignupDisabled }}{{ if $has_login_id }}
```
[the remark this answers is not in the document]
louischan-oursky: Dropped that commit

pkg/auth/webapp/service2.go
louischan-oursky: I realized I should use `EntryPointPost` in `select_account.go`. After switching to using it, I do not need this change.

louischan-oursky:
I realized we did not update the auth_time of refresh token. I would like to discuss how should we update that.

pkg/lib/oauth/redis/store.go
```diff
@@ -223,18 +298,24 @@ func (s *Store) ListOfflineGrants(userID string) ([]*oauth.OfflineGrant, error)
...
-				grants = append(grants, grant)
 			}
+
+			var grant *oauth.OfflineGrant
+			grant, err = s.unmarshalOfflineGrant(data)
+			if err != nil {
+				return err
+			}
+
+			grants = append(grants, grant)
```
carmenlau, line 318: The original `grants = append(grants, grant)` logic is in the `else` block, we didn't return in the `ErrGrantNotFound` case. So we need to put back `unmarshalOfflineGrant` logic in the `else` block, or put `continue` in the `ErrGrantNotFound` case?

pkg/auth/handler/webapp/panic_middleware.go
```diff
@@ -29,8 +30,16 @@ func (m *PanicMiddleware) Handle(next http.Handler) http.Handler {
...
+				// Only rethrow if the err is NOT api error.
+				rethrow := true
+				if err, ok := err.(error); ok {
+					if apierrors.IsAPIError(err) {
+						rethrow = false
+					}
+				}
+				if rethrow {
+					panic(err)
+				}
```
carmenlau, line 42: Would like to know why do we need this change?

carmenlau:
I just tried constructing the authorize url to do reauthentication, and got error `invalid authorization code`. Not sure if there is problem in my setup, let check it together.

pkg/auth/handler/webapp/panic_middleware.go
louischan-oursky: This is because we now detect `interaction.ErrNoAuthenticator` in `select_account.go` before we start interaction. We use `panic` to report such error. If we do not have this change, we will see panic log about `interaction.ErrNoAuthenticator`, which is actually an expected error.

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -682,6 +683,23 @@ func (h *TokenHandler) issueTokensForAuthorizationCode(
...
 	}
 
+	// Update auth_time of the offline grant if possible.
+	if sid := code.IDTokenHintSID; sid != "" {
+		if typ, sessionID, ok := oidc.DecodeSID(sid); ok && typ == session.TypeOfflineGrant {
+			offlineGrant, err := h.OfflineGrants.GetOfflineGrant(sessionID)
+			if err == nil {
+				if s.AuthenticatedAt.After(offlineGrant.AuthenticatedAt) {
+					offlineGrant.AuthenticatedAt = s.AuthenticatedAt
+					expiry := oauth.ComputeOfflineGrantExpiryWithClient(offlineGrant, client)
```
carmenlau, line 693: There is a side effect that the expiry of refresh token will be extended after re-auth, I think it is acceptable?

carmenlau (commented):
The updates look good to me.

pkg/lib/oauth/handler/handler_token.go
louischan-oursky: Actually it will not. If we look at `oauth.ComputeOfflineGrantExpiryWithClient`, the factors affecting the expiry is `CreatedAt`, `LastAccessAt` and the client config. We do not modify any of these in the reauthentication explicitly. Of course the last access at will be updated when the access token is used to refresh ID token.

PR #1334, opened 2021-07-09
Show back button as static link

### LINKED ISSUE #1306: Do not show back button if there is no logical back action to take
`window.history` is global to the browser itself. Therefore, it is hard to determine the length of the back stack just by inspecting `window.history.length` without sophisticated counting.

The back button should be shown when there is logical back action to take. There are 2 cases.

## Case 1: There is a ongoing web session

When there is ongoing web session, we can inspect the interaction graph to see if there exists any interactive nodes **BEFORE** the current node. If yes, then clicking the back button would bring the end-user back to the previous interactive node. If no, then we should not show the back button.

## Case 2: There is no ongoing web session, but the URL is hierarchical.

When the URL is hierarchical, then back button should be shown when the level of hierarchy is more than 1.
For example, when the end-user is at `/settings`, they should not see the back button. When they are at `/settings/mfa`, they should see the back button.

louischan-oursky:
fix #1306

### DIFF
--- a/authui/src/authgear.ts
+++ b/authui/src/authgear.ts
@@ -468,52 +468,6 @@ window.api.onLoad(() => {
   };
 });
 
-// Handle back button click.
-
-function handleBack(pathname: string): boolean {
-  const pathComponents = pathname.split("/").filter(c => c !== "");
-  if (pathComponents.length > 1 && pathComponents[0] === "settings") {
-    const newPathname = "/" + pathComponents.slice(0, pathComponents.length - 1).join("/");
-    Turbolinks.visit(newPathname, { action: "replace" });
-    return true;
-  }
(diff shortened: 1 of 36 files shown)


carmenlau:
After we removed the back button, seems look better if we add further 20px padding top to the pane? What do you think?


Before:
Screenshot 2021-07-12 at 12 16 39 PM

After:
Screenshot 2021-07-12 at 12 17 22 PM

fungc-io:
@carmenlau agree

carmenlau (commented):
@louischan-oursky Sorry one more comment, I think we need to remove `back-btn` related js code too?

louischan-oursky:
Add margin top 20px to most page titles.

PR #1348, opened 2021-07-14
Refactor portal resources manager

### LINKED ISSUE #1341: Refactor portal resource manager
Currently the config (e.g. SecretKeyAllowlist) are provided by the `AppFileView`. Refactor to add a portal resource manager layer, so that those config can be provided by the portal resource manager.

carmenlau:
fix #1341 

Instead of providing the config (e.g secretKeyAllowlist) by the caller, add a portal resource manager layer to obtain config through DI.


pkg/util/resource/view.go
```diff
@@ -20,12 +20,17 @@ type View interface {
 	view()
 }
 
+// ViewWithConfig is a wrapper that provides config to view
+type ViewWithConfig interface {
+	View
+	SecretKeyAllowlist() []string
+}
```
louischan-oursky, line 27: My original thought is that we do not even define this interface in this package, because this package is in `util` which should not be aware of secretkey allowlist.

carmenlau: If this is the case, where should we put the struct `AppFileWithConfig`? Seems putting it into package `pkg/lib/config/configsource` is not appropriate, since this is not only for descriptors in the configsource. See if you have any idea?

carmenlau: Updated!

PR #1350, opened 2021-07-15
Update portal using fluent ui Link component for translations string

carmenlau:
Otherwise the link style is missing

### DIFF
--- a/portal/src/ReactApp.tsx
+++ b/portal/src/ReactApp.tsx
@@ -27,10 +27,11 @@ import {
   instantiateSystemConfig,
   mergeSystemConfig,
 } from "./system-config";
-import { loadTheme } from "@fluentui/react";
+import { loadTheme, Link as FluentLink } from "@fluentui/react";
 import OnboardingConfigAppScreen from "./graphql/portal/OnboardingConfigAppScreen";
 import OnboardingCompletionScreen from "./graphql/portal/OnboardingCompletionScreen";
 import OnboardingRedirect from "./OnboardingRedirect";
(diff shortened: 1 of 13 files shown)


louischan-oursky:
Actually we still have a problem. If we use the bare Link from fluent, when the link is clicked, the app is reloaded.

To avoid reloading, we must use the Link from react-router-dom. I have built `src/ReactRouterLink.tsx` to make it work with the Link from fluent as we expect. You can read the source code of it to know the reason why it is there.

Finally, what we want here is to make [react-messageformat](https://github.com/oursky/react-messageformat) to support default `components`, so we do not need to provide `components` everywhere.

I suggest we add the default `components` in react-messageformat and then configure it to have a default `Link` component, which is backed by `ReactRouterLink`

carmenlau:
Updated using latest `react-messageformat`

portal/src/ReactApp.tsx
```diff
@@ -112,7 +119,14 @@ const ReactApp: React.FC = function ReactApp() {
...
   if (error != null) {
     return (
-      <LocaleProvider locale="en" messageByID={MESSAGES}>
+      <LocaleProvider
+        locale="en"
+        messageByID={MESSAGES}
+        defaultComponents={{
+          ReactRouterLink: PortalLink,
+          Link: FluentLink,
+        }}
```
louischan-oursky, line 128: The `defaultComponents` can be defined as static variable.

louischan-oursky:
Some of the usage of `Link` in `en.json` is missing `target="_blank"` and `rel="noreferrer"`. Can we add them back?

carmenlau:
Updated, see if it looks good!:)

PR #1373, opened 2021-07-20
 Limit audit logs retrieve days by feature config

carmenlau:
require #1372

### DIFF
--- a/pkg/lib/config/feature_audit_log.go
+++ b/pkg/lib/config/feature_audit_log.go
@@ -1,0 +1,22 @@
+package config
+
+var _ = FeatureConfigSchema.Add("AuditLogFeatureConfig", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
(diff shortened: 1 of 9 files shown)


pkg/lib/config/feature_audit_log.go
```diff
@@ -0,0 +1,22 @@
+package config
+
+var _ = FeatureConfigSchema.Add("AuditLogFeatureConfig", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"retrieve_days": { "type": "integer" }
```
louischan-oursky, line 8: `retrieval_days` seems more grammatically correct.

louischan-oursky (commented):
We should also appear the limit to the date range control.

carmenlau:
Updated the config name and added date boundaries to audit log date picker

carmenlau:
Updated to set the from query for full date range

PR #1385, opened 2021-07-23
Fix webhook identity payload created_at and updated_at are not up to date

carmenlau:
ref #1375

### DIFF
--- a/pkg/lib/authn/identity/service/service.go
+++ b/pkg/lib/authn/identity/service/service.go
@@ -347,24 +347,28 @@ func (s *Service) Create(info *identity.Info) error {
 		if err := s.LoginID.Create(i); err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/lib/authn/identity/service/service.go
```diff
@@ -347,24 +347,32 @@ func (s *Service) Create(info *identity.Info) error {
 		if err := s.LoginID.Create(i); err != nil {
 			return err
 		}
+		info.CreatedAt = i.CreatedAt
+		info.UpdatedAt = i.UpdatedAt
```
louischan-oursky, line 351: The pattern here should be

```
identity := functionThatTurnInfoIntoIdentity
err := s.Create(identity)
info = functionThatTurnIdentityIntoInfo(identity)
```

Issue #1384, opened 2021-07-23
Logo in UI Settings fallback to the default language


fungc-io:
Logo in UI Settings fallback to the default language chosen by the user

PR #1388, opened 2021-07-26
Update image descriptor to use fallback language in the app fs first

carmenlau:
ref #1384

### DIFF
--- a/pkg/lib/web/image_test.go
+++ b/pkg/lib/web/image_test.go
@@ -135,6 +135,29 @@ func TestTemplateResource(t *testing.T) {
 			So(asset.Data, ShouldResemble, pngA)
 		})
 
+		Convey("it should use fallback language in the app fs first", func() {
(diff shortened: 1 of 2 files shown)


pkg/lib/web/image_test.go
```diff
@@ -135,6 +135,29 @@ func TestTemplateResource(t *testing.T) {
...
+			writeFile(fsB, "jp", ".png", pngB)
+
+			asset, err := read(resource.EffectiveResource{
+				SupportedTags: []string{"en", "jp"},
+				DefaultTag:    "jp",
+				PreferredTags: []string{"en"},
+			})
+			So(err, ShouldBeNil)
+			So(asset.Path, ShouldEqual, "static/jp/myimage.png")
+			So(asset.Data, ShouldResemble, pngB)
```
louischan-oursky, line 149: This test case should resolve to `en`.

Issue #1110, opened 2021-04-01
Improve copy client id button


mr9dollars:
Show an `Copy` icon button next to the ClientId

Show ~a toast (`Snackbar` in Fluent UI)~ a `Tooltip` with the message "Copied to clipboard" when successfully copied, dismiss after 2 seconds

Design:
https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=4252%3A22663

Ref:
https://fabricweb.z5.web.core.windows.net/oufr/8.26.1/#/examples/tooltip
Tooltip with a closing delay

fungc-io:
@samchan0221 Updated description and included the design file

samchan0221:
time estimation: 3 hrs

fungc-io:
Updated the design to Tooltip to make use of Fluent UI components

fungc-io:
I've found a bug
- there are more than 2 applications in the project
- clicking the copy button on the second row
- The tooltip is shown above the 1st copy button

<img width="504" alt="Screenshot 2021-08-06 at 11 27 54" src="https://user-images.githubusercontent.com/15207470/128451723-af57fb06-366a-4adb-920c-30d22029c40d.png">

PR #1425, opened 2021-08-03
Improved UI on client_id copy

samchan0221:
ref #1110 
- [x] Added copy button
- [x] Added popup on copy
- [x] Remove old copy button logic
image

### DIFF
--- a/portal/src/graphql/portal/ApplicationsConfigurationScreen.module.scss
+++ b/portal/src/graphql/portal/ApplicationsConfigurationScreen.module.scss
@@ -55,13 +55,3 @@
   pointer-events: none;
   opacity: 0.5;
 }
-
-.callout {
-  padding: 8px;
-}
-
-.copyButtonRoot {
-  color: #323130;
-  background-color: #edebe9;
-  margin: 4px;
-}
--- a/portal/src/graphql/portal/ApplicationsConfigurationScreen.module.scss
+++ b/portal/src/graphql/portal/ApplicationsConfigurationScreen.module.scss
@@ -55,3 +55,13 @@
   pointer-events: none;
   opacity: 0.5;
 }
+
+.callout {
+  padding: 8px;
+}
+
+.copyButtonRoot {
+  color: #323130;
+  background-color: #edebe9;
+  margin: 4px;
+}
--- a/portal/src/graphql/portal/ApplicationsConfigurationScreen.tsx
+++ b/portal/src/graphql/portal/ApplicationsConfigurationScreen.tsx
@@ -10,9 +10,14 @@ import {
   Callout,
   DetailsList,
   DirectionalHint,
+  IButtonStyles,
+  ICalloutContentStyleProps,
+  ICalloutContentStyles,
   IColumn,
   ICommandBarItemProps,
   IconButton,
+  IIconProps,
+  IStyleFunctionOrObject,
   MessageBar,
   SelectionMode,
   Text,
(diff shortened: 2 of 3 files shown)


portal/src/graphql/portal/ApplicationsConfigurationScreen.tsx
```diff
@@ -83,33 +93,87 @@ function makeOAuthClientListColumns(
...
+    return (
+      <>
+        <span className={styles.cellContent}>{clientId}</span>
+        <IconButton
+          id={copyButtonId}
+          onClick={onCopyClick}
+          iconProps={{ iconName: "Copy" }}
+          title="Copy"
+          ariaLabel="Copy"
+          styles={{ root: styles.copyButtonRoot }}
```
louischan-oursky, line 140: This object can be lifted as module-level constant.

portal/src/graphql/portal/ApplicationsConfigurationScreen.tsx
```diff
@@ -83,33 +93,87 @@ function makeOAuthClientListColumns(
...
+        />
+        {isCalloutVisible && (
+          <Callout
+            target={`#${copyButtonId}`}
+            className={styles.callout}
+            directionalHint={DirectionalHint.topCenter}
+            onDismiss={handleCalloutDismiss}
+          >
+            <DelayedRender>
```
louischan-oursky, line 149: Can we avoid using DelayedRender?

portal/src/graphql/portal/ApplicationsConfigurationScreen.tsx
louischan-oursky: Can we add 1 more condition to dismiss the callout? The condition is after 0.5s when onMouseLeave fires on IconButton.

Can we tidy up the show/dismiss logic as a hook?

samchan0221:
@louischan-oursky 
I have committed the changes you requested
1. Moved icon props out as constant
2. Written a useDelayCallback hook
3. Removed the delay render wrapper

portal/src/graphql/portal/ApplicationsConfigurationScreen.tsx
```diff
@@ -102,31 +101,48 @@ function makeOAuthClientListColumns(
   ];
 }
 
+function useDelayedAction(delayed: () => void): (delay: number) => void {
+  const [delay, setDelay] = useState<[number] | null>(null);
```
louischan-oursky: Add a comment to explain why `delay` is a tuple of 1 element.

portal/src/graphql/portal/ApplicationsConfigurationScreen.tsx
```diff
@@ -101,24 +98,6 @@ function makeOAuthClientListColumns(
   ];
 }
 
-function useDelayedAction(delayed: () => void): (delay: number) => void {
```
louischan-oursky: The original version of the hook uses `useEffect` to correctly cleanup scheduled setTimeout when the component unmounts. I suggest we keep this version.

portal/src/graphql/portal/ApplicationsConfigurationScreen.module.scss
```diff
@@ -55,3 +55,19 @@
   pointer-events: none;
   opacity: 0.5;
 }
+
+.callout {
+  padding: 8px;
```
louischan-oursky, line 60: Use `styles` to custom the component itself. Use `className` to layout the component within its parent.

portal/src/graphql/portal/ApplicationsConfigurationScreen.module.scss
```diff
@@ -55,3 +55,19 @@
...
+}
+
+.copyButtonRoot {
+  margin: 4px;
```
louischan-oursky, line 64: For example, this should be referred as `className={styles.copyButton}` (No suffix Root because we do not assume the internal HTML structure of the component)

portal/src/graphql/portal/ApplicationsConfigurationScreen.module.scss
```diff
@@ -55,3 +55,19 @@
...
+}
+
+.copyButtonRoot:hover {
+  background-color: #d8d6d3;
+}
+
+.copyButtonRoot:active {
+  background-color: #c2c0be;
+}
```
louischan-oursky, line 73: Changing the hover and active color is of customizing the component. So this should be written using `styles` prop.

Issue #1320, opened 2021-07-07
[Web Portal, UI Settings] Custom theme colour should be saved transiently for quick colour comparison


stanleylhs:
ux/minor

## Background:
- Device / Browser (OS version): Chrome (90.0.4430.93) on macbook pro
- Build: Staging, 5 May 2021

## Description:
In UI Settings Light/ Dark mode tab, if any preset theme colour is selected, any custom theme colour which the user have picked will be lost. Therefore, it would be very inconvenient if users want to compare preset theme colour and desired custom theme colour.

## Steps to reproduce:
1. Login Authgear
2. Click any preexisting app (Create one if there is none)
3. Click UI Settings
4. Select a preset colour theme in Light/ Dark Mode
5. Click Custom 
6. Change Background, Text and primary colour 
7. Select a preset colour theme
8. Click Custom
9. See Background, Text and primary colour 

## Expected result:
It would be convenient if custom colour theme can be saved transiently as users might want to compare preset theme colour and custom theme colour.

## Screenshot:

https://user-images.githubusercontent.com/4337525/125577818-9884f0b6-b369-45e1-996c-58c78c929978.mp4

samchan0221:
time estimation: 5hrs

PR #1439, opened 2021-08-09
Add component-level memorization on custom color

samchan0221:
Ref #1320

### DIFF
--- a/portal/src/ThemeConfigurationWidget.tsx
+++ b/portal/src/ThemeConfigurationWidget.tsx
@@ -212,6 +212,49 @@ const ThemeConfigurationWidget: React.FC<ThemeConfigurationWidgetProps> =
       [onChangeDarkModeEnabled]
     );
 
+    const [customTheme, setCustomTheme] = useState<
+      Omit<LightTheme | DarkTheme, "isDarkTheme">
+    >(() => {
+      if (isDarkMode) {
+        if (darkTheme) {
+          const { isDarkTheme, ...colors } = darkTheme;
+          return colors;
+        }
+        const { isDarkTheme, ...colors } = DEFAULT_DARK_THEME;
+        return colors;
+      }
+      if (lightTheme) {
+        const { isDarkTheme, ...colors } = lightTheme;
+        return colors;
+      }
+      const { isDarkTheme, ...colors } = DEFAULT_LIGHT_THEME;
+      return colors;
+    });
+
+    const onPickerChangePrimaryColor = useCallback(
+      (color: string) => {
+        setCustomTheme((theme) => ({ ...theme, primaryColor: color }));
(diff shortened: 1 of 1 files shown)


portal/src/ThemeConfigurationWidget.tsx
```diff
@@ -212,6 +212,49 @@ const ThemeConfigurationWidget: React.FC<ThemeConfigurationWidgetProps> =
...
+        const { isDarkTheme, ...colors } = lightTheme;
+        return colors;
+      }
+      const { isDarkTheme, ...colors } = DEFAULT_LIGHT_THEME;
+      return colors;
+    });
+
+    const onPickerChangePrimaryColor = useCallback(
+      (color: string) => {
+        setCustomTheme({ ...customTheme, primaryColor: color });
```
louischan-oursky, line 236: setState can accept an update function. By using the update function, we can avoid depending on the current state.

portal/src/ThemeConfigurationWidget.tsx
```diff
@@ -212,6 +212,49 @@ const ThemeConfigurationWidget: React.FC<ThemeConfigurationWidgetProps> =
...
+        onChangePrimaryColor(color);
+      },
+      [customTheme, onChangePrimaryColor]
+    );
+
+    const onPickerChangeTextColor = useCallback(
+      (color: string) => {
+        setCustomTheme({ ...customTheme, textColor: color });
```
louischan-oursky, line 244: setState uses update function.

portal/src/ThemeConfigurationWidget.tsx
```diff
@@ -212,6 +212,49 @@ const ThemeConfigurationWidget: React.FC<ThemeConfigurationWidgetProps> =
...
+        onChangeTextColor(color);
+      },
+      [customTheme, onChangeTextColor]
+    );
+
+    const onPickerChangeBackgroundColor = useCallback(
+      (color: string) => {
+        setCustomTheme({ ...customTheme, backgroundColor: color });
```
louischan-oursky, line 252: setState uses update function.

louischan-oursky (commented):
Memoize the custom theme inside ThemeConfigurationWidget has one caveat. The discard change button does not discard the custom theme :( But fixing that caveat is not trivial so let ignore that for now.

samchan0221:
> Memoize the custom theme inside ThemeConfigurationWidget has one caveat. The discard change button does not discard the custom theme :( But fixing that caveat is not trivial so let ignore that for now.

I also thought about that. Passing state top-down or getting ref bottom-up are both not an elegant way to solve it. 
Maybe adding another useState in the configform hooks can do?

louischan-oursky:
Build failed

Issue #1331, opened 2021-07-09
Web session should not remain if the end-user quits


louischan-oursky:
Example

1. The end-user triggers reauthentication
1. The end-user quits reauthentication by clicking back button
1. The web session cookie remains in the browser
1. The end-user visit web UI by typing in the browser URL bar
1. The web session state is messed up

--- 

Proposed solution

Introduce a new endpoint for the authorization endpoint to enter web UI. That endpoint will redirect to `/select_account`.
Change `/` not to redirect to `/select_account`. It should redirect to `/login` or `/signup` only.

PR #1459, opened 2021-08-13
Change OAuth entrypoint from / to /_oauth_entrypoint

louischan-oursky:
ref #1331


chpapa:
Or just /_oauth? So the url is shorter in case we need to explain it in some doc like "assemble your own authorise url"?

louischan-oursky:
This is not the public facing endpoint. The public facing one is `/oauth2/authorize` and it did not change. This change is purely implementation details.

chpapa:
Icic

Issue #1438, opened 2021-08-06
OIDC Standard Claims: Email, Phone and Username


fungc-io:
Subtask of #1352 
Implement support for the claims that are populated from identities

- preferred_username
- email
- email verified
- phone_number
- phone_number_verified

PR #1498, opened 2021-09-07
Design of user profile

louischan-oursky:
ref #1438

### DIFF
--- a/docs/specs/user-profile/design.md
+++ b/docs/specs/user-profile/design.md
@@ -1,0 +1,329 @@
+# User profile
+
+User profile consists of standard attributes and custom attributes.
+
+## Standard attributes
+
+Standard attributes is a subset of [OIDC standard claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims).
+All standard attributes are optional.
+All standard attributes are readable and writable via the Admin API.
+All standard attributes are readable and writable in the portal.
+
+All standard attributes are by default readable and writable by the end-user in the settings page.
+The developer can restrict access to individual standard attribute.
+
+```yaml
+ui:
+  standard_attributes:
+  - pointer: /given_name
+    access_control: hidden
+  - pointer: /family_name
+    access_control: hidden
+```
+
+### email
+
+The primary email address of the end-user.
+The value always comes from one of the identity the end-user has.
+Therefore, the input control is a dropdown.
+It is subject to [automatic population](#automatic-population).
+
+### email\_verified
+
+Whether the primary email address is verified.
+This attribute is present when `email` is present.
+It is NOT writable.
+
+### phone\_number
+
+The primary phone number of the end-user.
+The value always from from one of the identity the end-user has.
+Therefore, the input control is a dropdown.
+It is subject to [automatic population](#automatic-population).
+
+### phone\_number\_verified
+
+Whether the primary phone number is verified.
+This attribute is present when `phone_number` is present.
+It is NOT writable.
+
+### preferred\_username
+
+The primary username of the end-user.
+The value always comes from one of the identity the end-user has.
+Therefore, the input control is a dropdown.
+It is subject to [automatic population](#automatic-population).
+
+### zoneinfo
+
+The preferred tz database name, e.g. `Asia/Hong_Kong`.
+The list of tz database names are constants, so the input control is a dropdown.
+
+### locale
+
+The preferred BCP47 tag, e.g. `zh-HK`.
+The input control is a dropdown, and the options are the supported languages of the project.
+
+### birthdate
+
+The supported format is `YYYY-MM-DD`, representing the birthdate, e.g. 1 Jan 1992.
+
+> The OIDC spec also allows `0000-MM-DD` for birthday and `YYYY` for year of birth.
+> We only support `YYYY-MM-DD` for simplicity.
+
+### given\_name
+
+The given name of the first name of the end-user.
+The input control is a simple text input without validation.
+It is subject to [automatic population](#automatic-population).
+
+### family\_name
+
+The family name or the last name of the end-user.
+The input control is a simple text input without validation.
+It is subject to [automatic population](#automatic-population).
+
+## Automatic population
+
+> Automatic population CANNOT be turned off!
+
+Some of the standard attributes are subject to automatic population.
+
+The automatic population runs when one of the following situation happens:
+
+- When the end-user signs up.
+- When the end-user adds an identity.
+- When the end-user updates an identity.
+- When the end-user removes an identity.
+
+The steps of the automatic population are:
+
+1. Generate a list of email address candidates from the identities, candidates from newer identities are ordered first.
+1. Generate a list of phone number candidates from the identities, candidates from newer identities are ordered first.
+1. Generate a list of username candidates from the identities, candidates from newer identities are ordered first.
+1. Generate a list of given name and family name candidates from the identities, candidates from newer identities are ordered first.
+1. If the `email` standard attribute does not refer to a candidate in the list, clear it.
+1. If the `phone_number` standard attribute does not refer to a candidate in the list, clear it.
+1. If the `preferred_username` standard attribute does not refer to a candidate in the list, clear it.
+1. If the `email` standard attribute is absent and the list of candidate is non-empty, set it to the first candidate in the list.
+1. If the `phone_number` standard attribute is absent and the list of candidate is non-empty, set it to the first candidate in the list.
+1. If the `preferred_username` standard attribute is absent and the list of candidate is non-empty, set it to the first candidate in the list.
+1. If both the `given_name` and the `family_name` standard attribute are absent and the list of candidate is non-empty, set it to the first candidate in the list.
+
+## Custom attributes
+
+In addition to standard attributes, the developer can define their own custom attributes.
+
+### Defining custom attributes
+
+The custom attributes are defined as a single JSON schema written against a subset of JSON schema 2019-09.
+
+Here is an example of the schema of the custom attributes.
+
+```JSON
+{
+  "properties": {
+    "app_user_role": {
+      "type": "string",
+      "enum": ["owner", "editor", "viewer"]
+    },
+    "stripe_customer_id": {
+      "type": "string"
+    }
+  }
+}
+```
+
+All changes made to custom attributes must be valid against the schema.
+
+#### Supported subset of JSON schema 2019-09
+
+- `type`: `boolean`, `string`, `number`, `integer`
+- `enum`
+- `multipleOf`
+- `maximum`
+- `exclusiveMaximum`
+- `minimum`
+- `exclusiveMinimum`
+- `maxLength`
+- `minLength`
+- `properties`
+
+### Custom attributes and the resolver
+
+The resolver originally can tell whether the request is authenticated.
+If the developer has defined a custom attribute to store the role of the user,
+the developer will want to know the role of the user as well.
+Then the backend server can do authentication and authorization by forwarding a subrequest to the resolver, without the overhead of calling the Admin API.
+
+The resolver includes the custom attributes of the end-user as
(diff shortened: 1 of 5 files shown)


docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,313 @@
...
+- `multipleOf`
+- `maximum`
+- `exclusiveMaximum`
+- `minimum`
+- `exclusiveMinimum`
+- `maxLength`
+- `minLength`
+- `properties`
+
+### Custom attributes and the resolver
```
fungc-io, line 152: **Discussion:** Should standard attributes also be included in the header of the resolver response?
Here we assume the users use custom attr for access control, returning them in the header will be helpful coz they dont need to call admin api in every request.

docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,329 @@
...
+All standard attributes are by default readable and writable by the end-user in the settings page.
+The developer can restrict access to individual standard attribute.
+
+```yaml
+ui:
+  standard_attributes:
+  - pointer: /given_name
+    access_control: hidden
+  - pointer: /family_name
+    access_control: hidden
```
chpapa, line 21: What are the options in `access_control`?

docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,329 @@
...
+    }
+  }
+}
+```
+
+All changes made to custom attributes must be valid against the schema.
+
+#### Supported subset of JSON schema 2019-09
+
+- `type`: `boolean`, `string`, `number`, `integer`
```
chpapa, line 141: I think we need way more than these types, at least if we look at the standard attributes, there are more types like `email`, `phone`, `url`, `photo`, `date`, `time`

docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,329 @@
...
+- `properties`
+
+### Custom attributes and the resolver
+
+The resolver originally can tell whether the request is authenticated.
+If the developer has defined a custom attribute to store the role of the user,
+the developer will want to know the role of the user as well.
+Then the backend server can do authentication and authorization by forwarding a subrequest to the resolver, without the overhead of calling the Admin API.
+
+The resolver includes the custom attributes of the end-user as
```
chpapa, line 159: would it be too hacky, or actually handy, if we have a special attributes `role` or `roles` that would be in the header alone...?

docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,329 @@
...
+```yaml
+ui:
+  custom_attributes:
+  - pointer: /app_user_role
+    access_control: readonly
+  - pointer: /hobby
+    access_control: readwrite
+```
+
+### Editing custom attributes in the settings page and in the portal
```
chpapa, line 210: if we can support more type's user friendly control here would be great

docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,329 @@
...
+  "family_name": "Doe",
+  "custom_attributes": {
+    "hobby": "reading"
+  }
+}
+```
+
+## Synchronization of user profile between Authgear and the backend server
+
+There is NO blocking nor non-blocking webhook when standard attributes and custom attributes
```
chpapa, line 245: As we discussed in meeting last time, what about if developer want to add server side validation? (Not a must in 1st implementation)

chpapa: and say in the case of `zoneinfo` or `locale`, have a webhook to know it is changed?

chpapa (changes requested):
Have a few comments and questions needs clarifications.

docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,329 @@
+# User profile
+
+User profile consists of standard attributes and custom attributes.
+
+## Standard attributes
+
+Standard attributes is a subset of [OIDC standard claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims).
```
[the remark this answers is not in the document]
louischan-oursky: If we support all the name attributes, namely given_name, family_name, middle_name, name and nickname,
there are too many name fields for the end-user to fill in.
For example, Google supports given_name, family_name and name only. https://developers.google.com/identity/protocols/oauth2/openid-connect

The profile attribute is just an URL.
Google supports it. https://developers.google.com/identity/protocols/oauth2/openid-connect
Facebook does NOT support it. https://developers.facebook.com/docs/graph-api/reference/user/#default-public-profile-fields

The picture attribute is just an URL.
Google supports it. https://developers.google.com/identity/protocols/oauth2/openid-connect
Facebook supports it. https://developers.facebook.com/docs/graph-api/reference/user/#default-public-profile-fields
I tend not to support allowing the end-user to upload their profile picture.
That would require us to maintain an object storage just for that purpose.
The simpliest way to have minimal support is support picture attribute with our automatic population.

The website attribute is just an URL.
Google does NOT support it. https://developers.google.com/identity/protocols/oauth2/openid-connect
Facebook does NOT support it. https://developers.facebook.com/docs/graph-api/reference/user/#default-public-profile-fields

The gender attribute is tricky.
On Facebook you can set it to values other than male or female.
Different application have different requirement on the possible values of this attribute.
It takes a lot of effort to support the gender attribute in a socially inclusive way.

The address attribute is hard to get it right.
Different country or region has different addressing system.
On Amazon, the end-user first choose the country or region, and then the address fields are changed.
It takes a lot of effort to get it right.
Amazon needs to get it right because it needs their customers addresses to ship goods, but we do not.
It is better to leave this to the application.

According to the OIDC spec, the updated_at is the time end-user information changes.
Since we do not have user profile now, we do not have updated_at.
Of course we can add it.

docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,329 @@
...
+
+### family\_name
+
+The family name or the last name of the end-user.
+The input control is a simple text input without validation.
+It is subject to [automatic population](#automatic-population).
+
+## Automatic population
+
+> Automatic population CANNOT be turned off!
```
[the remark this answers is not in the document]
louischan-oursky: I guess you might misunderstand my proposal on the automatic population.
The standard attributes subject to the automatic population are only populated when they are originally absent.
So if the end-user already has the email attribute present, it will not be changed because a new identity is added.

docs/specs/user-profile/design.md
louischan-oursky: JSON schema has a keyword called format.
We can define email, phone, url, and datetime as format.
email can just be rendered using a text input.
phone can be rendered with the existing library we are using.
url can just be rendered using a text input, with validation to ensure the value is an URL.
If there is an existing library for date, time, datetime, then we can support them.

docs/specs/user-profile/design.md
louischan-oursky: I tend not to impose a specific structure on the custom attributes.

docs/specs/user-profile/design.md
louischan-oursky: > As we discussed in meeting last time, what about if developer want to add server side validation? (Not a must in 1st implementation)

I would recommend they store that information themselves.

> and say in the case of zoneinfo or locale, have a webhook to know it is changed?

As I mentioned in the use case examples, if the developer wants to know when the value is changed,
they probably need that information for their business logic.
In that case I highly recommend they store zoneinfo and locale themselves,
and they tell us the new values so that localization of Authgear is consistent with their application.

docs/specs/user-profile/design.md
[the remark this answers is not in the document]
chpapa: So I don’t see any reason beyond it is a bit more difficult to implement…? 

Use case wise I think the minimal is we should add support for profile pic, which is pretty commonly expected when there is a user profile between multiple apps share same login

docs/specs/user-profile/design.md
chpapa: However I think we would probably need it for something like roles, otherwise user management portal design would be way more complicated

docs/specs/user-profile/design.md
chpapa: I think assuming all developers would not expect using Authgear custom attr is not the single source of truth is not realistic

docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,331 @@
...
+The preferred BCP47 tag, e.g. `zh-HK`.
+The input control is a dropdown, and the options are the supported languages of the project.
+When it is populated, the supported languages are taken into account.
+
+### birthdate
+
+The supported format is `YYYY-MM-DD`, representing the birthdate, e.g. 1 Jan 1992.
+
+> The OIDC spec also allows `0000-MM-DD` for birthday and `YYYY` for year of birth.
+> We only support `YYYY-MM-DD` for simplicity.
```
fungc-io, line 112: How much more complex will it be to support `0000-MM-DD` and `YYYY`?

louischan-oursky: The complexity is we do not know the need of the developer. We have to add configuration to let the developer tell us which format(s) they want their end-users to enter.

And we have to answer questions like:

- What if we receive a value from an external identity provider that is not in one of the allowed format.
- How do we present the UI to let the end-user know they can provide their birthdate / birthday / year of birth?

docs/specs/user-profile/design.md
```diff
@@ -0,0 +1,331 @@
...
+
+Possible values of `access_control` are:
+
+- `hidden`: The standard attribute is hidden from the end-user, the User Info endpoint and the resolver.
+- `internal`: The standard attribute is hidden in the settings page. But it is visible to the User Info endpoint and the resolver.
+- `readonly`: The standard attribute is visible to the end-user, the User Info endpoint and the resolver. The end-user can view but cannot edit it in the settings page.
+- `readwrite`: The standard attribute is visible to the end-user, the User Info endpoint and the resolver. The end-user can view and edit it in the settings page.
+
+The default value of `access_control` of all standard attributes is `readwrite`.
+
```
fungc-io, line 32: Recap discussion in standup:
May need a config to control visibility in the Admin Portal too. 
As we provide input fields for changing these values. When using the portal for user management, the staff will feel overwhelmed by the attributes which are not used by their app at all.

docs/specs/user-profile/design.md
fungc-io: I was thinking what if our UI always support YYYY-MM-DD only. But also allow `0000-MM-DD` and `YYYY` in Admin API.

fungc-io (approved):

docs/specs/user-profile/design.md
```diff
@@ -237,6 +191,86 @@ The UI control of the custom attribute is determined by the `type` and the `form
...
+|hidden|hidden|hidden|
+|hidden|hidden|readonly|
+|hidden|hidden|readwrite|
+|hidden|readonly|readonly|
+|hidden|readonly|readwrite|
+|hidden|readwrite|readwrite|
+|readonly|readonly|readonly|
+|readonly|readonly|readwrite|
+|readonly|readwrite|readwrite|
+|readwrite|readwrite|readwrite|
```
carmenlau, line 296: `readwrite` in the bearer means we will have api in the SDK to allow update user profile?
If yes, would be great if we can defined the SDK api in the spec.

carmenlau (commented):
Overall looks great!

docs/specs/user-profile/design.md
louischan-oursky: No. `readwrite` is equivalent to `readonly` in this case. See https://github.com/authgear/authgear-server/pull/1498/files#diff-299c7f857be8e61cccc8b09010cc5a0399272250bfeb8a4aeaff53292500b863R257

carmenlau: Thanks for clarifying, LGTM:)

carmenlau:
We started the development of standard attributes already, I am going to merge it if no further comments :)

carmenlau (approved):

PR #1512, opened 2021-09-13
Update OAuth identity on authentication

louischan-oursky:
ref #1507

### DIFF
--- a/pkg/lib/authn/identity/oauth/provider.go
+++ b/pkg/lib/authn/identity/oauth/provider.go
@@ -69,6 +69,24 @@ func (p *Provider) New(
 	return i
 }
 
+func (p *Provider) WithUpdate(
+	iden *Identity,
+	rawProfile map[string]interface{},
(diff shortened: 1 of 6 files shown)


carmenlau (commented):
@louischan-oursky This change is difficult to test, do you have the steps that can test it?

I tried manually update the db to test, but it fails with error `verification: expect login ID identity`. Here is my testing steps.

1. I manually update db data 
      - table `_auth_identity_oauth`, change claims's email to the new email
      - table `_auth_verified_claim` change the email to the new email
2. Login with google, I expect the email should be updated to the original email, but got the error.

Maybe we need to handle to mark the new email as verified? Or maybe the update is fine, only the problem of my testing flow. Please let me know how you test it or I should find one provider that can update email? :P

louischan-oursky:
Added a commit to fix this.

Your steps to test is correct. The reason for the error you saw is that I forgot we have to mark the updated email as verified.

Issue #1472, opened 2021-08-20
Analytics page on Portal


fungc-io:
It would be great if we can show some analytics in the portal to the users.
The analytics page can show the funnel stats and tenant usage:

Numbers:
- number of Total users

Charts + Numbers by period:
- number of Active users
- number of login pageview
- number of Sign up pageview
- number of Sign up
Breakdown by methods
- (Number only) Conversion percentage in the period, i.e. # of signup / # of signup pageview 

----
Data to collect
- Monthly active user
- Weekly active user
- Daily active user (Tracked, but not used by the chart)
- Total number of users (Daily)
- Number of signup (Daily)
- Number of signup page view (Daily)
    - unique (+1 only per web session)
    - total (+1 for every visit) (Tracked, but not used by the chart)
- Number of login page view (Daily)
    - unique (+1 only per web session)
    - total  (+1 for every visit)  (Tracked, but not used by the chart)
- Number of signup per provider (Daily)

---

TODO:
- [x] Refactor to connect analytic redis
- [x] Collect count with analytic redis
- [x] DB model and export count to db
- [x] Update portal graphql api for getting count data
- [ ] Portal frontend support display charts

carmenlau:
Update the api to fit the updated design:
- Separate the `signupSummary` query  to `totalUserCountChart`, `signupConversionRate` and `signupByMethodsChart`
- Add `AnalyticEpoch` config to portal to specific the date of start collecting data, so the user can only select date range from the epoch to yesterday
- Add validation, if query is selecting date range outside the range of epoch to yesterday. Query should return empty dataset.
- Add validation, if user is selecting current month / current week for monthly and weekly chart. Query should return empty dataset.

When we develop portal
- Should get `AnalyticEpoch` from system config, date picker should only allow selecting date from epoch to yesterday
- For line chart (Monthly, weekly active user / total number of user), if server returns empty dataset, show the empty state. The chart should still show the line, if data are all zero.
- Conversation rate chart and signup by methods chart should show empty state if data are all zero.

fungc-io:
Need updates on the design:

https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=4769%3A44729

- Designs for different numbers of signup methods, most users will only have 1-3 sign up methods
- Confirm the color codes used for signup methods

fungc-io:
@carmenlau 
For the percentage label, please use the primary blue as background with white border and border radius of 2px

[...]

PR #1514, opened 2021-09-14
Analytic collect counts to the audit db

carmenlau:
refs #1472

Due to performance considerations, I simplified the table schema to have type only and no tags. And the unique constraint is added. See if it looks good?

The counts to collect:
- [x] `total_number_of_user`
- [x] `monthly_active_user`
- [x] `weekly_active_user`
- [x] `daily_active_user`
- [x] `daily_signup`
- [x] `daily_signup_page_view`
- [x] `daily_signup_unique_page_view`
- [x] `daily_login_page_view`
- [x] `daily_login_unique_page_view`
- [x] `daily_signup_with_login_id_%s`
- [x] `daily_signup_with_oauth_%s`
- [x] `daily_signup_anonymously`


pkg/lib/analytic/count.go
```diff
@@ -0,0 +1,62 @@
...
+
+type Count struct {
+	ID    string
+	AppID string
+	Count int
+	Date  time.Time
+	Type  string
+}
+
+func NewCount(appID string, count int, date time.Time, _type string) *Count {
```
louischan-oursky, line 35: The convention is to use `typ`. `_type` looks like it is an unused variable.

cmd/portal/analytic_collect_count.go
```diff
@@ -54,11 +57,19 @@ var cmdAnalyticCollectCount = &cobra.Command{
...
 		case analytic.CollectIntervalTypeDaily:
-			fmt.Println(dbCredentials, auditDBCredentials)
-			return fmt.Errorf("TODO: collect analytic count daily")
+			log.Println("Start collecting daily analytic count")
+			yesterday := time.Now().UTC().AddDate(0, 0, -1)
+			err := countCollector.CollectDaily(yesterday)
+			if err != nil {
+				return err
+			}
+			log.Println("Done")
```
louischan-oursky, line 72: Can we make this command to support the following synopsis?

```
# Collect the count for yesterday
collect-count yesterday
# Collect the count for last week
collect-count last-week
# Collect the count for last month
collect-count last-month
# Collect the daily count for the specific date
collect-count 2021-09-14
# Collect the weekly count for the specific week
collect-count 2021W40
# Collect the monthly count for the specific month
collect-count 2021-09
```

And it should be possible to run `collect-count` multiple times. The command should be smart enough to upsert the count.

pkg/lib/analytic/count_collector.go
```diff
@@ -84,5 +101,145 @@ func (c *CountCollector) CollectDailyCountForApp(appID string, date time.Time, n
...
+			return nil, err
+		}
+
+		// Termination condition
+		if len(events) == 0 {
+			return result, nil
+		}
+
+		after = lastCursor
+		for _, e := range events {
```
louischan-oursky, line 178: I am thinking whether we should perform the aggregation directly in the SQL query.

louischan-oursky (commented):
Why do we leave out the tag in table schema?

pkg/lib/analytic/count_collector.go
carmenlau: After discussion, we will keep the original implementation. Since in the sql approach, will need to check all the login methods even for the apps that don't have much activities. 

May consider use the sql approach if there are apps have many activities in the future.

carmenlau:
Updated:
- Rename count type
- Support upsert count to db
- Support periodical argument in the collect-count command
- Collect daily count from redis
- Collect weekly and monthly count

WIP:
- Set expire to key in the redis

PR #1518, opened 2021-09-15
Fetch OIDC standard claims of OAuth providers

louischan-oursky:
ref #1513


pkg/lib/interaction/nodes/use_identity_oauth_user_info.go
```diff
@@ -85,14 +85,15 @@ func (e *EdgeUseIdentityOAuthUserInfo) Instantiate(ctx *interaction.Context, gra
...
+	providerID := providerConfig.ProviderID()
 	spec := &identity.Spec{
 		Type: authn.IdentityTypeOAuth,
 		Claims: map[string]interface{}{
 			identity.IdentityClaimOAuthProviderKeys: providerID.Claims(),
-			identity.IdentityClaimOAuthSubjectID:    userInfo.ProviderUserInfo.ID,
+			identity.IdentityClaimOAuthSubjectID:    userInfo.StandardAttributes.Sub(),
 			identity.IdentityClaimOAuthProfile:      userInfo.ProviderRawProfile,
-			identity.IdentityClaimOAuthClaims:       userInfo.ProviderUserInfo.ClaimsValue(),
+			identity.IdentityClaimOAuthClaims:       userInfo.StandardAttributes.ToClaims(),
```
carmenlau, line 96: In the OAuth Claims, the original data from `ProviderUserInfo.ClaimsValue()` doesn't have `sub`. But after update, `ToClaims()` returns all values which include `sub`. For data consistent, maybe we should 

1. Remove `sub` in the `ToClaims()`
2. Handle it in the migration task later, add back `sub` for the old data

louischan-oursky: Actually the `sub` claim is not very important, since the column `provider_user_id` is used instead. I prefer we handle data migration in a later task.

carmenlau (commented):
The PR looks good, tested most of the providers except `apple` and `adfs`. They work great!
Only login with facebook, the picture is removed during normalization. The url return from facebook failed the `CheckFormat` check as it contains trailing slash.

louischan-oursky:
Let me fix the URL normalization then.

louischan-oursky:
Updated.

PR #1520, opened 2021-09-17
Refactor sql builder

carmenlau:
Refactor SQLBuilder to SQLBuilder and SQLBuilderApp

So the stores have flexibility to use SQLBuilder and provide app id in runtime

### DIFF
--- a/pkg/lib/event/store.go
+++ b/pkg/lib/event/store.go
@@ -7,12 +7,12 @@ import (
 )
 
 type StoreImpl struct {
-	SQLBuilder  *appdb.SQLBuilderApp
+	SQLBuilder  *appdb.SQLBuilder
 	SQLExecutor *appdb.SQLExecutor
 }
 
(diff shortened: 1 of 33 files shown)


pkg/lib/event/store.go
```diff
@@ -7,12 +7,12 @@ import (
...
 
 type StoreImpl struct {
-	SQLBuilder  *appdb.SQLBuilder
+	SQLBuilder  *appdb.SQLBuilderApp
 	SQLExecutor *appdb.SQLExecutor
 }
 
 func (s *StoreImpl) NextSequenceNumber() (seq int64, err error) {
-	builder := s.SQLBuilder.Global().
+	builder := s.SQLBuilder.
```
louischan-oursky, line 15: This seems incorrect. Select the next value from the sequence should not be app-aware.

carmenlau: Updated!

Issue #1492, opened 2021-09-02
Force update non-compliant password after changing policy


fungc-io:
Case: After strictiening the password policy, old passwords are still valid and now there is no way to force the end-user to update the non-compliant password.

Add an option in Password Policy, so that Authgear will show a screen that ask users can’t comply with new password policy to update their password once they logged in

Alternatively:
Every password can optionally have an expiration time. If the password is used for authentication, after the authentication success, prompt (or force) the user to change the password. For this incident, just set the expiration to time in the past for those affected users.

fungc-io:
duplicate of #129

chpapa:
I think this issue is different from #129

louischan-oursky:
- Password policy applies to primary password and secondary password
- After verifying password, if the password violates the current policy, add a interaction step to force the end-user to update their password.

fungc-io:
Copy:
Same with `/settings/change_password` with the following changes:
> # Change Password
> 
> Your password doesn't comply with the requirements, please enter a new password.
> (remove current password field)

fungc-io:
(updated the above copy to sound more natural)

louischan-oursky:
Design option for disable/enable/skip ...

fungc-io:
Your password does not meet the requirements -> Your password does not meet the password policy requirements.

fungc-io:
Show option on the Portal in Password Policy as:
"Force the end-user to change password on next login if it does not meet the policy requirements"
default = checked

PR #1523, opened 2021-09-19
Force change password if current password does not meet the policy

louischan-oursky:
ref #1492


pkg/auth/handler/webapp/change_password.go
```diff
@@ -80,6 +118,39 @@ func (h *ChangePasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
 	})
 
 	ctrl.PostAction("", func() error {
+		if maybeWebSession != nil {
```
carmenlau, line 121: Seems we cannot use the existence of web session to determine this is from setting page / force changing password during authentication. For some cases in the setting page there is web session too. To reproduce the problem:
case 1. Submit wrong password in the change password setting page, web session will be started.
case 2. Do other actions before entering the change password setting page, for example, create totp.

Is it correct if we change to check the request path instead?

louischan-oursky:
Updated.

PR #1524, opened 2021-09-20
Analytic report active user

carmenlau:
- Breaking change: Update `analytic report` command with period flag and it is required. The commands will be like this,
        - `go run ./cmd/portal analytic report user --period=last-week`
        - `go run ./cmd/portal analytic report project --period=last-week`
        - `go run ./cmd/portal analytic report project --period=last-month`
- Add active users count to project weekly and monthly report


user-weekly-report.csv
```diff
@@ -0,0 +1,2 @@
+Year,Week,Range From,Range To,Number of new users,Number of new users who have created project
```
louischan-oursky, line 1: Is this file committed by mistake?

cmd/portal/analytic_report.go
```diff
@@ -146,14 +124,24 @@ var cmdAnalyticReport = &cobra.Command{
...
+		}
+
 		var data *analyticlib.ReportData
 		dbPool := db.NewPool()
 		switch reportType {
 		case analytic.ReportTypeUserWeeklyReport:
-			year, week, err := getISOWeek()
-			if err != nil {
-				return err
+			if periodicalType != periodical.Weekly {
```
louischan-oursky, line 141: I have another idea of  changing the CLI interface.

```sh
go run ./cmd/portal analytic report user --period=last-week
go run ./cmd/portal analytic report project --period=last-month
```

Then we eliminate the problem of duplicating the period in the report type argument.

pkg/lib/feature/verification/store_pq.go
```diff
@@ -35,7 +35,7 @@ func (s *StorePQ) scan(scn db.Scanner) (*Claim, error) {
 		&c.CreatedAt,
 	)
 	if errors.Is(err, sql.ErrNoRows) {
-		return nil, ErrClaimUnverified
+		return nil, nil
```
louischan-oursky, line 38: The value of `ErrClaimUnverified` is used in VerificationService. It seems that we have to return ErrClaimUnverified to signify the claim is not verified (or not found in the verified claim table).

carmenlau: Sorry, updated it accidentally, removed it:(

cmd/portal/analytic_report.go
carmenlau: Updated!

louischan-oursky:
Conflicted :(

carmenlau:
Rebased!

PR #1679, opened 2021-12-01
Simplify the specification of custom attributes

louischan-oursky:
@carmenlau Could you take a look on this if you have time?

### DIFF
--- a/docs/specs/user-profile/design.md
+++ b/docs/specs/user-profile/design.md
@@ -196,6 +196,23 @@ For example, if the developer wants to define a custom attribute of non-negative
 
 The UI control of it is a text field restricted to integers.
 
+#### Custom Attribute type `number`
+
+The custom attribute is of type number.
+
+Optionally, the developer can define the minimum and the maximum allowed value.
+
+For example, if the developer wants to define a custom attribute of non-negative number, they write
+
+```
+- pointer: /hourly_wage
+  type: number
+  minimum: 0.0
+  maximum: 100.0
+```
+
+The UI control of it is a text field restricted to numbers.
+
 #### Custom Attribute type `enum`
 
 `enum` is a string, but can only take one of the values defined by the developer.
--- a/docs/specs/user-profile/design.md
+++ b/docs/specs/user-profile/design.md
@@ -166,7 +166,7 @@ Each custom attribute MUST have an unique pointer and of one of the defined type
 The pointer of custom attribute MUST have exactly ONE level, as in `/this_is_one_level`.
 The pointer MUST also be non-empty, so `/` is not a valid pointer.
 
-Only `a-z`, `A-Z` and `_` character are allowed in the pointer of custom attribute.
+Only `a-z`, `A-Z`, `0-9` and `_` character are allowed in the pointer of custom attribute.
 
 The pointer MUST NOT conflict with the pointer of any standard attributes,
 so the developer CANNOT define a custom attribute with pointer `/email`.
--- a/docs/specs/user-profile/design.md
+++ b/docs/specs/user-profile/design.md
@@ -166,6 +166,8 @@ Each custom attribute MUST have an unique pointer and of one of the defined type
 The pointer of custom attribute MUST have exactly ONE level, as in `/this_is_one_level`.
 The pointer MUST also be non-empty, so `/` is not a valid pointer.
 
+Only `a-z`, `A-Z` and `_` character are allowed in the pointer of custom attribute.
+
 The pointer MUST NOT conflict with the pointer of any standard attributes,
 so the developer CANNOT define a custom attribute with pointer `/email`.
 
@@ -174,6 +176,10 @@ This means if the developer change the pointer, existing data WILL BE lost.
 
 The label of the custom attribute can be localized with the translation key `custom-attribute-label-{pointer}`.
 For example, if the custom attribute is `/x_email`, then the translation key is `custom-attribute-label-/x_email`.
+
+If the label of the custom attribute is not localized, a default label is generated based on the pointer.
+`_` are replaced by space character, and the first character of each word is capitalized.
+For example, the default label of `/job_title` is `Job Title`.
 
 #### Custom Attribute type `string`
 
--- a/docs/specs/user-profile/design.md
+++ b/docs/specs/user-profile/design.md
@@ -155,13 +155,24 @@ Here is an example of the configuration.
 user_profile:
   custom_attributes:
     attributes:
-    - pointer: /x_phone_number
+    - id: "0001"
+      pointer: /x_phone_number
       type: phone_number
-    - pointer: /x_email
+    - id: "0002"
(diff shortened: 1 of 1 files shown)


fungc-io:
I think we missed something and didn't discuss this before.
How will the end-user facing attribute labels be stored?
In the "Add New Claims" dialog,
Shall we allow `[a-zA-Z0-9 ._-]` in the `Field Name` and normalize it into `[a-z0-9._-]` for `pointer`?
and let the user provide translation in `translation.json`?

louischan-oursky:
I prefer we do not perform any processing on the field name. It will be used directly as the JSON pointer by treating it as the first level of the pointer. For example, if the developer enters `age`, then the pointer is `/age`. The translation of course should be provided in `translation.json`. The problem is that the current translation UI is not very friendly :(

docs/specs/user-profile/design.md
```diff
@@ -147,49 +147,98 @@ In addition to standard attributes, the developer can define their own custom at
...
+Each custom attribute MUST have an unique pointer and of one of the defined types.
+
+The pointer of custom attribute MUST have exactly ONE level, as in `/this_is_one_level`.
+The pointer MUST also be non-empty, so `/` is not a valid pointer.
+
+The pointer MUST NOT conflict with the pointer of any standard attributes,
+so the developer CANNOT define a custom attribute with pointer `/email`.
+
+NO data migration is ever done when the configuration is changed.
+This means if the developer change the pointer, existing data WILL BE lost.
```
carmenlau, line 173: How about changing the type and other validation config of an existing field?
We will do the validation during read/write or write only?

docs/specs/user-profile/design.md
```diff
@@ -147,49 +147,98 @@ In addition to standard attributes, the developer can define their own custom at
...
+#### Custom Attribute type `integer`
+
+The custom attribute is of type integer.
+Optionally, the developer can define the minimum and the maximum allowed value.
+
+For example, if the developer wants to define a custom attribute of non-negative integer, they write
+
+```
+- pointer: /x_age
+  type: integer
```
carmenlau, line 192: Will we support number in this stage? Or later?

carmenlau (commented):
The changes look great! 

Btw, for access control, it should be `admin_user` or `portal_ui`?

docs/specs/user-profile/design.md
louischan-oursky: It is possible to change the type if the underlying type is still the same. We validate the whole custom attributes object so if the underlying type has been changed, the validation will fail. 

For example, let say the developer initially define `a: integer` and `b: string`. Some users have `a` being set. The developer then change to `a: string`. The underlying type has been changed. So when a user attempts to change `b`, `a` will fail the validation.

An example of changing type is safe is type widening, like change from `a: enum` to `a: string`. It is because the old type is a subset of the new type so validation must succeed.

docs/specs/user-profile/design.md
louischan-oursky: Let me add `number` too.

docs/specs/user-profile/design.md
louischan-oursky: It is `portal_ui`.

carmenlau: > So when a user attempts to change b, a will fail the validation.

That means user will need to correct `a` before changing `b`. Right? 

There is a rare case that if it is type narrowing / changing (string to enum, string to number), the initial value will not be able to show in the ui and the ui field should be in empty state. User will need to fill in value in order to save the form. 

I think it is acceptable, given that type changing is not a common operation. Maybe add warning when developer changes the custom attributes config in the portal?

> It is portal_ui.

Please update the table `Exhaustive list of access control combination`:)

louischan-oursky: The editing UI for standard attribute is editing one standard attribute at a time. Custom attributes should follow this design. When the end-user wants to change `b`, they can only see the control for changing `b`. There is no way for them to be aware that `a` needs correction. Anyway, I think it should not be the responsibility of the end-user to correct `a` themselves.

Actually there is another case that can change the type of a custom attribute. It is by deleting the custom attribute definition first, followed by defining a new one with an incompatible type and the same name :(

carmenlau: > It is by deleting the custom attribute definition first, followed by defining a new one with an incompatible type and the same name.

yes, I thought of it too :(

> So when a user attempts to change `b`, `a` will fail the validation.

> The editing UI for standard attribute is editing one standard attribute at a time.

Then how can we resolve this case? Change the api to allow user to update one attribute and only that attribute will be validated? So `a` will keep the initial value when user change `b`. Or any other ideas?

louischan-oursky: Discussed with Fung, I am going to follow [the approach of Amazon Cognito](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html). Custom attributes, once defined, cannot be removed or changed. With this limitation introduced, this problem will be eliminated. I am going to update the spec to specify how this approach works with our own design.

louischan-oursky:
@carmenlau I further amended the spec to allow custom attributes to have their pointer changed.

docs/specs/user-profile/design.md
```diff
@@ -172,15 +183,18 @@ The pointer MUST NOT conflict with the pointer of any standard attributes,
...
+> a new custom attribute with the same ID but a incompatible type.
 > This will result in a situation that is very complicated to handle.
 > Given that it is extremely easy to define custom attribute with just a few clicks,
 > a few clicks should never lead to such a complicated scenario.
 
-Other than `pointer` and `type`, the developer is allowed to freely change other supplementary configuration of a particular type of custom attribute.
+The `pointer` can be changed as long as the developer is aware of the consequence of the rename.
+They have to be prepared for receiving custom attributes shown in the new pointer.
+
+The developer is allowed to freely change other supplementary configuration of a particular type of custom attribute.
```
carmenlau, line 197: How do we handle the case if the custom attribute had a negative value, and developer add `minimum: 0.0` to the config?

carmenlau (commented):
Using `id` looks good!

docs/specs/user-profile/design.md
louischan-oursky: During read, we do not perform validation so the read will succeed.
The input control will be filled with the invalid value.
When the value is written with the value just read, the write will fail.
So error will delay until the custom attribute is written.

carmenlau: If the field `a` fall into this situation, then user change field `b` value in the settings (given that the ui only allow update single field). Can user save the field `b` successfully?

louischan-oursky: Yes, correct. Though in the portal, the admin user will likely see all fields at the same time, they need to correct `a` in order to save everything successfully.

Issue #1685, opened 2021-12-03
Anonymous user support for websites


fungc-io:
Currently "Anonymous user" is only supported in mobile apps, we are extending this feature to support websites as well.
It should support both  `sessionType` `"refresh_token"` and `"cookie"`

Major change: Instead of using a private key to promote the end-user, on websites, due to the lack of key storage mechanism, the session itself will be used for promotion.

On websites, the anonymous user will be irrecoverable once the session becomes invalid, e.g. cookies expired, cookies were cleared, switched to another browser.

out of bounds: 
It's easier to go to a website than to install an app. While in an app, the expected logic will be creating an anonymous user upon first opening the app. If such behavior is adopted in a website, many orphan users will be created. We shall advise the devs to create anonymous users only when necessary. This is up to their app design and we will assume they will handle this issue by themselves.

TODO:
- [x] Update anonymous user specification
- [x] Change key-pair to optional in the anonymous user table
- [x] Add api for Web SDK to sign up new anonymous user
    - [x] Update anonymous user signup api, client_id should be optional when session type is cookie
- [x] Add api for Web SDK to request promotion code
- [x] Add promotion code support in the authorization endpoint
- [x] Add rate limit for creating new anonymous user
- [x] Update web SDK to support authenticate and promote anonymous user
- [x] Block anonymous user from visiting settings page
- [x] Update portal to show warning, the anonymous user lifetime will be the same as the session as re-login is not allowed

fungc-io:
In the Portal > "Anonymous Users", tips should be added about the session lifetime to remind the dev to avoid accidentally creating too much orphan users.


![Screenshot 2022-01-10 at 18 45 14](https://user-images.githubusercontent.com/15207470/148753695-dd0c3f4a-3c49-4ed0-bac7-93abc23d7122.png)

Add the following inside the pane of "Anonymous Users"

> ## User Lifetime
> 
> For **web applications**, a new anonymous user will be created when their previous session or token becomes invalid.
> 
> **Cookie-based applications**
> {idle timeout | session lifetime} seconds
> 
> **Token-based applications**
> 
> > (list of application)
>  > {application name}
> > {Refresh Token Idle Timeout | Refresh Token Lifetime} seconds
> 
> Change the timeout settings in the \[Applications]\(configuration/apps) page.

PR #1708, opened 2021-12-20
Update anonymous user specification to add web support

carmenlau:
ref #1685

### DIFF
--- a/docs/specs/api.md
+++ b/docs/specs/api.md
@@ -2,6 +2,28 @@
 
 Here are the endpoints that are not part of OIDC, but essential to provide some features of Authgear.
 
+If the api request is success, the response of the API will be a JSON body with key `result`.
+
+```json
+{
+  "result": { /* the result object */ }
+}
+```
+
+If the api request is failed, the response of the API will be a JSON body with key `error`.
+
+```json
+{
+  "error": {
+    "name": "Invalid",
+    "reason": "ValidationFailed",
+    "message": "invalid request body",
+    "code": 400,
+    "info": { /* the error info */ }
+  }
+}
+```
+
 ## /oauth2/challenge
 
 This endpoint is for requesting a one-time use, short-lived challenge.
@@ -20,7 +42,7 @@ The challenge is used in anonymous user and biometric authentication.
 }
 ```
 
-### Response Schema
+### Response Result Object Schema
 
 ```json
 {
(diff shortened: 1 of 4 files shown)


fungc-io (approved):

docs/specs/oidc.md
```diff
@@ -176,14 +176,18 @@ The following are recognized query parameters:
...
 For examples:
 - To login with email `user@example.com`:
     `https://authgear.com/login_hint?type=login_id&email=user%40example.com`
 - To login with Google OAuth provider:
     `https://authgear.com/login_hint?oauth_provider=google`
-- To signup/login as anonymous user:
+- To auth or promote an anonymous user with the jwt:
     `https://authgear.com/login_hint?type=anonymous&jwt=...`
+- To promote an anonymous user with the promotion code:
+    `https://authgear.com/login_hint?type=anonymous&pr=...`
```
louischan-oursky, line 189: Is `pr` the actual name we are going to accept?

carmenlau: It should be `promotion_code`, I corrected it!

louischan-oursky (approved):

carmenlau:
Updated the spec to describe the signup api behavior.

docs/specs/api.md
```diff
@@ -59,6 +59,20 @@ The challenge is used in anonymous user and biometric authentication.
...
+1. The api will check the cookie.
+1. If there is no logged in session, new user will be created. The server will return the `Set-Cookie` header and the result object will be an empty object.
+1. If the logged in user is an anonymous user, a success response will be returned. No new cookie header will be return and the result object will be an empty object.
+1. If the logged in user is a normal user, an error will be returned.
+
+For the `refresh_token` app,
+
+1. If the user has logged in, `refresh_token` will be included in the request.
+1. If there is no logged in session, new user will be created. `refresh_token` and `access_token` will be issued in the result object.
+1. If the logged in user is an anonymous user, a success response will be returned. New `access_token` will be issued in the result object.
```
louischan-oursky, line 73: I am thinking for both cookie and refresh token case, should we extend the lifetime of the session if the developer calls authenticateAnonymously again. If yes, this behavior should be stated here.

carmenlau: I think it is not very useful since if developer use the api correctly, they should only call `authenticateAnonymously` when there is no logged in session. What do you think?

PR #1710, opened 2021-12-23
Signup anonymous user without key api

carmenlau:
refs #1685

### DIFF
--- a/pkg/auth/handler/api/anonymous_user_signup.go
+++ b/pkg/auth/handler/api/anonymous_user_signup.go
@@ -37,7 +37,8 @@ var AnonymousUserSignupAPIRequestSchema = validation.NewSimpleSchema(`
 			"session_type": {
 				"type": "string",
 				"enum": ["cookie", "refresh_token"]
-			}
+			},
+			"refresh_token": { "type": "string" }
 		},
 		"required": ["client_id", "session_type"]
 	}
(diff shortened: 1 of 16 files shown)


pkg/auth/handler/api/anonymous_user_signup.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (p *AnonymousUserSignupAPIRequest) Validate(ctx *validation.Context) {
+	if !p.SessionType.IsValid() {
+		ctx.Child("session_type").EmitErrorMessage("unknown session type")
+	}
+}
+
+type JSONResponseWriter interface {
+	WriteResponse(rw http.ResponseWriter, resp *api.Response)
```
louischan-oursky, line 73: If we use JSONResponseWriter to write response, the response will be wrapped under the `result` key, which does NOT match the declared response schema.

pkg/util/httputil/json.go
```diff
@@ -77,6 +77,13 @@ type JSONResponseWriter struct {
 	Logger JSONResponseWriterLogger
 }
 
+func (w *JSONResponseWriter) WriteResponseWithCookies(rw http.ResponseWriter, resp *api.Response, cookies []*http.Cookie) {
```
louischan-oursky, line 80: This method is not very useful. The caller can just call UpdateCookie in a loop directly.

pkg/auth/handler/api/anonymous_user_signup.go
carmenlau: This is intended, since we had a convention for the api response, the success result should has key `result`. Updated the general API spec to mention it. see https://github.com/authgear/authgear-server/pull/1708

louischan-oursky:
I created a branch `anonymous-user` for this feature.

Issue #1709, opened 2021-12-21
Allow turn off "My Profile" Section in `/settings`


fungc-io:
Some developers implement their own profile management in the projects.
They are using authgear solely for authentication-related features.
Even if they hide all the attributes, the "My profile" is still there because of the "email", "username", and "phone_number" fields.
This is not desirable coz the developer never intended to utilize the "Profile" feature.

## Solution
Disabling the Profile feature to end-users should be an option.
On the standard attribute configuration page, add the atttributes "email", "username", and "phone_number"

If all of them are "Hidden" to end-users, the "My Profile" section should be hidden, and `/settings/profile` will be redirected to `/settings`

PR #1711, opened 2022-01-03
Respect identity aware standard attributes access control

louischan-oursky:
ref #1709


pkg/lib/config/user_profile.go
```diff
@@ -224,6 +221,18 @@ func (c *StandardAttributesConfig) GetAccessControl() accesscontrol.T {
 	return t
 }
 
+func (c *StandardAttributesConfig) IsEndUserAllHidden() bool {
+	if len(c.AccessControl) <= 0 {
```
carmenlau, line 225: After updating the `SetDefaults` implementation, the length of `AccessControl` will never be zero?

Issue #1705, opened 2021-12-17
2FA config turn on automatically


matthew-oursky:
The 2FA option comes back with no reason after changing other settings in the Authgear portal.

#### Environment
Staging

#### Expected Results
Changing other settings should not turn on 2FA.

#### Actual Results
2FA comes back with no reason.

#### Steps to reproduce
1. Disable all "Secondary Authenticator" options in "Authentication > Authenticators"
2. Change other settings in the Authgear portal.
3. See "Authentication > Authenticators > Authenticator Apps / Devices" option turned on with no reason.

louischan-oursky:
The root cause of this bug is due to the the behavior of `omitempty`. In this case, an omitted `secondary_authenticators` and an empty slice of `secondary_authenticators` mean different things, but `omitempty` treat both the same. Currently it is possible to set `secondary_authenticators` to an empty slice. However, when it is serialized, it is omitted by `encoding/json`, leading to the ambiguity.

Since fixing `secondary_authenticators` is impossible, I propose we add a new variant `disabled` to `secondary_authentication_mode` to turn off MFA explicitly.

@carmenlau @fungc-io Thoughts?

louischan-oursky:
I just realized we have an hidden, unimplemented option call `if_requested`. Let's take this change to remove `if_requested` and add `disabled`. What do you guys think?

louischan-oursky:
The issue will be solved in two phases.

In the first phase, the following changes are included:

- Add `disabled` as a new option
- Deprecate `if_requested`. It becomes equivalent to `disabled`. The option becomes hidden in the dropdown. Existing value is shown as `Disabled (Deprecated)`.
- (TBC) Revamp the configuration UI so that identity is more closely related to primary authenticators, while secondary authenticators are relevant only when mode is NOT `disabled`
- Add a re-runnable migration command to turn `if_requested` to `disabled`, and set mode to `disabled` if `secondary_authenticators` is an empty slice.

In the second phase, the following changes are included:

- Remove `if_requested`
- Do NOT allow empty slice in `secondary_authenticators`

fungc-io:
The `emitempty` bug mentioned in https://github.com/authgear/authgear-server/issues/1705#issuecomment-1003904462 found to be fixable.

In our discussion, we found the re-arrangement proposed in #1715 is not a good solution to improve IA and UX. Further re-design is needed to improve the configuration experience. Which should be put into the next big UI overhaul project.

[...]

PR #1713, opened 2022-01-03
[Do NOT merge] Add `disabled` and deprecate `if_requested` for secondary authentication mode

louischan-oursky:
ref #1705

### DIFF
--- a/portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
+++ b/portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
@@ -543,9 +543,9 @@ const SecondaryAuthenticationModeContent: React.FC<SecondaryAuthenticationModeCo
     const options: IDropdownOption[] = useMemo(
       () => [
         {
-          key: "required",
+          key: "disabled",
           text: renderToString(
-            "Onboarding.secondary-authentication-mode.required"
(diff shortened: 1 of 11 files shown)


portal/src/graphql/portal/OnboardingConfigAppScreen.tsx
```diff
@@ -543,9 +543,9 @@ const SecondaryAuthenticationModeContent: React.FC<SecondaryAuthenticationModeCo
     const options: IDropdownOption[] = useMemo(
       () => [
         {
-          key: "required",
+          key: "disabled",
           text: renderToString(
-            "Onboarding.secondary-authentication-mode.required"
+            "Onboarding.secondary-authentication-mode.disabled"
```
carmenlau, line 548: We may need to implement remove `secondary_authenticators` slices when user select `disabled`. Or we will implement it in the next clean up PR?

carmenlau: Also in the on-boarding page, should we hide the mfa question when user select disabled? Or we will do it in the next ui update PR?

louischan-oursky: I will do both of those in this PR, but still pending the new design of the identity/primary authenticators/secondary authenticators configuration screen.

PR #1716, opened 2022-01-04
Update deps 2022-01-04

louischan-oursky:


louischan-oursky:
@carmenlau Could you have a look on this?

carmenlau:
The changes look good to me, but when I want to try the portal with the bundled build. We had a setup that we can loaded the bundled build by visiting http://portal.localhost:8010, but the setup is broken after commit e68939d. We may want to revert the `nginx.conf` and `docker-compose.yaml` files?

Since the problem is not related this PR and the bundled build is also working, I will merge this PR first.

louischan-oursky:
That setup is working fine for me. I updated the `make vendor` make target in 17e9245f60fe138a85c8b36211c511b0f2da7a05

Try the following steps and see if they work

```
docker-compose down
docker-compose up -d
make vendor
make start
make start-portal
open http://portal.localhost:8010
```

PR #1726, opened 2022-01-10
Add signup anonymous user rate limit

carmenlau:
refs #1685

### DIFF
--- a/pkg/lib/interaction/nodes/do_create_user.go
+++ b/pkg/lib/interaction/nodes/do_create_user.go
@@ -2,6 +2,7 @@ package nodes
 
 import (
 	"github.com/authgear/authgear-server/pkg/api/apierrors"
+	"github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/lib/interaction"
 	"github.com/authgear/authgear-server/pkg/lib/ratelimit"
 	"github.com/authgear/authgear-server/pkg/util/httputil"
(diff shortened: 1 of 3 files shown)


pkg/lib/interaction/nodes/do_create_user.go
```diff
@@ -79,6 +80,26 @@ func (n *NodeDoCreateUser) GetEffects() ([]interaction.Effect, error) {
...
+			isAnonymous := false
+			for _, i := range newIdentities {
+				if i.Type == model.IdentityTypeAnonymous {
+					isAnonymous = true
+				}
+			}
+
+			if !n.BypassRateLimit && isAnonymous {
+				// check the rate limit only before running the effects
+				pass, _, err := ctx.RateLimiter.CheckToken(interaction.SignupAnonymousUserRateLimitBucket(ip))
```
louischan-oursky, line 94: Should we follow our convention to call CheckToken in Instantiate?

carmenlau: The cases for normal signup and anonymous user rate limit are a bit different.

For normal signup, we called `CheckToken` in `Instantiate` since we want to block the user as soon as we could and user will be blocked in the first step of the auth ui flow. So user could have better UX, it is annoying if user is blocked after completing the whole auth ui flow.

For signup anonymous user check, we use `graph.GetUserNewIdentities` to check if there is any anonymous identity. We need to ensure the node `NodeDoCreateIdentity` has been run, so we need to do the check in the `GetEffects` functions. Also signup anonymous user won't have multiple steps, check it in the `GetEffects` should not have UX problem.

Let me add some comments to the code.

PR #1729, opened 2022-01-11
Anonymous user settings page

carmenlau:
refs #1685 

Show placeholder settings page to anonymous user in /settings, and redirect all settings sub-routes to /settings

### DIFF
--- a/pkg/auth/webapp/settings_sub_routes_middleware.go
+++ b/pkg/auth/webapp/settings_sub_routes_middleware.go
@@ -1,0 +1,55 @@
+package webapp
+
+import (
+	"net/http"
+
+	"github.com/authgear/authgear-server/pkg/api/model"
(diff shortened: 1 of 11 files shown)


pkg/auth/webapp/settings_sub_routes_middleware.go
```diff
@@ -0,0 +1,51 @@
...
+					}
+				}
+				return nil
+			})
+			if err != nil {
+				panic(err)
+			}
+		}
+
+		if userID == nil || isAnonymous {
```
louischan-oursky, line 45: This middleware should not handle the case of userID is nil. That case should be handled by another middleware.

```go
userID := session.GetUserID(r.Context())
if userID == nil {
  next.ServeHTTP(w, r)
  return
}
// Continue our logic...
```

PR #1730, opened 2022-01-11
Add user lifetime description to the portal anonymous user setting page

carmenlau:
refs #1685


louischan-oursky (approved):
LGTM. Could you fix the build error?

Issue #1722, opened 2022-01-07
Respect post_login_redirect_uri when / is visited


louischan-oursky:
```diff
diff --git a/pkg/auth/webapp/auth_entry_point_middleware.go b/pkg/auth/webapp/auth_entry_point_middleware.go
index 233628746..a0953d37f 100644
--- a/pkg/auth/webapp/auth_entry_point_middleware.go
+++ b/pkg/auth/webapp/auth_entry_point_middleware.go
@@ -9,6 +9,7 @@ import (
 
 type AuthEntryPointMiddleware struct {
 	TrustProxy config.TrustProxy
+	UIConfig   *config.UIConfig
 }
 
 func (m AuthEntryPointMiddleware) Handle(next http.Handler) http.Handler {
@@ -23,7 +24,7 @@ func (m AuthEntryPointMiddleware) Handle(next http.Handler) http.Handler {
 		}
 
 		if userID != nil && !fromAuthzEndpoint {
-			defaultRedirectURI := "/settings"
+			defaultRedirectURI := DefaultPostLoginRedirectURI(m.UIConfig)
 			redirectURI := GetRedirectURI(r, bool(m.TrustProxy), defaultRedirectURI)
 
 			http.Redirect(w, r, redirectURI, http.StatusFound)
diff --git a/pkg/auth/wire_gen.go b/pkg/auth/wire_gen.go
index 36e93bfb3..089d01b13 100644
--- a/pkg/auth/wire_gen.go
+++ b/pkg/auth/wire_gen.go
@@ -28644,8 +28644,12 @@ func newAuthEntryPointMiddleware(p *deps.RequestProvider) httproute.Middleware {
 	rootProvider := appProvider.RootProvider
 	environmentConfig := rootProvider.EnvironmentConfig
 	trustProxy := environmentConfig.TrustProxy
+	config := appProvider.Config
+	appConfig := config.AppConfig
+	uiConfig := appConfig.UI
 	authEntryPointMiddleware := &webapp.AuthEntryPointMiddleware{
 		TrustProxy: trustProxy,
+		UIConfig:   uiConfig,
 	}
 	return authEntryPointMiddleware
 }
```

PR #1733, opened 2022-01-12
Respect default_redirect_uri in entrypoints

louischan-oursky:
ref #1722

### DIFF
--- a/pkg/auth/webapp/auth_entry_point_middleware.go
+++ b/pkg/auth/webapp/auth_entry_point_middleware.go
@@ -9,6 +9,7 @@ import (
 
 type AuthEntryPointMiddleware struct {
 	TrustProxy config.TrustProxy
+	UIConfig   *config.UIConfig
 }
 
 func (m AuthEntryPointMiddleware) Handle(next http.Handler) http.Handler {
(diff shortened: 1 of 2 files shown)


pkg/auth/webapp/auth_entry_point_middleware.go
```diff
@@ -23,7 +24,7 @@ func (m AuthEntryPointMiddleware) Handle(next http.Handler) http.Handler {
 		}
 
 		if userID != nil && !fromAuthzEndpoint {
-			defaultRedirectURI := "/settings"
+			defaultRedirectURI := DefaultPostLoginRedirectURI(m.UIConfig)
```
carmenlau, line 27: Just thinking if developer inputs the authgear url into the DefaultPostLoginRedirectURI, it will cause too many redirect here. Do you think it is necessary to check and avoid developer to have such config?

louischan-oursky: To detect infinite redirection, we need to maintain a separate list of routes that will go through AuthEntryPointMiddleware, in addition to adding the routes in `pkg/auth/routes.go`. We also need to parse the redirect URI. First we need to check if the host is custom domain or default domain. If it is neither, then there is properly no infinite redirection. However, it is possible that the redirection first goes to `https://myapp.com/redirect` and then `https://myapp.authgearapps.com/login`, which we can only detect by introducing some state to keep track of the redirection.

Issue #1739, opened 2022-01-17
Terminate all sessions when a user is disabled


louischan-oursky:
- [x] Terminate all sessions when user is disabled
- [x] Remove the code that check if the user is disabled in session validating middleware

PR #1740, opened 2022-01-17
Terminate all sessions when user is disabled

louischan-oursky:
ref #1739


pkg/lib/oauth/handler/handler_token.go
```diff
@@ -275,16 +269,10 @@ func (h *TokenHandler) parseRefreshToken(token string) (*oauth.Authorization, *o
 		return nil, nil, err
 	}
 
+	// We used to check if the user is disabled here.
+	// But now disabling user will terminate all sessions.
+	// So the checking is no longer needed.
```
carmenlau, line 274: Removing the checking may affect the existing app? Since terminating all session when disabling a user is newly introduced, so the old disabled user may hold a refresh token which has not been deleted.

louischan-oursky: `pkg/lib/session/middleware.go:50` will remove the session cookie. So previously when our code still have the checking, it will result in ErrInvalidSession and the session cookie will be removed. The SDK will also remove the refresh token if it encounters that error.

When we delete a user, we will also remove all sessions. So this only affects disabled users, and only those who are inactive when they were disabled.

louischan-oursky: I dropped the commits involving the code removal. Instead, I added a commit to align the session checking behavior.

PR #1749, opened 2022-01-19
Add `disabled` and deprecate `if_requested` for secondary authentication mode

### LINKED ISSUE #1748: [Portal] Misleading description in Password Policy
One change to be kept from #1715 ,
The description in the Password Policy page is misleading as the policy applies to both primary and additional password.

Change from:
> Configure the password policy if Password is configurate as primary authenticators.

to: 
> Configure the password policy for Password and Additional Password in MFA.

carmenlau:
ref #1705 
fix #1748


louischan-oursky:
According to our discussion, do we need to move Password Policy to be under Authentication? Also "Enable Password as Authenticator" seems duplicates with our primary authenticator list.

fungc-io:
No, we are keeping "Password Policy" as is as a standalone page. Keep the primary authenticators list and  do not introduce the "Enable Password as Authenticator" toggle.

carmenlau:
Oh, sorry seems I submitted a wrong branch 🙈
#1750 is the correct one, I will recreate a PR that doesn’t include the cleanup later 🙇🏻‍♀️ Closing this one first

Issue #1742, opened 2022-01-18
Introduce a new command to run periodic jobs


louischan-oursky:
The first use case is to scan for mature scheduled account deletion and perform the deletion.

This is a new command because this daemon is supposed to have one instance only, as opposed to other commands like `authgear start` which is horizontally scalable.

The name of the new command is `authgear background`.

Each periodic job runs in its own goroutine. If we have more jobs in the future, more goroutine is created.
Currently there is only one job, which is the scheduled account deletion.

---

The periodic job is not time.Timer based because if the job runs longer than the tick period, the job will keep running.
[Demonstration](https://go.dev/play/p/pM1OW9BSxp9)

Instead the periodic job is time.Sleep based so that it is guaranteed that the job pauses for some duration.
[Demonstration](https://go.dev/play/p/NzsVAic5GT4)

---

- [x] Add columns `is_deactivated` and `delete_at` to user; Add index on `delete_at`
- [x] Implement a long-running goroutine to look for mature `delete_at` and delete user
  - This task may need to figure out how to initialize the user service **without** an incoming HTTP request; Refactoring may be needed

PR #1752, opened 2022-01-20
Add command `authgear background` to process scheduled account deletions

louischan-oursky:
ref #1742


carmenlau (approved):
The PR looks good to me. Seems we still need to inject the request, so commit 023a535 is for tidy up only?

louischan-oursky:
We still need to inject request for reading or writing cookie. Quite hard to remove request from dependency entirely.

PR #1756, opened 2022-01-24
Support account deletion in settings

louischan-oursky:
ref #1755


pkg/auth/handler/webapp/settings_delete_account.go
```diff
@@ -19,16 +22,31 @@ func ConfigureSettingsDeleteAccountRoute(route httproute.Route) httproute.Route
 		WithPathPattern("/settings/delete_account")
```
carmenlau, line 25: Better to block the account deletion flow if it is disabled in the config

resources/authgear/templates/zh-TW/translation.json
```diff
@@ -62,8 +62,10 @@
   "error-webhook-delivery-timeout": "因Webhook逾時，操作已被禁止",
   "error-disabled-user": "你的帳號已被管理員停用。",
   "error-disabled-user-reason": "原因： {reason}",
-  "error-disabled-user-no-reason": "請向管理員了解詳情。",
-  "error-blocked-user": "你的帳號已被禁止登入。",
+  "error-deactivated-user": "你已停用你的帳戶。",
+  "error-scheduled-deletion-by-admin": "你的帳戶已被管理員停用，而你的帳戶將在{date, date, long}被刪除。",
+  "error-scheduled-deletion-by-end-user": "你已停用你的帳戶，而你的帳戶將在{date, date, long}被刪除。",
```
carmenlau, line 67: Maybe we should use the JS code to convert the date to the user's timezone?

louischan-oursky: I just realized the `data-date` is for relative datetime formatting. Relative datetime formatting is not a good choice here because we should report the exact date on which the account will be deleted.

For legacy reason, let's keep `data-date` for relative date formatting and introduce `data-exact-date` for absolute date (without time) formatting.

PR #1758, opened 2022-01-25
Configure account deletion

louischan-oursky:
ref #1757


carmenlau:
Merged to branch `account-deletion` via commit 9135421

PR #1763, opened 2022-01-26
Fix admin adding identity to anonymous user

### LINKED ISSUE #1762: Fix adding identity to anonymous user in admin portal
Should remove the anonymous user identity and the user will become a normal user

carmenlau:
fix #1762 

- Remove anonymous identity when admin add identity to anonymous user 
- Trigger promote non-blocking event
- Fix not able to add email / phone through portal in the follow scenarios
    - when OOB OTP is disabled 
    - when OOB OTP is enabled but priority is lower than password


louischan-oursky:
We now fire `identity.x.added` and `user.anonymous.promoted` at the same time if the user is promoted via admin API. Should we omit `identity.x.added` in this case?

PR #1768, opened 2022-02-11
Analytic portal part 1

carmenlau:
refs #1472 

- no. of active user chart
- no. of total user chart

Todo:
- signup conversion
- user signup methods


portal/package.json
```diff
@@ -46,6 +46,7 @@
     "@monaco-editor/react": "3.7.2",
     "@oursky/react-messageformat": "2.0.2",
     "base64-js": "1.5.1",
+    "chart.js": "^3.7.0",
```
louischan-oursky, line 49: Let's stick with our convention here to use exact version.

Issue #1784, opened 2022-02-14
[User profile, Settings] '0' is not shown for integer/number custom attribute in profile preview


stanleylhs:
ux/minor

## Background:
- Device / Browser (OS version): Chrome on win 10
- Version Tag: 2022-02-08.0
- Reported by: Jasonns 

## Description:
No values are shown for integer/number fields with 0 as their contents, which may confuse users with those fields that have literally no values.

## Steps to reproduce:
1. Log in to the app portal > Go to the Settings page
2. Edit the integer/number fields in the profile and set it to zero
3. See no values are shown in the profile preview page (though values are retrieved when using UserInfo api)
4. Remove the values of those fields
5. See no values are shown in the profile preview page, while no values are retrieved when using UserInfo api

## Expected result:
Should show the assigned values for those profile fields if '0' value is assigned

## Screenshot:

https://user-images.githubusercontent.com/4337525/153841939-040f6173-7aff-4728-9312-3a232bb660c1.mp4

louischan-oursky:
Problem of zero value. We need to introduce `isNil` to check whether a value is nil explicitly.

PR #1800, opened 2022-02-15
Fix displaying custom attribute zero value in the settings page

carmenlau:
refs #1784

### DIFF
--- a/pkg/util/template/func_test.go
+++ b/pkg/util/template/func_test.go
@@ -27,4 +27,19 @@ func TestRFC3339(t *testing.T) {
 			So(RFC3339([]struct{}{}), ShouldEqual, "INVALID_DATE")
 		})
 	})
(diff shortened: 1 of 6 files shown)


pkg/util/template/func_test.go
```diff
@@ -27,4 +27,21 @@ func TestRFC3339(t *testing.T) {
 			So(RFC3339([]struct{}{}), ShouldEqual, "INVALID_DATE")
 		})
 	})
+
+	Convey("IsNil", t, func() {
+		Convey("it supports time.Time", func() {
```
louischan-oursky, line 32: This Convey call can be removed. The label is irrelevant anyway.

carmenlau: Removed 🙇🏻‍♀️

Issue #1818, opened 2022-02-21
[Audit log] Incorrect user ID in the summary section of log details page


stanleylhs:
bug/minor

## Background:
- Device / Browser (OS version): Chrome on win 10
- Build: Version Tag: 2022-02-08.0
- Reported by: Jasonns

## Description:
The user ID shown in the summary section is the same for all log details pages.

## Steps to reproduce:
1. Log in to the Authgear portal and select any project
2. Go to the 'Audit Log' page and select any audit log with user ID
3. See the user ID shown on the summary section is not the same as the one in the raw event log

## Expected result:
Should show the correct user ID in the summary section

## Screenshot:
<img width="1099" alt="image" src="https://user-images.githubusercontent.com/4337525/154935207-68ba87c1-381b-4db8-8e0e-484ff7bd758d.png">

PR #1828, opened 2022-02-22
Display raw user id in the audit log entry page

carmenlau:
ref #1818


fungc-io:
Could you explain to me what the `VXNlcjp...` ID is? Why are there two different IDs? 🙏

louischan-oursky:
In GraphQL, every entity must have a unique ID across all types. A common way to do that is to `BASE64(TYPE_NAME ":" ID)`. For example, it is common to use sequence number to generate ID. So a Product of ID `1` conflicts with a Order of ID `1`. Under the common scheme, the GraphQL ID of that product becomes `BASE64("Product:1") = UHJvZHVjdDox`

fungc-io:
@louischan-oursky 
got it, thanks for your explanation!

Issue #1819, opened 2022-02-21
[User management, Profile] Missing account name in the profile header


stanleylhs:
bug/minor

## Background:
- Device / Browser (OS version): Chrome on win 10
- Build: Version Tag: 2022-02-08.0
- Reported by: Jasonns

## Description:
The account name is not shown in the profile header, though in the design the account name is also shown.

## Steps to reproduce:
1. Log in to the Authgear Portal and select any project
2. Go to the User management page and view the profile page of any user with name (Given name + family name)
3. See only email/username is shown in the profile header, but not the given name + family name

## Expected result:
Should also show the name + family name in the profile header

## Screenshots:
![image](https://user-images.githubusercontent.com/4337525/154936033-6d063008-3a36-467c-b894-c829c4e3410f.png)
design:
![image](https://user-images.githubusercontent.com/4337525/154936099-3cda8f54-f851-43ed-b228-350cd7a46909.png)

app portal:
![image](https://user-images.githubusercontent.com/4337525/154936084-9e0c9255-5a34-456d-a1fa-50eda88730ef.png)

PR #1829, opened 2022-02-22
Show user formatted name in the user profile detail page

carmenlau:
refs #1819


carmenlau:
@louischan-oursky I added user's formatted name in  the admin api, see if it looks good for you?

Issue #1810, opened 2022-02-21
[Account deletion] Can reach "delete account" success page via direct URL access


stanleylhs:
(Describe the issue here)

- Authgear Server Date/Version:
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Steps to reproduce

1.
2.
3.

stanleylhs:
Same with forget password success page

AppBaseURL/reset_password/success

<img width="641" alt="image" src="https://user-images.githubusercontent.com/4337525/154908773-5b6faaa9-d0aa-447b-8c1e-fdac35547610.png">

PR #1831, opened 2022-02-23
Allow access successful pages once only

carmenlau:
refs #1810

### DIFF
--- a/pkg/auth/handler/webapp/settings_delete_account.go
+++ b/pkg/auth/handler/webapp/settings_delete_account.go
@@ -9,6 +9,7 @@ import (
 	"github.com/authgear/authgear-server/pkg/auth/webapp"
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/lib/session"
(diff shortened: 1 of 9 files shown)


louischan-oursky (commented):
I have another idea. What about we introduce a new middleware for those success pages. The idea is similar to FlashMessage

```golang
type FlashPage struct { ... }
func (f *FlashPage) Flash(r *http.Request, rw http.ResponseWriter) {
  // Set a cookie to indicate any success page can be reached in subsequent request.
  // The cookie should have a max-age of 1 minutes.
}

func (f *FlashPage) Clear(r *http.Request, rw http.ResponseWriter) {
  // Clear the cookie unconditionally.
  // Intended to be called by the middleware
}

func setupRoutes() {
  webappSuccessPageChain := httproute.Chain(
    webappPageChain,
    webapp.SuccessPageMiddleware{},
  )
  webappSuccessPageRoute := httproute.Route{Middleware: webappSuccessPageChain}
  // Configure the success pages to use the new route
}

// In the handler that will redirect to success page, call FlashPage.Flash
```

Issue #1827, opened 2022-02-22
Azure B2C integration


fungc-io:
- Support Azure B2C via OAuth2/OIDC
- Support single policy
- https://docs.microsoft.com/en-us/azure/active-directory-b2c/openid-connect
- https://docs.microsoft.com/en-us/azure/active-directory-b2c/tokens-overview#claims
- Update docs.authgear.com

fungc-io:
Discussed yesterday and @louischan-oursky mentioned that:

The `email` claims is not guaranteed from the IdP Azure B2C, and it's not enabled by default.
However it is recommended for the dev to turn that on to avoid collision with the email user ID in Authgear. 
Our docs should guide the dev to turn on the `email` claims in for the "policy" Azure B2C dashboard.

In the default configuration, Authgear will attempt to get the `email` value upon a login of end-user with Azure B2C, an ID token without the `email` claim will result in a failure message reminding the developer of the misconfiguration.

Note that although rare, there are possible scenarios that the `email` claim is irrelevant. Therefore in the config pane of Azure B2C connection in the Portal, we will add a checkbox to opt to ignore the claim, and thus changing the aforementioned behaviour.

@louischan-oursky @carmenlau What do you think about this copy?

> (checkbox - default on) Obtain the email attribute from connection

fungc-io:
Please update the doc and add a page regarding the integration of azure ad b2c

PR #1834, opened 2022-02-25
Support Azure AD B2C

louischan-oursky:
ref #1827


carmenlau (commented):
The PR looks good to me, only one minor question. 

When I edit the SSO setting in the portal but not touching any `claims.email.assume_verified` or `claims.email.required` checkbox, those fields will have value set in the config. Is it expected?

carmenlau (approved):

louischan-oursky:
> When I edit the SSO setting in the portal but not touching any claims.email.assume_verified or claims.email.required checkbox, those fields will have value set in the config. Is it expected?

Our config mechanism is not smart enough to default values (or structs) and eliminate them :(

PR #1838, opened 2022-02-28
Add command `authgear import`

louischan-oursky:


cmd/authgear/importer/importer.go
```diff
@@ -0,0 +1,189 @@
...
+				userID,
+				passwordHash,
+			).Suffix("ON CONFLICT (id) DO UPDATE SET password_hash = EXCLUDED.password_hash"),
+		}
+
+		err = i.Handle.WithTx(func() (err error) {
+			for _, stmt := range insertStmts {
+				_, err = i.SQLExecutor.ExecWith(stmt)
+				if err != nil {
+					return
```
carmenlau, line 164: We return nil here instead of error, is it because we want to skip the problematic row rather than stop the whole import?

It would be better if the importer can print out a simple import summary like number of successful import, the problematic row number? So that when there are any problems during import, we can spot the problematic data easily.

cmd/authgear/importer/importer.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return err
+	}
+
+	headerSkipped := false
+	r := csv.NewReader(f)
+	for {
+		record, err := r.Read()
+		if errors.Is(err, io.EOF) {
+			break
```
carmenlau, line 40: If we are going to have report summary, we need to add this row to the problematic rows.

cmd/authgear/importer/importer.go
louischan-oursky: We do return error here. `err` is declared in the function signature.

louischan-oursky: A transaction is opened per row in the CSV. The program stops as soon as it encounter an error. I agree that we should skip problematic row. I suggest we print the problematic record to stderr, and also print a summary to the stdout at the end of the import.

louischan-oursky:
Updated.

Issue #1848, opened 2022-03-02
Transform standard attribute picture from authgearimages:// to https://


carmenlau:

louischan-oursky:
@carmenlau I have assigned this to myself.

PR #1849, opened 2022-03-02
[WIP] Transform picture attribute from authgearimages:// to https://

carmenlau:
refs #1848

Completed transforming picture attribute from `authgearimages://` to `https://`, but two problems are find.

1. Now we cannot set `authgearimages:///OBJECT_ID` to the picture, as the validation requires the uri to be absolute that means the host cannot be empty.
    1. Add a new validation format that allows empty host and use it in picture attribute
    2. Or we add a dummy host to the authgear images url. e.g. `authgearimages://host/OBJECT_ID`
1. When I set the picture url to `authgearimages://host/OBJECT_ID` in the portal, the url will be transformed to `https`. When I click save again, the `https` value will be saved and the `authgearimages` url cannot be preserved.
    1. Maybe when we update an attribute, we run the processor on the original value and check if it has been changed. If the value is the same, don't update that attribute?

### DIFF
--- a/cmd/authgear/background/wire_gen.go
+++ b/cmd/authgear/background/wire_gen.go
@@ -347,12 +347,19 @@ func newUserService(ctx context.Context, p *deps.BackgroundProvider, appID strin
 		ClaimStore:        storePQ,
 		RateLimiter:       limiter,
 	}
+	request := NewDummyHTTPRequest()
+	imagesCDNHost := environmentConfig.ImagesCDNHost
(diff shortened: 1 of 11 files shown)


louischan-oursky:
- I prefer 1.i.
- A note on the change checking, if the URL is changed slightly, for example from `https://authgearappsimages.com/app/object/varianta` to `https://authgearappsimages.com/app/object/variantb`, we will be confused that the new URL is a non-authgearimages URL :(

carmenlau:
After discussion, we will update the processor to do both transform and reverse, so when user update the attribute with `https` url, we will reverse it back to the `authgearimages` url.

PR #1858, opened 2022-03-08
Implement images server GET endpoint

louischan-oursky:
ref #1845


pkg/images/deps/provider.go
```diff
@@ -38,11 +41,16 @@ func NewRootProvider(
 		sentry.NewLogHookFromHub(sentryHub),
 	)
 
+	// We do not have a chance to close this yet :(
+	// But it is not harmful not to close this.
+	vipsDaemon := vipsutil.OpenDaemon(runtime.NumCPU())
```
carmenlau, line 46: How about we open daemon it in the controller `./cmd/authgear/images/server/server.go` and pass it to the RootProvider constructor?

carmenlau (approved):

carmenlau:
@louischan-oursky How about we open the vipsDaemon in the controller (`./cmd/authgear/images/server/server.go` ) and pass it to the RootProvider constructor? The current PR looks good to me, i will merge it first.

PR #1867, opened 2022-03-09
Add images cors middleware

carmenlau:
refs #1862 

~I just realized that we cannot use the original configsource implementation in the images server, as it resolves the app by domain. If we enabled CDN, the domain will become the CDN domain. So we need to update to resolve the app by path:(
You can still use this for development locally.~

Updated to support resolve app id by path

The portal will need the CORS middleware


louischan-oursky:
In this PR https://github.com/authgear/authgear-server/pull/1872 I have changed the local development setup so CORS is not required.

Issue #1871, opened 2022-03-10
Upload and remove profile picture in Auth UI


louischan-oursky:
- [x] Make form submission button disabled while loading
- [x] Show loading indicator when form is being submitted
  - [x] fetch does not support upload progress event; Will switch to axios
- [x] The final image size is at most 1024x1024.
- [x] Confirmation when removing image

PR #1872, opened 2022-03-10
Upload and remove standard attribute picture

louischan-oursky:
ref #1871


carmenlau:
As there is image upload, better to have loading state for the button.

PR #1873, opened 2022-03-11
Support authgearimages URI

louischan-oursky:
ref #1848


pkg/lib/feature/stdattrs/transformer_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+		Convey("transform from representation form to storage form", func() {
+			test := func(repr string, storage string) {
+				actual, err := t.RepresentationFormToStorageForm(stdattrs.Picture, repr)
+				So(err, ShouldBeNil)
+				So(actual, ShouldEqual, storage)
+			}
+
+			test("https://authgearappscdn.com/_images/app/objectid/profile", "authgearimages:///app/objectid")
+			test("https://authgearappscdn.com/_images/app/objectid/", "authgearimages:///app/objectid")
```
carmenlau, line 71: Currently in the spec, the `authgearimages` url  is `authgearimages:///objectid` but not `authgearimages:///app/objectid`. But I think it is ok to have the app id in the url, I will update the spec and upload url for that

PR #1880, opened 2022-03-15
Improve form submission UX

louischan-oursky:
ref #1871

### DIFF
--- a/authui/src/form.ts
+++ b/authui/src/form.ts
@@ -1,4 +1,5 @@
 import Turbolinks from "turbolinks";
+import axios, { Method } from "axios";
 import { disableAllButtons } from "./loading";
 
 // Handle click link to submit form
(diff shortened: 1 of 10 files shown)


authui/src/form.ts
```diff
@@ -78,42 +74,43 @@ export function xhrSubmitForm(): () => void {
...
+        method: form.method,
+        headers: {
+          "Content-Type": "application/x-www-form-urlencoded;charset=UTF-8",
+          "X-Authgear-XHR": "true",
+        },
+        credentials: "same-origin",
+        body: params,
       });
+      if (resp.status < 200 || resp.status >= 300) {
+        setServerError();
```
carmenlau, line 89: Seems we should call revert here as well.

carmenlau (approved):

PR #1883, opened 2022-03-15
Adjust the cropped image size to at most 1024x1024

carmenlau:
refs #1871 

I found that there are options `maxWidth` and `minWidth` for `getCroppedCanvas` which is suitable for our case, see the [doc](https://github.com/fengyuanchen/cropperjs#getcroppedcanvasoptions). When I tried using the placeholder images, it works as expected. But when I tried a very large image and crop an area which should be larger than 1024x1024, I expected I should be able to get an image with 1024x1024, but the result image is smaller. No idea yet :(

The placeholder images:

The large images:
153853716-065103b0-f4f8-4cfb-9fb5-d857b9e2fe09

### DIFF
--- a/authui/src/imagepicker.ts
+++ b/authui/src/imagepicker.ts
@@ -86,9 +86,34 @@ function onClickSave(e: Event) {
     return;
   }
 
+  const maxDimensions = 1024;
+  const dimensionsOptions = (function () {
+    const imageData = cropper.getImageData();
+    const cropBoxData = cropper.getCropBoxData();
+    // assume the cropped area is square
+    if (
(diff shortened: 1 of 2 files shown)


louischan-oursky:
https://github.com/fengyuanchen/cropperjs/blob/main/src/js/methods.js#L681
https://github.com/fengyuanchen/cropperjs/blob/main/src/js/utilities.js#L613

I haven't looked into the details of the source code. At first sight, I guess `width` and `height` should do the job already.

carmenlau:
Updated to calculate the dimensions by ourselves

Issue #1674, opened 2021-11-30
The default UI is confusing for newly created project


fungc-io:
For a new signup and newly created project, 
after clicking the "Try it out now!" button,
the first thing the user sees is a page identical to the Authgear sign up page with the heading "Sign up to Authgear"
It's very confusing because it's hard to understand from the UI that it's their tenant login page.

**Solution**
- Change the default `app.name` to the tenant id 
- Do not show the big Authgear Logo if the logo is not set
  - In the Signup/Login page, if logo is not set, don't show logo and keep the top and bottom padding at 16px
  - In the settings page, don't show logo and use 0px for top and bottom paddings

fungc-io:
@carmenlau @louischan-oursky 
Updated logos can be found here:
https://www.figma.com/file/swXJYkdXuPsLyA5q0aHu1r/Authgear-UI---Fung?node-id=54%3A2007

PR #1886, opened 2022-03-16
Update new app default UI

carmenlau:
refs #1674 

- Allow no logo by default 
- Use app id as the default app name when creating new app

Important notes:
- When deploying this change, we need to upload the logo for the `accounts` app


louischan-oursky (commented):
The padding does not seem to align with the description in the issue

carmenlau:
Sorry, stupid mistake, updated 🙈

louischan-oursky:
I added a commit to update the portal logo as well.

Issue #1825, opened 2022-02-22
[Portal] Better user identification in the "User Management" list


fungc-io:
## Problem
On the user list in the User Management page of the Portal. The end-users are currently identified by the `sub`, which is not optimal for human.

## Objective
As we are already storing the profile pic and names in the profile. These can be shown on the list as a better identification.

## Appetite
0.5 weeks

fungc-io:
Design:
https://www.figma.com/file/swXJYkdXuPsLyA5q0aHu1r/Authgear-UI---Fung?node-id=4%3A243

![authgear-user-management-list](https://user-images.githubusercontent.com/15207470/155278753-ff2297b4-e976-45ea-8010-78e1bded3474.png)

PR #1910, opened 2022-03-24
Show formattedName and picture in UsersList

louischan-oursky:
ref #1825


carmenlau (commented):
In most of the cases, the formatted name are missing for the users which makes the ui looks like the user id misalign. Should we center align the id when the name is missing? 

Screenshot 2022-03-29 at 12 13 05 PM

fungc-io:
Can we show the Login ID in the formatted name field, e.g. email, user id or phone number when the name is not available?
Similar to the "My Profile" in `/settings`

carmenlau:
And show "Anonymous user" label for anonymous user.

Issue #1856, opened 2022-03-08
Improve duplicated identity error message


louischan-oursky:
Suppose there is an account with email login ID `user@gmail.com`. When the end-user tries to sign in with Google with `user@gmail.com`, they will see an error message like `This email address has been registered. Maybe you want to sign in?`.
The action the end-user actually needs to take is to sign in with their email + password first, and then connect their Google account. The error message should point this out clearly.

fungc-io:
## Current Messages

1. User already signed up with Google and try to login with the same email address

> The account vou entered cannot be found. Please ensure the account is correct

<img width="1225" alt="Screenshot 2022-03-16 at 15 00 29" src="https://user-images.githubusercontent.com/15207470/158540912-e20d8668-b4c3-4568-a751-8fa7a6fa6737.png">

2. User already signed up with email, and try to login with Google account of the same email address

> Email already exists, maybe you want to login?

<img width="1226" alt="Screenshot 2022-03-16 at 14 59 24" src="https://user-images.githubusercontent.com/15207470/158541180-31c286f4-08b8-4169-8cad-83e3516f01fa.png">


## Change to

Suggest to change the error messages to:

1.
> This email is associated with an account. Do you want to login with Google?

2.
> This email is already registered. You can link it with your Google account via the setting page after login.

PR #1914, opened 2022-03-28
Improve sign in error message

louischan-oursky:
ref #1856


pkg/lib/interaction/nodes/do_update_identity.go
```diff
@@ -47,7 +47,9 @@ func (n *NodeDoUpdateIdentity) GetEffects() ([]interaction.Effect, error) {
 		interaction.EffectRun(func(ctx *interaction.Context, graph *interaction.Graph, nodeIndex int) error {
 			if _, err := ctx.Identities.CheckDuplicated(n.IdentityAfterUpdate); err != nil {
 				if errors.Is(err, identity.ErrIdentityAlreadyExists) {
-					return n.IdentityAfterUpdate.FillDetails(interaction.ErrDuplicatedIdentity)
+					s1 := n.IdentityBeforeUpdate.ToSpec()
+					s2 := n.IdentityAfterUpdate.ToSpec()
+					return identityFillDetails(interaction.ErrDuplicatedIdentity, &s1, &s2)
```
carmenlau, line 52: Technically, the incoming identity should be ` s2`, and we don't need the existing identity for this case?

louischan-oursky:
I rewrote the error message composition logic.

louischan-oursky:
Fixed.

PR #1941, opened 2022-03-30
Support searching user by oauth provider user id

carmenlau:
refs #1909


louischan-oursky:
Technically it is a breaking change because the schema of the index has been updated. Should we aggregate this change and the subsequent changes (search by standard attributes) into a new branch?

Issue #1928, opened 2022-03-29
[Profile pic] Shows "device is offline" error when the error is actually "too many request"


anthonyfhj:
ux/major

## Background:
- Device / Browser (OS version): Chrome 99.0.4844.82/Firefox 98.0.1/Edge 99.0.1150.46 on Win 10
- Build: Staging, demo-enterprise-25032022, demo-webapp-cookie-25032022, demo-webapp-refresh-token-250322
- Version Tag: staging-2022-03-01.1
- Reported by: Jasonns

## Description:
The rate limit of uploading profile pictures is configured to 10 photos per hour per user. If the user attempts to upload and save a profile picture again after the quota is exceeded, a 'device is offline' error will be shown even though the error is actually 'too many requests'.

## Steps to reproduce:
1. Log in to the demo-enterprise app
2. Edit personal profile > Edit profile picture
3. Upload and save the profile pictures until all quotas are taken
4. See the 'device is offline' error, though the error is actually 'too many requests' according to the console log

## Expected result:
Should display correct error messages

## Screenshot:

https://user-images.githubusercontent.com/93981572/160550793-a2c3e2f9-b679-42e8-93ab-da05596cf18d.mp4

PR #1949, opened 2022-03-31
Handle ajax error more robustly

louischan-oursky:
ref #1928

### DIFF
--- a/authui/src/error.ts
+++ b/authui/src/error.ts
@@ -1,0 +1,47 @@
+import { AxiosResponse } from "axios";
+
+export function handleAxiosError(e: unknown) {
+  const err = e as any;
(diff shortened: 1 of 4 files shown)


authui/src/error.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (err.response != null) {
+    const response: AxiosResponse = err.response;
+    const reason = response.data?.error.reason;
+    if (reason === "RateLimited") {
+      showErrorMessage("error-message-rate-limited");
+    }
+  } else if (err.request != null) {
+    showErrorMessage("error-message-network");
+  } else {
+    // programming error
```
carmenlau, line 14: Should we show `error-message-server` here rather than no visual feedback?

PR #1954, opened 2022-04-03
User search standard attributes

carmenlau:
refs #1908

### DIFF
--- a/pkg/api/model/user.go
+++ b/pkg/api/model/user.go
@@ -23,16 +23,17 @@ type UserRef struct {
 }
 
 type ElasticsearchUserRaw struct {
-	ID                string
-	AppID             string
-	CreatedAt         time.Time
-	UpdatedAt         time.Time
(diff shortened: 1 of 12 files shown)


pkg/api/model/user.go
```diff
@@ -65,4 +66,20 @@ type ElasticsearchUserSource struct {
...
+
+	FamilyName    string `json:"family_name,omitempty"`
+	GivenName     string `json:"given_name,omitempty"`
+	MiddleName    string `json:"middle_name,omitempty"`
+	Name          string `json:"name,omitempty"`
+	Nickname      string `json:"nickname,omitempty"`
+	Gender        string `json:"gender,omitempty"`
+	Zoneinfo      string `json:"zoneinfo,omitempty"`
+	Locale        string `json:"locale,omitempty"`
+	Address       string `json:"address,omitempty"`
```
louischan-oursky, line 78: `address` is an JSON object. We only need its components.

Issue #1898, opened 2022-03-22
Change Project creation wizard


fungc-io:
- [ ] The create app title is different when the developer has no apps

fungc-io:
I've made the mockup from the wireframe for the wizard
https://www.figma.com/file/swXJYkdXuPsLyA5q0aHu1r/Authgear-UI---Fung?node-id=120%3A2266

Please use this as reference

fungc-io:
@louischan-oursky 
On staging, for non-first project, the wizard is still showing `Create your first project` in the first step.
Please change it to `Create a new project` in this case

PR #1962, opened 2022-04-07
Use a slightly different copy for developers without any apps

louischan-oursky:
ref #1898

### DIFF
--- a/portal/src/locale-data/en.json
+++ b/portal/src/locale-data/en.json
@@ -1016,7 +1016,7 @@
 
   "WizardContentLayout.skip.label": "Skip and I’ll set up my project manually",
 
-  "CreateProjectScreen.title": "Create your first project",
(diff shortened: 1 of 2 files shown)


portal/src/locale-data/en.json
```diff
@@ -1016,7 +1016,7 @@
 
   "WizardContentLayout.skip.label": "Skip and I’ll set up my project manually",
 
-  "CreateProjectScreen.title": "Create your first project",
+  "CreateProjectScreen.title": "{apps, plural, zero{Create your first project} other{Create a new project}}",
```
carmenlau, line 1019: Seems `zero` is not supported. Maybe we should change the copy to `{apps, plural, =0{Create your first project} other{Create a new project}}`?

louischan-oursky: English recognizes `one` and `other` only. So you're right, we have to use explicit value here.

Issue #1930, opened 2022-03-29
Support Login with GitHub


fungc-io:
Add a login provider "GitHub" and support this in the "portal" app

ref:
https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps

fungc-io:
@louischan-oursky 
please use #171515 to represent GitHub in the Portal
https://www.figma.com/file/swXJYkdXuPsLyA5q0aHu1r/Authgear-UI---Fung?node-id=14%3A1006

fungc-io:
In the GitHub config pane
Add the toggle: `Require the email attribute`

If this is enabled and the end-user do not have a email claim
Show error message:
> Please add a "Public Email" into your GitHub profile. Manage your profile here: \[https://github.com/settings/profile](https://github.com/settings/profile)

louischan-oursky:
- [ ] Use github username to populate name and given_name

PR #1967, opened 2022-04-11
Populate name and given_name with github username

louischan-oursky:
ref #1930

### DIFF
--- a/pkg/lib/authn/sso/github.go
+++ b/pkg/lib/authn/sso/github.go
@@ -63,7 +63,7 @@ func (g *GithubImpl) NonOpenIDConnectGetAuthInfo(r OAuthAuthorizationResponse, _
 
 	idJSONNumber, _ := userProfile["id"].(json.Number)
 	email, _ := userProfile["email"].(string)
-	name, _ := userProfile["name"].(string)
+	login, _ := userProfile["login"].(string)
 	picture, _ := userProfile["avatar_url"].(string)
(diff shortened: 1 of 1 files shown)


pkg/lib/authn/sso/github.go
```diff
@@ -63,18 +63,19 @@ func (g *GithubImpl) NonOpenIDConnectGetAuthInfo(r OAuthAuthorizationResponse, _
...
 
 	authInfo.ProviderUserID = id
 	stdAttrs, err := stdattrs.Extract(map[string]interface{}{
-		stdattrs.Email:   email,
-		stdattrs.Name:    name,
-		stdattrs.Picture: picture,
-		stdattrs.Profile: profile,
+		stdattrs.Email:     email,
+		stdattrs.Name:      login,
+		stdattrs.GivenName: login,
```
carmenlau, line 76: Using github username as the name looks good to me, but given name should be the first name of the user seems using username is not accurate?

louischan-oursky: This is because given_name is default read-write and name is default hidden https://github.com/authgear/authgear-server/blob/master/pkg/lib/config/user_profile.go#L146
If we do not populate the github username as given_name, a new user signed up with github does not have visible "name" by default.

fungc-io: I agree with using the github username in given_name for the reason @louischan-oursky mentioned.

If there is no other problem, shall we merge this and deploy to staging?

Issue #1963, opened 2022-04-07
Support WhatsApp OTP to login


fungc-io:
## User Stories

- As an end-user, I can use my phone number to login to Authgear by sending an OTP via WhatsApp.
- As a developer, I can send use WhatsApp to send OTP to verify a phone number and authenticate the end-user.

## Design

**Business**

We will support the use of the official "WhatsApp Business API" via Twillio.

A separate API account will be used by each tenant. The user do not need to provide their account, we will help them manage the account at the back if they subscribe to this feature. This feature is not included in the standard pricing plans, it must be manually enabled by us.

By default it has a limit of 3000 Login via WhatsApp per day.

**Login flow**

1. End-user enter their own phone number.
2. The screen will show an OTP and a WhatsApp number. With a [`wa.me`](https://faq.whatsapp.com/general/chats/how-to-use-click-to-chat/) link. e.g. `https://wa.me/1XXXXXXXXXX?text=654092`
3. The end-user send the OTP to the number using their WhatsApp account
4. The system will check if the number and the OTP match. The end-user is authenticated if they match.

In step 4, there is a button saying "Send OTP to me with SMS"
By clicking this, it will fallback to the traditional SMS OTP flow.

## Goal

Create a project for us to demo this feature.

## Appetite 

2 weeks

## Task Tracker

- Add config to phone otp authenticator #1982
    - SMS only
    - Whatsapp + SMS fallback
    - Whatsapp only
- Update interaction for whatsapp otp flow
  - [x] Update signup flow with whatsapp otp #1982
  - [x] Update add authenticator flow via settings, admin portal with whatsapp otp #1982
  - [x] Update phone verification flow with whatsapp otp #1987
  - [x] Implement whatsapp otp provider to generate and verify OTP #1990
  - [x] Update whatsapp otp page design #1995
  - [x] Update login flow with whatsapp otp #2000 
  - [x] Check promote user flow
  - [x] Refactor sms fallback using existing alternatives handler #2018
  - [x] Fix OTP is renew and won't resend during authenticator creation when switching different authenticator type #2025
  - [x] Fix switch to verification code again within 1 min will failed due to rate limit #2025
  - [x] Update to new UI #2030
  - [x] Support device token checkbox in whatsapp otp page #2031 
  - [ ] Update button with gtm event tag
  - [ ] Update copywriting in AuthUI and portal to replace the term "SMS" with "Phone" wherever appropriate
- Webhook #2003 
    - verify webhook 
[...]

PR #1990, opened 2022-05-10
Implement whatsapp otp store and provider

carmenlau:
refs #1963 

after #1987 

To test the otp store, we can use the dummy handler here to simulate the user input https://github.com/carmenlau/authgear-server/commit/b229f4fd8d7a6a8894b20baf83a4f65fdf738997

### DIFF
--- a/pkg/lib/authn/authenticator/whatsapp/provider.go
+++ b/pkg/lib/authn/authenticator/whatsapp/provider.go
@@ -1,0 +1,85 @@
+package whatsapp
+
+import (
+	"errors"
+
+	"github.com/authgear/authgear-server/pkg/util/clock"
+	"github.com/authgear/authgear-server/pkg/util/log"
+	"github.com/authgear/authgear-server/pkg/util/secretcode"
+)
+
+type Logger struct{ *log.Logger }
+
+func NewLogger(lf *log.Factory) Logger { return Logger{lf.New("whatsapp")} }
+
+type Provider struct {
+	CodeStore *StoreRedis
+	Clock     clock.Clock
+	Logger    Logger
+}
+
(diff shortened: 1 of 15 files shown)


pkg/lib/authn/authenticator/whatsapp/store_redis.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+type StoreRedis struct {
+	Redis *appredis.Handle
+	AppID config.AppID
+	Clock clock.Clock
+}
+
+func (s *StoreRedis) Create(code *Code) error {
+	ctx := context.Background()
```
louischan-oursky, line 23: It's better to inject the request context here, so that redis operation can have the same deadline of the request.

pkg/lib/authn/authenticator/whatsapp/provider.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	if !secretcode.OOBOTPSecretCode.Compare(codeModel.UserInputtedCode, codeModel.Code) {
+		return nil, ErrInvalidCode
+	}
+
+	return codeModel, nil
+}
+
+func (p *Provider) DeleteCode(phone string) {
```
louischan-oursky, line 65: This method should not be exposed. Instead, in VerifyCode, if the code is correct, delete code from the store directly. This can ensure code is always one-time use

carmenlau: Since we will need to verify the code without consuming for switching the state in the UI, so I separate it. Let me change the verify function to `VerifyCode(phone string, webSessionID string, consume bool)`, so we don't have to expose the delete function

PR #1999, opened 2022-05-16
Fix tag based CI job not running

roxk:
Need to add `tags: '*'` at the top first. Tested working in xamarin sdk.


louischan-oursky:
Applied in e671bdd52ba95db80de72358b405e0663f26d512

PR #2000, opened 2022-05-16
Support login with whatsapp otp

carmenlau:
refs #1963 

after #1995


louischan-oursky (commented):
I got a question I always forgot to ask. Why don't we allow the end-user to switch back to whatsapp once they opted for SMS?

carmenlau:
> I got a question I always forgot to ask. Why don't we allow the end-user to switch back to whatsapp once they opted for SMS?

Currently if user wants to switch back to whatsapp, they can simply back to the previous page to continue the whatsapp flow.

Or that would be better to add a link explicitly in the sms page to switch back to whatsapp?

louischan-oursky:
> Or that would be better to add a link explicitly in the sms page to switch back to whatsapp?

Yes

Issue #1951, opened 2022-03-31
Replace vanilla JavaScript with Hotwire Stimulus


louischan-oursky:
part 1:
- [x] copy button (2hrs)

part 2: 
- [x] prevent double tap (15mins)
- [x] account deletion (1hrs)
- [x] Resend button (2hrs)
- [x] message bar (2hrs)

part 3:
- [x] intl-tel-input (1hrs)
- [x] form controls (2hrs)
- [x] image picker (1.5hrs)

part 4:
- [x] date format (1hrs)
- [x] websocket (2.5hrs)
  - [x] Implement exponential backoff for reconnection. The pattern is 0s -> 1s -> 2s -> 4s. The cap is 4s. The backoff is reset upon successful connection.

part 5:
- [x] modal (2hrs)
- [x] form submission (2.5hrs)

PR #2002, opened 2022-05-17
Migrate to Stimulus part 2

andychow326:
Ref #1951

### DIFF
--- a/authui/src/messageBar.ts
+++ b/authui/src/messageBar.ts
@@ -1,46 +1,30 @@
-export function setupMessageBar(): () => void {
-  const disposers: Array<() => void> = [];
-  const closeButtons = document.querySelectorAll("[data-close-button-target]");
+import { Controller } from "@hotwired/stimulus";
 
-  for (let i = 0; i < closeButtons.length; i++) {
-    const closeButton = closeButtons[i];
+export class MessageBarController extends Controller {
+  static targets = ["button", "bar"];
 
-    const targetID = closeButton.getAttribute("data-close-button-target");
-    if (targetID == null) {
-      continue;
-    }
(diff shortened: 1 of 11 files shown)


authui/src/resendButton.ts
```diff
@@ -1,40 +1,45 @@
...
+        this.animHandle = requestAnimationFrame(tick);
+      }
+    };
+
+    this.animHandle = requestAnimationFrame(tick);
+  }
+
+  disconnect() {
+    if (this.animHandle != null) {
+      cancelAnimationFrame(this.animHandle);
```
louischan-oursky, line 42: We should set `this.animHandle` to `null` after cancelling the animation frame. Generally speaking, a cleanup method may be called more than once. The cleanup method should be able to be run more than once, and running it the second time should have no effect. Here, the same animation handle will be freed twice and it is a problem.

authui/src/messageBar.ts
```diff
@@ -1,46 +1,30 @@
...
-
-    const onCloseButtonClick = (e: Event) => {
-      e.preventDefault();
-      e.stopPropagation();
-      target.classList.add("hidden");
-    };
-
-    // Close the message bar before cache the page.
-    // So that the cached page does not have the message bar shown.
-    // See https://github.com/authgear/authgear-server/issues/1424
```
louischan-oursky, line 26: Keep this comment

authui/src/resendButton.ts
andychow326: Done, please take a look

authui/src/messageBar.ts
andychow326: Done, please take a look

Issue #1988, opened 2022-05-06
Allow setting locale and theme when opening settings page in SDKs


fungc-io:
In `authgear.open(Page.Settings)`, the theme and language in the settings page opened are depending on the user browser.

However some apps allow changing light/dark theme and locale regardless of the system settings.
Developer should be able to force the setting page to show a locale and theme when calling this page from the SDK.

Update SDK
- [x] RN
- [x] Web
- [x] iOS
- [ ] Android
- [x] Flutter
- [x] Xamarin

louischan-oursky:
@carmenlau @fungc-io 

https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/

Our current implementation of dark theme does not support explicit theme selection. If we change the implementation to follow the approach mentioned in the above blogpost, we will break the apps with custom dark themes :(

louischan-oursky:
- https://github.com/authgear/authgear-sdk-flutter/issues/1
- https://github.com/authgear/authgear-sdk-js/issues/181
- https://github.com/authgear/authgear-sdk-ios/issues/80
- https://github.com/authgear/authgear-sdk-android/issues/97
- https://github.com/authgear/authgear-sdk-xamarin/issues/27

Issue #1902, opened 2022-03-22
Update the colors of Dark Mode of the Default theme


fungc-io:
(Details to be provided)

Background:
Pane Card Background:
Text Color:

Screenshot:

karinyu:
Background color > #212121

**Log in/ Sign up Card**
Background > #353535
Border > #3E3E3E
Elevation > x=0, y=4, b=8, s=0 #000000, 10%

**Text Color**
Primary > #E5E5E5
Secondary/ Placeholder > #A6A6A6
Tertiary > #828282

Divider > #595653

https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=5954%3A110466
<img width="354" alt="Screenshot 2022-04-04 at 3 27 44 PM" src="https://user-images.githubusercontent.com/83946723/161494409-aef0afa1-dd76-489b-a3af-5f7f8ac661d2.png">

karinyu:
**User account settings**
Items' background on card > #F8F8F8, 20%
<img width="678" alt="Screenshot 2022-04-08 at 3 15 34 PM" src="https://user-images.githubusercontent.com/83946723/162384263-9f5f0fa0-4200-45a5-acad-060f9b778366.png">

https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=6008%3A110821

fungc-io:
The background color for all the preset themes ("Default", "Tropical",...) should be `#212121`

![Screenshot 2022-05-17 at 15 27 40](https://user-images.githubusercontent.com/15207470/168755215-a3c79ad3-8f08-4fd0-a7ea-89218fd08581.png)

For the existing project, if they are using the preset theme, can we migrate their current color into "Custom"?

PR #2004, opened 2022-05-18
Switch to class-based dark theme implementation

louischan-oursky:
ref #1988 
ref #1902 

~~This PR is missing a migration script to convert old-style media-query-based CSS to class-based CSS~~


carmenlau:
In the portal preview, the pane color should be `#353535`. The other part looks great!

PR #2006, opened 2022-05-19
Migrate to Stimulus part 4

andychow326:
ref #1951

### DIFF
--- a/authui/src/date.ts
+++ b/authui/src/date.ts
@@ -3,6 +3,7 @@ import {
   intlRelativeTimeFormatIsSupported,
 } from "./feature";
 import { DateTime } from "luxon";
+import { Controller } from "@hotwired/stimulus";
 
 // In order to be backward compatible,
 // <span data-date> is equivalent to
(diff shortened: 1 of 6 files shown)


authui/src/websocket.ts
```diff
@@ -13,6 +13,7 @@ function refreshPage() {
 }
 
 export class WebSocketController extends Controller {
+  timeDelay: number = -1;
```
louischan-oursky, line 16: ## Point 1

It's better to name this `backoffIndex`.
The initial value should be 0 (more intuitive than -1).

## Point 2
The setTimeout should be canceled in case the controller disconnects.

## Point 3
The `backoffIndex` should be reset to 0 when connection can be successfully established.

```typescript
reconnectSetTimeoutHandle: number | null = null;

dispose = () => {
  // ...
  // Point 2
  if (this.reconnectSetTimeoutHandle != null) {
    clearTimeout(this.reconnectSetTimeoutHandle);
  }
  this.reconnectSetTimeoutHandle = null;
}

reconnectWebSocket = () => {
  // Point 1
  const index = this.backoffIndex;
  if (this.backoffIndex < 2) {
    this.backoffIndex += 1;
  }
  
  // Point 2
  this.reconnectSetTimeoutHandle = setTimeout(() => {
    this.reconnectSetTimeoutHandle = null;
    this.connectWebSocket();
  }, Math.pow(2, index) * 1000);
};

connectWebSocket = () => {
  // ...
  this.ws.onopen = (e) => {
    // ...
    // Point 3
    this.backoffIndex = 0;
  }
}
```

authui/src/date.ts
```diff
@@ -130,15 +131,18 @@ export function formatDateRelative() {
...
-    const dateSpan = dateSpans[i];
+  declare inputDateTarget: HTMLSpanElement;
+
+  connect() {
+    const hasAbs = intlDateTimeFormatIsSupported();
+    if (!hasAbs) {
+      return;
+    }
+
+    const dateSpan = this.inputDateTarget;
```
louischan-oursky, line 147: This original content should be restored when disconnect. Otherwise, connecting twice will permanently lose the original content.

authui/src/date.ts
```diff
@@ -54,73 +54,75 @@ function parseTimeStyle(s: string | null): TimeStyle | undefined {
...
         }
-      } else {
-        if (abs != null) {
-          dateSpan.textContent = abs;
+
+        if (dateType === "relative") {
+          // Prefer showing relative date time,
+          // and fallback to absolute date time.
+          if (rel != null) {
+            dateSpan.textContent = rel;
```
louischan-oursky, line 118: Restore the original text content in disconnect.

Issue #2012, opened 2022-05-23
Remove core.ts


louischan-oursky:
- [ ] Port back-btn to Stimulus
- [ ] Remove core.ts

PR #2013, opened 2022-05-23
Remove core.ts

andychow326:
ref #2012


authui/src/back.ts
```diff
@@ -0,0 +1,7 @@
+import { Controller } from "@hotwired/stimulus";
+
+export class BackButtonController extends Controller {
+  connect() {
+    this.element.classList.add("js");
```
louischan-oursky, line 5: Instead of adding the `js` class, port the behavior. That is, show the back button in `connect`. Remove the corresponding CSS that reference `.js` class.

louischan-oursky (approved):

resources/authgear/templates/en/web/__nav_bar.html
```diff
@@ -2,10 +2,12 @@
 <div class="flex flex-row">
 	{{ if . }}
 	<a
-		class="not-a btn back-btn icon-btn py-5"
+		class="not-a btn icon-btn py-5"
```
louischan-oursky, line 5: Add the class `invisible` by default.

authui/src/back.ts
```diff
@@ -2,6 +2,8 @@ import { Controller } from "@hotwired/stimulus";
 
 export class BackButtonController extends Controller {
   connect() {
-    this.element.classList.add("js");
+    const button = this.element as HTMLButtonElement;
+
+    button.style.visibility = "visible";
```
louischan-oursky, line 7: `this.element.classList.remove("invisible")`

Issue #1991, opened 2022-05-11
Build and release binary


louischan-oursky:

fungc-io:
To be considered done after deploying "staging-2022-05-12.0" to Production and successfully build the binary

PR #2014, opened 2022-05-23
Build lite binary

louischan-oursky:
ref #1991


carmenlau (commented):
The PR works great! I tried in a clean container. But will we remove the unrelated commands from the lite version? e.g. start.. etc.

p.s. We need keep`authgear images database`

Issue #1889, opened 2022-03-21
[User Portal] `Terminate all sessions` will still keep current session


stanleylhs:
ux/trivial 

## Background:
- Device / Browser (OS version): Chrome on Mac
- Build: Version Tag: 2022-02-24.0

## Description:
`Terminate all sessions` will still keep current session.

## Steps to reproduce:
1. Log in to the User Portal
2. Go to, Security, Signed in Sessions
3. Click  `Terminate all sessions`

## Expected result:
Should Clear up the text description. 
Should change text label to Terminate all **other** sessions
In the dialog box it should be specified that "you won't be logged out of this current sessions".

## Screenshots:

<img width="569" alt="image" src="https://user-images.githubusercontent.com/4337525/159213958-a2337ca3-85a0-4fce-afd4-f6417076b899.png">

https://user-images.githubusercontent.com/4337525/159215100-0a61e82c-99ee-4ffb-b4c4-f233d1db4df1.mov

fungc-io:
Change button copy to "Terminate all other sessions"

PR #2021, opened 2022-05-25
Rename 'Terminate all sessions' message

andychow326:
ref #1889


louischan-oursky:
@fungc-io When we do this issue, we found out that some of the copywritings are incorrect. We have corrected them. Let us know if you want further improvement.

fungc-io:
looks good to me

Issue #1474, opened 2021-08-23
Email verification code can't be verified with copy and paste 


chpapa:
Step to reproduce:

1. Use Email as Login ID
2. Use Passwordless via Email as Authenticator
3. Try to sign up / login
4. At the OTP verification page, copy/paste the email's verification code to the field
5. It shows "Invalid Credentials" (which is a bad error message btw)

Alternatively, if type in the numbers manually, it is okay.

carmenlau:
@chpapa We suspected that there may be extra space during copy and paste, but I just tried and I cannot reproduce the problem that you mentioned in both mobile and desktop. Could you help to try again and see what have you copied?

Desktop:

<img width="697" alt="Screenshot 2021-08-25 at 2 44 54 PM" src="https://user-images.githubusercontent.com/388892/130728678-a6586101-7d79-4aa2-8b38-003b502800b8.png">
 

Mobile:

![IMG_1323](https://user-images.githubusercontent.com/388892/130728662-87892698-a244-4ee9-993e-3745c28571df.PNG)

fungc-io:
TODO: improve error message to tell the end-user that the code should consist of digits only.

chpapa:
Oh I just encounter it again, I found that there are space at the beginning of the OTP code!

louischan-oursky:
The email client Superhuman is used in this case. Not sure if it is related.

louischan-oursky:
I opened the HTML email with Safari (macos and iOS), Chrome (macos) and Gmail web client. There is no space at the beginning of the OTP code. Is it a Superhuman specific problem?

chpapa:
Can't reproduce

chpapa:
I just encountered it again, and I'm pretty sure it is becoz of the empty space before the OTP code, maybe add a logic to trim space?

fungc-io:
trim spaces in otp and recovery codes before verification

PR #2023, opened 2022-05-25
Remove leading and trailing whitespace in OTP code

andychow326:
ref #1474

### DIFF
--- a/pkg/lib/authn/otp/totp.go
+++ b/pkg/lib/authn/otp/totp.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pquerna/otp"
@@ -49,7 +50,8 @@ func GenerateTOTPSecret() (string, error) {
 
 // ValidateTOTP validates the TOTP code against the secret at the given time t.
 func ValidateTOTP(secret string, code string, t time.Time, opts ValidateOpts) bool {
-	ok, err := totp.ValidateCustom(code, secret, t, opts)
+	formattedCode := strings.TrimSpace(code)
+	ok, err := totp.ValidateCustom(formattedCode, secret, t, opts)
 	if err != nil {
 		return false
 	}
--- a/pkg/util/secretcode/oob_otp.go
+++ b/pkg/util/secretcode/oob_otp.go
@@ -3,6 +3,7 @@ package secretcode
 import (
 	"crypto/subtle"
 	"fmt"
+	"strings"
 
 	"github.com/authgear/authgear-server/pkg/util/rand"
 )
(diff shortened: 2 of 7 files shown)


pkg/auth/handler/webapp/verify_identity.go
```diff
@@ -175,6 +176,7 @@ func (h *VerifyIdentityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
 		}
 
 		code := r.Form.Get("x_verification_code")
+		code = strings.TrimSpace(code)
```
louischan-oursky, line 179: Instead of trimming the space here, find out where is the code verified and modify the verification logic.

Some hints on how to locate the places:
- We are dealing with 6-digit OTP so which `"format"` correspond to 6-digit OTP?
- Where are the corresponding `"format"` being used?
- 6-digit OTP is being in verification, primary authentication and secondary authentication.

pkg/util/secretcode/oob_otp.go
```diff
@@ -17,7 +18,8 @@ func (OOBOTPSecretCodeType) Generate() string {
 }
 
 func (OOBOTPSecretCodeType) Compare(a, b string) bool {
-	return subtle.ConstantTimeCompare([]byte(a), []byte(b)) == 1
+	formatedCode := strings.TrimSpace(a)
```
louischan-oursky, line 21: This method treats its argument commutatively. Trim both `a` and `b`.

`s/formatedCode/formattedCode`

pkg/lib/authn/otp/totp.go
```diff
@@ -49,7 +50,8 @@ func GenerateTOTPSecret() (string, error) {
 
 // ValidateTOTP validates the TOTP code against the secret at the given time t.
 func ValidateTOTP(secret string, code string, t time.Time, opts ValidateOpts) bool {
-	ok, err := totp.ValidateCustom(code, secret, t, opts)
+	formatedCode := strings.TrimSpace(code)
```
louischan-oursky, line 68: `s/formatedCode/formattedCode`

pkg/util/secretcode/totp.go
```diff
@@ -13,6 +13,16 @@ import (
...
+var TOTPSecretCode = TOTPSecretCodeType{}
+
+type TOTPSecretCodeType struct{}
+
+type MakeTOTPKeyOptions struct {
+	Issuer      string
+	AccountName string
+	Secret      string
+}
+
```
louischan-oursky, line 25: - Remove `type ValidateOpts`
- Make `ValidateOptsTOTP` private
- Redesign the API of TOTP

```golang
type TOTP struct {
  Secret string
}

type QRCodeImageOptions struct {
  Issuer string
  AccountName string
  Widget int
  Height int
}

func NewTOTPFromRNG() (*TOTP, error)
func NewTOTPFromSecret(secret string) *TOTP
func (c *TOTP) GenerateCode(t time.Time) (string, error)
func (c *TOTP) ValidateCode(t time.Time, code string) bool
func (c *TOTP) QRCodeImage(opts QRCodeImageOptions) (image.Image, error)

func TOTPCheckFormat(value interface{}) error
```

The rationale of the redesign is that TOTPSecretCodeType is not a very useful type. It contains no data. Its methods can simply be turned into ordinary functions.

But TOTP is different. It contains the secret. It encapsulates the secret and exposes 3 methods for us to interact with it. The caller cannot misuse the concept of `secret` and `code` here. To use TOTP, we have to choose from NewTOTPFromRNG or NewTOTPFromSecret. From the design of the API, the caller can learn that TOTP contains secret. GenerateCode and ValidateCode advertise that TOTP is related to time, as both methods take a time argument.

Issue #2011, opened 2022-05-20
Ensure google tag manager script can be loaded


louischan-oursky:
- Make script source nonce relatively stable with a session cookie. When the nonce is not stable, a new page fetch by turbolinks will fail due to the mismatch nonce of the current document.
- Add frame-src so that google tag manager frame can be injected.

PR #2024, opened 2022-05-25
Fix GTM script cannot be loaded

louischan-oursky:
ref #2011


carmenlau (commented):
LGTM! Please help update the test

louischan-oursky:
Updated.

Issue #1696, opened 2021-12-09
Support `passwordrules` HTML attributes


louischan-oursky:
https://developer.apple.com/password-rules/

Generate the `passwordrules` attribute with the password policy

PR #2029, opened 2022-05-26
Support PasswordRules for Apple

andychow326:
ref #1696

### DIFF
--- a/pkg/auth/handler/webapp/change_password.go
+++ b/pkg/auth/handler/webapp/change_password.go
@@ -47,6 +47,7 @@ func (h *ForceChangePasswordHandler) GetData(r *http.Request, rw http.ResponseWr
 	baseViewModel := h.BaseViewModel.ViewModel(r, rw)
 	passwordPolicyViewModel := viewmodels.NewPasswordPolicyViewModel(
 		h.PasswordPolicy.PasswordPolicy(),
+		h.PasswordPolicy.PasswordRules(),
 		baseViewModel.RawError,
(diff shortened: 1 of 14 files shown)


pkg/lib/authn/authenticator/password/checker.go
```diff
@@ -284,6 +287,25 @@ func (pc *Checker) ValidateCurrentPassword(plainPassword string) error {
...
+	}
+	if pc.PwUppercaseRequired {
+		builder.WriteString("required: upper; ")
+	}
+	if pc.PwLowercaseRequired {
+		builder.WriteString("required: lower; ")
+	}
+	if pc.PwDigitRequired {
+		builder.WriteString("required: digit; ")
+	}
```
louischan-oursky, line 304: We can also support `required: special; ` if symbol is required.

pkg/auth/handler/webapp/change_password.go
```diff
@@ -53,7 +54,8 @@ func (h *ForceChangePasswordHandler) GetData(r *http.Request, rw http.ResponseWr
 	viewmodels.Embed(data, baseViewModel)
 	viewmodels.Embed(data, passwordPolicyViewModel)
 	viewmodels.Embed(data, ChangePasswordViewModel{
-		Force: true,
+		Force:               true,
```
louischan-oursky, line 57: Instead of adding PasswordRulesString to individual viewmodel, add it to passwordPolicyViewModel

PR #2036, opened 2022-05-30
Update copywriting to replace sms to phone and add gtm tag to whatsapp otp page

carmenlau:
refs #1963 

@louischan-oursky Not sure if it is appropriate to add `data-authgear-event` to the `a` tag?


louischan-oursky:
We can add `data-authgear-event` to `a`. In GTM, we have to configure a form trigger and a link click trigger to make things work.

Issue #1820, opened 2022-02-21
[Authentication, Login ID] Unable to remove excluded keywords if duplicated items exist in the field


stanleylhs:
bug/minor

## Background:
- Device / Browser (OS version): Chrome on win 10
- Build: Version Tag: 2022-02-08.0
- Reported by: Jasonns

## Description:
If duplicated items exist in the excluded keywords field, upon deleting some unique items, the deleted item will become another duplication of that duplicated items, and cannot be deleted unless the original duplication is deleted.

Related: See video of #1543 deleted button become wonky in the same way

## Steps to reproduce:
1. Log in to the Authgear portal and select any project
2. Go to the 'Authentication > Login ID' page and scroll down to the excluded keywords under the username section
3. Input any set of unique and duplicated entries (e.g. "test", "t", "e", "s", "t")
4. Remove any unique value from the field
5. See the unique value becomes duplication of duplicated entries and cannot be deleted

## Expected result:
Should not allow duplicated value for the "excluded keyword" fields

## Screenshot:

https://user-images.githubusercontent.com/4337525/154938294-ccd9ca4e-d670-423a-8a5b-d0635018c113.mp4

PR #2053, opened 2022-05-31
Don't allow adding same items into the tag picker

carmenlau:
ref #1820


louischan-oursky (commented):
I look into our usage of TagPicker can find out some problems we can fix as well.

- We should NOT use TagPicker. We should use CustomTagPicker everywhere.
- The third argument `suggestionList` of `useTagPickerWithNewTags` is never used. We can remove support for it.
- `useTagPickerWithNewTags` should always deduplicate. If we ever need to allow duplicate, we can add an option to `useTagPickerWithNewTags` to allow that in the future. So every usage of CustomTagPicker is correct without manually calling `deduplicate`

Issue #2033, opened 2022-05-27
Fix sourcemap


louischan-oursky:
The sourcemap cannot be loaded correctly because of mismatch hash

PR #2057, opened 2022-06-01
Fix Source Map

andychow326:
ref #2033

### DIFF
--- a/authui/packages/parcel-namer-authgear/src/index.js
+++ b/authui/packages/parcel-namer-authgear/src/index.js
@@ -1,0 +1,32 @@
+const { Namer } = require("@parcel/plugin");
+const path = require("path");
+
+const NAMING_LIST = [
+  { asset: "authgear", newName: "authgear", type: "js" },
+  { asset: "tailwind", newName: "tailwind", type: "css" },
+];
+
+module.exports = new Namer({
+  name({ bundle }) {
+    const asset = bundle.getMainEntry() ?? bundle.getEntryAssets()[0];
+    const assetPath = asset.filePath;
+    const nameFromSuper = path.basename(assetPath, path.extname(assetPath));
+
+    for (const item of NAMING_LIST) {
+      if (item.asset === nameFromSuper && item.type === bundle.type) {
+        let name = item.newName;
+        if (bundle.type == "js") {
+          if (bundle.env.outputFormat == "esmodule") {
+            name = item.newName + "-module";
+          }
+          if (bundle.env.outputFormat == "global") {
+            name = item.newName + "-classic";
+          }
+        }
+        return name + "." + bundle.hashReference + "." + item.type;
+      }
+    }
+
+    return null;
+  },
+});
--- a/pkg/lib/web/source_map.go
+++ b/pkg/lib/web/source_map.go
@@ -1,0 +1,87 @@
+package web
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/authgear/authgear-server/pkg/util/resource"
+)
+
+type SourceMapDescriptor struct {
+	Path string
+}
+
+var _ resource.Descriptor = SourceMapDescriptor{}
+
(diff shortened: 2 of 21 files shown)


devtools/parcel/main.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+		return nil
+	}
+	traverse(doc)
+
+	return m
+}
+
+func main() {
+	htmlText, err := readHtmlFromFile("./resources/authgear/static/build.html")
```
louischan-oursky, line 105: The convention of a well-behaved command line program is to exit with 1 in case of error.

```typescript
func main() {
  err := do()
  if err != nil {
    fmt.Fprintf(os.Stderr, "%v\n", err)
    os.Exit(1)
  }
}
```

devtools/parcel/main.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		log.Printf("failed to read from file: %s", err)
+		return
+	}
+
+	f, err := os.Create("pkg/lib/webparcel/parcel_gen.go")
+	if err != nil {
+		log.Printf("failed to generate file: %s", err)
+		return
+	}
```
louischan-oursky, line 115: The proper way to handle resource in go looks like

```golang
resource, err := openResource()
if err != nil {
  return nil, err
}
defer resource.Close()
```

pkg/lib/web/javascript.go
```diff
@@ -5,18 +5,25 @@ import (
...
-	Path string
+	Path     string
+	IsHashed bool
 }
 
 var _ resource.Descriptor = JavaScriptDescriptor{}
 
-func (d JavaScriptDescriptor) MatchResource(path string) (*resource.Match, bool) {
-	if path == d.Path {
+func (d JavaScriptDescriptor) MatchResource(resourcePath string) (*resource.Match, bool) {
```
louischan-oursky, line 21: What if we do not make changes to JavaScript?

pkg/lib/web/source_map.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+var _ resource.Descriptor = SourceMapDescriptor{}
+
+func (d SourceMapDescriptor) MatchResource(resourcePath string) (*resource.Match, bool) {
+	if d.IsHashed && path.Ext(resourcePath) == path.Ext(d.Path) && resourcePath[:strings.IndexByte(resourcePath, '.')] == d.Path[:strings.IndexByte(d.Path, '.')] {
+		return &resource.Match{}, true
+	}
+
+	if resourcePath == d.Path {
```
louischan-oursky, line 26: Would simple string comparison work for all cases?

pkg/lib/web/source_map.go
```diff
@@ -0,0 +1,103 @@
...
+	"context"
+	"fmt"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/authgear/authgear-server/pkg/util/resource"
+)
+
+type SourceMapDescriptor struct {
```
louischan-oursky, line 14: See if we can use SimpleDescriptor instead.

louischan-oursky (commented):
See if we can write a parcel plugin that is simultaneously a Namer, Packager, etc to rename, fix source map reference, and generate necessary go code.

authui/packages/parcel-namer-authgear/src/index.js
```diff
@@ -14,7 +14,16 @@ module.exports = new Namer({
...
+        let name = item.newName;
+        if (bundle.type == "js") {
+          if (bundle.env.outputFormat == "esmodule") {
+            name = item.newName + "-module";
+          }
+          if (bundle.env.outputFormat == "global") {
+            name = item.newName + "-classic";
+          }
+        }
+        return name + "." + bundle.hashReference + "." + item.type;
```
louischan-oursky, line 26: This change should belong to the first commit.

Issue #1529, opened 2021-09-21
Allow recovery code to be turned off


louischan-oursky:
Auth0 now allows the developer to turn off recovery code. https://support.auth0.com/notifications/60be5b436c8cef000c69903c

Read that article for the reason for disabling recovery code.

- Add an toggle to `/configuration/authentication/authenticators` > Recovery code, to make the feature optional
  - Default enabled
- If off, no recovery code screen in 2FA setup and no "use recovery code instead" in 2fa screen

louischan-oursky:
![Screenshot 2022-06-07 at 6 15 34 PM](https://user-images.githubusercontent.com/15082564/172356174-5682ce17-da9d-49f3-aa31-1718ac71be94.png)
![Screenshot 2022-06-07 at 6 15 29 PM](https://user-images.githubusercontent.com/15082564/172356190-f7eceace-4a49-4ebd-91df-26c40af369f7.png)

Does it look good? @fungc-io

PR #2065, opened 2022-06-06
Allow recovery code to be disabled

louischan-oursky:
ref #1529

### DIFF
--- a/resources/authgear/templates/en/translation.json
+++ b/resources/authgear/templates/en/translation.json
@@ -87,7 +87,7 @@
   "next-button-label": "Next",
   "copy-button-label": "Copy",
   "copied-button-label": "Copied",
-  "get-new-code-button-label": "Get new code",
+  "get-new-recovery-codes-button-label": "Get new codes",
   "download-button-label": "Download",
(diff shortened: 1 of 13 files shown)


carmenlau (commented):
Screenshot 2022-06-08 at 12 15 38 PM

In the recovery code listing page, we may want to handle when the user doesn't have the recovery code yet. Maybe showing different copywriting?

Steps to reproduce:
- Disable recovery code
- User signup with mfa
- App enable recovery code
- User visit recovery code listing page

louischan-oursky:
Perhaps just say you dont have any recovery yet, and just show the get new code button?

louischan-oursky:
Screenshot 2022-06-09 at 1 38 01 PM

resources/authgear/templates/en/translation.json
```diff
@@ -88,6 +88,7 @@
   "copy-button-label": "Copy",
   "copied-button-label": "Copied",
   "get-new-code-button-label": "Get new code",
```
carmenlau, line 90: Seems it is replaced by `get-new-recovery-codes-button-label` and it is not used. Remove it?

Issue #2058, opened 2022-06-02
Code splitting


louischan-oursky:
Goals

- Code splitting
- No longer need to manually declare dependencies in `static_assets.go`
- Manage third dependencies with `authui/package.json`
- Rebuild `authui` does not require restart of server to serve newer assets
- Do NOT need to change our code to use dynamic import

Idea

- Introduce GeneratedAssetDescriptor.
- It reads a generated manifest JSON file to serve ALL generated assets
- It does NOT read the manifest every time.
- It watches the change of the manifest, update its state

louischan-oursky:
Example of including classic script and CSS

```html
<!DOCTYPE html>
<html>
<head>
	<link rel="stylesheet" href="npm:intl-tel-input/build/css/intlTelInput.css">
	<link rel="stylesheet" href="npm:@tabler/icons/iconfont/tabler-icons.min.css">
	<script src="npm:intl-tel-input/build/js/intlTelInput.js"></script>
	<script src="npm:intl-tel-input/build/js/utils.js"></script>
	<script type="module" src="./pageA.ts"></script>
</head>
</html>

```

louischan-oursky:
Problems we have encountered

## Parcel requires either `import()` or multiple page to enable code splitting

Our solution is to create a dummy page that refers a dummy script. The dummy script contains `import()` calls to force code splitting on third party modules

```html
<!DOCTYPE html>
<html>
  <head>
    <script type="module" src="./dummy.ts"></script>
  </head>
</html>
```

```typescript
import("library-a");
import("library-b");
```

```sh
$ parcel build ./src/*.html
```

## FOUC after code splitting

Our color scheme is class-based so it requires JavaScript to work. The color scheme JavaScript was initially bundled together with other non-rendering essential code. It runs only after every script has been loaded. However, there is a gap between all blocking CSS have been loaded and all deferred scripts have been fetched. Once blocking CSS have been loaded, the page renders immediately. At that time, the color scheme JavaScript did not kick in yet, thus the page is rendered in light. After all deferred scripts have been executed, the page is rendered in dark. This light to dark change is the FOUC. We solve this by extracting the color scheme JavaScript from the main bundle, making it a blocking classic script. So it is executed BEFORE the page is rendered.

PR #2070, opened 2022-06-07
Code Splitting

andychow326:
ref #2058


authui/src/colorscheme.ts
```diff
@@ -1,43 +1,31 @@
...
-      return;
-    }
-
-    let explicitColorScheme = "";
-    const metaElement = document.querySelector("meta[name=x-color-scheme]");
-    if (metaElement instanceof HTMLMetaElement) {
-      explicitColorScheme = metaElement.content;
-    }
+function onChange() {
+  const htmlElement = document.getElementsByTagName("html")[0];
```
louischan-oursky, line 2: `document.documentElement` is the [HTMLHtmlElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHtmlElement) you want.

authui/src/colorscheme.ts
```diff
@@ -1,43 +1,31 @@
...
-  disconnect() {
-    this.queryResult.removeEventListener("change", this.onChange);
+  if (colorScheme === "dark") {
+    htmlElement.classList.add("dark");
+  } else {
+    htmlElement.classList.remove("dark");
   }
 }
+
+const queryResult = window.matchMedia("(prefers-color-scheme: dark)");
```
louischan-oursky, line 29: This should appear before `onChange` because it is being referenced by `onChange`.

authui/src/colorscheme.ts
```diff
@@ -1,43 +1,31 @@
...
+queryResult.addEventListener("change", onChange);
+onChange();
```
louischan-oursky, line 31: So now `colorscheme.ts` is a classic script. It is being executed in the global scope. After evaluating it, we now have polluted the global scope with two global variables `onChange` and `queryResult`. Classic script usually takes the form of [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE). Convert this classic script into IIFE to avoid polluting the global scope.

pkg/lib/web/generated_asset.go
```diff
@@ -0,0 +1,209 @@
...
+	"gopkg.in/fsnotify.v1"
+
+	"github.com/authgear/authgear-server/pkg/util/log"
+	"github.com/authgear/authgear-server/pkg/util/resource"
+)
+
+type LocalFSLogger struct{ *log.Logger }
+
+type GeneratedAssetDescriptor struct {
+	Logger LocalFSLogger
```
louischan-oursky, line 21: The logger is not injected and it is nil. Let's not to log anything for now.

pkg/lib/web/generated_asset.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func (d *GeneratedAssetDescriptor) FindResources(fs resource.Fs) ([]resource.Location, error) {
+	manifest := d.manifest.Load().(*ManifestContext).content
+	locations := make([]resource.Location, 0)
+
+	for _, value := range manifest {
+		location := resource.Location{
+			Fs:   fs,
+			Path: fmt.Sprintf("static/%s", value),
```
louischan-oursky, line 142: Use [path.Join](https://pkg.go.dev/path#Join) to build path. We do not use [filepath.Join](https://pkg.go.dev/path/filepath#Join) because we always deal with forward-slash path.

pkg/lib/web/generated_asset.go
```diff
@@ -0,0 +1,209 @@
...
+		_, err := resource.ReadLocation(location)
+
+		if os.IsNotExist(err) || err != nil {
+			continue
+		}
+		locations = append(locations, location)
+
+		locationMap := resource.Location{
+			Fs:   fs,
+			Path: fmt.Sprintf("static/%s.map", value),
```
louischan-oursky, line 153: Use [path.Join](https://pkg.go.dev/path#Join) to build path.

pkg/lib/web/generated_asset.go
```diff
@@ -0,0 +1,209 @@
...
+		if IsSourceMapPath(key) {
+			isMap = true
+			key = strings.TrimSuffix(key, ".map")
+		}
+		assetPath := fmt.Sprintf("%s%s", StaticAssetResourcePrefix, manifest[key])
+		if isMap {
+			assetPath += ".map"
+		}
+
+		fmt.Printf("r.Location.Path: %s, assetPath: %s, key: %s\n", r.Location.Path, assetPath, key)
```
louischan-oursky, line 198: Do not print log in this descriptor.

pkg/lib/web/generated_asset.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func (d *GeneratedAssetDescriptor) viewResources(resources []resource.ResourceFile, rawView resource.View) ([]byte, error) {
+	if len(resources) == 0 {
+		return nil, resource.ErrResourceNotFound
+	}
+
+	manifest := d.manifest.Load().(*ManifestContext).content
+	for _, r := range resources {
+		key := strings.TrimPrefix(rawView.(resource.EffectiveFile).Path, StaticAssetResourcePrefix)
```
louischan-oursky, line 187: This cast will crash if rawView is not resource.EffectiveFile. If we do not know how to handle other view, return resource.ErrResourceNotFound

pkg/lib/web/generated_asset.go
```diff
@@ -0,0 +1,209 @@
...
+		isMap := false
```
louischan-oursky, line 188: Would a simple suffix match on the view path and the resource path suffice?

pkg/lib/web/static_asset_resolver.go
```diff
@@ -14,20 +15,8 @@ import (
...
-	TailwindCSS:          true,
-	TailwindCSSMap:       true,
-}
-
 var StaticAssetResources = map[string]resource.Descriptor{
-	"authgear-classic.js":     AuthgearClassicJS,
-	"authgear-classic.js.map": AuthgearClassicJSMap,
-	"authgear-module.js":      AuthgearModuleJS,
-	"authgear-module.js.map":  AuthgearModuleJSMap,
+	"generated-asset": GeneratedAsset,
```
louischan-oursky, line 19: No need to add this.

pkg/lib/web/static_asset_resolver.go
```diff
@@ -102,6 +67,14 @@ func (r *StaticAssetResolver) StaticAssetURL(id string) (string, error) {
 	return staticAssetURL(r.Config.PublicOrigin, string(r.StaticAssetsPrefix), hashPath)
 }
 
+func (r *StaticAssetResolver) GeneratedStaticAssetURL(key string) (string, error) {
+	desc, _ := StaticAssetResources["generated-asset"]
```
louischan-oursky, line 71: We can just refer the descriptor with `GeneratedAsset`, as we did in line 19. We can save a cast in line 73 too.

pkg/lib/web/static_asset_resolver.go
```diff
@@ -102,6 +67,14 @@ func (r *StaticAssetResolver) StaticAssetURL(id string) (string, error) {
...
+
+	manifest := desc.(*GeneratedAssetDescriptor).manifest.Load().(*ManifestContext).content
```
louischan-oursky, line 73: Instead of accessing the private field `manifest`, expose a method on GeneratedAssetDescriptor `GetAssetPathForKey(key string) (string, error)`. Propagate the error.

pkg/lib/web/static_asset_resolver.go
```diff
@@ -149,13 +122,8 @@ func ParsePathWithHash(hashedPath string) (filePath string, hash string) {
...
 }
 
-func IsAssetPathHashed(assetPath string) bool {
-	id := strings.TrimPrefix(assetPath, StaticAssetResourcePrefix)
-	desc, ok := StaticAssetResources[id]
-	if !ok {
-		return false
-	}
-	return StaticAssetResourcesIsHashed[desc]
+func IsAssetPathHashed(hash string) bool {
```
louischan-oursky, line 125: Do not alter the meaning of an existing function. Delete the original function and create a new one with a name telling what it does. Here, `LookLikeAHash(s string) bool` would be a good name.

louischan-oursky (changes requested):

louischan-oursky (commented):
Make the following case work

1. `make authui`
2. Start server
3. See it works
4. Make change to `authgear.ts`, for example adding a console.log
5. `make authui`
6. Without restarting the server, see if refresh the page can pick up the changed bundle. It should print the newly added console.log message.

PR #2081, opened 2022-06-10
Switch to graphql-codegen

andychow326:
ref #1721

PR #2084, opened 2022-06-10
Track non-blocking event `whatsapp.otp.verified`

carmenlau:
ref #2083

### DIFF
--- a/pkg/admin/graphql/audit_log.go
+++ b/pkg/admin/graphql/audit_log.go
@@ -113,6 +113,9 @@ var auditLogActivityType = graphql.NewEnum(graphql.EnumConfig{
 		"SMS_SENT": &graphql.EnumValueConfig{
 			Value: "sms.sent",
 		},
+		"WHATSAPP_OTP_VERIFIED": &graphql.EnumValueConfig{
+			Value: "whatsapp.otp.verified",
+		},
 	},
 })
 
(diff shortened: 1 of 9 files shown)


pkg/lib/interaction/nodes/create_authenticator_whatsapp_otp.go
```diff
@@ -29,6 +30,12 @@ func (e *EdgeCreateAuthenticatorWhatsappOTP) Instantiate(ctx *interaction.Contex
 	if err != nil {
 		return nil, err
 	}
+	err = ctx.Events.DispatchEvent(&nonblocking.WhatsappOTPVerifiedEventPayload{
```
louischan-oursky, line 33: If we add the DispatchEvent call in WhatsappCodeProvider.VerifyCode, it would be less likely we will forget to DispatchEvent in the future, when we have an additional place to call WhatsappCodeProvider.VerifyCode

carmenlau:
1. Moved `DispatchEvent` to `WhatsappCodeProvider.VerifyCode` in commit c853da1.
2. I rebased the `billing` branch and run `npm run gentype` in this PR (d63ac72),  the `globalTypes.generated.ts` is updated. But when I search, there are similar `AuditLogActivityType` definition in files `AuditLogEntryScreen.generated.ts`, `AuditLogScreen.generated.ts` and `UsersScreen.generated.ts` which are not updated. Is it expected, or did i miss something?

louischan-oursky:
You did discover a problem. The codegen config has gone through several modification https://github.com/authgear/authgear-server/pull/2081#pullrequestreview-1002577977 This modification corrected the config so that global types are placed in globalTypes.generated.ts while query- and mutation-specific types are colocated with the .graphql files. The colocated files should not contain global types. I guess when we correct the config, we did not perform `find . -name `*.generated.ts` -delete` and graphql-codegen is smart enough to only touch the part it knows. So the duplicate code is not removed when we rerun `npm run gentype`.

Issue #2101, opened 2022-06-13
Restoring phone value in the signup / login form should not pre-filled the country code


carmenlau:
**Describe the bug**
When user switches from the login page to the signup page, the country code will be pre-filled in the phone input which is unexpected. 

**Authgear Version**
 - Version: SaaS 13 Jun 2022

**To Reproduce**
1. Select **Use phone number instead** in the login page, the input is empty and a country is selected.
2. Click signup tab and switch to the signup page
3. The previous selected country is restored, but the country code is also pre-filled to the input. 

**Expected behavior**
The previous selected country should be restored, but the input field should retain empty.

**Client Env (if applicable, please complete the following information):**
The problem happens in all environments, but it is more serious in android chrome. Since in the android chrome when user switches between the pages, the input cursor focus on the incorrect position.

**Video**

https://user-images.githubusercontent.com/388892/173390413-651b9833-4ba1-473c-9189-c0f48f8ae4c3.mov

PR #2109, opened 2022-06-15
Restore phone value without country code

andychow326:
ref #2101

### DIFF
--- a/authui/src/intlTelInput.ts
+++ b/authui/src/intlTelInput.ts
@@ -31,7 +31,7 @@ export class IntlTelInputController extends Controller {
   ignoreConnect: boolean = false;
   ignoreDisconnect: boolean = false;
 
-  input() {
+  input(fromConnect: boolean) {
     const instance = this.instance;
     if (instance == null) {
       return;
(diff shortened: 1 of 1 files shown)


resources/authgear/templates/en/web/signup.html
```diff
@@ -67,7 +67,7 @@ <h1 class="m-0 primary-txt text-center text-xl font-bold">
 						placeholder="{{ template "phone-number-placeholder" }}"
 						data-controller="intl-tel-input"
 						data-intl-tel-input-target="input"
-						data-action="input->intl-tel-input#input countrychange->intl-tel-input#input intl-tel-input:input->retain-form-form#input"
+						data-action="input->intl-tel-input#onChange countrychange->intl-tel-input#onChange intl-tel-input:input->retain-form-form#input"
```
louischan-oursky, line 70: Do not rename the event handler.

authui/src/intlTelInput.ts
```diff
@@ -52,8 +52,14 @@ export class IntlTelInputController extends Controller {
         });
       }
     } else {
-      const { dialCode } = instance.getSelectedCountryData();
-      if (dialCode != null) {
+      const { iso2, dialCode } = instance.getSelectedCountryData();
+      if (fromConnect && iso2 && dialCode) {
+        this.inputElement.value =
+          this.inputElement.value.split(dialCode).pop() ?? "";
```
louischan-oursky, line 58: Write code that is easier to understand.

```typescript
if (fromConnect && iso2 != null && dialCode != null) {
  const countryCallingCode = `+${dialCode}`;
  this.inputElement.value = this.inputElement.value.replace(countryCallingCode, "");
  instance.setCountry(iso2);
}
```

Issue #2086, opened 2022-06-10
Store usage record


carmenlau:
- [x] Create the `_portal_usage_record` table
- [x] Add command to aggregate active user count into `_portal_usage_record`
- [x] Add command to aggregate sms count by region into `_portal_usage_record`
- [x] Add command to aggregate email count into `_portal_usage_record`
- [x] Add command to aggregate whatsapp otp success count into `_portal_usage_record`

carmenlau:
Initial planning for the commands. Seems we only need daily count for sms, email and whatsapp, so only active user count consist of weekly and monthly data. 

- `go run ./cmd/portal usage collect-count active-user (yesterday|last-month|last-week|2006-01-02|2006-01|2006-W37)`
- `go run ./cmd/portal usage collect-count sms-sent (yesterday|2006-01-02)`
- `go run ./cmd/portal usage collect-count email-sent (yesterday|2006-01-02)`
- `go run ./cmd/portal usage collect-count whatsapp-otp-verified (yesterday|2006-01-02)`

cc @louischan-oursky

PR #2112, opened 2022-06-15
Collect sms sent usage record

carmenlau:
refs #2086


carmenlau:
Instead of using `sms-count` as the usage record name, I changed the names to be more specific as follow:

https://github.com/authgear/authgear-server/blob/cb6fd2c83a1a42ab113a8709b47f61708e07bae7/pkg/lib/usage/usage_record.go#L12-L19

See if it looks good? If yes, I will update the spec in #2059

Issue #1275, opened 2021-06-18
Sort by timestamp in Audit log and detail page back to current page


fungc-io:
It'd be useful if the audit log
- [x] allow changing the sorting direction by timestamp
- [x] default sort by decreasing timestamp, i.e. recent entries on the top
- [x] after viewing activity details / user details, back navigation goes to the original page of the audit log table
- [x] Refresh button

louischan-oursky:
The problem of sorting by decreasing timestamp is that in a buzy app, when you navigate to the next page, you probably will see very unstable result. It is because there are new items being prepended to the list constantly.

chpapa:
The problem Louis mentioned should be able to solve by adopting some pagination best practices? like "after=[monotonic incremental id of event"

louischan-oursky:
It should be `from=<audit-log-from-bound>&to=<the-timestamp-when-I-first-visit-this-page>&order_by=desc&page=[1|2|3|...]`. But the user has to somehow refresh / re-enter the page in order to be able to see a "newer world", not sure if this is an UX problem.

fungc-io:
A refresh button can be added to the action menu on the top to get the newer entries. Press the refresh button to get the latest log and go to the first page.

fungc-io:
@andychow326 updated design here: 
https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=6258%3A97185


It is similar to Google Cloud Storage browser
<img width="234" alt="Screenshot 2022-06-13 at 17 33 13" src="https://user-images.githubusercontent.com/15207470/173324655-86dc8fe6-eace-4a41-983f-f93530e64c06.png">

fungc-io:
@andychow326 please implement "back to original page"

PR #2125, opened 2022-06-17
Support restore page by search params

andychow326:
ref #1275

### DIFF
--- a/portal/src/graphql/adminapi/AuditLogEntryScreen.tsx
+++ b/portal/src/graphql/adminapi/AuditLogEntryScreen.tsx
@@ -1,5 +1,5 @@
 import React, { useMemo, useContext } from "react";
-import { useParams } from "react-router-dom";
+import { useParams, useSearchParams } from "react-router-dom";
 import { Text, Label } from "@fluentui/react";
 import { FormattedMessage, Context } from "@oursky/react-messageformat";
 import { useQuery } from "@apollo/client";
@@ -54,15 +54,19 @@ function SummaryText(props: { children: React.ReactNode; light?: boolean }) {
 // eslint-disable-next-line complexity
 const AuditLogEntryScreen: React.FC = function AuditLogEntryScreen() {
   const { logID } = useParams();
+  const [searchParams] = useSearchParams();
 
   const { renderToString, locale } = useContext(Context);
 
   const navBreadcrumbItems = useMemo(() => {
     return [
-      { to: "../../", label: <FormattedMessage id="AuditLogScreen.title" /> },
+      {
+        to: `../..?${searchParams.toString()}`,
+        label: <FormattedMessage id="AuditLogScreen.title" />,
+      },
       { to: "./", label: <FormattedMessage id="AuditLogEntryScreen.title" /> },
     ];
-  }, []);
+  }, [searchParams]);
 
   const { data, loading, error, refetch } = useQuery<
     AuditLogEntryQueryQuery,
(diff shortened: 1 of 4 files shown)


portal/src/graphql/adminapi/AuditLogScreen.tsx
```diff
@@ -149,13 +174,36 @@ const AuditLogScreen: React.FC = function AuditLogScreen() {
...
+    return lastUpdatedAt.toISOString();
   }, [rangeTo, lastUpdatedAt]);
 
+  const location = useLocation();
+
   const isCustomDateRange = rangeFrom != null || rangeTo != null;
 
   const { renderToString } = useContext(Context);
 
+  useEffect(() => {
```
louischan-oursky, line 186: Instead of syncing searchParams and state, can we eliminate state and rely only on searchParams as a single source of truth?

portal/src/graphql/adminapi/AuditLogList.tsx
```diff
@@ -141,7 +143,7 @@ const AuditLogList: React.FC<AuditLogListProps> = function AuditLogList(props) {
       let href: string | null = null;
       switch (column?.key) {
         case "activityType":
-          href = `./${item.id}/details/`;
+          href = `${item.id}/details${searchParam}`;
```
louischan-oursky, line 146: It is unnecessary to pass the query around. For example, we did not pass the query to the user details page link. But we still can restore the search state when we pop the user details page.

portal/src/graphql/adminapi/AuditLogEntryScreen.tsx
```diff
@@ -54,15 +54,19 @@ function SummaryText(props: { children: React.ReactNode; light?: boolean }) {
...
     return [
-      { to: "../../", label: <FormattedMessage id="AuditLogScreen.title" /> },
+      {
+        to: `../..?${searchParams.toString()}`,
+        label: <FormattedMessage id="AuditLogScreen.title" />,
+      },
       { to: "./", label: <FormattedMessage id="AuditLogEntryScreen.title" /> },
     ];
-  }, []);
+  }, [searchParams]);
```
louischan-oursky, line 69: So we do not need to change anything in this screen

portal/src/graphql/adminapi/AuditLogScreen.tsx
```diff
@@ -81,13 +92,24 @@ function RefreshButton(props: ICommandBarItemProps) {
...
+  const selectedKeyParam = searchParams.get("activity_type");
+
+  const [offset, setOffset] = useState(offsetParam);
+  const [selectedKey, setSelectedKey] = useState(selectedKeyParam ?? "ALL");
   const [dateRangeDialogHidden, setDateRangeDialogHidden] = useState(true);
-  const [sortDirection, setSortDirection] = useState<SortDirection>(
-    SortDirection.Desc
-  );
+  const [sortDirection, setSortDirection] = useState<SortDirection>(orderParam);
   const [lastUpdatedAt, setLastUpdatedAt] = useState(new Date());
```
louischan-oursky, line 113: We need to persist this as well. Otherwise, when the log page is visited again, there could be new items being prepended to the front of the list. Persist this in numeric format, i.e. the milliseconds from the epoch.

portal/src/graphql/adminapi/AuditLogScreen.tsx
```diff
@@ -149,13 +175,44 @@ const AuditLogScreen: React.FC = function AuditLogScreen() {
...
+  }, [
+    offset,
+    sortDirection,
+    isCustomDateRange,
+    rangeTo,
+    rangeFrom,
+    lastUpdatedAt,
+    setSearchParams,
+    selectedKey,
+  ]);
```
louischan-oursky, line 214: When the UI is showing "Full Date Range" then there should be no "to" param appearing in the query.

PR #2133, opened 2022-06-22
Run npm test for portal in ci

andychow326:
ref #2131


portal/jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  preset: "ts-jest",
+  testEnvironment: "node",
+  transform: {
+    "^.+\\.ts?$": "ts-jest",
```
louischan-oursky, line 5: `^.+\\.tsx?$`: `ts-jest`

PR #2137, opened 2022-06-23
Upload usage to Stripe

louischan-oursky:
ref #2134
requires #2132


cmd/portal/cmd/cmdpricing/stripe_service.go
```diff
@@ -0,0 +1,345 @@
...
+			"count",
+			"stripe_timestamp",
+		).
+			From(s.SQLBuilder.TableName("_portal_usage_record")).
+			// We have two conditions here.
+			// 1st condition is to retrieve usage records that have not been uploaded.
+			// 2nd condition is to retrieve usage records that have been uploaded the same day, so that if this command
+			// is ever re-run, we still sum up to a correct quantity.
+			Where(
+				"app_id = ? AND name = ? AND period = ? AND ((stripe_timestamp IS NULL AND end_time < ?) OR stripe_timestamp IS NOT NULL AND stripe_timestamp = ?)",
```
carmenlau, line 142: Should we include the midnight in this query, so it should be `end_time <= ?` ?

cmd/portal/cmd/cmdpricing/stripe_service.go
```diff
@@ -0,0 +1,345 @@
...
+		si := subscriptionItem
+		return si, true
+	}
+	return nil, false
+}
+
+func (s *StripeService) uploadUsage(ctx context.Context, appID string) (err error) {
+	midnight := timeutil.TruncateToDate(s.Clock.NowUTC())
+
+	stripeSubscriptionID, err := s.getStripeSubscriptionID(appID)
```
carmenlau, line 280: We should handle record not found of `getStripeSubscriptionID` gracefully, since most of the apps should not have subscriptions and the command should not return error.

carmenlau (commented):
Please rebase!

cmd/portal/cmd/cmdpricing/stripe_service.go
louischan-oursky: It is already in the condition `(stripe_timestamp IS NULL AND end_time < ?)`

carmenlau: I mean make the mightnight inclusive `<=`, see does it make sense?

louischan-oursky: Yes, you're right.

carmenlau (approved):

carmenlau:
Please update the PR, the enum moved to the model package.

louischan-oursky:
Updated.

PR #2177, opened 2022-06-29
Fix the input field style in mobile safari

carmenlau:
## Before
IMG_6265

## After
IMG_6266


louischan-oursky:
@carmenlau I further tidied up the input control related styles.

Issue #2166, opened 2022-06-28
~ add MAU metered billing


fungc-io:

louischan-oursky:
@fungc-io Should we round down or round up when calculating the cost of additional MAU?

For example, an app subscribing to a plan with 30000 MAU included has MAU of 34000, the additional MAU is 4000. The price is USD50/mo per 5000 additional MAU. In this case, should the quantity be `floor(4000 / 5000) = 0 (round down) => USD 0 additional cost` or `ceil(4000 / 5000) = 1 (round up) => USD50 additional cost`?

louischan-oursky:
BTW, the price created in the Stripe dashboard is always round up. I cannot find a way in the dashboard to make it round down.

PR #2185, opened 2022-06-30
Add MAU billing

louischan-oursky:
ref #2166 
ref #2167


cmd/portal/cmd/cmdpricing/stripe_service.go
```diff
@@ -252,6 +272,70 @@ func (s *StripeService) uploadUsageRecordToSubscriptionItem(
...
+
+	// Skip when there are no records.
+	if len(records) <= 0 {
+		return
+	}
+
+	quantity := 0
+	for _, record := range records {
+		quantity += record.Count
+	}
```
carmenlau, line 305: I think we should have one record only here?

Issue #2170, opened 2022-06-28
Server performance enhancement


andychow326:
## Problem

The AuthUI is too slow and the signup/login page sometimes takes more than 30 seconds to load up even under 1000Mbps network. 
This is problematic to the UX and conversions as longer the load time, poorer the bounce rate. [See research.](https://www.thinkwithgoogle.com/marketing-strategies/app-and-mobile/mobile-page-speed-new-industry-benchmarks/)

## Goal

On a reasonably fast computer/mobile phone, at 50Mbps (typical cellular/home internet speed), the end-users should be able to interact with the signup/login page in under *3 seconds* after starting.

## Todo

Use `pprof` to profile the requests and find the area we can optimize so that Authgear is more reliable and faster.

- [ ] Reduce the memory usage of each request
- [ ] Reduce the latency of AuthUI

fungc-io:
@andychow326 
The issue description is updated

cc: @louischan-oursky @carmenlau

PR #2204, opened 2022-07-08
Introduce new resource manager for serving embedded resources

andychow326:
ref #2170


pkg/auth/handler/webapp/static_assets.go
```diff
@@ -60,15 +60,13 @@ func (h *StaticAssetsHandler) Open(name string) (http.File, error) {
 		return nil, os.ErrNotExist
 	}
 
-	resolvePath := filePath
-
 	// Use GlobalEmbeddedResourceManager to check if the static asset is belong to it
-	if asset, ok := h.EmbeddedResources.Resolve(resolvePath); ok {
+	if asset, ok := h.EmbeddedResources.Resolve(filePath, hashInPath); ok {
```
louischan-oursky, line 83: The new manager need not to support hash checking. It should resolve `p` directly.

Issue #130, opened 2020-08-06
WebAuthn


chpapa:
## Problem/Goal

- Biometric login on browser become feasible as the increased support for WebAuthn standard
- To provide more passwordless option for the users, e.g. Yubikey
- To support the use of passwordless on ios/android devices, and websites via Chromium, Safari, Firefox.

## Solution
- End-users can use WebAuthn to authenticate themselve, for registration and (~)MFA
- If it is used as primary authenticator, user will first input their login ID, i.e. phone/email/username, and then auth via webauthn.

Platforms:
* Windows Desktop (Firefox / Edge / Chrome): Windows Hello / FIDO2 (Yubikey) / U2F (Yubikey)
* Mac Desktop (Firefox / Chrome / Safari): FIDO2 (Yubikey) / U2F (Yubikey) / TouchID
* iOS (Safari): Yubikey 5Ci NFC, Lightening, FaceID/TouchID
* Android (Chrome): Yubikey NFC, USB-C

Figma: https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=6752%3A101280

Reference:
- Apple Passkey https://9to5mac.com/2022/06/07/passkeys-passwordless-sign-in-ios-16/
- Chrome https://developer.chrome.com/docs/devtools/webauthn/

## Appetite
4 weeks

tonyhhyip:
Any update on this feature?

fungc-io:
User flow for end-user:
1. Sign up with a login ID and an authenticator, e.g. email + password
2. Prompted to setup WebAuthn on its device/hardware key
3. User can set it up or skip
4. If set up, the end-user can login directly using the device or key, 
without typing their login ID and password.

louischan-oursky:
## Desktop Chrome

if the user choose "This Device", the credential is stored in Chrome. It can be clear with `Settings -> Privacy and security -> Clear browsing data -> Advanced -> Check password and other sign-in data`. Note that there is no way to delete individual credential. The "This Device" option corresponds to "platform" [AuthenticationAttachment](https://www.w3.org/TR/webauthn-2/#enumdef-authenticatorattachment). If we set the attachment to be "cross-platform", the "This Device" option is gone. The user must choose cross-platform authenticator.

fungc-io:
Design needed:
- "Add Passkey"
- "Use Passkey", if autofill is not available
- Passkey management in `/settings`
- Passkey icon that fits into tablericons / fluentui
[...]

PR #2216, opened 2022-07-19
Specify WebAuthn

louischan-oursky:
ref #130 

You can view the rendered document [here](https://github.com/louischan-oursky/authgear-server/blob/130-webauthn-spec/docs/specs/webauthn.md)


louischan-oursky:
The use cases have been updated. Passkeys are now treated as replacement of passwords. During sign-up, the user either sets up a passkey, or a password, but not both. The user will still have a Login ID as usual.

docs/specs/webauthn.md
```diff
@@ -0,0 +1,346 @@
...
+## Configuration
+
+The following configuration are added.
+
+```yaml
+authentication:
+    identities:
+    - login_id
+    - oauth
+    - webauthn
```
carmenlau, line 176: I thought `webauthn` identity should be removed after update, so the authentication will become `login_id` + `webauthn`? 

But if `webauthn` identity is removed, what will happen if user add one more `login_id` in the setting. Which `login_id` will be used if the user authenticates with `webauthn` key?

Would like to know your idea about the model.

louischan-oursky: > I thought webauthn identity should be removed after update, so the authentication will become login_id + webauthn?

A passkey is still considered as an identity and a primary authenticator. The reason is mentioned in "WebAuthn credential is both an Identity and a Primary Authenticator"

> But if webauthn identity is removed, what will happen if user add one more login_id in the setting. Which login_id will be used if the user authenticates with webauthn key?

We treat passkey as identity so this question does not apply.

PR #2219, opened 2022-07-19
Support x_oauth_provider_alias to redirect the user to the oauth provider

carmenlau:
ref #2218


louischan-oursky:
LGTM. In addition to adding support in JS SDK, we should adding it in the Xamarin SDK.

Issue #2211, opened 2022-07-13
Fix horizontal overflow in the Portal


fungc-io:
when the horizontal space is small, make sure:
- the command bar and message bar should still occupy 100% width of the right frame and do not scroll out of view
- components that are aligned right do not overflow out of view.

Example:
"UI Settings"
<img width="1273" alt="Screenshot 2022-07-13 at 14 04 22" src="https://user-images.githubusercontent.com/15207470/178662934-025f827a-a243-4ae2-b558-60e502441080.png">

PR #2233, opened 2022-07-25
Fix horizontal overflow in Portal

andychow326:
ref #2211

### DIFF
--- a/portal/src/ScreenLayout.module.css
+++ b/portal/src/ScreenLayout.module.css
@@ -27,6 +27,7 @@
   min-width: 0;
   min-height: 0;
 
-  overflow-x: auto;
-  overflow-y: auto;
+  overflow: hidden;
+  display: flex;
+  flex-direction: column;
 }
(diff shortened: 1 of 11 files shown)


portal/src/ScreenLayout.module.css
```diff
@@ -27,6 +27,7 @@
   min-width: 0;
   min-height: 0;
 
-  overflow-x: auto;
-  overflow-y: auto;
+  overflow: hidden;
```
louischan-oursky, line 30: No need to set hidden if the scrollable container has correct overflow value.

portal/src/graphql/portal/AdminAPIConfigurationScreen.tsx
```diff
@@ -261,10 +261,12 @@ const AdminAPIConfigurationScreen: React.FC =
...
-      <AdminAPIConfigurationScreenContent
-        appID={appID}
-        queryResult={queryResult}
-      />
+      <div className={styles.content}>
+        <AdminAPIConfigurationScreenContent
+          appID={appID}
+          queryResult={queryResult}
+        />
+      </div>
```
louischan-oursky, line 269: Teach ScreenContent to support `scrollable` which is false by default. For screens that do not use CommandBarContainer, make them to use `<ScreenContent scrollable={true} />`

Issue #2103, opened 2022-06-14
Improve "Application" setup flow


fungc-io:
## Problem
The application set up is not intuitive and users often get confused by the instructions.
The goal of this change is to simplify the setup for common usage and provide better copywriting to explain the config.

## Solution
- Introduce "Application Type", this translates to token or cookie based behind the scene. Different options are shown based on the type.
- Move Allowed Origins to per application level, group together with redirect URIs & post logout redirect URIs
- Collapse "Cookie Session Settings" as an advance setting
- Change copy `Invalidate ... after idling` -> `Expire after idling` and `... Idle Timeout` -> `Idling Lifetime`
- See below for detailed copy writing changes

### Breadboard:
![Anth gear Applications](https://user-images.githubusercontent.com/15207470/173523599-91ab612b-a136-4141-a7f4-28e2f71ab074.png)

### Copywriting
----
# Applications Page

`Description`: Configure applications to use this project as authentication.

## Advanced Cookie Settings

`Description`: Configure the cookie session settings for (domain). If you only use token-based authentication, you can simply ignore this section.

- (toggle) Persistent cookie
  - `Description:` The cookie persists after the browser is closed
- Cookie Lifetime
  - `Description:` The cookie session will expire after the lifetime
- (Toggle) Expire after idling
  - `Description:` The cookie session will expire after the idling lifetime
- Idling Lifetime
  - `Description:` The idling lifetime of a cookie

# Create Application Page

- Application Name
- Application Type
    - Native App
    - SPA
    - Traditional Web Application (Cookie based)

# Application Details Page

- Basic Info
    - Name
    - Client ID (Copy Btn)
    - Application Type
- URIs
    - Authorized Redirect URIs
        - `Description:` After users are authenticated, users will be redirected to these URI. They should include the protocol like `http://`,`https://`, or custom URI scheme for native clients.
    - Post Logout Redirect URIs (only when app type = Traditional Web Application)
        - `Description:` Users will be redirected to this path after they have logged out.
    - Allowed Origins
      - Description:` Configure cross-origin resource sharing (CORS), allowing authentication requests from these domains.
- Access Token (only when app type = SPA / Native)
    - Access Token Lifetime
        - `Description:` The `access_token` will expire after the lifetime.
    - (Check box) Issue JWT as access token
- Refresh Token (only when app type = SPA / Native)
    - Refresh token Lifetime
[...]

PR #2248, opened 2022-07-29
Application config UI optimization

andychow326:
ref #2103


portal/src/graphql/portal/CreateOAuthClientScreen.tsx
```diff
@@ -134,6 +140,13 @@ const CreateOAuthClientContent: React.FC<CreateOAuthClientContentProps> =
       [setState]
     );
 
+    useEffect(() => {
+      if (client.name == null) {
+        onClientConfigChange(updateClientConfig(client, "name", "My App"));
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
```
louischan-oursky, line 148: Do not use useEffect to set initial value.

Issue #2256, opened 2022-07-30
Verify by "WhatsApp or SMS" by default


fungc-io:
In Verification > Phone number verification,
Change the default to WhatsApp or SMS instead of "SMS" Only

PR #2260, opened 2022-08-01
Set default phone_otp_mode to whatsapp or sms

andychow326:
ref #2256


portal/src/graphql/portal/VerificationConfigurationScreen.tsx
```diff
@@ -127,7 +128,7 @@ function constructConfig(
       delete v.criteria;
     }
 
-    if (a.oob_otp.sms.phone_otp_mode === "sms") {
```
carmenlau, line 130: Better to define a constant for the default value, it can be a constant in the same file. It is easy to miss if we need to update more than 1 place in the future.

Issue #2249, opened 2022-07-30
Button in Discard confirmation box is ambiguous


fungc-io:
It is ambiguous that if the primary button confirm the "discard" action or confirms the "changes", esp in the unsaved warning when navigating to another page

Change:
- In "Discard and leave" scenario, the button text to "Discard and leave this page", and change color to red.
- In "Discard Changes" confirmation, the button text to "Yes, discard"

![Screenshot 2022-07-30 at 19 02 06](https://user-images.githubusercontent.com/15207470/181908037-14aab90a-536a-45fd-985a-127627390d81.png)

![Screenshot 2022-07-30 at 19 02 14](https://user-images.githubusercontent.com/15207470/181908034-76d7e733-1a48-4e73-8745-cd98a04ef06f.png)

PR #2262, opened 2022-08-01
Adjust discard confirm dialog button design

andychow326:
ref #2249


carmenlau (commented):
Better to adjust the dialog width so that the buttons can be aligned in the same row.

Screenshot 2022-08-01 at 4 04 05 PM

louischan-oursky:
What about shortening the copy to "Discard and leave"? @fungc-io

fungc-io:
@louischan-oursky good idea.

andychow326:
Now the dialog looks like this
image

PR #2265, opened 2022-08-01
Show the quickstart list after creating new oauth client

carmenlau:
ref #2103


portal/src/graphql/portal/EditOAuthClientScreen.tsx
```diff
@@ -97,22 +97,80 @@ function constructConfig(
...
+        return true;
+      }
+
+      // show open tutorial label when hover
+      return !isHovering;
+    }, [showOpenTutorialLabelWhenHover, isHovering]);
+
+    return (
+      <Link
+        onMouseOver={onMouseOver}
```
louischan-oursky, line 136: The hover state will be canceled when the "Open tutorial" label is hovered

Issue #2286, opened 2022-08-04
Failed to load template due to too many opened files


louischan-oursky:
I guess our manifest watcher has some problems with its watching machanism. During development, I made changes and run `make authui` very frequently. After more than 20 times of repetition, I start seeing "too many opened files" error. The problem persists even I restart the server or reboot my machine.

PR #2290, opened 2022-08-05
Fix too many open files error

andychow326:
ref #2286

### DIFF
--- a/pkg/lib/web/embedded_resource.go
+++ b/pkg/lib/web/embedded_resource.go
@@ -40,7 +40,6 @@ func NewDefaultGlobalEmbeddedResourceManager() (*GlobalEmbeddedResourceManager, 
 		ResourceDir:    DefaultResourceDir,
 		ResourcePrefix: DefaultResourcePrefix,
 		Name:           DefaultManifestName,
-		FilePath:       DefaultResourceDir + DefaultResourcePrefix + DefaultManifestName,
 	})
 }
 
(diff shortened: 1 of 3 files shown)


pkg/lib/web/embedded_resource.go
```diff
@@ -101,12 +101,15 @@ func (m *GlobalEmbeddedResourceManager) watch() {
 				break
 			}
 
-			if event.Op&fsnotify.Write != fsnotify.Write && event.Op&fsnotify.Create != fsnotify.Create {
-				break
+			switch event.Op {
```
louischan-oursky, line 103: Refactor so that `watcher.Add` and `watcher.Remove` only appear in one method only. The initial setup code and the `watch()` function should use this method.

pkg/lib/web/embedded_resource.go
```diff
@@ -74,7 +74,7 @@ func NewGlobalEmbeddedResourceManager(manifest *Manifest) (*GlobalEmbeddedResour
 
 func (m *GlobalEmbeddedResourceManager) loadManifest() (map[string]string, error) {
 	jsonFile, err := os.Open(m.ManifestFilePath())
-	if err != nil {
+	if err != nil && !os.IsNotExist(err) {
```
louischan-oursky, line 77: if err is NotExist, we end up calling io.ReadAll with a nil file handle.

Issue #2258, opened 2022-08-01
Make paywall indicator more eye-catching


fungc-io:
The grey "Upgrade your project plan." message bar is easy to be missed. 

~Add a warning toast when clicked on the greyed out features, so for the users it is easier to understand the reason of unable to access that particular feature.~

~> Your subscription does not support this feature. Please upgrade the project plan.~

fungc-io:
blocked feature should not have hover state and the text should be greyed.

See "Passwordless via Phone"
![Screenshot 2022-08-01 at 20 06 00](https://user-images.githubusercontent.com/15207470/182144266-f7ed9591-201a-4e31-9836-19590c38187a.png)

fungc-io:
After discuss with others, seems opening a dialog box is the most intuitive

When clicked:

open modal

> ## Upgrade plan to [use this feature]
> This feature is not supported by the [current plan name] plan, please upgrade your project plan.
> [Upgrade in Billing] [Dismiss]

fungc-io:
@carmenlau 
change the message bar design to:
https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=6912%3A104488

fungc-io:
The color of text and link part should be in the same color:

Staging:
![Screenshot 2022-08-09 at 15 46 29](https://user-images.githubusercontent.com/15207470/183593991-d78bdc85-7037-49a2-94f2-ffec7457a4b8.png)

Design:
![Screenshot 2022-08-09 at 15 47 00](https://user-images.githubusercontent.com/15207470/183594044-e82e8ab9-ae95-49a9-8996-41d07c9f097f.png)

PR #2295, opened 2022-08-09
Force the color of link and text to be the same in the billing message bar

carmenlau:
ref #2258


louischan-oursky:
I changed to override the style with PartialTheme. It is more robust. See the commit message for details.

PR #2297, opened 2022-08-10
Configure passkey

louischan-oursky:
ref #2296


carmenlau (commented):
Currently the passkey enabled state is controlled by the toggle, but not the checkbox. So if I check the passkey checkbox and save, there is no effect. Should we change the checkbox to be disabled?

Screenshot 2022-08-10 at 11 27 37 AM

louischan-oursky:
Let me hide the checkbox

Issue #2206, opened 2022-07-11
Define onboarding funnel and add analytics


fungc-io:
Events:
- [x] create project
- [x] click on getting started items
- completed getting started items
  - [x] create application
  - [x] invite admin
  - ~~1st user signup~~
  - [x] added login provider
- [x] clicked on doc links
- project setup wizard, per step
  - [x] summited (pressed "Next") and the option selected
  - [x] pressed "Skip" at that step
- [x] Set user email to mixpanel
- [x] Provider user's role (owner/ editor) in event
- [x] page view in portal
     -  ~~Change index.html to be server-side render~~
- [ ] track user ownership on project 
---
Connected via GTM

fungc-io:
updated the issue to include setup wizard

PR #2299, opened 2022-08-10
Setup GTM in portal

carmenlau:
ref #2206 

WIP:
Identity user in mixpanel will be handled in another PR

Setup:
https://github.com/carmenlau/authgear-server/blob/2206-portal-gtm/docs/portal/gtm.md

### DIFF
--- a/portal/src/GTMProvider.tsx
+++ b/portal/src/GTMProvider.tsx
@@ -5,7 +5,7 @@ import {
 } from "@elgorditosalsero/react-gtm-hook";
 import { useAppContext } from "./context/AppContext";
 
-export type TrackValue = string | string[] | undefined;
+export type TrackValue = string | string[] | boolean | undefined;
 export type EventData = Record<string, TrackValue>;
 
 export interface AuthgearGTMEvent {
@@ -21,6 +21,8 @@ export enum AuthgearGTMEventType {
   InviteAdmin = "authgear.inviteAdmin",
   AddSSOProviders = "authgear.addSSOProviders",
   ClickDocLink = "authgear.clickDocLink",
+  ClickNextInProjectWizard = "authgear.clickNextInProjectWizard",
+  ClickSkipInProjectWizard = "authgear.clickSkipInProjectWizard",
 }
 
 interface AuthgearGTMEventParams {
--- a/portal/src/GTMProvider.tsx
+++ b/portal/src/GTMProvider.tsx
@@ -20,6 +20,7 @@ export enum AuthgearGTMEventType {
   CreateApplication = "authgear.createApplication",
   InviteAdmin = "authgear.inviteAdmin",
   AddSSOProviders = "authgear.addSSOProviders",
+  ClickDocLink = "authgear.clickDocLink",
 }
 
 interface AuthgearGTMEventParams {
--- a/portal/src/GTMProvider.tsx
+++ b/portal/src/GTMProvider.tsx
@@ -14,6 +14,7 @@ export interface AuthgearGTMEvent {
 export enum AuthgearGTMEventType {
   CreateProject = "authgear.createProject",
   ClickGetStarted = "authgear.clickGetStarted",
+  CreateApplication = "authgear.createApplication",
 }
 
 interface AuthgearGTMEventParams {
--- a/portal/src/GTMProvider.tsx
+++ b/portal/src/GTMProvider.tsx
@@ -3,13 +3,41 @@ import {
   GTMProvider as ReactHookGTMProvider,
   useGTMDispatch as useReactHookGTMDispatch,
 } from "@elgorditosalsero/react-gtm-hook";
+import { useAppContext } from "./context/AppContext";
 
 export interface AuthgearGTMEvent {
+  event: AuthgearGTMEventType;
+  appID?: string;
+  value1?: string;
+}
+
+export enum AuthgearGTMEventType {
+  CreateProject = "authgear.createProject",
+}
+
+interface AuthgearGTMEventParams {
   event: AuthgearGTMEventType;
   value1?: string;
 }
 
-export enum AuthgearGTMEventType {}
+export function useAuthgearGTMEvent({
+  event,
+  value1,
+}: AuthgearGTMEventParams): AuthgearGTMEvent {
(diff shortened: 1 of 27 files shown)


portal/src/GTMProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export interface GTMHookProviderProps {
+  containerID?: string;
+  children: React.ReactNode;
+}
+
+function GTMProvider(pros: GTMHookProviderProps): JSX.Element {
+  if (pros.containerID) {
+    return (
+      <ReactHookGTMProvider state={{ id: pros.containerID }}>
```
louischan-oursky, line 35: Use useMemo to make state relatively stable.

portal/src/graphql/portal/GetStartedScreen.tsx
```diff
@@ -247,6 +261,7 @@ function Card(props: CardProps) {
           to={internalHref}
           component={Link}
           className={styles.cardActionButton}
+          onClick={onClickActionButton}
```
louischan-oursky, line 264: Per our discussion, link tracking should be done with Page View trigger or History Change trigger.

carmenlau:
After more testing, found that historyChange events should be fired right after the page is loaded if I inject the GTM script into the html. The initial historyChange events are missed if I load the script in react. May need to consider server side render the html :(

louischan-oursky:
If we really switch to server-side render index.html, maybe we could eliminate /system-config.json altogether

louischan-oursky (commented):
> https://github.com/carmenlau/authgear-server/blob/2206-portal-gtm/docs/portal/gtm.md

The setup is very detailed! I have one question though. In our discussion we talked about introducing a data layer variable `{{ gtm }}`. Is it useful to us?

carmenlau:
> The setup is very detailed! I have one question though. In our discussion we talked about introducing a data layer variable {{ gtm }}. Is it useful to us?

`{{ gtm }}` is useful! But now I only use it to discover different variables e.g. `gtm.element`..etc. Or do you have any idea how can we use it?:)

louischan-oursky:
https://help.mixpanel.com/hc/en-us/articles/115004613766-Default-Properties-Collected-by-Mixpanel

Mixpanel is able to collect some default properties, so I guess we do not need to give them `{{ gtm }}`. What about replacing `value1` with `mixpanel`?

The custom HTML of track look like

```
var event = {{ event }};
var mixpanelValue = {{ mixpanel }};
var appID = {{appID}};
mixpanel.track(event, {appID, ...mixpanelValue });
```

Then we can pass any mixpanel specific value at call-site.

portal/src/ReactApp.tsx
```diff
@@ -178,9 +181,28 @@ function ExternalLink(props: ILinkProps) {
...
+      {...gtmEventDataAttributes}
+      {...props}
+    />
+  );
+};
+
 const defaultComponents = {
   ExternalLink,
   ReactRouterLink: PortalLink,
+  DocLink,
```
louischan-oursky, line 205: Smart approach!

portal/src/GTMProvider.tsx
```diff
@@ -0,0 +1,121 @@
...
+  ClickedNextInProjectWizard = "ag.event.clickedNextInProjectWizard",
+  ClickedSkipInProjectWizard = "ag.event.clickedSkipInProjectWizard",
+}
+
+interface AuthgearGTMEventParams {
+  event: AuthgearGTMEventType;
+  eventData?: EventData;
+}
+
+export function useAuthgearGTMEvent({
```
louischan-oursky, line 34: Instead of requiring the caller to provide a full event, I suggest we return an object that can aid constructing a full event. The benefit is that if we need to send multiple events, we need not call the hook multiple times.

```diff
diff --git a/portal/src/GTMProvider.tsx b/portal/src/GTMProvider.tsx
index 62ecd58bf..ec0ad5ca9 100644
--- a/portal/src/GTMProvider.tsx
+++ b/portal/src/GTMProvider.tsx
@@ -8,13 +8,16 @@ import { useAppContext } from "./context/AppContext";
 export type TrackValue = string | string[] | boolean | undefined;
 export type EventData = Record<string, TrackValue>;
 
-export interface AuthgearGTMEvent {
-  event: AuthgearGTMEventType;
+export interface AuthgearGTMEventBase {
   app_id?: string;
-  event_data?: EventData;
   _clear: boolean;
 }
 
+export interface AuthgearGTMEvent extends AuthgearGTMEventBase {
+  event: AuthgearGTMEventType;
+  event_data?: EventData;
+}
+
 export enum AuthgearGTMEventType {
   CreatedProject = "ag.event.createdProject",
   ClickedGetStarted = "ag.event.clickedGetStarted",
@@ -26,15 +29,7 @@ export enum AuthgearGTMEventType {
   ClickedSkipInProjectWizard = "ag.event.clickedSkipInProjectWizard",
 }
 
-interface AuthgearGTMEventParams {
-  event: AuthgearGTMEventType;
-  eventData?: EventData;
-}
-
-export function useAuthgearGTMEvent({
-  event,
-  eventData,
-}: AuthgearGTMEventParams): AuthgearGTMEvent {
+export function useAuthgearGTMEventBase(): AuthgearGTMEventBase {
   let appContextID: string | undefined;
   try {
     const appContext = useAppContext();
@@ -43,16 +38,27 @@ export function useAuthgearGTMEvent({
 
   return useMemo(() => {
     return {
-      event: event,
       app_id: appContextID,
-      event_data: eventData,
       // Prevent GTM recursive merge event data object
       // https://github.com/google/data-layer-helper#preventing-default-recursive-merge
       _clear: true,
     };
-  }, [event, eventData, appContextID]);
+  }, [appContextID]);
 }
 
+// Usage
+// const send = useGTMDispatch();
+// const base = useAuthgearGTMEventBase();
+//
+// send({
+//    ...base,
+//    event: AuthgearGTMEventType.CreatedProject,
+// });
+// send({
+//    ...base,
+//    event: ClickedGetStarted,
+// });
+
 interface AuthgearGTMEventDataAttributesParams {
   event: AuthgearGTMEventType;
   eventDataAttributes?: Record<string, string>;

```

Issue #2302, opened 2022-08-12
[Passkey] See Invalid input message and unable to save after toggle on passkey


haretorii:
## Background
Device: Google Chrome on Mac
Build: Staging
Testing account & project: fung+qa@oursky.com / qa-passkey

## Description
When user toggle on passkey feature in authentication and save, he will see `Invalid input at /authentication/identities/0: const` error message and cannot save the auth setting

## Steps
1. Login to the testing account and go to testing project
2. Go to Authentication > Authenticators
3. Toggle on 'Login with Passkey'
4. Click Save
5. See `Invalid input at /authentication/identities/0: const` error message
6. Click on other pages
7. See discard changes dialogue

## Expected result
Should allow user to toggle on Passkey when required primary (and 2FA) authenticator enabled

## Screenshot

https://user-images.githubusercontent.com/69806274/184300721-43c00d99-068b-4d50-974f-8726fe44cc7c.mov

louischan-oursky:
I think that app has no login ID enabled, that is email, phone and username are ALL turned off.

fungc-io:
Fix:
when no login id disable the toggle and tooltip show `You must enable at least 1 Login ID to use passkey`

joyzoursky:
Tried with a project with Login ID (a new project), and it also shows this error when trying to turn on the toggle and save.

@fungc-io @louischan-oursky Is the issue already fixed? Could we have it pushed first so we can try out the feature? Thanks.

<img width="1515" alt="Screenshot 2022-08-15 at 7 31 52 PM" src="https://user-images.githubusercontent.com/12539486/184627775-f5ce87ec-74c6-4457-b63b-c5da599e6ac2.png">
<img width="1515" alt="Screenshot 2022-08-15 at 7 31 58 PM" src="https://user-images.githubusercontent.com/12539486/184627789-e6620a10-1709-427b-ae86-6df054c5dfeb.png">

louischan-oursky:
Try this workaround first

1. Go to Login ID and disable everything. And then enable email.
2. You should be able to turn on passkey now

joyzoursky:
@louischan-oursky thanks, it works

PR #2304, opened 2022-08-12
Disable passkey toggle if login ID is disabled

louischan-oursky:
ref #2302


carmenlau (commented):
We also need to customize the error when the passkey is on, the developer goes to the login id page and disables all the login IDs.

PR #2312, opened 2022-08-15
Remove defaults

louischan-oursky:


carmenlau (commented):
If the developer intentionally sets the config the same as the current default, the behavior should not be changed. If we are going to remove defaults when saving the config, we need to have a commitment that we won't change any of the default settings in the future. Thoughts?

Issue #2320, opened 2022-08-16
More robust passkey error handling


louischan-oursky:
## Scenario 1

- Create an account with passkey on another device
- Sign in the account on Chrome desktop
- Choose "This device"
- Chrome shows an error message in the modal dialog with "Your identity couldn't be verified"
- The end user has no choice but to click cancel to close the modal dialog

### Cause

Chrome does not have the passkey. And it is not smart enough to hide the "This device" option.

### Suggested solution

In /use_password page, offer some **static** help text to explain **possible** errors

## Scenario 2 (Fixed in https://github.com/authgear/authgear-server/pull/2322)

- Create an account with passkey on another device
- Sign in the account on Firefox desktop
- Firefox only supports security key so use security key
- navigator.credentials.get will be rejected with `DOMException(name="InvalidStateError", message="An attempt was made to use an object that is not, or is no longer, usable")`.

### Cause

This message could mean 2 things. The security key indeed contained the passkey in the past but it has been removed. The security key never contains the passkey, the user switches device.

### Suggested solution

Literally tell the user the 2 possible situation.

## Scenario 3 (Fixed in https://github.com/authgear/authgear-server/pull/2321)

- Sign up a new account on Chrome desktop
- Choose "Add a new Android phone" or a named device in the list
- Go through the process as instructed by the phone
- See failed message on phone, and Chrome shows an error message in the modal dialog with "Your identity couldn't be verified"
- The end user has no choice but to click cancel to close the modal dialog

### Cause

We require resident key but Android does not support it yet.

### Suggested solution

Relax [ResidentKeyRequirement](https://www.w3.org/TR/webauthn-2/#enumdef-residentkeyrequirement) to preferred

## Scenario 4

- Sign up a new account on iOS 16 with passkey
- Sign in the account
- Choose use security key instead
- See passkey not found message in the modal dialog. The system asks the user to retry
- We do not need to do anything

## Scenario 5 (Fixed in https://github.com/authgear/authgear-server/pull/2324)

- Sign up a new account on iOS 16 with passkey
- Sign in the account
- Choose use Android phone instead
- Safari throws `DOMException(name="NotAllowedError", message="Operation failed.")`

### Suggested solution

The exception is distinguishable if we also take message into account.

louischan-oursky:
@carmenlau @fungc-io Any thoughts on addressing Scenario 1?

carmenlau:
[...]

PR #2322, opened 2022-08-16
Handle passkey not found on Firefox

louischan-oursky:
This PR addresses the scenario 2 of https://github.com/authgear/authgear-server/issues/2320

ref #2320

### DIFF
--- a/authui/src/passkey.ts
+++ b/authui/src/passkey.ts
@@ -1,6 +1,6 @@
 import { Controller } from "@hotwired/stimulus";
 import axios from "axios";
-import { handleAxiosError } from "./messageBar";
+import { handleAxiosError, showErrorMessage } from "./messageBar";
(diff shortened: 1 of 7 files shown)


louischan-oursky:
@carmenlau @fungc-io Any suggestion on the error message? Currently I just tell the user the password is not found. But what actionable suggestions should we give them?

fungc-io:
How about

If app enabled password/otp
> Passkey is not found on your [device / security key]. Login with [password / one-time-password] instead?
(show password / one-time-password according to priority)

If disabled password/otp
> Passkey is not found on your [device / security key]. Try another device?


If we can detect device vs yubikey, show the corresponding word. 
If not, I think we can just use "device".
Coz "device" can mean both the computer and the physical key.

see what do u think.

PR #2323, opened 2022-08-16
[WIP] Identify portal user in mixpanel tracking

carmenlau:
ref #2206 
require  #2299

My original idea is to call `mixpanel.reset` after user authenticated (`finishedAuthentication`), but it won't work if the user login via `/login`.

In Mixpanel best practices, they suggest that `mixpanel.reset()` should be called on log out. I tried in commit 0e8dd36e5ee31f504b41bff759b3ca6297db4fc5, but there is still some problems.

1. `mixpanel.reset()` needs to be called directly, as the user will be redirected after clicking logout. There is not enough time for GTM to trigger Mixpanel script. 
2. It cannot handle the case when the user's session expires.

Still brainstorming a more robust solution, welcome for any ideas.

ref: https://help.mixpanel.com/hc/en-us/articles/115004497803-Identity-Management-Best-Practices

WIP:
- Discover some problems that need to be fixed
    - user id returned from the viewer query is encoded twice
    - email is missing in the viewer query for new user


portal/src/ReactApp.tsx
```diff
@@ -205,6 +210,35 @@ const defaultComponents = {
...
+          ...gtmEvent,
+          event_data: {
+            // fixme: userid encoded twice
+            user_id: extractRawID(extractRawID(viewer.id)),
+            email: viewer.email ?? undefined,
+          },
+        };
+        sendDataToGTM(event);
+      }
+    }, [viewer, gtmEvent, sendDataToGTM]);
```
louischan-oursky, line 237: My suggestion.

We only handle user identification here. The reason is that this place is the most accurate place of knowing which account is signed in. We always fire the lifecycle event here. In the custom HTML, we always call reset(). If the viewer is present, we call identify(). The only downside is that refresh the portal will create a new mixpanel session, but the mixpanel session will always be identified correctly.

PR #2324, opened 2022-08-17
Handle passkey not found error when using Android phone on iOS

louischan-oursky:
ref #2320

### DIFF
--- a/authui/src/passkey.ts
+++ b/authui/src/passkey.ts
@@ -134,7 +134,7 @@ function isSafariAndroidError(err: unknown) {
     err instanceof DOMException &&
     err.name === "NotAllowedError" &&
     // "Operation failed."
-    /fail/i.test(err.message)
+    /operation.*fail/i.test(err.message)
   );
 }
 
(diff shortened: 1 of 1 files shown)


authui/src/passkey.ts
```diff
@@ -107,21 +107,91 @@ function serializeAssertionResponse(credential: PublicKeyCredential) {
...
+function isSafariAndroidError(err: unknown) {
+  // This happens when the user chooses to use Android phone to scan QR code.
+  // And the passkey is not found on that Android phone.
+  // If security key is used, the error message is shown in the modal dialog and the dialog will not close.
+  // Thus we cannot detect security key error.
+  return (
+    err instanceof DOMException &&
+    err.name === "NotAllowedError" &&
+    // "Operation failed."
+    /fail/i.test(err.message)
```
carmenlau, line 137: If we need to use some keywords in the message to guess the scenario. I think we should avoid using some words that are too generic (e.g. fail), seems it is easier to match some unexpected cases. I think the word `cancel` or `abort` are fine since they have some user cancel meaning. If the error message is already too general, maybe we could match a longer phrase to avoid matching some unexpected cases. What do you think?

louischan-oursky: Updated to use longer phrase for matching.

Issue #2333, opened 2022-08-21
Post login link doesn't work


carmenlau:
**Describe the bug**
Failed to redirect to **post login link** after login

**Authgear Version**
 - Version: staging-2022-08-17.1

**To Reproduce**
1. Go to the portal and set **post login link**
2. Login in the authui
3. User stucks at the end of the login flow and fails to redirect to the **post login link**

**Error Log**
```
Error invoking action "turbo:submit-end@document->turboform#submitEnd"

TypeError: Cannot read properties of undefined (reading 'location')
    at t.submitEnd (form.ts:12:53)
    at g.invokeWithEvent (stimulus.js:278:25)
    at g.handleEvent (stimulus.js:260:18)
    at s.handleEvent (stimulus.js:31:25)
    at P (turbo.es2017-esm.js:312:16)
    at V.requestFinished (turbo.es2017-esm.js:700:9)
    at B.perform (turbo.es2017-esm.js:440:27)
```

PR #2338, opened 2022-08-22
Do not let Turbo to submit forms

louischan-oursky:
ref #2333

### DIFF
--- a/authui/src/form.ts
+++ b/authui/src/form.ts
@@ -1,25 +1,150 @@
+import { visit, clearCache } from "@hotwired/turbo";
 import { Controller } from "@hotwired/stimulus";
+import axios, { Method } from "axios";
+import {
+  disableAllButtons,
+  hideProgressBar,
+  showProgressBar,
+  progressEventHandler,
+} from "./loading";
+import { handleAxiosError } from "./messageBar";
 
-// We used to disable buttons during form submission.
-// It is now supported by Turbo natively.
-// See https://turbo.hotwired.dev/handbook/drive#form-submissions
-export class TurboformController extends Controller {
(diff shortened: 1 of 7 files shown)


authui/src/form.ts
```diff
@@ -1,25 +1,150 @@
...
+  };
+
+  async submitForm(e: Event) {
+    const form = e.currentTarget as HTMLFormElement;
+
+    if (form.querySelector('[data-turbo="false"]')) {
+      return;
+    }
+
+    // Do not stop propagation so that GTM can recognize the event as Form Submission trigger.
```
carmenlau, line 57: In the original `XHRSubmitFormController` implementation, we had the following code here. Do we need to add it back?
```
if (e.defaultPrevented) {
      return;
}
e.preventDefault();
```

authui/src/form.ts
```diff
@@ -1,25 +1,150 @@
...
+      form.addEventListener("submit", this.onSubmitCapture, true);
+      form.addEventListener("submit", this.onSubmit);
+    }
+
+    document.addEventListener("turbo:before-cache", this.beforeCache);
+  }
+
+  disconnect() {
+    for (const form of this.forms) {
+      form.addEventListener("submit", this.onSubmitCapture, true);
```
carmenlau, line 143: Should be `removeEventListener`

authui/src/form.ts
louischan-oursky: Intentionally removed because we always call preventDefault() in the capture event listener. If we were to keep it, we always early return.

louischan-oursky:
Updated.

PR #2344, opened 2022-08-22
Introduce app context in GTM event and provide collaborator role

carmenlau:
ref #2206 

I added the field `currentCollaboratorRole` role in `appAndSecretConfigQuery` to avoid additional request when the page is loaded. See if it looks good for you.

### DIFF
--- a/pkg/portal/graphql/app.go
+++ b/pkg/portal/graphql/app.go
@@ -8,6 +8,7 @@ import (
 	relay "github.com/authgear/graphql-go-relay"
 	"github.com/graphql-go/graphql"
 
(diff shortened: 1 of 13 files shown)


pkg/portal/graphql/app.go
```diff
@@ -322,6 +322,19 @@ var nodeApp = node(
 					return entry, nil
 				},
 			},
+			"currentCollaboratorRole": &graphql.Field{
```
louischan-oursky, line 325: What about

```graphql
type App implements Node {
  """"""
  collaborators: [Collaborator!]!

  viewer: Collaborator!
}
```

carmenlau: Updated!

Issue #2365, opened 2022-08-24
[Select Project] Text overflows when select project on Safari


joyzoursky:
## Environment
- Browser: Safari on Macbook Version 15.6.1 (17613.3.9.1.16)
- Server: Staging

## Description
When click and hold on the project on select project page (app list), the text will overflow out of the box. Seems only happening on Safari browsers.

## Steps to reproduce
1. Login to the portal on Safari
2. Create some apps with long app url
3. On the select project page, click on the apps with long url
4. See the url text overflow when it is on click 

## Expected result
Text should not overflow

## Video
https://user-images.githubusercontent.com/12539486/186416106-1c3f7aad-c1e2-45b8-b5a9-1d6bb9dd7f6b.mov

PR #2373, opened 2022-08-25
Fix text overflow in AppsScreen

andychow326:
ref #2365


portal/src/graphql/portal/AppsScreen.module.css
```diff
@@ -46,5 +46,5 @@
   color: #176df3;
   margin-bottom: 20px;
   font-size: 22px;
-  overflow-wrap: break-word;
+  word-break: break-word;
```
louischan-oursky, line 49: This usage is deprecated https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

Issue #2366, opened 2022-08-24
[UI Setting] End-user app "Log in" and "Sign up" buttons do not follow primary color in dark mode


joyzoursky:
## Environment
- Browser: Safari, Chrom on Macbook Pro
- Server: Staging

## Description
In dark mode, the Log in / Sign up button color does not follow the Primary color in UI Settings as in light mode.

## Steps to reproduce
1. Go to UI Settings in portal
2. Select Custom color and set the primary color for both dark and light mode
3. Go to the web endpoint of the app
4. Change the device theme to dark mode
5. See the Log in / Sign up page, but the buttons color are not the primary color as in light mode

## Expected result
The  Log in / Sign up buttons should use primary color in both light and dark mode as shown in the portal preview.

## Screenshot

<img width="1440" alt="Screenshot 2022-08-24 at 8 55 56 PM" src="https://user-images.githubusercontent.com/12539486/186425070-c71c162a-7e27-4c70-8813-c8be97a6d2aa.png">

<img width="1440" alt="Screenshot 2022-08-24 at 8 56 10 PM" src="https://user-images.githubusercontent.com/12539486/186425146-728ab3cd-cf37-4618-86ea-560ff98cb286.png">

andychow326:
@fungc-io Should the color of Log in / Sign up buttons be changed to primary?

fungc-io:
@karinyu 
For dark mode, I think the Login/Signup buttons in white looks better than in the theme color. i.e. the real page looks better than the preview. What do you think?

karinyu:
@fungc-io Agree.

I checked with the old design, it's using white instead of the theme color. Let's follow.

<img width="234" alt="Screenshot 2022-08-26 at 6 36 08 PM" src="https://user-images.githubusercontent.com/83946723/186886600-69ec9717-5324-40ee-a526-46edf3411ea0.png">

joyzoursky:
@andychow326 In this case please update the "Preview Mode" on the portal instead, thanks.

@fungc-io @karinyu And the light mode should keep using primary color for the buttons in active state right?

Also notice that in light mode, the "Sign up" text looks too light now on the web site, in the preview mode it is black color, seems the preview side is correct?

fungc-io:
@andychow326 the color of "Sign up" is correct in `demo-enterprise`. Can you check what's making it too light on `refuse-anchor-618`?

[...]

PR #2380, opened 2022-08-29
UI setting screen preview mode minor fix

andychow326:
ref #2366

### DIFF
--- a/portal/src/ThemePreviewWidget.module.css
+++ b/portal/src/ThemePreviewWidget.module.css
@@ -126,6 +126,14 @@
   color: var(--color-text-unshaded);
 }
 
+.secondaryTxt {
+  color: var(--color-text-shaded-1);
+}
+
+.dark .secondaryTxt {
+  color: var(--color-text-shaded-6);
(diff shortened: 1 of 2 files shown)


portal/src/ThemePreviewWidget.module.css
```diff
@@ -187,6 +187,11 @@
   color: var(--color-primary-unshaded);
 }
 
+.signinSignupLink.currentDarkMode {
```
louischan-oursky, line 194: We use `.dark` class-based approach to alter between light/dark theme here.

portal/src/ThemePreviewWidget.module.css
```diff
@@ -126,6 +126,10 @@
   color: var(--color-text-unshaded);
 }
 
+.secondaryTxt {
+  color: var(--color-text-shaded-1);
```
louischan-oursky, line 130: In dark theme, secondaryTxt is --color-text-shaded-6 according to authgear-dark-theme.css.

Issue #2358, opened 2022-08-24
[User Management] Values too long get chopped on Users table


joyzoursky:
## Environment
- Browser: Safari, Chrome on Macbook
- Server: Staging

## Description
Long values (e.g. email) on the Users table will get chopped directly, user may misunderstand it as the full display of the value.

## Steps to reproduce
1. Sign up for a user with long email on an app, e.g. joyz+mynameisstevejobs@oursky.com
2. Login to portal
3. Go to "User Management"
4. See the long email gets chopped on the table, becomes like joyz+mynameisstevejobs@ou

## Expected result
Should have better display for values too long, maybe truncate and show "..." on the table.

## Screenshot
<img width="1515" alt="Screenshot 2022-08-24 at 5 04 03 PM" src="https://user-images.githubusercontent.com/12539486/186378770-d8b43f17-d675-456d-ab84-76b18d36f3c2.png">

PR #2384, opened 2022-08-30
Truncate long text in the portal user list

carmenlau:
ref #2358

### DIFF
--- a/portal/src/graphql/adminapi/UsersList.tsx
+++ b/portal/src/graphql/adminapi/UsersList.tsx
@@ -289,7 +289,10 @@ const UsersList: React.VFC<UsersListProps> = function UsersList(props) {
         default:
           return (
             <div className={styles.cell}>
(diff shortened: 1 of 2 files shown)


portal/src/graphql/adminapi/UsersList.tsx
```diff
@@ -289,7 +289,10 @@ const UsersList: React.VFC<UsersListProps> = function UsersList(props) {
         default:
           return (
             <div className={styles.cell}>
-              {item[column?.key as keyof UserListItem] ?? USER_LIST_PLACEHOLDER}
+              <div className={styles.cellText}>
+                {item[column?.key as keyof UserListItem] ??
+                  USER_LIST_PLACEHOLDER}
+              </div>
```
louischan-oursky, line 295: Can we remove this extra div and write `cn(styles.cell, styles.cellText)` instead?

carmenlau: `text-overflow: ellipsis` doesn't work with `display: flex`, so the extra div is needed.

Issue #2351, opened 2022-08-23
Investigate how to add CDN for portal


louischan-oursky:
- [x] Make system-config.json cache-friendly
- [x] Make index.html always revalidate
- [x] Make all other assets cache-friendly

PR #2392, opened 2022-09-01
Make portal static assets cache friendly

louischan-oursky:
ref #2351


pkg/util/httputil/filesystem_cache.go
```diff
@@ -0,0 +1,123 @@
...
+
+	originalPath := r.URL.Path
+	filename := filepathutil.EscapePath(originalPath)
+
+	c.mutexForMapping.RLock()
+	mappedFilename, ok := c.mapping[filename]
+	c.mutexForMapping.RUnlock()
+
+	handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if ok {
```
carmenlau, line 80: The `ok` here indicates whether the filename exists in the cache map, so it will be false for the first time I visit the page. 

So when I call the system config, the first time will be 404.

Issue #2412, opened 2022-09-07
Add "Login with Passkey" to the login page


fungc-io:
The autofill function does not work in WebKit as expected (#2368). It will be useful if the login page shows a login with Passkey button

After clicking that button, all available passkeys on the device should be shown (provided by the platform). User can just pick one to login.

fungc-io:
@louischan-oursky 
https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=7553%3A113821

light:
![authUI-login-with-passkey-mobile](https://user-images.githubusercontent.com/15207470/189093628-502e428b-c62b-47db-9acf-175a3faf42a1.png)

dark:
![authUI-login-with-passkey-macbook-formx](https://user-images.githubusercontent.com/15207470/189093618-12b1f35a-23d0-4b6d-b93d-719a4f3f129b.png)

PR #2420, opened 2022-09-09
Add sign in with passkey button

louischan-oursky:
ref #2412


louischan-oursky:
Capitalized the word passkey in "Sign in with Passkey"

Issue #2417, opened 2022-09-08
Revise Widget to support extend and collapse


louischan-oursky:
- [x] Support extend and collapse in Widget
- [x] Support transition in extend/collapse
- [x] Replace existing usage of ExtendableWidget with Widget
- [x] Remove ExtendableWidget

carmenlau:
<img width="1460" alt="screenshot_2022-09-14_at_15 28 13" src="https://user-images.githubusercontent.com/388892/190171714-abc9b23e-a296-46ca-a9f5-76559de63e9c.png">

carmenlau:
@louischan-oursky Please take a look at PR #2439 when free.

PR #2426, opened 2022-09-13
Support extend/collapse in Widget

louischan-oursky:
ref #2417


portal/src/Widget.tsx
```diff
@@ -1,23 +1,80 @@
...
+import { IconButton, DefaultEffects } from "@fluentui/react";
 import cn from "classnames";
-import { DefaultEffects } from "@fluentui/react";
+
 import styles from "./Widget.module.css";
 
-export interface WidgetProps {
+interface WidgetProps {
   className?: string;
+  extendable?: boolean;
```
carmenlau, line 15: Currently, when `extendable` is empty, it implies the widget is non-extendable. And the `true` or `false` value of `extendable` is the initial value of `extended`? I think it is a bit confusing.

carmenlau (commented):
Now the SSO widget is not extensible when it is disabled initially. The steps to reproduce:

1. `apple` is disabled before entering the SSO page.
2. Enter the SSO page and click the toggle to enable `apple`.
3. The `apple` widget is not extensible.

portal/src/Widget.tsx
```diff
@@ -1,23 +1,80 @@
...
+  iconName: "ChevronDown",
+};
+
+// 16px top padding + 32px icon button height + 16px bottom padding
+const COLLAPSED_HEIGHT = 64;
+
 const Widget: React.VFC<WidgetProps> = function Widget(props: WidgetProps) {
-  const { className, children } = props;
+  const { className, extendable, children } = props;
+  const [extended, setExtended] = useState(extendable ?? true);
```
carmenlau, line 28: Currently, when `extendable` is empty, it implies the widget is non-extendable. And the `true` or `false` value of `extendable` is the initial value of `extended`? I think it is a bit confusing.

carmenlau (approved):

PR #2437, opened 2022-09-14
Update oauth client config to support third party client

carmenlau:
ref #2435

### DIFF
--- a/pkg/lib/config/secret_data.go
+++ b/pkg/lib/config/secret_data.go
@@ -441,3 +441,71 @@ func (c *WATICredentials) SensitiveStrings() []string {
 		c.WebhookAuth,
 	}
 }
+
+var _ = SecretConfigSchema.Add("OAuthClientCredentials", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"items": {
+			"type": "array",
+			"items": {
+				"type": "object",
+				"additionalProperties": false,
+				"properties": {
+					"client_id": {
+						"type": "string"
+					},
+					"keys": {
+						"type": "array",
+						"items": { "$ref": "#/$defs/JWK" },
+						"minItems": 1
+					}
+				},
+				"required": ["client_id", "keys"]
+			}
+		}
+	},
+	"required": ["items"]
+}
+`)
+
+type OAuthClientCredentials struct {
+	Items []OAuthClientCredentialsItem `json:"items,omitempty"`
+}
+
+func (c *OAuthClientCredentials) Lookup(clientID string) (*OAuthClientCredentialsItem, bool) {
+	for _, item := range c.Items {
+		if item.ClientID == clientID {
+			ii := item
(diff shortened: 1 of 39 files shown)


pkg/lib/config/secret_data.go
```diff
@@ -441,3 +441,71 @@ func (c *WATICredentials) SensitiveStrings() []string {
...
+	jwk.Set
+}
+
+var _ json.Marshaler = &OAuthClientCredentialsKeySet{}
+var _ json.Unmarshaler = &OAuthClientCredentialsKeySet{}
+
+func (c *OAuthClientCredentialsKeySet) MarshalJSON() ([]byte, error) {
+	return c.Set.(interface{}).(json.Marshaler).MarshalJSON()
+}
+func (c *OAuthClientCredentialsKeySet) UnmarshalJSON(b []byte) error {
```
louischan-oursky, line 552: Any specific reason why we need to implement custom JSON marshaling? jwk.Set should support JSON marshaling out-of-the-box https://pkg.go.dev/github.com/lestrrat-go/jwx/jwk#readme-get-the-json-representation-of-a-key

carmenlau: Using struct to wrap the interface doesn't work:( Seems we may need to customize the JSON marshaling in our case.

```
package main

import (
	"encoding/json"
	"fmt"
)

type A interface {
}

type AStruct struct {
	A
}

func (a *AStruct) MarshalJSON() ([]byte, error) {
	return a.A.(interface{}).(json.Marshaler).MarshalJSON()
}
func (a *AStruct) UnmarshalJSON(b []byte) error {
	return a.A.(interface{}).(json.Unmarshaler).UnmarshalJSON(b)
}

func main() {
	type B struct {
		Keys []string `json:"keys"`
	}

	type C struct {
		// Str string `json:"str"`
		AStruct
	}

	var a A
	a = B{
		Keys: []string{"a"},
	}

	bytes, err := json.Marshal(C{
		//		Str: "test",
		AStruct: AStruct{
			A: a,
		},
	})
	fmt.Println(string(bytes), err)
}
```

pkg/lib/config/secret_data.go
```diff
@@ -441,3 +441,71 @@ func (c *WATICredentials) SensitiveStrings() []string {
...
+		if item.ClientID == clientID {
+			ii := item
+			return &ii, true
+		}
+	}
+	return nil, false
+}
+
+func (c *OAuthClientCredentials) SensitiveStrings() []string {
+	return nil
```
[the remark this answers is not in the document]
carmenlau: Per our discussion, we are going to use the decoded value as the client secret, so I updated it to return the decoded values for `SensitiveStrings`. I also updated the other oct key set. See [f3de6f2](https://github.com/authgear/authgear-server/pull/2437/commits/f3de6f2815d4b4124b4c1359a27b35ecde9cecc0)

PR #2441, opened 2022-09-14
Update authz and token endpoints to support client secret for third party app

carmenlau:
ref #2438 
ref #2440

### DIFF
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -23,6 +23,7 @@ import (
 	"github.com/authgear/authgear-server/pkg/util/clock"
 	"github.com/authgear/authgear-server/pkg/util/duration"
 	"github.com/authgear/authgear-server/pkg/util/httputil"
+	"github.com/authgear/authgear-server/pkg/util/jwkutil"
 	"github.com/authgear/authgear-server/pkg/util/jwtutil"
 	"github.com/authgear/authgear-server/pkg/util/log"
 )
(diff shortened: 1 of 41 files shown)


pkg/lib/oauth/handler/handler_token.go
```diff
@@ -150,14 +151,21 @@ func (h *TokenHandler) doHandle(
...
 		}
-		if r.CodeVerifier() == "" {
-			return protocol.NewError("invalid_request", "PKCE code verifier is required")
+		// either client secret or PKCE code verifier is required
+		if client.IsFirstParty() {
+			if r.CodeVerifier() == "" {
+				return protocol.NewError("invalid_request", "PKCE code verifier is required")
+			}
+		} else {
+			if r.CodeVerifier() == "" && r.ClientSecret() == "" {
```
louischan-oursky, line 166: Third party client MUST provide client_secret. Third party MAY provide code_verifier

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -207,8 +215,28 @@ func (h *TokenHandler) handleAuthorizationCode(
...
+		}
+	}
+
+	// verify client secret
+	if r.ClientSecret() != "" {
+		credentialsItem, ok := h.OAuthClientCredentials.Lookup(client.ClientID)
+		if !ok {
+			return nil, protocol.NewError("invalid_request", "client secret is not supported for the client")
+		}
+		key, _ := credentialsItem.Set.Get(0)
```
louischan-oursky, line 231: To support secret rotation, we should accept any client secret, not just the first one.

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -207,8 +215,28 @@ func (h *TokenHandler) handleAuthorizationCode(
...
+		var expectedClientSecret []byte
+		err = key.Raw(&expectedClientSecret)
+		if err != nil {
+			return nil, err
+		}
+		if subtle.ConstantTimeCompare([]byte(r.ClientSecret()), []byte(expectedClientSecret)) != 1 {
```
louischan-oursky, line 237: We will show `k` to the developer, which is a base64URL encoded string. `key.Raw` will decode `k` into bytes. So they are not going to match.

carmenlau: Per our discussion. Since the decoded oct key is printable, we are going to use the decoded value as the client secret, like the webhook secret key.

PR #2449, opened 2022-09-15
Fix the grid layout in various pages

carmenlau:
ref #2417 

Some pages layout are broken after adding the content wrapper (74aca12bde3c615712fd1687540a3b3aa2d4f670)

Screenshot 2022-09-15 at 5 54 27 pm


louischan-oursky:
I added a prop to control the content layout instead.

Issue #2461, opened 2022-09-16
[Android Chrome, Login] No response upon clicking "Sign in with Passkey" 


JasonNS1425:
bug major, reported by qa

## Background
Build: Staging
Device: Chrome, Firefox on win 11, S21 FE (Android 12), iPhone 12 mini (iOS 16), Samsung note 9 (Android 12)

## Steps
1. Log in to the admin portal and enable the passkey feature of the application
2. Go to the login page of that application using Android Chrome browsers
3. Click the "Sign in with Passkey" option and see no response
4. Input the account email address to proceed to the password input screen
5. Click "Use Passkey instead" and see the passkey dialog can be triggered successfully

## Expected behavior
Should trigger the passkey dialog, or show proper error messages if the current device is not supported

## Screenshots

https://user-images.githubusercontent.com/67095825/190607758-2e5ff5a2-43b5-4098-b894-3fc16f9889cc.mp4

louischan-oursky:
Any error in console or network tab?

JasonNS1425:
@louischan-oursky Yes, I see ``passkey.ts:173 DOMException: Use of an empty `allowCredentials` list is not supported on this device.``

fungc-io:
Show an error message if empty `allowCredentials` list is not supported on the browser
> Sign in with passkey is not fully supported on this device. Please sign in with another method.

PR #2470, opened 2022-09-16
Handle unsupported empty allowCredentials error

louischan-oursky:
ref #2461

### DIFF
--- a/resources/authgear/templates/en/web/__error.html
+++ b/resources/authgear/templates/en/web/__error.html
@@ -58,22 +58,10 @@ $display_error = true }} {{ if eq .Error.reason "PasswordPolicyViolated" }}
 
     {{ if $display_error }} {{ $info := .Error.info }} {{ if eq .Error.reason
     "ValidationFailed" }} {{ range .Error.info.causes }} {{ if (eq .kind
-    "required") }} {{ if (and (call $.SliceContains .details.missing
-    "x_login_id" ) ($.x_login_id_input_type)) }} {{ if (eq
-    $.x_login_id_input_type "phone") }}
+    "required") }}
+    {{ if (call $.SliceContains .details.missing "x_login_id") }}
     <li>
-      {{ template "error-login-id-required" (dict "variant"
-      $.x_login_id_input_type) }}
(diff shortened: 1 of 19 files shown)


fungc-io:
@carmenlau can you take a look at this PR on Monday?
I want this UX get fixed asap.

resources/authgear/templates/en/web/__error.html
```diff
@@ -40,6 +40,9 @@
       {{ template "error-no-passkey" (dict "type"
       $.FirstNonPasskeyPrimaryAuthenticatorType) }}
     </li>
+    <li id="error-message-passkey-empty-allow-credentials" class="hidden">
+      {{ template "error-passkey-empty-allow-credentials" (dict "variant" $.x_login_id_input_type) }}
```
carmenlau, line 44: The value of `x_login_id_input_type` are `email`, `phone` and `text`. https://github.com/authgear/authgear-server/blob/d47e0223dc3e557dcb07f5bd156ad8bf0a320cdb/pkg/auth/handler/webapp/util.go#L52
So when the config is username only, the error message will be wrong.

Should we use `LoginIDInputVariant` if `x_login_id_input_type` is not `phone`, like the error `error-login-id-required`?

https://github.com/authgear/authgear-server/blob/d47e0223dc3e557dcb07f5bd156ad8bf0a320cdb/resources/authgear/templates/en/web/__error.html#L67

carmenlau (commented):
Besides the error message problem, the other part looks good. I think you can merge it after fixing the message.

resources/authgear/templates/en/web/__error.html
louischan-oursky: I revamped this part. Now error-login-id-required is greatly simplified. error-passkey-empty-allow-credentials use the same values as error-login-id-required. They now should display the contextual login ID type.

fungc-io: @carmenlau ,
@louischan-oursky has added more commits, please check when your are available. Thanks 🙏

carmenlau (approved):

Issue #2478, opened 2022-09-19
Add Passkey Screen and 2FA screen


louischan-oursky:
- [x] Add passkey screen
- [x] Remove passkey configuration from Authenticator screen
- [x] Add 2FA screen
- [x] Remove 2FA configuration from Authenticator screen

PR #2479, opened 2022-09-19
Add passkey screen and 2FA screen

louischan-oursky:
ref #2478


carmenlau (commented):
1. In the authenticators screen, the text color is incorrect when the phone login id is disabled in the feature config. 
2. Since there are more changes on the screen, see if we want to merge it to master or create a new feature branch for the ongoing changes.

louischan-oursky:
Fixed point 1 and changed the base branch to something else.

PR #2481, opened 2022-09-20
Introduce full-userinfo scope for third party app

carmenlau:
refs #2442

### DIFF
--- a/pkg/auth/handler/oauth/userinfo.go
+++ b/pkg/auth/handler/oauth/userinfo.go
@@ -4,7 +4,9 @@ import (
 	"encoding/json"
 	"net/http"
 
+	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/lib/infra/db/appdb"
+	"github.com/authgear/authgear-server/pkg/lib/oauth"
 	"github.com/authgear/authgear-server/pkg/lib/session"
 	"github.com/authgear/authgear-server/pkg/util/httproute"
 	"github.com/authgear/authgear-server/pkg/util/log"
@@ -17,7 +19,7 @@ func ConfigureUserInfoRoute(route httproute.Route) httproute.Route {
 }
 
 type ProtocolUserInfoProvider interface {
-	GetUserInfo(userID string) (map[string]interface{}, error)
+	GetUserInfo(userID string, clientLike *oauth.ClientLike) (map[string]interface{}, error)
 }
 
 type UserInfoHandlerLogger struct{ *log.Logger }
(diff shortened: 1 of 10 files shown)


carmenlau:
Only include essential standard attributes (email, email_verified, phone, phone_number_verified, preferred_username) in ID token

pkg/auth/handler/oauth/userinfo.go
```diff
@@ -30,13 +32,17 @@ type UserInfoHandler struct {
...
 	Database         *appdb.Handle
 	UserInfoProvider ProtocolUserInfoProvider
+	OAuth            *config.OAuthConfig
 }
 
 func (h *UserInfoHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	s := session.GetSession(r.Context())
+	scopes := oauth.SessionScopes(s)
+	clientID := oauth.ClientID(s)
+	client, _ := h.OAuth.GetClient(clientID)
```
louischan-oursky, line 42: client could also be nil if the third party client is deleted. We now assume a nil client is IDP session. If the client is deleted, we treat it as third party with no scopes.

pkg/lib/oauth/oidc/id_token.go
```diff
@@ -239,7 +239,7 @@ func (ti *IDTokenIssuer) PopulateUserClaims(token jwt.Token, userID string, nonP
 	return nil
 }
 
-func (ti *IDTokenIssuer) GetUserInfo(userID string) (map[string]interface{}, error) {
+func (ti *IDTokenIssuer) GetUserInfo(userID string, client *config.OAuthClientConfig, scopes []string) (map[string]interface{}, error) {
```
louischan-oursky, line 253: Maybe we can introduce a data structure to represent the following states

1. IDP session
2. First party client
3. Third party client

```golang
type ClientParty string
const ClientPartyFirst ClientParty = "first_party"
const clientPartyThird ClientParty = "third_party"

type ClientLike struct {
  ClientParty ClientParty
  Scopes []string
}

var ClientLikeIDPSession = &ClientLike{
  ClientParty: ClientPartyFirst,
  Scopes: []string{FullAccessScope},
}

var ClientLikeNotFound = &ClientLike{
  ClientParty: ClientPartyThird,
}

func (s *IDPSession) ClientLike() *ClientLike {
  return ClientLikeIDPSession
}

func (s *OfflineGrant) ClientLike(c *config.OAuthConfig) *ClientLike {
  client, ok := c.GetClient(s.ClientID)
  if !ok {
    return ClientLikeNotFound
  }
  return &ClientLike{
    ClientParty: client.ClientParty(),
    Scopes: s.Scopes,
  }
}
```

We can remove `IsFirstParty()` and use `ClientParty()` instead.

GetUserInfo should takes `(userID string, client *ClientLike)`

pkg/auth/handler/oauth/userinfo.go
carmenlau: The session should become invalid if the client is deleted, so I didn't handle this case.

Updated to introduce `ClientLike` to handle the case :)

pkg/lib/oauth/oidc/id_token.go
carmenlau: Rewrite the commits to introduce `ClientLike`, and see if it looks better!

Issue #2516, opened 2022-09-29
[Login, Ethereum & NFT] Can still access the MetaMask sign-in screen and send login request when the Ethereum & NFT feature is disabled


JasonNS1425:
bug minor, reported by qa

Build: Staging-29Sep2022
Device: Chrome, Firefox, Edge on win 11

Steps:
1. Visit any sign-in page of a project that has disabled the ETH login method
2. Append the following path to the workspace URL : "\<workspace URL\>/confirm_web3_account"
3. Connect to the MetaMask account to proceed
4. See the login request is successfully sent but no error message is shown (in the console, the response indicates an internal server error)

Expected behavior:
Should forbid users from accessing the MetaMask sign-in screen if the ETH login method is disabled

Screenshots:

https://user-images.githubusercontent.com/67095825/193027235-151afd75-85cc-45ae-9ce6-da9336e90f6e.mp4

Issue #2513, opened 2022-09-29
[Ethereum & NFT, Login] Incorrect alignment of the MetaMask icon on the "Install Wallet" screen


JasonNS1425:
ui must, reported by qa

Build: Staging-29Sep2022
Device: Chrome, Firefox, Edge on win 11

Steps:
1. Uninstall the MetaMask extension of the using browser (if any)
2. Go to any ETH login-enabled sign-in portal
3. Continue with MetaMask to proceed to the Install Wallet screen
4. See the Install MetaMask icon of the continue button is not left-aligned as shown in the design

Expected behavior:
Should follow the design and left-align the MetaMask icon

Screenshots:
![image](https://user-images.githubusercontent.com/67095825/193023336-344964ae-ebf0-493d-aa77-8bfcf5596d04.png)
![image](https://user-images.githubusercontent.com/67095825/193023383-f6dccb5e-ea05-44c9-9390-d3dd190d8c1c.png)

hochiw:
Handled via https://github.com/authgear/authgear-server/pull/2525

PR #2525, opened 2022-10-03
Safe-guard web3 pages when SIWE is not enabled

hochiw:
refs #2513 #2516

### DIFF
--- a/pkg/auth/handler/webapp/confirm_web3_account.go
+++ b/pkg/auth/handler/webapp/confirm_web3_account.go
@@ -80,11 +80,6 @@ func (h *ConfirmWeb3AccountHandler) ServeHTTP(w http.ResponseWriter, r *http.Req
 		return
 	}
 	identities := h.AuthenticationConfig.Identities
-	if identities == nil {
-		http.Error(w, "Failed to parse identity config", http.StatusInternalServerError)
-		return
-	}
-
 	isSIWEEnabled := false
 	for _, i := range identities {
 		if i == model.IdentityTypeSIWE {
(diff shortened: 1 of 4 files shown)


pkg/auth/handler/webapp/confirm_web3_account.go
```diff
@@ -76,6 +79,24 @@ func (h *ConfirmWeb3AccountHandler) ServeHTTP(w http.ResponseWriter, r *http.Req
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
+	identities := h.AuthenticationConfig.Identities
+	if identities == nil {
+		http.Error(w, "Failed to parse identity config", http.StatusInternalServerError)
+		return
+	}
```
louischan-oursky, line 86: No need to check nil, `range nil` is perfectly fine.

pkg/auth/handler/webapp/confirm_web3_account.go
```diff
@@ -76,6 +79,24 @@ func (h *ConfirmWeb3AccountHandler) ServeHTTP(w http.ResponseWriter, r *http.Req
...
+	isSIWEEnabled := false
+	for _, i := range identities {
+		if i == model.IdentityTypeSIWE {
+			isSIWEEnabled = true
+			break
+		}
+	}
+
+	if !isSIWEEnabled {
+		http.Error(w, "404 page not found", http.StatusNotFound)
```
louischan-oursky, line 97: Do not return 404. Use `http.Redirect(w, r, "/", http.StatusFond)` instead.

pkg/auth/handler/webapp/missing_web3_wallet.go
```diff
@@ -57,6 +60,25 @@ func (h *MissingWeb3WalletHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
...
+	isSIWEEnabled := false
+	for _, i := range identities {
+		if i == model.IdentityTypeSIWE {
+			isSIWEEnabled = true
+			break
+		}
+	}
+
+	if !isSIWEEnabled {
+		http.Error(w, "404 page not found", http.StatusNotFound)
```
louischan-oursky, line 79: Do not return 404. Use `http.Redirect(w, r, "/", http.StatusFond)` instead.

PR #2536, opened 2022-10-06
Add back missing case check for SIWE identity type

hochiw:
- Prevent yielding error after https://github.com/authgear/authgear-nft-indexer/pull/25#issuecomment-1267896019 due to no nft collection
- Added missing case check for SIWE identity in AllStandardClaims


louischan-oursky (commented):
Drop the commit that checks =0 case, we are going to make the API server to return empty.

hochiw:
dropped :adore2x:

Issue #2480, opened 2022-09-20
Add LoginMethodConfigurationScreen


louischan-oursky:
- [x] All widgets not collapsable
- [x] Review Toggle inlineLabel
- [x] Show Password settings in MFA
- [x] Implement new design
  - [x] Disabled options are put at the end

fungc-io:
@louischan-oursky 
The design is changed as follow: https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=7709%3A116632

When the user click into the login methods page from the nav bar,
It will show the "Configuration" with the blue box on top.

When the user clicked "Change login method",
It will show the matrix page,
after they chosen the method and click next,
it will go back to the config screen with the relevant settings.
The whole thing is saved when they press the save button on top.

PR #2538, opened 2022-10-06
Rewrite login method chooser

louischan-oursky:
ref #2480


louischan-oursky:
Applied opacity to the whole button

Issue #2539, opened 2022-10-06
Avoid injecting CSP for wallet confirmation page


hochiw:
- Firefox and Safari block metamask injection
- Drop CSP configuration specifically for wallet confirmation page

hochiw:
Dupe #2507

Issue #2507, opened 2022-09-29
[Firefox, Ethereum & NFT, Login] Unable to detect installed MetaMask Firefox Extension during sign-in/up


JasonNS1425:
bug critical, reported by qa

Build: Staging-29Sep2022
Device: Chrome, Firefox, Edge on win 11

Steps:
1. Install the MetaMask extension in the Firefox browser (e.g. 
2. Visit any Ethereum-login-enabled login portal in the browser
3. Log in using the MetaMask account
4. See the MetaMask extension not detected error

Expected behavior:
Should be able to log in using the MetaMask account on all browsers supported by MetaMask

Screenshots:

https://user-images.githubusercontent.com/67095825/193016662-9b646d70-243d-4a31-bd33-58447e5229a2.mp4

hochiw:
This is due to the strict CSP we have for authgear to prevent script injection. This also happens with Safari (i.e metamask mobile), see how we want to handle this. @louischan-oursky 

refs 
https://github.com/MetaMask/detect-provider/issues/31
https://github.com/authgear/authgear-server/blob/master/pkg/lib/web/csp.go

louischan-oursky:
The team decided to make the following changes

- Relax CSP (or disable CSP, if necessary) on SIWE pages
- Keep login page CSP enabled

hochiw:
Interesting found on the CSP [draft standard](https://w3c.github.io/webappsec-csp/#allow-all-inline),

`
If expression matches the nonce-source or hash-source grammar, return "Does Not Allow".
`
Since nonce is provided for authgear's CSP, `unsafe-inline` will be ignored, so we will need to not provide nonce for the SIWE pages.

PR #2546, opened 2022-10-07
Prevent CSP violation for MetaMask on SIWE pages

hochiw:
# Notes
- Show dash and dummy logo if no account has been connected yet 
- If account connection is rejected, prompt again for connection when Next button is clicked and will proceed directly to the SIWE flow

@fungc-io @louischan-oursky see if these changes make sense to you

Screenshot 2022-10-07 at 10 23 00 AM

refs #2507 #2539

### DIFF
--- a/authui/src/web3.ts
+++ b/authui/src/web3.ts
@@ -111,51 +111,6 @@ function handleError(err: unknown) {
   }
   return;
 }
-
-export class WalletConnectionController extends Controller {
-  static targets = ["button"];
-  static values = {
-    provider: String,
-  };
-
-  declare buttonTarget: HTMLButtonElement;
-
-  declare providerValue: string;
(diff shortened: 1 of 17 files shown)


authui/src/web3.ts
```diff
@@ -192,6 +192,7 @@ export class WalletConfirmationController extends Controller {
   }
 
   connect() {
+    this.displayedTarget.textContent = "-";
```
louischan-oursky, line 195: Better to put the dash in the template.

And we should be consistent in where to set this value. For example, when I disconnect, it does not show a dash again. Ideally, the textContent should be updated in the same function. This would eliminate a lot of potential inconsistent UI state changes.

```typescript
function _getAccount() {
  this.displayedTarget.textContent = "-";
  // ... remaining code
  // textContent must either be "-" or some address after this function returns.
}
```

Issue #2514, opened 2022-09-29
[User details, Identities] Inconsistent width of the remove button for email and username identity


JasonNS1425:
ui must, reported by qa

Build: Staging-29Sep2022
Device: Chrome, Firefox, Edge on win 11

Steps:
1. Log in to the admin portal and select any project space
2. Go to the User Management page and select  any user with both username and email login ids
3. Switch to the Identities tab
4. See the remove buttons of the email and the username identity have different widths

Expected behavior:
Should have consistent width for the same type of button

Screenshots:
![image](https://user-images.githubusercontent.com/67095825/193025101-f2222cc2-47d9-4c53-8ea8-a3ee9867967c.png)

PR #2550, opened 2022-10-07
Align remove button to the right in UserDetailsIdentities page

andychow326:
ref #2514


louischan-oursky (commented):
Try an approach that does not require a hidden button

andychow326:
> Try an approach that does not require a hidden button

PR #2552, opened 2022-10-10
Separate trimmed zero hexstring parser and make ContractID support parameters

hochiw:
- Apparently we have to keep the leading zeroes for the contract addresses
- Added query params for contractID to handle putting token id information into the url, this respects the `parameters` field for the eip681 [spec](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-681.md)

### DIFF
--- a/pkg/util/web3/contractid.go
+++ b/pkg/util/web3/contractid.go
@@ -12,9 +12,10 @@ type ContractID struct {
 	Blockchain string
 	Network    string
 	Address    string
+	Query      *url.Values
 }
 
-func NewContractID(blockchain string, network string, address string) (*ContractID, error) {
+func NewContractID(blockchain string, network string, address string, query *url.Values) (*ContractID, error) {
 	hexaddr, err := hexstring.Parse(address)
 	if err != nil {
 		return nil, err
@@ -24,6 +25,7 @@ func NewContractID(blockchain string, network string, address string) (*Contract
 		Blockchain: blockchain,
 		Network:    network,
 		Address:    hexaddr.String(),
+		Query:      query,
 	}, nil
 }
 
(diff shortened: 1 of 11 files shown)


pkg/util/web3/eip681.go
```diff
@@ -13,17 +13,42 @@ import (
...
 	ChainID int
 	Address string
+	Query   *url.Values
 }
 
 func (eip681 *EIP681) URL() *url.URL {
+	query := ""
+	if eip681.Query != nil {
+		query = eip681.Query.Encode()
+	}
```
louischan-oursky, line 23: Call Encode() directly. No need check nil. https://go.dev/play/p/mLQbvdoPYXx

pkg/util/web3/eip681.go
```diff
@@ -13,17 +13,42 @@ import (
...
+	return &EIP681{
+		ChainID: chainID,
+		Address: addrHex.String(),
+		Query:   query,
+	}, nil
+}
+
 func ParseEIP681(uri string) (*EIP681, error) {
-	protocolURI := strings.Split(uri, ":")
+	queryString := strings.Split(uri, "?")
```
louischan-oursky, line 50: Do not parse url manually. Use url.Parse

portal/src/util/eip681.ts
```diff
@@ -3,13 +3,15 @@ const ETHEREUM_ADDRESS_REGEX = /^0x[a-fA-F0-9]{40}$/;
...
+  query?: URLSearchParams;
 }
 
 export function parseEIP681(
   url: string,
   skipAddressCheck: boolean = false
 ): EIP681 {
-  const protocolURI = url.split(":");
+  const queryURI = url.split("?");
+  const protocolURI = queryURI[0].split(":");
```
louischan-oursky, line 14: Use `new URL()` to parse URI. Do not parse URL manually.

pkg/util/web3/contractid.go
```diff
@@ -12,9 +12,10 @@ type ContractID struct {
 	Blockchain string
 	Network    string
 	Address    string
+	Query      *url.Values
```
louischan-oursky, line 15: Use `url.Values` instead of `*url.Values`. `url.Values` is a map so it is a pointer type already. `*url.Values` is a pointer to a pointer, too complicated.

pkg/util/web3/eip681.go
```diff
@@ -13,17 +13,42 @@ import (
 type EIP681 struct {
 	ChainID int
 	Address string
+	Query   *url.Values
```
louischan-oursky, line 16: Use `url.Values` directly.

hochiw:

PR #2556, opened 2022-10-10
Remove legacy is_first_party flag from the OAuth config

carmenlau:
ref #2555


louischan-oursky:
I think this can be merged now.

Issue #2554, opened 2022-10-10
Update `authgear init` to make it deployment friendly


carmenlau:
- Combine `init authgear.yaml` and `init authgear.secrets.yaml` to `init`
- Add `--for-helm-chart=true` to generate config for helm chart deployment
- Update questions of command
    - When `--for-helm-chart=true`
        -  APP ID
        - HTTP origin of Authgear
        - HTTP origin of Portal
        - Turn off email verification?
    - When `--for-helm-chart=false` (default)
        -  APP ID
        - HTTP origin of Authgear
        - HTTP origin of Portal
        - DB credentials
        - redis credentials
        - audit log credentials
        - elasticsearch
- When `--for-helm-chart=true`, don't include the credentials in `authgear.secrets.yaml`. They should be provided via the helm chart vendors config.
- Generate OAuth client for portal with application type traditional web app

PR #2567, opened 2022-10-11
Update `authgear init` to make it deployment friendly

carmenlau:
ref #2554


cmd/authgear/config/marshal.go
```diff
@@ -3,21 +3,24 @@ package config
...
 		return err
 	}
 
-	if outputPath == "-" {
+	if outputFolderPath == "-" {
 		_, err = os.Stdout.Write(yaml)
 		return err
 	}
 
+	outputPath := filepath.Join(outputFolderPath, fileName)
```
louischan-oursky, line 17: The command should be smart enough to create the directory if it does not exist. Use os.MkdirAll to mimic `mkdir -p` in case the argument has more than 1 level of directories.

carmenlau:
Updated:)

Issue #2569, opened 2022-10-11
Cannot sign-in the second time with SIWE


hochiw:
Seems the server is trying to insert an identical row to the identity table

PR #2570, opened 2022-10-12
Align ethereum address format to use EIP-55 in front end level

hochiw:
refs #2569

### DIFF
--- a/pkg/util/web3/eip55.go
+++ b/pkg/util/web3/eip55.go
@@ -1,0 +1,54 @@
+package web3
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/authgear/authgear-server/pkg/util/hexstring"
+	"golang.org/x/crypto/sha3"
+)
+
+// https://eips.ethereum.org/EIPS/eip-55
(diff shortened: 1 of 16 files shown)


pkg/util/web3/eip55.go
```diff
@@ -0,0 +1,54 @@
...
+var parseRegExp = regexp.MustCompile(`^(0x)0*([0-9a-fA-F]+)$`)
+
+func NewEIP55(s string) (EIP55, error) {
+	if !parseRegExp.MatchString(s) {
+		return "", fmt.Errorf("hex string must match the regexp %q", parseRegExp)
+	}
+
+	strippedHex := strings.ToLower(strings.ReplaceAll(s, "0x", ""))
+
+	sha := sha3.NewLegacyKeccak256()
```
louischan-oursky, line 24: Can we use go-ethereum to help us implement the parsing? We still want to keep `type EIP55 string`, only use go-ethererum for the implementation.

louischan-oursky:
change the type of pkg/api/model.SIWEWallet.Address to EIP55

PR #2572, opened 2022-10-12
Support non-http origin

### LINKED ISSUE #2571: Allow origin to be non-http protocol
Some technology uses non-http protocol (such as https://tauri.app/, which uses `tauri://localhost`). The origin is `tauri://localhost` and it is unsupported at the moment.

- [ ] Derive allowed origin for any protocol, not just http: and https:
- [ ] Update the underlying library originmatcher to support any protocol.

louischan-oursky:
fix #2571


pkg/lib/infra/middleware/cors.go
```diff
@@ -30,7 +30,7 @@ func (m *CORSMiddleware) Handle(next http.Handler) http.Handler {
 		allowedOrigins := m.Config.AllowedOrigins
 		allowedOrigins = append(allowedOrigins, m.CORSAllowedOrigins.List()...)
 		for _, oauthClient := range m.OAuthConfig.Clients {
-			allowedOrigins = append(allowedOrigins, oauthClient.RedirectURIHosts()...)
```
carmenlau, line 33: The PR looks good with one minor comment.

We updated to add `RedirectURIs` directly into `allowedOrigins` directly without trimming the path here. I checked the updates in [originmatcher](https://github.com/iawaknahc/originmatcher) it should be fine. But it is better to add test case in `originmatcher` to indicate we accept origin with path. So when we want to update `originmatcher` in the future, we remember to keep this behavior.

Thought?

carmenlau (approved):

pkg/lib/infra/middleware/cors.go
louischan-oursky: Sure.

Issue #2576, opened 2022-10-13
Self-heal from insufficient fund of subscription


louischan-oursky:
The event we receive is in this sequence.

- checkout.session.completed
- customer.subscription.created

The status is `incomplete`. According to the Stripe, the status will become `incomplete_expired` and this is the terminal state.

So I think we should stick to our current handling of `incomplete`, which is do nothing. When we see `incomplete_expired` we should let the developer to retry.

louischan-oursky:
- Research if we can offer a link to retry the payment only.

PR #2577, opened 2022-10-13
Handle stuck incomplete subscription

louischan-oursky:
ref #2576

### DIFF
--- a/pkg/portal/transport/stripe_webhook_handler.go
+++ b/pkg/portal/transport/stripe_webhook_handler.go
@@ -153,16 +153,56 @@ func (h *StripeWebhookHandler) handleCheckoutSessionCompletedEvent(event *libstr
 }
 
 func (h *StripeWebhookHandler) handleCustomerSubscriptionEvent(event *libstripe.CustomerSubscriptionEvent) error {
-	if !event.IsSubscriptionActive() {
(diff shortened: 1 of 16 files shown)


pkg/portal/transport/stripe_webhook_handler.go
```diff
@@ -153,16 +153,56 @@ func (h *StripeWebhookHandler) handleCheckoutSessionCompletedEvent(event *libstr
...
+		WithField("stripe_subscription_id", event.StripeSubscriptionID).
+		WithField("stripe_subscription_status", event.StripeSubscriptionStatus).
+		Info("unhandled subscription status")
+	return nil
+}
+
+func (h *StripeWebhookHandler) handleIncompleteExpiredSubscriptionEvent(event *libstripe.CustomerSubscriptionEvent) error {
+	err := h.Database.WithTx(func() error {
+		// Mark checkout session as cancelled
+		err := h.Subscriptions.UpdatedSubscriptionCheckoutStatusToCancelled(
```
carmenlau, line 183: For diagnostic, is it better to introduce a new status for this case instead of using cancel? Maybe "subscription_incomplete_expired"?

louischan-oursky: Added `expired`

PR #2583, opened 2022-10-14
Block SIWE if user signs in with another network

hochiw:
Screenshot 2022-10-14 at 6 28 23 PM

refs #2582


pkg/lib/feature/siwe/errors.go
```diff
@@ -5,5 +5,7 @@ import (
 )
 
 var InvalidNonce = apierrors.Forbidden.WithReason("InvalidNonce")
+var InvalidNetwork = apierrors.BadRequest.WithReason("InvalidNetwork")
 
 var ErrNonceNotFound = InvalidNonce.NewWithCause("nonce is expired or invalid", apierrors.StringCause("NonceNotFound"))
+var ErrMismatchNetwork = InvalidNetwork.NewWithCause("network does not match expected network", apierrors.StringCause("MismatchNetwork"))
```
louischan-oursky, line 11: Remove this.

pkg/lib/feature/siwe/errors.go
```diff
@@ -5,5 +5,7 @@ import (
 )
 
 var InvalidNonce = apierrors.Forbidden.WithReason("InvalidNonce")
+var InvalidNetwork = apierrors.BadRequest.WithReason("InvalidNetwork")
```
louischan-oursky, line 8: Construct this error with details. The details should contain the expected network. In UI, tell the user the expected network (both the numeric ID and the name)

Issue #2578, opened 2022-10-13
Show incomplete subscription in Portal


fungc-io:
Continue from the problem in #2576 

## Problem

Currently when a subscription is created and before the payment goes through,
The status of the subscription is `incomplete` and the billing page will be showing "Processing payment"
<img width="396" alt="image" src="https://user-images.githubusercontent.com/15207470/195637990-312811d1-a225-4cc6-9a9b-a74c2102e603.png">

If there is a problem in the payment process, let's say the payment card do not have sufficient credit. The billing page is stuck until we got the webhook of `incomplete_expired` 24 hours later from Stripe.

## Todo

1. See if the Stripe have any retry mechanism in case of payment failure
2. We the user visits the billing page, we can query Stripe for the latest status instead of waiting for the expiry webhook.
3. Change the UI so that the user understand the subscription is pending for payment, and provide affordance for retry and cancel.
    - See if Stripe provides a way to retry the payment/re-enter the payment method for that particular subscription. 
    - Allow the user to cancel and retry when the process takes too long. The previous subscription must be canceled first to avoid double charge.

fungc-io:
@louischan-oursky 
I've updated the design for "Processing payment"
https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?node-id=8190%3A92757

see if there is any problem?

fungc-io:
Thanks @carmenlau ,
I've tested with the Stripe card of `Decline after attaching` (`4000 0000 0000 0341` Attaching this card to a [Customer](https://stripe.com/docs/api/customers) object succeeds, but attempts to charge the customer fail.)
It is working as expected.

But then I tried with another test case:
`3DS2 Required`: `4000000000003220`
It's now stuck at the incomplete state. 

On Stripe Dashboard:
<img width="503" alt="image" src="https://user-images.githubusercontent.com/15207470/199179469-98c759fc-a213-472c-b16e-f26b9d9ffc75.png">
<img width="1028" alt="image" src="https://user-images.githubusercontent.com/15207470/199179561-02fc04ff-7781-43be-9d87-ad4d4578aa14.png">


Any advise on the design?

fungc-io:
Perhaps after the payment intent is created 5 minutes,
If the payment is still `incomplete`,
show the "Cancel" button alongside the "Payment in progress" message

PR #2613, opened 2022-10-25
Allow cancel failed subscription

carmenlau:
refs #2578 

1. The selected plan title is not shown in the processing payment screen since we didn't store this information. See if it is acceptable. (See screenshots)
2. I only allow users to cancel the payment themself when the error is`card_declined` (https://stripe.com/docs/error-codes
). For the other errors, the support screen will be shown instead. Just afraid some unknown case may cause double charging. Open to discussion.

**Processing payment**

Screenshot 2022-10-25 at 5 30 13 pm

**card_declined error**

Screenshot 2022-10-25 at 5 30 31 pm

**other errors**

Screenshot 2022-10-25 at 5 30 45 pm


pkg/portal/graphql/app.go
```diff
@@ -298,6 +299,12 @@ var nodeApp = node(
 
 					stripeError, err := ctx.StripeService.GetLastPaymentError(*customerID)
 					if err != nil {
+						if errors.Is(err, libstripe.ErrNoSubscription) {
```
louischan-oursky, line 302: I think GetLastPaymentError should return (nil, nil) for this case. So this checking is not needed here.

portal/src/types.ts
```diff
@@ -833,3 +833,10 @@ export interface TutorialStatusData {
     invite?: boolean;
   };
 }
+
+// https://stripe.com/docs/error-codes
+export type StripeErrorCode = "card_declined";
+
+export interface StripeError {
```
louischan-oursky, line 840: We can use `GQL_StripeError` directly. It is an alias of stripe.Error

PR #2623, opened 2022-10-31
GraphiQL

### LINKED ISSUE #2622: Change page title of GraphiQL explorer
The page title of the GraphiQL explorer of the Admin API is called "GraphiQL"
Change it to "GraphiQL: Admin API - Authgear" for more context.

louischan-oursky:
fix #2622

### DIFF
--- a/pkg/util/graphqlutil/graphiql.go
+++ b/pkg/util/graphqlutil/graphiql.go
@@ -1,0 +1,65 @@
+package graphqlutil
+
+import (
+	htmltemplate "html/template"
+	"net/http"
+)
+
+var graphiqlTemplate = htmltemplate.Must(htmltemplate.New("graphiql").Parse(`<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<title>{{ .Title }}</title>
+	<style>
+		body {
+			padding: 0;
+			margin: 0;
+			min-height: 100vh;
+		}
+		#root {
+			height: 100vh;
+		}
+	</style>
+	<script
+		crossorigin
+		src="https://unpkg.com/react@17.0.2/umd/react.production.min.js"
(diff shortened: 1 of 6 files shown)


pkg/util/graphqlutil/graphiql.go
```diff
@@ -0,0 +1,65 @@
...
+		crossorigin
+		src="https://unpkg.com/graphiql@2.0.13/graphiql.min.js"
+	></script>
+	<script>
+		ReactDOM.render(
+			React.createElement(GraphiQL, {
+				fetcher: GraphiQL.createFetcher({
+					url: "",
+				}),
+				query: (new URLSearchParams(window.location.search)).get("query") || "",
```
carmenlau, line 46: Currently, the query function is partly broken. When `query`, `variables`, and `operationName` are provided, only `query` is pre-filled. When I updated the `query` in the UI, the URL query was not updated either. 

I checked the source code of `graphql-go` and found they have some logic to sync the query parameter.
refs: https://github.com/graphql-go/handler/blob/f96ffdde846be75dd40541aebd1dba604f274817/graphiql.go#L21

So I think maybe we should either:
1. Drop query parameter support in the new version? We may need to discuss on this
2. Repeat what `graphql-go` is doing, so the query is synced in 2 ways.

Thought?

carmenlau: I tried repeating what `graphql-go` is doing and it seems working well. See if it looks good for you?

https://github.com/carmenlau/authgear-server/commits/2622-graphiql

louischan-oursky: I did not drop it because we need `query` to support a use case in the portal. In the Admin API page in the portal, we have an URL to open GraphiQL with a default query. I did not copy graphql-go was doing because GraphiQL now have tabs, which makes sync query with an arbitrary tab looks weird. So in this sense the meaning of `query` changes, it means initial query now.

Issue #2618, opened 2022-10-28
Detect alchemy error


louischan-oursky:

fungc-io:
@louischan-oursky 
Please change it such that:

1. Shorten the timeout to 5 seconds
2. Show a dialog saying:
> ## Contact us 
> 
> The collection you entered cannot be added.
> Please contact us if this is the contract you want to use it in your project.
> 
> [Contact us] [Cancel]

Thanks! 🙏

PR #2624, opened 2022-10-31
Show error in dialog

louischan-oursky:
ref #2618

### DIFF
--- a/portal/src/graphql/portal/Web3ConfigurationAddCollectionSection.tsx
+++ b/portal/src/graphql/portal/Web3ConfigurationAddCollectionSection.tsx
@@ -184,47 +184,49 @@ const Web3ConfigurationAddCollectionForm: React.VFC<AddCollectionSectionProps> =
     );
 
     return (
-      <form
-        className={cn(styles.addCollection, className)}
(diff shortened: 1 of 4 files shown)


portal/src/graphql/portal/Web3ConfigurationAddCollectionSection.tsx
```diff
@@ -222,6 +234,10 @@ const Web3ConfigurationAddCollectionForm: React.VFC<AddCollectionSectionProps> =
           onContinue={handleAddCollection}
           onDismiss={dismissDialogs}
         />
+        <Web3ConfigurationContactUsDialog
+          isVisible={!contactUsDialogHidden}
+          onDismiss={onDismissContactUsDialog}
+        />
```
carmenlau, line 240: One minor ui comment. The `form` element is `display:flex` with `gap`. Maybe we should put the dialog outside the `form`. 

The problem is when the dialog is shown, 1 more element is added to the `form` which makes the form move.

Before the dialog is shown:
Screenshot 2022-10-31 at 10 57 41 am

After the dialog is shown:
Screenshot 2022-10-31 at 10 57 33 am

PR #2627, opened 2022-10-31
Specify single sign on with Authgear

carmenlau:
- The first part of the spec is copied from the discussion note, except **1st first-party OIDC-Compatible client** is removed.
- **Implementation details** section is newly added to specify the implementation details.


docs/specs/oidc-sso.md
```diff
@@ -0,0 +1,176 @@
...
+        - If the refresh token's `sso_enabled` is `false` (default), revoke the refresh token only
+
+- "Continue as" screen
+    - When the enduser chooses **Login with another account**, the existing IdP session (including its refresh token with `sso_enabled=true`) will be revoked after the new authentication is completed.
+    - Add help text: `After you have signed in another account, your current session as user@example.com will terminate`.
+
+- Update application config
+    - For `x_application_type=spa`, change `post_logout_uris` to optional. Change the label in the Portal to **Post Logout Redirect URI (Legacy)**. See [Web SDK](#web-sdk) for details.
+
+- Session listing
```
fungc-io, line 156: - SSO enabled: 1 entry per device
- Without SSO: 1 entry for each app on each device
- Session created by biometric will be handled separately.

docs/specs/oidc-sso.md
```diff
@@ -0,0 +1,176 @@
...
+    - Revoke refresh token
+        - If the refresh token's `sso_enabled` is `true`, invalidates its IdP session and all its siblings with `sso_enabled=true`
+        - If the refresh token's `sso_enabled` is `false` (default), revoke the refresh token only
+
+- "Continue as" screen
+    - When the enduser chooses **Login with another account**, the existing IdP session (including its refresh token with `sso_enabled=true`) will be revoked after the new authentication is completed.
+    - Add help text: `After you have signed in another account, your current session as user@example.com will terminate`.
+
+- Update application config
+    - For `x_application_type=spa`, change `post_logout_uris` to optional. Change the label in the Portal to **Post Logout Redirect URI (Legacy)**. See [Web SDK](#web-sdk) for details.
```
fungc-io, line 154: Show a description of which optional from which SDK version

docs/specs/oidc-sso.md
louischan-oursky: I suggest this.

Without SSO, we group refresh tokens for each app on each device and show the group as 1 entry. The IDP session is not grouped. For existing apps, the IDP session will be shown as 1 entry. For new apps, we do not create IDP session so it is not an issue.

This means visually we do not have group in the session list.

chpapa: > Without SSO, we group refresh tokens for each app on each device and show the group as 1 entry.

But can we match device reliably? If not, I think if an app only calls login without logout will have multiple entry, is fine...? (Most apps won't keep logout users right?)

carmenlau:
@louischan-oursky @fungc-io 
Updated the event like this, see if it looks good. e03055e80be60d42549981e0deaddc07ca069e75

docs/specs/oidc-sso.md
carmenlau: Conclusion: SSO disabled refresh token will not be grouped.

Issue #2628, opened 2022-11-01
Include the relevant information as param in the login link


fungc-io:
- Preserve `client_id` and `redirect_uri` query during the whole login flow.
    - After authentication, redirect the enduser based on following whitelisted
        1. Redirect URL in query param (must be whitelisted) 
        2. Default redirect URL of the client
        3. Post-login URL
        4. `/settings`
    - Default redirect URL of the client
        - In the config, the first item of **redirect_uris** is the default implicitly
        - Update the Portal UI to allow setting the default (#2629)

PR #2648, opened 2022-11-02
Redirect the user based on query after authentication

carmenlau:
ref #2628 

----

**Some cases to test**

1. When the unauthenticated user visits an auth required path (e.g. "/settings/identity"),
they should be redirected back to the original path after login. (This was broken and it is fixed in this PR)
1. When the unauthenticated user visits "/", they should be redirected to "/settings" after login.
1. Start authentication via SDK, the enduser should be redirected back to the client app.
1. The enduser visits login link with `client_id` and `redirect_uri`, they should be redirected back to `redirect_uri`.
1. The enduser visits login link with `client_id` only, they should be redirected back to default redirect uri.
1. The enduser visits login link with `redirect_uri` only, they should be redirected back to "/settings".

More cases are welcome!!

### DIFF
--- a/pkg/auth/endpoints.go
+++ b/pkg/auth/endpoints.go
@@ -44,8 +44,17 @@ func (p *EndpointsProvider) RevokeEndpointURL() *url.URL     { return p.urlOf("o
 func (p *EndpointsProvider) JWKSEndpointURL() *url.URL       { return p.urlOf("oauth2/jwks") }
 func (p *EndpointsProvider) UserInfoEndpointURL() *url.URL   { return p.urlOf("oauth2/userinfo") }
 func (p *EndpointsProvider) EndSessionEndpointURL() *url.URL { return p.urlOf("oauth2/end_session") }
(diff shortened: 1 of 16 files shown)


louischan-oursky (commented):
I think we can remove ClientIDMiddleware as well. It is unused.

carmenlau:
Todo:
- Remove client id from cookie
  - check if the client id is preserved in the /settings paths
  - Remove ClientIDMiddleware and read client id from query
- Remove `x_step` which the interaction is completed

carmenlau:
@louischan-oursky When I remove the `ClientIDMiddleware`, I found that we use it in the `event.Service`. We had the client id in the webhook event. If we remove the client id from the cookie. When the user visits the page by inputting the URL without the client id. The client id will be removed from the webhook event. So it seems better to keep it, thought?

I removed it from this branch, FYI: https://github.com/carmenlau/authgear-server/tree/2628-client-id-redirect-uri-query-2

louischan-oursky:
OK. Let's keep it.

Issue #2649, opened 2022-11-03
Disable "Create Project" Button when creating project


fungc-io:
The button is still active when the project is being created.

PR #2651, opened 2022-11-04
Disable "Create Project" Button when creating project

andychow326:
ref #2649


louischan-oursky:
Do not use ButtonWithLoading. Simply disable the button while loading

andychow326:
@louischan-oursky Done! please take a look :adore2x:

Issue #2641, opened 2022-11-01
Synchronize IdP session and SSO enabled refresh token lifetime


carmenlau:
For SSO-enabled refresh token:

- Refresh tokens and IdP sessions lifetime are managed separately. The only difference is the refresh token's lifetime will be shorter than or equal to the IdP session. e.g. If the IdP session will expire in 1 day, the refresh token will only have 1 day lifetime when it is issued.
- When updating the refresh token's last access time, its IdP session's last access time will also be updated. So that the IdP session won't be invalidated due to inactivity.

Issue #2638, opened 2022-11-01
Update authz endpoint to support `x_sso_enabled`


carmenlau:
- New query parameter `x_sso_enabled`
    - When `x_sso_enabled=false`
        - Equal to `x_suppress_idp_session_cookie=true`.
        - The existing IdP session in cookie is ignored, and login will NOT create IdP Session.
        - Mark refresh token's `sso_enabled` to `false`.
    - When `x_sso_enabled=true`
        - Equal to `x_suppress_idp_session_cookie=false`.
        - Mark refresh token's `sso_enabled` to `true`.
- `x_suppress_idp_session_cookie` remains unchanged for backward compatibility, but the new SDKs won't send this parameter anymore. It is replaced by `x_sso_enabled`.
- When the developer uses `x_sso_enabled=true` + `prompt=login`, the enduser will need to log in again. The new refresh token will be linked to the new IdP session, and the old IdP session will be overwritten and revoked. (handle in #2639)

PR #2660, opened 2022-11-08
[Do not merge] Update authz endpoint to support x_sso_enabled and sync the sessions expiry 

carmenlau:
refs #2638 
refs #2641


carmenlau:
I am thinking should we do "SSO enabled refresh token requires valid IDP session when checking its validity" instead of syncing their lifetime.

Seems it is difficult to sync their lifetime perfectly. Given that we will access the IDP session when accessing the refresh token, we cannot do it backward. For the following scenario:

1. IDP session is idle time-out enabled, but the refresh token isn't.
2. When the user accesses the refresh token, both the IDP session and refresh token will be accessed. Both sessions will have the same expiry (last access + idle timeout duration). 
3. The user accesses the IDP session and the IDP session expiry extends.
4. Refresh token will become invalid due to inactivity, even if the IDP session is accessed. But the refresh token is not idle time-out enabled.

So I am thinking maybe we should:
1. Access the IDP session when accessing the refresh token session (keep)
2. When checking the validity of the refresh token, check its IDP session too. If it is SSO enabled. (add)
3. Remove the refresh token and IDP session lifetime sync (remove)

The cons are some of the invalid refresh tokens will be left in the store, but should not be harmful.

Any thoughts? @louischan-oursky

louischan-oursky:
Your new approach is better.

PR #2662, opened 2022-11-09
Support create session via Admin API

louischan-oursky:
ref #2654


pkg/admin/facade/oauth.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type OAuthFacade struct {
+	Config         *config.OAuthConfig
+	Authorizations OAuthAuthorizationService
+	Tokens         OAuthTokenService
+	Clock          clock.Clock
+}
+
+func (f *OAuthFacade) CreateSession(clientID string, userID string) (protocol.TokenResponse, error) {
```
carmenlau, line 45: Better check if the user exists before generating the token, when I input an invalid user id I got the following error.

`pq: insert or update on table \"_auth_oauth_authorization\" violates foreign key constraint \"_auth_oauth_authorization_user_id_fkey\"`

pkg/admin/facade/oauth.go
```diff
@@ -0,0 +1,98 @@
...
+		AuthenticatedAt: f.Clock.NowUTC(),
+	}
+	deviceInfo := make(map[string]interface{})
+
+	client, ok := f.Config.GetClient(clientID)
+	if !ok {
+		return nil, apierrors.NewInvalid("invalid client ID")
+	}
+
+	authz, err := f.Authorizations.CheckAndGrant(
```
carmenlau, line 62: Should allow check and grant only when the client is first-party. I think we could reject client third-party client here.

carmenlau (commented):
The other parts good good!

PR #2663, opened 2022-11-09
Update authz endpoint to support x_sso_enabled and check refresh token validity based on idp session

carmenlau:
refs #2638
refs #2641


pkg/lib/oauth/resolver.go
```diff
@@ -172,20 +165,44 @@ func (re *Resolver) resolveCookie(r *http.Request) (session.Session, error) {
...
 
-	event := access.NewEvent(re.Clock.NowUTC(), re.RemoteIP, re.UserAgentString)
-	offlineGrant, err = re.OfflineGrants.AccessWithID(offlineGrant.ID, event, expiry)
+	offlineGrant, err = re.OfflineGrants.AccessWithID(offlineGrant.ID, accessEvent, expiry)
 	if err != nil {
 		return nil, err
 	}
 
-	return offlineGrant, nil
+	return offlineGrant, err
```
louischan-oursky, line 214: Keep this line.

PR #2673, opened 2022-11-16
Support DenoHook

louischan-oursky:
ref #2670


pkg/lib/hook/resources.go
```diff
@@ -47,10 +47,10 @@ func (d DenoFileDescriptor) FindResources(fs resource.Fs) ([]resource.Location,
 func (DenoFileDescriptor) ViewResources(resources []resource.ResourceFile, rawView resource.View) (interface{}, error) {
 	output := bytes.Buffer{}
 
-	app := func() error {
+	app := func(view resource.AppFileView) error {
 		var target *resource.ResourceFile
 		for _, resrc := range resources {
-			if resrc.Location.Fs.GetFsLevel() == resource.FsLevelApp {
+			if resrc.Location.Fs.GetFsLevel() == resource.FsLevelApp && resrc.Location.Path == view.AppFilePath() {
```
carmenlau, line 53: I tried but the matching seems doesn't work. `view.AppFilePath()` returns an absolute path, but  `resrc.Location.Path` is relative. 

**My setup:**

In `authgear.yaml`:
```
hook:
  blocking_handlers:
  - event: user.pre_create
    url: authgeardeno:///deno/precreate.ts
```

Add `./deno/precreate.ts` to the config source directory.

Issue #2639, opened 2022-11-01
In the Continue screen: Change “Login with another account” to do logout as well


carmenlau:
refs #2639 

We had to logic to clean up the unreachable IDP session already. refs: https://github.com/authgear/authgear-server/blob/36b1da967a0b466ffc7e7b9a9877da2eb8d783aa/pkg/lib/interaction/nodes/do_ensure_session.go#L203

In this issue, we update the cleanup to remove refresh tokens in the same SSO group

PR #2675, opened 2022-11-16
Cleanup whole SSO group when cleaning up unreachable IDP session

carmenlau:
refs #2639 

We had to logic to clean up the unreachable IDP session already. refs: https://github.com/authgear/authgear-server/blob/36b1da967a0b466ffc7e7b9a9877da2eb8d783aa/pkg/lib/interaction/nodes/do_ensure_session.go#L203

In this PR, I updated the cleanup to remove refresh tokens in the same SSO group

### DIFF
--- a/pkg/lib/session/manager.go
+++ b/pkg/lib/session/manager.go
@@ -51,7 +51,7 @@ func (m *Manager) resolveManagementProvider(session Session) ManagementService {
 	}
 }
 
-func (m *Manager) invalidate(session Session, reason DeleteReason, isAdminAPI bool) (ManagementService, error) {
+func (m *Manager) invalidate(session Session, isAdminAPI bool) (ManagementService, error) {
 	sessions, err := m.List(session.GetAuthenticationInfo().UserID)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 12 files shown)


pkg/lib/session/manager.go
```diff
@@ -138,7 +144,16 @@ func (m *Manager) Logout(session Session, rw http.ResponseWriter) error {
...
-	_, err := m.invalidate(session, isAdminAPI)
+	_, err := m.invalidate(session, &signedOutEventOption{IsAdminAPI: isAdminAPI})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *Manager) Delete(session Session) error {
```
louischan-oursky, line 155: It is better to include NoEvent and WithEvent in the method to tell the difference. 

```golang
RevokeWithEvent(session Session, isAdminAPI bool) error
RevokeNoEvent(session Session, isAdminAPI bool) error
```

louischan-oursky (commented):
Now when the user revokes a SSO group, multiple events will be generated. That does not seem very good from the developer's point of view. They hardly realize what actually happened. Should we generate one event only?

PR #2683, opened 2022-11-18
Update session listing settings page for SSO enabled

carmenlau:
refs #2643 

Done
- Updated session listing for SSO group
- Handle revoke all

Todo
- ~~Update admin API to use SessionListingService~~ in #2691


pkg/auth/handler/webapp/settings.go
```diff
@@ -60,6 +60,10 @@ type SettingsAuthorizationService interface {
 	Delete(a *oauth.Authorization) error
 }
 
+type SettingsSessionListingService interface {
+	FilterForDisplay(sessions []session.Session) ([]*model.Session, map[string]interface{}, error)
```
louischan-oursky, line 64: Returning an extra map specific for authui seems strange. What about returning a new model?

```
type Session2 struct {
  *model.Session
  IsSSOEnable `json:"-"`
}
```

louischan-oursky (commented):
1. We can remove pkg/api/model/session_group.go

pkg/auth/handler/webapp/settings.go
carmenlau: Currently, we show the device icon for
- all the refresh tokens
- The IDP session that has SSO enabled refresh tokens

So the device icon is not for SSO enabled sessions only.
I amend the commit this way, see if it makes sense. fae1ec0

louischan-oursky:
Screenshot 2022-11-29 at 11 53 32 AM

When the terminate all other sessions button is not shown, the spacing in the bottom edge is too small.

carmenlau:
Updated!

Issue #2688, opened 2022-11-21
[Backport] Improve elasticsearch reindex command


louischan-oursky:
- Handle deleted user
- Print timestamp
- Print number of users indexed
- Print number of users deleted.
- Do not print every user. Print summary instead.

PR #2690, opened 2022-11-21
Improve reindex command

louischan-oursky:
ref #2688

---

To test deleted user,

1. Shutdown elasticsearch
2. Delete the user immediately
3. Restart elasticsearch
4. Reindex. We will see `1 deleted` in the command output.

### DIFF
--- a/cmd/authgear/elasticsearch/reindex.go
+++ b/cmd/authgear/elasticsearch/reindex.go
@@ -2,12 +2,15 @@ package elasticsearch
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
-	"io"
+	"log"
+	"os"
 
 	"github.com/elastic/go-elasticsearch/v7"
 	"github.com/elastic/go-elasticsearch/v7/esapi"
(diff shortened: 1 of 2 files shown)


cmd/authgear/elasticsearch/reindex.go
```diff
@@ -126,69 +174,128 @@ func (q *Reindexer) Reindex(es *elasticsearch.Client) (err error) {
...
+	size := 1000
+	var searchAfter interface{}
 
-	_, err = buf.Write(actionBytes)
-	if err != nil {
-		return
-	}
+	underscoreIDsToDelete := make(map[string]struct{})
+	for {
+		bodyJSONValue := map[string]interface{}{
```
carmenlau, line 225: I think we need to add app_id to the search?

carmenlau (commented):
The other part looks fine.

cmd/authgear/elasticsearch/reindex.go
louischan-oursky: Fixed.

PR #2709, opened 2022-12-06
Add Deno Hooks to portal

louischan-oursky:
ref #2627


carmenlau (commented):
When I click **Finish editing** without editing, the script will be reverted.

Steps:
1. Update the script and save it normally.
2. Reload the page and click **Edit Script**, I can see the script that I saved.
3. Click **Finish editing** without editing.
4. Click **Edit Script** again.
5. The script reverted to the default value.

louischan-oursky:
Screenshot 2022-12-09 at 5 27 03 PM

@fungc-io

Issue #2710, opened 2022-12-06
Add blocking hook for customize JWT payload


louischan-oursky:
- Refactor mutation
  - Remove `GenerateFullMutations()`
  - Add `PerformEffects(ctx)`
- Add event `user.session.jwt.pre_create`. It only affects access token encoded in JWT

PR #2714, opened 2022-12-07
Add user.session.jwt.pre_create event

louischan-oursky:
ref #2710


pkg/lib/oauth/token_encoding.go
```diff
@@ -53,6 +61,30 @@ func (e *AccessTokenEncoding) EncodeAccessToken(client *config.OAuthClientConfig
...
+			Meta: model.Meta{
+				ID: userID,
+			},
+		},
+		Payload: claimsMap,
+	}
+
+	err = e.Events.DispatchEvent(eventPayload)
+	if err != nil {
+		return "", err
```
carmenlau, line 80: When I return `is_allowed` to `false`, the SDK got the internal server error from the SDK. See if we can improve the error handling?

pkg/lib/oauth/token_encoding.go
```diff
@@ -53,6 +61,30 @@ func (e *AccessTokenEncoding) EncodeAccessToken(client *config.OAuthClientConfig
...
+	}
+
+	claims, err = jwtutil.BuildFromMap(eventPayload.Payload)
```
carmenlau, line 83: Instead of letting the user edit the jwt freely, should we restrict the user from editing some mandatory fields which will affect the jwt validity (e.g. `sub`, `iss`... etc)? Or the developer should take responsibility for that? 

Currently, if the hook returns a response that doesn't have those fields. The server will generate an invalid jwt. 

Open to discussion.

docs/specs/event.md
```diff
@@ -145,6 +146,24 @@ Occurs right before the end-user schedules account deletion.
...
+#### user.session.jwt.pre_create
+
+Occurs right before the access token is issued.
+Use this event to add custom fields to the JWT access token.
+
+```json5
+{
+  "payload": {
+    "user": { /* ... */ },
+    "payload": {
```
carmenlau, line 158: `payload` in `payload` looks a bit strange.

How about using `jwt`, or `payload` inside `jwt`?

carmenlau (approved):
Please rebase!

Issue #2733, opened 2022-12-19
[Concurrent Session] Limit Concurrent Session to 1 per oauth client


carmenlau:
## Problem

- The developer has 3 apps and they hope the user of each app should simultaneously be logged in on 1 device only.
    - Let's say a user `user@example.com` is logged in with app A and app B on an iPad.
    - When they log in with App A on their iPhone, the app A on iPad is logged out, while the app B in iPad is still logged in.
- When the user login on another device, all existing sessions will be revoked
    - per client (application) setting
        - Only sessions created same client application are revoked

## Implementations

- Add config `x_max_concurrent_session` to `OAuthClientConfig`, default `0`
    - add validation to allow `0` or `1` only (json schema)

- When `x_max_concurrent_session` is `1`, should check and terminate other refresh token from the same oauth client
    - Hints:
        - Should be done during authorization code exchange
           - https://github.com/authgear/authgear-server/blob/6f11bd10ea7880835f0d1f9aee7e75ee4bcdc35b/pkg/lib/oauth/handler/handler_token.go#L203
        - Update Session manager to support terminating sessions by client id
            - https://github.com/authgear/authgear-server/blob/master/pkg/lib/session/manager.go
- Screen to warn the end-user that other sessions will be terminated
     - In regular login, if cancelled, go back to login screen
     - In continue screen, if cancelled, go back to continue screen
- Update portal to configure `x_max_concurrent_session`
    - In “Applications” > individual Application
        
        ```markdown
        **Expire when login on another device**
        [ ] On/off
        If user login on a new device, the existing token of this application on another device is expired and the user will be logged out.
        ```
         ![Untitled](https://user-images.githubusercontent.com/388892/208474256-324e1e48-d488-480c-98b0-d8fe7aaad1ee.png)

## Todo

- [x] #2756
- [x] #2757
- [x] #2758
- [x] #2759

louischan-oursky:
> Should fire user.session.terminated with termination_type=concurrent_session_disabled

I think we do not need to fire this event. As we have `user.authenticated` being fired already.

fungc-io:
Need a UI to warn the end-user that other sessions will be terminated.

fungc-io:
Est for UI update: 2 days

PR #2734, opened 2022-12-19
Specify concurrent sessions

carmenlau:
Listed the implementations in #2733 

after #2627

### DIFF
--- a/docs/specs/sessions.md
+++ b/docs/specs/sessions.md
@@ -96,25 +96,10 @@ When it is specified, sessions will be considered **in terms of refresh token**.
 
 *Client-wise concurrent sessions* cannot restrict the number of sessions for the cookies-based website.
 
+Enabling *client-wise concurrent sessions* with SSO enabled is not allowed.
+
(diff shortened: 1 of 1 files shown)


carmenlau:
Updated the spec to introduce **Project-wise** and **Client-wise** concurrent sessions, and see if they can cover most of the use cases. 

We will only do **Client-wise** concurrent sessions at this stage.

https://github.com/authgear/authgear-server/blob/4c2d1a2b1f458bd4fc0485de96002152a12bb8e2/docs/specs/sessions.md#concurrent-sessions

carmenlau:
@chpapa @fungc-io 

I updated the concurrent sessions spec to introduce **Project-wise** and **Client-wise** concurrent sessions with use cases. I hope it covers most of the use cases. Let me know if it needs to be clarified.

In this stage, we will only do the **Client-wise concurrent sessions** and @tung2744 is working on it.

Please take a look at it when you have time. 🙏🏻

cc @louischan-oursky

Issue #2735, opened 2022-12-19
Support adding unverified mfa authenticators


carmenlau:
## Problem
Allow end-users to add email/phone during signup, and verify it in the next login

#### Supported secondary authenticators:
- Email OOB OTP
- Photo OOB OTP
- Email magic link (Not implemented yet)

PR #2737, opened 2022-12-19
Specify mfa updates

carmenlau:
ref #2735 
ref #2736

### DIFF
--- a/docs/specs/user-model.md
+++ b/docs/specs/user-model.md
@@ -296,6 +296,8 @@ Primary authenticators authenticate the identity.
 
 ### Secondary Authenticator
 
+#### Secondary authentication mode
+
 Secondary authenticators are additional authentication methods to ensure higher degree of confidence in authenticity.
 
 The mode of secondary authentication is configurable. They are as follows:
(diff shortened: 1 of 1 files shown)


docs/specs/user-model.md
```diff
@@ -306,6 +308,25 @@ The mode of secondary authentication is configurable. They are as follows:
...
+authentication:
+  secondary_authentication_mode: (disabled|required|if_exists)
+```
+
+#### Secondary Authenticator Create Policy
+
+The secondary authenticator creation policy specifies how the secondary authenticators are set up during signup.
+
+- `any`: When `secondary_authentication_mode` is `required`, users have to set up one authenticator during signup. If multiple secondary authenticators are enabled in the project, users can choose one of them from the list.
+- `all`: When `secondary_authentication_mode` is `required` or `if_exists`, users need to set up all the enabled secondary authenticators based on the configuration order during signup.
```
louischan-oursky, line 321: > When `secondary_authentication_mode` is `required` or `if_exists`

When the mode is `if_exists`, I think we should not force the user to set up any MFA during signup.

Issue #2672, opened 2022-11-14
Add more context to invitation email


fungc-io:
## Current design problem

The current email contain little context on what it's doing. Without looking at the sender address, it's impossible to understand it's from Authgear before inspecting the link.

Subject: `Collaborator invitation`
Content:
```
# Join [project] as a collaborator

You have been invited to join [project] as a collaborator. Please click the link below to view the invitation.

[View invitation]

If you want to reject the invitation please ignore this email.
```

## Change

Subject: `You are invited to collaborate on [project] in Authgear`
Content:
```
# Join [project] as a collaborator

You have been invited to join as a collaborator to manage the [project] account in Authgear. Please click the link below to view the invitation.

[View Invitation]

Authgear is a user identity and authentication software that helps you manage users and log them in securely.

If you want to reject the invitation please ignore this email.

```

Later when we make the name of user mandatory with a proper welcome screen, add the name of the person who invites to the subject too.

`[User] has invited you to collaborate on [project] in Authgear`

louischan-oursky:
> You have been invited to join as a collaborator to manage the [project] account in Authgear.

I suggest changing it to `You have been invited to join as a collaborator to manage the project "myapp" in Authgear` @fungc-io . The rationale is "myapp" is not an account, it is a project.

fungc-io:
@louischan-oursky ok 👍

PR #2747, opened 2022-12-22
Update wordings for collaborator invitation email

IniZio:
ref: #2672

**Preview**

> The project name is "accounts"

**Subject**

`"You are invited to collaborate on "accounts" in Authgear"`

**Text body**

```
Join \"accounts\" as a collaborator\n\nYou have been invited to join as a collaborator to manage the project \"accounts\" in Authgear. Please click the link below to view the invitation.\n\nhttp://portal.localhost:8000/collaborators/invitation?code=RMW389CRP5N7WHKPJF5860EB166JJ71H\n\nAuthgear is a user identity and authentication software that helps you manage users and log them in securely.\n\nIf you want to reject the invitation please ignore this email.\n
```

**HTML preview**

Screen Shot 2023-01-03 at 15 44 17

### DIFF
--- a/resources/portal/templates/en/messages/collaborator_invitation_email.mjml
+++ b/resources/portal/templates/en/messages/collaborator_invitation_email.mjml
@@ -8,10 +8,13 @@
 <mj-body>
   <mj-section>
     <mj-column>
-      <mj-text font-weight="bold" font-size="24px" padding="20px">Join {{ .AppName }} as a collaborator</mj-text>
-      <mj-text font-size="16px" padding="20px">You have been invited to join {{ .AppName }} as a collaborator. Please click the link below to view the invitation.</mj-text>
+      <mj-text font-weight="bold" font-size="24px" padding="20px">Join "{{ .AppName }}" as a collaborator</mj-text>
+      <mj-text font-size="16px" padding="20px">You have been invited to join as a collaborator to manage the project <b>{{ .AppName }}</b> in Authgear. Please click the link below to view the invitation.</mj-text>
       <mj-button href="{{ .Link }}" padding="20px">View invitation</mj-button>
-      <mj-text font-weight="light" font-size="12px" padding="20px">If you want to reject the invitation please ignore this email.</mj-text>
+      <mj-text font-weight="light" font-size="12px" padding="20px" padding-top="10px">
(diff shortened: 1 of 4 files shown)


resources/portal/templates/en/messages/collaborator_invitation_email.mjml
```diff
@@ -9,9 +9,12 @@
   <mj-section>
     <mj-column>
       <mj-text font-weight="bold" font-size="24px" padding="20px">Join {{ .AppName }} as a collaborator</mj-text>
-      <mj-text font-size="16px" padding="20px">You have been invited to join {{ .AppName }} as a collaborator. Please click the link below to view the invitation.</mj-text>
+      <mj-text font-size="16px" padding="20px">You have been invited to join {{ .AppName }} as a collaborator to manage the [project] account in Authgear. Please click the link below to view the invitation.</mj-text>
```
louischan-oursky, line 12: `[project]` actually means substituting the app ID here. 😂 

`{{ .AppName }}` does not appear after `join` too.

Another thing to put attention is that we want to see the app ID here. Please double check if `{{ .AppName }}` refers to the app ID or the app name. We have both. The app name is defined in the translation.json under the key `app.name`.

louischan-oursky:
You only need to update the .mjml files. The .html is generated by running `make html-email`.

resources/portal/templates/en/messages/collaborator_invitation_email.mjml
IniZio: Currently the render uses project id, pasted an example for reference. Also fixed `[project]` 🤦

IniZio:
Yes I did generate the html with `make html-email`

louischan-oursky:
@fungc-io Do you think we need to apply some style to the app ID to make it more visually prominent in the subject and in the body?

fungc-io:
yea, bold it will do

louischan-oursky:
I think we can make the app ID bold in the body, but not in the subject. Maybe we can quote it in the subject. Which delimiter should we use?

fungc-io:
double quote `" U+0022`

IniZio:
Updated to bold the project ID in html body, added quotes for other places where bold style cannot be applied.

PR #2752, opened 2023-01-03
Limit Concurrent Session to 1 per oauth client #2733

tung2744:
ref #2733 #2756 #2757 #2758 #2759

Changes:
  - [x] Add x_max_concurrent_session to client config
  - [x] Revoke session of the same client when max concurrent session is set
  - [x] Screen to warn the end-user that other sessions will be terminated
  - [x] Update portal to configure x_max_concurrent_session
Screenshot 2023-01-05 at 7 36 54 PM

https://user-images.githubusercontent.com/34942210/210772473-f9a44732-b236-4c3e-a4b9-1e2a23913932.mov


pkg/lib/interaction/intents/authenticate.go
```diff
@@ -357,6 +357,10 @@ func (i *IntentAuthenticate) DeriveEdgesForNode(graph *interaction.Graph, node i
 			panic(fmt.Errorf("interaction: unexpected authentication stage: %v", node.Stage))
 		}
 	case *nodes.NodePromptCreatePasskeyEnd:
+		return []interaction.Edge{
+			&nodes.EdgeConfirmTerminateOtherSessionsBegin{},
+		}, nil
```
carmenlau, line 363: Currently, when I try to promote an anonymous user in the mobile app, I will see this warning screen. Since a refresh token was issued when the user is an anonymous user. Maybe we should only show this when login?

tung2744:
TODO:

- [ ] Block using sso login with concurrent session config in authorize endpoint
- [ ] clear steps in session
- [ ] redirect user back to select account

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -701,6 +709,22 @@ func (h *TokenHandler) handleIDToken(
...
+	userID string) error {
+	offlineGrants, err := h.OfflineGrants.ListOfflineGrants(userID)
+	if err != nil {
+		return err
+	}
+	for _, offlineGrant := range offlineGrants {
+		if offlineGrant.ClientID != client.ClientID {
+			continue
+		}
+		h.SessionManager.RevokeWithEvent(offlineGrant, false, false)
```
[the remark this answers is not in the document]
carmenlau: Discussion conclusion, having `x_max_concurrent_session` in the client + sso enabled is rare. We will block it from the authz endpoint.

tung2744:
@carmenlau Updated, thx!

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -789,14 +808,7 @@ func (h *TokenHandler) issueTokensForAuthorizationCode(
...
 	if issueRefreshToken {
-		// First revoke existing refresh tokens if MaxConcurrentSession == 1
-		if client.MaxConcurrentSession == 1 {
-			err := h.revokeClientOfflineGrants(client, code.AuthenticationInfo.UserID)
-			if err != nil {
-				return nil, err
-			}
-		}
-		offlineGrant, err := h.TokenService.IssueOfflineGrant(client, opts, resp)
+		offlineGrant, err := h.issueOfflineGrant(client, code, resp, opts)
```
carmenlau, line 811: In the `TokenHandler`, we have multiple places that call `h.TokenService.IssueOfflineGrant` to issue the refresh token. Should we also update those places to use `h.issueOfflineGrant` instead? (e.g. `handleBiometricAuthenticate`, to restrict biometric login concurrent sessions ...etc.) If yes, please check the functionality after updating

pkg/auth/webapp/service2.go
```diff
@@ -332,16 +332,7 @@ func (s *Service2) doPost(
...
-	switch graph.CurrentNode().(type) {
-	case *nodes.NodeCancelInteraction:
-		isCancelled = true
-	}
-	if isCancelled {
-		err = nil
-	}
-
-	err = s.afterPost(result, session, graph, err, isCancelled, isFinished, isNewGraph)
+	err = s.afterPost(result, session, graph, err, isFinished, isNewGraph)
```
carmenlau, line 335: It is okay for now. But if we want to revert some unwanted changes, it would be better to do it in a separate commit or even update the commit through rebase. So that we can have a clear commit history.

carmenlau (commented):
Please rebase

pkg/lib/oauth/handler/handler_token.go
tung2744: I guess yes, let me check it

tung2744:
@carmenlau Updated. Also checked biometric login is working. Thx!

carmenlau:
Please fix the build, thanks @tung2744!!

tung2744:
@carmenlau Fixed and rebased again, thx!

Issue #2740, opened 2022-12-20
[OTP] Create or retrieve OTP via Admin API


louischan-oursky:
# Use case

Sometimes OTP code can fail to deliver to the audience. It would be helpful if the admin staff can tell the end-user the OTP code when the end-user asks for support.

# Technical background

Currently OTP code in Authgear is handled by different packages.

When sending OTP code for signing in, the OTP code is bound to an authenticator ID. See https://github.com/authgear/authgear-server/blob/master/pkg/lib/authn/authenticator/oob/provider.go#L87

When sending OTP code for signing up or verification, the OTP is bound to an identity ID and a web session. See https://github.com/authgear/authgear-server/blob/master/pkg/lib/feature/verification/service.go#L190 

We would like to unify OTP code such that the OTP code is bound to the target itself (email address or phone number).

# Implementation details

- Introduce a new struct in `pkg/lib/authn/otp` to generate and verify OTP code.
```golang
// pseudocode!!!

type Code struct {
  // 6 digit code
  Code string
  ExpireAt time.Time
  // Code will be persisted so it does not contain target.
  // Add any other necessary fields...
}

type OTPCodeService interface {
  // GenerateCode generates a new Code or return an existing Code.
  GenerateCode(target string) (*Code, error)
  // VerifyCode verifies if code matches an existing Code.
  // The error it returns should follow existing implementation.
  VerifyCode(target string, code string) error
}
```

- [x] #2760
- [x] #2761
- [x] #2762
-  ~(**Design Required**) (TBC) Find a place in the portal to trigger this mutation. @fungc-io Thoughts?~
- ~Display the generated code in a [Dialog](https://developer.microsoft.com/en-us/fluentui#/controls/web/dialog). The title is `OTP code for [email or phone]`. The description is `The end-user can use this code to sign in if they did not receive the code`. The only button is a cancel button which dismisses the dialog. Click outside the dialog should also dismiss the dialog.~
- [x] #2763

fungc-io:
Need to refactor the OTP generation package for signup, verification and 2FA

louischan-oursky:
@Rubikium I have updated the description to reflect the actual changes we want to make. Feel free to ask if you are not sure what to do.

Rubikium:
> Introduce a new struct in pkg/lib/authn/otp to generate and verify OTP code.

> Remove the old implementation in pkg/lib/authn/authenticator/oob and pkg/lib/feature/verification

[...]

PR #2753, opened 2023-01-03
Unify OTP implementations

Rubikium:
implements #2760, #2761 
ref #2740


pkg/lib/authn/otp/errors.go
```diff
@@ -0,0 +1,10 @@
+package otp
+
+import (
+	"github.com/authgear/authgear-server/pkg/api/apierrors"
+)
+
+var InvalidOTPCode = apierrors.Forbidden.WithReason("InvalidOTPCode")
+
+var ErrCodeNotFound = InvalidOTPCode.NewWithCause("otp code is expired or invalid", apierrors.StringCause("CodeNotFound"))
+var ErrInvalidCode = InvalidOTPCode.NewWithCause("invalid otp code", apierrors.StringCause("InvalidOTPCode"))
```
louischan-oursky, line 10: We added new errors here. This implies new code returns new errors. Existing code checks old errors. `errors.Is` WILL NOT report the old ones and new ones equal. We have bugs here.

This is one refactoring technique I want to share.

> When we add new things, make sure we remove the old ones. 

This ensures existing code is 100% converted to use new code. If we do not do this, we could face mysterious bugs that take hours to diagnose.

---

- [ ] Remove pkg/lib/authn/authenticator/whatsapp/errors.go
- [ ] Remove pkg/lib/authn/authenticator/oob/errors.go
- [ ] We cannot simply remove pkg/lib/feature/verification/errors.go because if we search for "InvalidVerificationCode", we can see one usage in __error.html. To preserve the behavior, we have to translate the error. That is, we have to make ensure we still let __error.html see  the old errors.

PR #2793, opened 2023-01-10
Resolve authgear event type hints for deno hook editor

IniZio:
Demo: https://codesandbox.io/s/focused-kalam-8jxnfl

### Intent

Currently Deno hook editor cannot resolve external dependencies, so user only know if types are correct on save.

This PR downloads imports and include them in monaco-editor, providing better type-hint for users.

### Limitations and Reasons

1. Assumes downloaded file does not have relative imports, falls back to `any` otherwise

	- Typescript ignores path alias if the imported path has `.ts` extension
		- `"moduleResolution": "bundler"`([related pr](https://github.com/microsoft/TypeScript/pull/51669)) might help fix `.ts` extension handling, however monaco-editor's `monaco.languages.typescript.ModuleResolutionKind` does not support it yet
	- monaco-editor cannot install vscode extension e.g. https://marketplace.visualstudio.com/items?itemName=denoland.vscode-deno so need to fake with nodejs module resolution

2. Does not support recursive import

	- Deno does not provide a reliable way to know all imported modules. Therefore recursive import has to be done linearly, which is too expensive in this use case.

### DIFF
--- a/portal/src/CodeEditor.tsx
+++ b/portal/src/CodeEditor.tsx
@@ -1,16 +1,124 @@
-import React from "react";
-import ControlledEditor, { EditorProps } from "@monaco-editor/react";
+import React, { useCallback, useRef } from "react";
+import ControlledEditor, {
+  EditorProps,
+  Monaco,
+  OnChange,
+  OnMount,
+} from "@monaco-editor/react";
 
 export interface CodeEditorProps extends EditorProps {
   className?: string;
+  packages?: string[];
+}
+
+const DEFAULT_WHITELISTED_PACKAGES = [
+  "https://deno.land/std",
+  "https://deno.land/x/authgear_deno_hook",
+];
+
+const REGEX_DETECT_IMPORT =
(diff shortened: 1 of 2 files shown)


portal/src/CodeEditor.tsx
```diff
@@ -1,16 +1,95 @@
...
+    }
+  }, []);
+
+  const handleEditorMount = useCallback<OnMount>(
+    async (editor, monaco) => {
+      monacoRef.current = monaco;
+
+      const options =
+        monaco.languages.typescript.typescriptDefaults.getCompilerOptions();
+      options.strictNullChecks = true;
```
louischan-oursky, line 59: You can turn on the strictest compiler warnings.

portal/src/CodeEditor.tsx
```diff
@@ -1,16 +1,95 @@
...
+  Monaco,
+  OnChange,
+  OnMount,
+} from "@monaco-editor/react";
 
 export interface CodeEditorProps extends EditorProps {
   className?: string;
 }
 
+const REGEX_DETECT_IMPORT =
```
louischan-oursky, line 13: Per our offline discussion, please document the approach here we are using here. Like the assumption that the imported module looks like an ambient .d.ts. 

And we should only detect the import of authgear_deno_hook. Ignore other unrecognized modules.

Issue #2781, opened 2023-01-07
[OTP] Configurable Cooldown period


fungc-io:
## Requirement

- The current cooldown is set to 60s. Make it configurable per project so that user can choose 60s or 120s

### AuthUI

![Image](https://user-images.githubusercontent.com/15207470/211144709-81598628-df9d-4873-9d8b-4dfdf0d5d46e.png)

### Portal

![Image](https://user-images.githubusercontent.com/15207470/211144775-03c88402-1802-4a75-a7f1-062052a992fc.png)

Add "Cooldown duration before resending OTP" and a dropdown of 60 seconds / 120 seconds

louischan-oursky:
## Add the new config `verification.code_resend_cooldown_reconds`

Since we have `verification.code_expiry_seconds`, we are going to add `verification.code_resend_cooldown_reconds`. The JSON schema of it should be

```
{
  "type": "integer",
  "enum": [60, 120]
}
```

In Go, it should be a `DurationSeconds`. Implement SetDefaults() such that if it is zero, the default value is 60.

## Use `verification.code_resend_cooldown_reconds` for cooldown

We now have AntiSpamSendVerificationCodeBucket and AntiSpamSendOOBCodeBucket, let's unify them into AntiSpamOTPCodeBucket. Use `otp-code:%s` for the key.

Now that the bucket depends on the config, it is reasonable to use dependency injection to get the bucket. Since wire injects by type, we cannot provide `ratelimit.Bucket` directly. Doing so means we can ever inject this and only this bucket. It is very simple to circumvent this limitation. We introduce a new type to contain a ratelimit.Bucket.

```golang
type OTPCodeBucket {
  Bucket ratelimit.Bucket
}

func ProvideOTPCodeBucket(cfg *config.VerificationConfig) OTPCodeBucket {
  // use cfg to construct the bucket
}

var DependencySet = wire.NewSet(
  // other deps
  ProvideOTPCodeBucket,
)
```

Issue #2780, opened 2023-01-07
[OTP] SMS OTP Limiting


fungc-io:
This are the new requirements for SMS OTP, details are written in each sub-issue.

### **Configurable Cooldown period**

- The current cooldown is set to 60s. Make it configurable per project so that user can choose 60s or 120s
- [x] https://github.com/authgear/authgear-server/issues/2781

### **Generate OTP on every resend**

- Regardless of the OTP valid lifetime, every resend should generate a new OTP. And the old one is immediately invalidated.
- [x] authgear/authgear-server#2782

### **Revoke current OTP after X attempts** 

- Let the default value of X be 5
- After 5 attempts of entering the OTP, the OTP is revoked and the enduser must generate a new one (resend) in order to continue
- The UI need to know there is no OTP available and block further input until the enduser triggers resending the OTP
- [x] authgear/authgear-server#2783

### **SMS send limit**

- There will be a daily send limit for each phone number. If the send limit is exceed the enduser cannot receive a new OTP until a day later. 
- Effectively this will block the user from logging in.
- [x] authgear/authgear-server#2784

fungc-io:
@louischan-oursky Can you help do some technical planning on each sub-issue?
Thanks!

louischan-oursky:
The shape of the new config

```
sms:
  ratelimit:
    per_phone:
      size: 20
      reset_period: 1day
    resend_cooldown_seconds: 60s
email:
  ratelimit:
    resend_cooldown_seconds: 60s
otp:
  ratelimit:
    failed_attempt:
      size: 5
      reset_period: 20min

```

PR #2794, opened 2023-01-10
Add cooldown config on OTP SMS

Rubikium:
impl #2781
ref #2780


pkg/lib/config/otp.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+var _ = Schema.Add("OTPSMSConfig", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"resend_cooldown_seconds": {
+			"type": "integer",
+			"enum": [60, 120]
```
louischan-oursky, line 24: Add test for this config. See pkg/lib/config/testdata/account_deletion_tests.yaml for an example.

pkg/lib/interaction/ratelimit.go
```diff
@@ -63,3 +64,22 @@ func AntiSpamSendOOBCodeBucket(oobType OOBType, target string) ratelimit.Bucket
...
+			Size:        1,
+			ResetPeriod: cfg.SMS.ResendCooldownSeconds.Duration(),
+		},
+	}
+}
+
+func AntiSpamOTPCodeBucket(target string) OTPCodeBucket {
+	b := OTPCodeBucket{}
+	b.Bucket.Key = fmt.Sprintf("otp-code:%s", target)
+	return b
```
louischan-oursky: This is incorrect. The Size and ResetPeriod of b is zero value. The correct way here is to introduce a struct that can return a ratelimit.Bucket.

```go
type AntiSpamOTPCodeBucketMaker struct {
  Config *config.OTPConfig // This field is dependency injected.
}

func (m *AntiSpamOTPCodeBucketMaker) MakeBucket(target string) ratelimit.Bucket {
  return ratelimit.Bucket{
    Key: fmt.Sprintf("otp-code:%s", target),
    Size: 1,
    ResetPeriod: m.Config.Blahblahblah,
  }
}
```

```go
// Other places that need the bucket inject the maker.

type AntiSpamOTPCodeBucketMaker interface {
  MakeBucket(target string) ratelimit.Bucket
}
```

louischan-oursky (changes requested):
Again, please delete the old code, AntiSpamSendOOBCodeBucket and AntiSpamSendVerificationCodeBucket

pkg/lib/config/config.go
```diff
@@ -41,7 +41,8 @@ var _ = Schema.Add("AppConfig", `
 		"forgot_password": { "$ref": "#/$defs/ForgotPasswordConfig" },
 		"welcome_message": { "$ref": "#/$defs/WelcomeMessageConfig" },
 		"verification": { "$ref": "#/$defs/VerificationConfig" },
-		"otp": { "$ref": "#/$defs/OTPConfig" },
+		"sms": { "$ref": "#/$defs/SMSConfig" },
+		"email": { "$ref": "#/$defs/EmailConfig" },
```
louischan-oursky, line 45: I just realized we have MessagingConfig. Please move SMSConfig and EmailConfig inside MessageConfig.

pkg/lib/config/email.go
```diff
@@ -0,0 +1,38 @@
...
+		"resend_cooldown_seconds": {
+			"type": "string",
+			"enum": ["60s", "120s"]
+		}
+	}
+}
+`)
+
+type EmailRatelimitConfig struct {
+	ResendCooldownSeconds DurationString `json:"resend_cooldown_seconds,omitempty"`
```
louischan-oursky, line 32: This should be of type `*DurationSeconds` instead of DurationString.

pkg/lib/config/sms.go
```diff
@@ -0,0 +1,38 @@
...
+		"resend_cooldown_seconds": {
+			"type": "string",
+			"enum": ["60s", "120s"]
+		}
+	}
+}
+`)
+
+type SMSRatelimitConfig struct {
+	ResendCooldownSeconds DurationString `json:"resend_cooldown_seconds,omitempty"`
```
louischan-oursky, line 32: This should be of type `*DurationSeconds` instead of `DurationString`.

pkg/lib/config/sms.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+var _ = Schema.Add("SMSRatelimitConfig", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"resend_cooldown_seconds": {
+			"type": "string",
+			"enum": ["60s", "120s"]
```
louischan-oursky, line 25:
```
"resend_cooldown_seconds": {
  "$ref": "#/$defs/DurationSeconds",
  "enum": [60, 120]
}
```

louischan-oursky (changes requested):
So in this PR we need not introduce DurationString. DurationString can be introduced in another PR.

louischan-oursky (changes requested):
Screenshot 2023-01-16 at 11 00 12 AM

Some issues here.

- The cooldown label for email is incorrect. It says "via SMS".
- There are 3 sections here. The first section is about code expiry and verification criteria. The second is about email. The third is about SMS. But there is no separator in between which makes them very difficult to read. Please follow the password policy pane, and add separators and subheadings to make this pane more easier to follow.

Issue #2778, opened 2023-01-06
Admin to change a user identity


fungc-io:
Currently the enduser can change their identity (email/phone number/username) via `/settings`

However it's not supported in Admin Portal and Admin API.

- Add a mutation to update identity
- Add affordance in portal
- The validation and duplication checking still applies

### Design:

Merge the buttons into a menu (https://fabricweb.z5.web.core.windows.net/oufr/8.104.2/#/examples/contextualmenu)

![image](https://user-images.githubusercontent.com/15207470/210925195-db0d2625-eaf6-492f-81a6-f9b4f899e521.png)

```
[ Actions 🔽 ]
[ Mark as verified/unverified ]
[ Edit ]
[ Remove ]
```

The edit page is similar to the "Add" page with the following changes:
![image](https://user-images.githubusercontent.com/15207470/210925132-73080744-c676-4bb4-9f68-eefcc348ec0a.png)

## Todo

- [x] Update action buttons to context menu
- [x] Add edit username screen
- [x] Add edit phone screen
- [x] Add edit email screen
- [x] Add mutation to update identity
- [x] Integrate update identity mutation
- [ ] Check new identity isVerified logic

louischan-oursky:
No caveats.

tung2744:
@fungc-io For oauth identity, we won't have edit right?

fungc-io:
Only phone, email, username are editable

tung2744:
- [x] Update action buttons to context menu
- [ ] Add edit username screen
- [ ] Add edit phone screen
- [ ] Add edit email screen
- [ ] Add mutation to update identity
- [ ] Integrate update identity mutation

tung2744:
@carmenlau @louischan-oursky 
I found out the verified state is actually due to authenticator being created when new identity added. When I add an email identity, with password disabled, an oob_otp_email (Or oob_otp_sms if it is phone) authenticator will be created, and as a side effect it set the identity as verified. However, if password is enabled, this authenticator will not be created and therefore the identity becomes verified. Would like to confirm is this the expected behavoir?

louischan-oursky:
No, creating OOB OTP authenticator via admin API should not mark the phone / email address as verified.

tung2744:
ok. Then as a result new and edited identity will always be unverified.

PR #2797, opened 2023-01-11
Support editing login id in portal

tung2744:
ref #2778 

https://user-images.githubusercontent.com/34942210/211722968-173adf25-2320-4116-92e4-baa7cb2fb808.mov


carmenlau (commented):
💯

PR #2798, opened 2023-01-11
Add OTP retry rate limit

### LINKED ISSUE #2783: [OTP] Revoke current OTP after X attempts
This will drastically reduce the chance successful brute force attacks on OTP

## Requirements

- Let the default value of X be 5
- After 5 attempts of entering the OTP, the OTP is revoked and the enduser must generate a new one (resend) in order to continue
- The UI need to know there is no OTP available and block further input until the enduser triggers resending the OTP

## Design

Image

- Disable input field
- Show message 
  - > You have exceeded the maximum of attempts for entering the one-time password. Please get a new one-time password. [Resend](#)

Rubikium:
fix #2783
ref #2780

### DIFF
--- a/pkg/lib/config/otp.go
+++ b/pkg/lib/config/otp.go
@@ -37,10 +37,7 @@ var _ = Schema.Add("OTPFailedAttemptConfig", `
 			"type": "integer",
 			"minimum": 1
 		},
-		"reset_period": {
-			"$ref": "#/$defs/DurationString",
-			"format": "x_duration_string"
-		}
+		"reset_period": { "$ref": "#/$defs/DurationString" }
 	}
 }
 `)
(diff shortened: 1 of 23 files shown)


pkg/lib/config/otp.go
```diff
@@ -0,0 +1,59 @@
...
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"size": {
+			"type": "int"
+		},
+		"reset_period": {
+			"$ref": "#/$defs/DurationString",
+			"format": "x_duration_string"
```
louischan-oursky, line 42: format should belongs to DurationString.

louischan-oursky (changes requested):
Screenshot 2023-01-16 at 12 20 11 PM

We should disable the primary button here as well. So that the end-user knows they have to wait and press the resend button.

louischan-oursky:
@Rubikium Please resolve the conflict.

Issue #2767, opened 2023-01-04
[Magic Link] Implement approval and close window screen


IniZio:

IniZio:
@fungc-io Screenshots & wordings for current implementation

### Setup authenticator

**Enter email**

![Screen Shot 2023-01-16 at 14 48 07](https://user-images.githubusercontent.com/3119725/212615708-9bc5fa55-07e5-483a-a576-930d6ebba3bc.png)

**Waiting for email verification**

![Screen Shot 2023-01-16 at 14 48 17](https://user-images.githubusercontent.com/3119725/212615813-b841fae2-9dcc-4762-8e07-5c2c5583346f.png)

**Resend email flash message**

![Screen Shot 2023-01-16 at 14 49 43](https://user-images.githubusercontent.com/3119725/212615854-e9157410-f17d-4b6c-9e0a-6cb43935848b.png)

### Login

**Login screen**

![Screen Shot 2023-01-16 at 14 51 19](https://user-images.githubusercontent.com/3119725/212616422-ff8fbd72-f9f3-41b0-92de-95464779257f.png)

### Magic link verification

**Email for verify magic link as new authenticater**

![Screen Shot 2023-01-16 at 14 50 12](https://user-images.githubusercontent.com/3119725/212615911-a7c45e2d-2c8a-4aa1-920d-857b0b60dc40.png)

**Email for verify magic link to login**

![Screen Shot 2023-01-16 at 14 51 39](https://user-images.githubusercontent.com/3119725/212616465-91aea8fe-0e8b-499f-9112-b64d99d04cb2.png)

**Approval screen for magic link as authenticator**

![Screen Shot 2023-01-16 at 14 50 31](https://user-images.githubusercontent.com/3119725/212616018-fa22c6b6-b171-49d8-bb63-d600d5f9da97.png)

**Approval screen for magic link as login**

![Screen Shot 2023-01-16 at 14 51 53](https://user-images.githubusercontent.com/3119725/212616687-2cede5e2-2fd7-4834-b1f8-dbd809420985.png)

**Success, close window screen**

![Screen Shot 2023-01-16 at 14 50 40](https://user-images.githubusercontent.com/3119725/212616103-0a32bb9a-5b88-43d7-8c82-c5a18e0f57c0.png)

**Success at original session screen**

[...]

PR #2802, opened 2023-01-12
Implement authenticate with magic link

IniZio:
ref: #2767

### Changes

- Flow for create & verify magic link authenticator
- Flow for authenticate with magic link
- Resend magic link
- Handle Magic link in alternatives

### Not covered

- Open App
- Allow magic Link to be unverified on setup `secondary_allow_unverified`

### DIFF
--- a/pkg/auth/handler/webapp/magic_link_otp.go
+++ b/pkg/auth/handler/webapp/magic_link_otp.go
@@ -22,11 +22,11 @@ func ConfigureMagicLinkOTPRoute(route httproute.Route) httproute.Route {
 }
 
 type MagicLinkOTPNode interface {
-	GetMagicLinkOTP() string
+	GetMagicLinkOTPTarget() string
 }
 
-type MagicLinkOTPAuthnNode interface {
-	GetAuthenticatorIndex() int
+type MagicLinkOTPViewModel struct {
+	Target string
 }
 
 type MagicLinkOTPHandler struct {
@@ -39,8 +39,13 @@ type MagicLinkOTPHandler struct {
 func (h *MagicLinkOTPHandler) GetData(r *http.Request, rw http.ResponseWriter, session *webapp.Session, graph *interaction.Graph) (map[string]interface{}, error) {
 	data := map[string]interface{}{}
 	baseViewModel := h.BaseViewModel.ViewModel(r, rw)
-
+	viewModel := MagicLinkOTPViewModel{}
+	var n MagicLinkOTPNode
+	if graph.FindLastNode(&n) {
+		viewModel.Target = mail.MaskAddress(n.GetMagicLinkOTPTarget())
+	}
 	viewmodels.Embed(data, baseViewModel)
+	viewmodels.Embed(data, viewModel)
 	return data, nil
 }
 
--- a/pkg/auth/handler/webapp/magic_link_otp.go
+++ b/pkg/auth/handler/webapp/magic_link_otp.go
@@ -1,12 +1,14 @@
 package webapp
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
 
 	"github.com/authgear/authgear-server/pkg/auth/handler/webapp/viewmodels"
 	"github.com/authgear/authgear-server/pkg/auth/webapp"
+	"github.com/authgear/authgear-server/pkg/lib/authn/otp"
 	"github.com/authgear/authgear-server/pkg/lib/infra/mail"
 	"github.com/authgear/authgear-server/pkg/lib/interaction"
 	"github.com/authgear/authgear-server/pkg/lib/interaction/nodes"
@@ -38,6 +40,7 @@ type MagicLinkOTPViewModel struct {
 }
 
 type MagicLinkOTPHandler struct {
+	MagicLinkOTPCodeService   otp.Service
 	ControllerFactory         ControllerFactory
 	BaseViewModel             *viewmodels.BaseViewModeler
 	AlternativeStepsViewModel *viewmodels.AlternativeStepsViewModeler
@@ -140,14 +143,46 @@ func (h *MagicLinkOTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) 
 		return nil
 	})
 
+	getTargetFromGraph := func() (string, error) {
+		graph, err := ctrl.InteractionGet()
+		if err != nil {
+			return "", err
+		}
+		var target string
+		var n MagicLinkOTPNode
+		if graph.FindLastNode(&n) {
+			target = n.GetMagicLinkOTPTarget()
+		} else {
+			panic(fmt.Errorf("webapp: unexpected node for magic link: %T", n))
+		}
+
+		return target, nil
+	}
+
 	ctrl.PostAction("matched", func() error {
-		u := url.URL{}
-		u.Path = r.URL.Path
-		q := r.URL.Query()
-		q.Set(MagicLinkOTPPageQueryStateKey, string(MagicLinkOTPPageQueryStateMatched))
-		u.RawQuery = q.Encode()
+		var state MagicLinkOTPPageQueryState = MagicLinkOTPPageQueryStateInitial
+
+		target, err := getTargetFromGraph()
+		if err != nil {
+			return err
+		}
+
+		_, err = h.MagicLinkOTPCodeService.VerifyMagicLinkCode(target, false)
+		if err == nil {
+			state = MagicLinkOTPPageQueryStateMatched
+		} else if errors.Is(err, otp.ErrInvalidCode) {
+			state = MagicLinkOTPPageQueryStateInvalidCode
+		} else {
+			return err
+		}
+
+		url := url.URL{Path: r.URL.Path}
+		query := r.URL.Query()
+		query.Set(MagicLinkOTPPageQueryStateKey, string(state))
+		url.RawQuery = query.Encode()
+
 		result := webapp.Result{
-			RedirectURI:      u.String(),
+			RedirectURI:      url.String(),
 			NavigationAction: "replace",
 		}
 		result.WriteResponse(w, r)
@@ -156,22 +191,6 @@ func (h *MagicLinkOTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) 
 
 	ctrl.PostAction("next", func() error {
 		deviceToken := r.Form.Get("x_device_token") == "true"
-
-		getTargetFromGraph := func() (string, error) {
-			graph, err := ctrl.InteractionGet()
-			if err != nil {
-				return "", err
-			}
-			var target string
-			var n MagicLinkOTPNode
-			if graph.FindLastNode(&n) {
-				target = n.GetMagicLinkOTPTarget()
-			} else {
-				panic(fmt.Errorf("webapp: unexpected node for magic link: %T", n))
-			}
-
-			return target, nil
-		}
 
 		result, err := ctrl.InteractionPost(func() (input interface{}, err error) {
 			target, err := getTargetFromGraph()
--- a/pkg/auth/handler/webapp/magic_link_otp.go
+++ b/pkg/auth/handler/webapp/magic_link_otp.go
@@ -157,8 +157,30 @@ func (h *MagicLinkOTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) 
 	ctrl.PostAction("next", func() error {
 		deviceToken := r.Form.Get("x_device_token") == "true"
 
+		getTargetFromGraph := func() (string, error) {
+			graph, err := ctrl.InteractionGet()
+			if err != nil {
+				return "", err
+			}
+			var target string
+			var n MagicLinkOTPNode
+			if graph.FindLastNode(&n) {
+				target = n.GetMagicLinkOTPTarget()
+			} else {
+				panic(fmt.Errorf("webapp: unexpected node for magic link: %T", n))
+			}
+
+			return target, nil
+		}
+
 		result, err := ctrl.InteractionPost(func() (input interface{}, err error) {
+			target, err := getTargetFromGraph()
+			if err != nil {
+				return
+			}
+
 			input = &InputVerifyMagicLinkOTP{
+				Target:      target,
 				DeviceToken: deviceToken,
 			}
 			return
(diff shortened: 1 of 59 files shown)


pkg/lib/interaction/nodes/create_authenticator_magic_link_otp.go
```diff
@@ -0,0 +1,55 @@
...
+	Authenticator *authenticator.Info
+}
+
+func (e *EdgeCreateAuthenticatorMagicLinkOTP) Instantiate(ctx *interaction.Context, graph *interaction.Graph, rawInput interface{}) (interaction.Node, error) {
+	var input InputCreateAuthenticatorMagicLinkOTP
+	if !interaction.Input(rawInput, &input) {
+		return nil, interaction.ErrIncompatibleInput
+	}
+
+	// Unlike Whatsapp, magic link is already verified before this edge, so we directly go to next node
```
carmenlau, line 28: We cannot skip the verification here, and let the user simply submit the `matched` action and go to the next step. There will be a vulnerability that users don't need the code and can bypass the checking easily.

We can check the code in the email, but it needs to be dry run. And we must check the code again in the original web session.

In whatsapp otp implementation, we store the user inputted code from the webhook and trigger web session refresh. In the original web session, it will compare the user inputted code with the real code to ensure a correct code is received. When auto-refresh happens it is dry run only, and the code is consumed when the user clicks `Next`.

Please find me to further explain if it is not clear.

carmenlau (commented):
- The magic link authentication page is broken, see screenshot
- We need to discuss how to add `Do not need to ask again` checkbox on this page. This page doesn't have the submit button, does it make sense to have the checkbox here?
- Setup magic link flow is broken
 
Screenshot 2023-01-13 at 17 06 59

pkg/lib/interaction/nodes/create_authenticator_magic_link_otp_setup.go
```diff
@@ -0,0 +1,141 @@
...
+		return &NodeCreateAuthenticatorMagicLinkOTP{Stage: e.Stage, Authenticator: info}, nil
+	}
+
+	aStatus, err := ctx.Verification.GetAuthenticatorVerificationStatus(info)
+	if err != nil {
+		return nil, err
+	}
+
+	if aStatus == verification.AuthenticatorStatusVerified {
+		return &NodeCreateAuthenticatorOOB{Stage: e.Stage, Authenticator: info}, nil
```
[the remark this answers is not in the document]
IniZio: Given that `NodeCreateAuthenticatorMagicLinkOTPSetup` is pretty different from other OTP flows, I think it would be better to separate other nodes e.g in this case `NodeCreateAuthenticatorMagicLinkOTP` as well?

pkg/lib/authn/otp/service.go
```diff
@@ -31,12 +31,19 @@ func (s *Service) getCode(target string) (*Code, error) {
...
-func (s *Service) createCode(target string, codeModel *Code) (*Code, error) {
+func (s *Service) createCode(target string, otpMode OTPMode, codeModel *Code) (*Code, error) {
 	if codeModel == nil {
 		codeModel = &Code{}
 	}
 	codeModel.Target = target
-	codeModel.Code = secretcode.OOBOTPSecretCode.Generate()
+	switch otpMode {
+	case OTPModeMagicLink:
+		codeModel.Code = secretcode.MagicLinkOTPSecretCode.Generate()
```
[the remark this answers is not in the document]
IniZio: Updated, but I think the reason to hashing code for forgot password was the relatively longer expiration time (20 min) compared to login (~5 min)?

pkg/auth/handler/webapp/setup_magic_link_otp.go
```diff
@@ -0,0 +1,112 @@
...
+	ControllerFactory ControllerFactory
+	BaseViewModel     *viewmodels.BaseViewModeler
+	Renderer          Renderer
+}
+
+func (h *SetupMagicLinkOTPHandler) GetData(r *http.Request, rw http.ResponseWriter, session *webapp.Session, graph *interaction.Graph) (map[string]interface{}, error) {
+	data := map[string]interface{}{}
+	baseViewModel := h.BaseViewModel.ViewModel(r, rw)
+
+	viewmodels.Embed(data, baseViewModel)
```
carmenlau, line 49: We should show the alternatives list when setting up magic link, like setting up SMS 2fa. (signup only, not adding magic link in the settings page)

Screenshot 2023-01-16 at 17 56 02

Screenshot 2023-01-16 at 17 56 09

pkg/lib/authn/otp/sender.go
```diff
@@ -70,6 +72,16 @@ func (s *MessageSender) SendEmail(email string, opts SendOptions) error {
 	}
 	data.Email = email
 
+	if opts.OTPMode == OTPModeMagicLink {
+		url := s.Endpoints.MagicLinkVerificationEndpointURL()
+		query := url.Query()
+		query.Set("token", data.Code)
+		query.Set("target", data.Email)
```
carmenlau, line 79: Please be aware that we should not include any PII in query parameters. That information may leave in the browser history/server log...etc. 

ref: https://owasp.org/www-community/vulnerabilities/Information_exposure_through_query_strings_in_url

resources/authgear/templates/en/web/magic_link_otp.html
```diff
@@ -0,0 +1,82 @@
+<!-- TODO(newman): Implement websocket listener to continue on magic link verified -->
```
carmenlau, line 1: Can remove it?

carmenlau (commented):
Thanks for your hard work @IniZio.

And I have some friendly advice on submitting a PR.
1. Please try to create a PR with a smaller scope next time. The PR now has 4xxx lines of code that are difficult to follow. The reviewers may miss the details easily.
2. Try to create atomic commits and the commit should only do what the commit message mentioned. Sometimes there are some commit that do more than what it mentioned. e.g. 7855540d653761a520c839e9a6d3f21648a90996. That can help to make a clean commit history and easier for the reviewer to follow.

Thank you 🙇🏻‍♀️

carmenlau:
When I added more than 1 magic link authenticator on the settings page. I am not able to choose to use another one when login.


Screenshot 2023-01-16 at 18 29 29

pkg/auth/handler/webapp/setup_magic_link_otp.go
IniZio: Thanks updated 🙏 

Screen Shot 2023-01-17 at 11 32 05

carmenlau (commented):
The other part looks fine for me, thanks @IniZio

pkg/auth/handler/webapp/magic_link_otp.go
```diff
@@ -140,14 +143,46 @@ func (h *MagicLinkOTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+		if graph.FindLastNode(&n) {
+			target = n.GetMagicLinkOTPTarget()
+		} else {
+			panic(fmt.Errorf("webapp: unexpected node for magic link: %T", n))
+		}
+
+		return target, nil
+	}
+
 	ctrl.PostAction("matched", func() error {
```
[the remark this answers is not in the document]
IniZio: Sure updated 🙏

pkg/lib/authn/otp/service.go
```diff
@@ -36,19 +36,21 @@ func (s *Service) createCode(target string, otpMode OTPMode, codeModel *Code) (*
 		codeModel = &Code{}
 	}
 	codeModel.Target = target
+	codeModel.ExpireAt = s.Clock.NowUTC().Add(duration.UserInteraction)
+
 	switch otpMode {
 	case OTPModeMagicLink:
 		codeModel.Code = secretcode.MagicLinkOTPSecretCode.Generate()
-		break
+		err := s.CodeStore.Create(codeModel.Code, codeModel)
```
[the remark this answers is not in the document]
IniZio: After discussion, have concluded to store 2-way i.e. one record uses code as key while the other uses target.

On validation / resend, will always use target key as source of truth.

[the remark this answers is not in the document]
IniZio: @carmenlau Have updated codestore implementation for magic link 🙏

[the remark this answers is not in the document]
carmenlau: @IniZio Sorry, I think you misunderstood what I mean. Duplicating the records in the existing codestore seems not a good idea when is confusing, we should always use the target as the key in the existing codestore. Instead, we should use another key to store the code to email mapping.

For example,

redis key `app:{APP_ID}:otp-code:{EMAIL}` store the code model
redis key `app:{APP_ID}:magic-link-email:{CODE}` store the email string

So when we check, we use the code to fetch `app:{APP_ID}:magic-link-email:{CODE}` to get the email, then use the email to perform the original checking, get the code model from `app:{APP_ID}:magic-link-email:{CODE}`, and check the code.

See if it is clear and what do you think?

Let me know if further explanation is needed, take your time.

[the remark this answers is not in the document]
IniZio: Sorry, updated to use separate store name instead of code itself 🙇

pkg/lib/authn/otp/service.go
```diff
@@ -16,15 +16,26 @@ type CodeStore interface {
 	Delete(target string) error
 }
 
+type MagicLinkStore interface {
+	Create(token string, code *Code) error
```
carmenlau, line 20: I think MagicLinkStore should only store the email but not the code model. The reason is to avoid confusion, the original code store should be the single source of truth.

pkg/lib/authn/otp/service.go
```diff
@@ -41,7 +52,11 @@ func (s *Service) createCode(target string, otpMode OTPMode, codeModel *Code) (*
 	switch otpMode {
 	case OTPModeMagicLink:
 		codeModel.Code = secretcode.MagicLinkOTPSecretCode.Generate()
-		err := s.CodeStore.Create(codeModel.Code, codeModel)
+		err := s.MagicLinkStore.Create(codeModel.Code, codeModel)
+		if err != nil {
+			return nil, err
+		}
+		err = s.CodeStore.Update(target, codeModel)
```
carmenlau, line 59: This should be `create`?

carmenlau (commented):
Thanks a lot, @IniZio 🙇🏻‍♀️

pkg/lib/authn/otp/service.go
IniZio: Updated

pkg/lib/authn/otp/magic_link_store_redis.go
```diff
@@ -0,0 +1,89 @@
...
+	data, err := json.Marshal(target)
+	if err != nil {
+		return err
+	}
+
+	return s.Redis.WithConn(func(conn *goredis.Conn) error {
+		key := redisMagicLinkKey(s.AppID, token)
+		ttl := expireAt.Sub(s.Clock.NowUTC())
+
+		_, err := conn.SetEX(ctx, key, data, ttl).Result()
```
carmenlau, line 34: We should use `SetNX` for this case, we should not allow overwritten this mapping. Although the magic link code is relatively long, we cannot guarantee its uniqueness. If we generate a duplicated code, the original mapping will be overwritten and the verification will go to an unexpected state.

So if it is duplicated, we should return err which is done in the next 2 lines.

pkg/lib/authn/otp/magic_link_store_redis.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+	return codeModel, err
+}
+
+func (s *MagicLinkStoreRedis) Update(token string, target string, expireAt time.Time) error {
```
carmenlau, line 71: MagicLinkStoreRedis should not allow updating? We don't have to implement it if we don't need it.

pkg/lib/authn/otp/service.go
```diff
@@ -146,13 +170,20 @@ func (s *Service) SetUserInputtedCode(target string, userInputtedCode string) (*
 }
 
 func (s *Service) SetUserInputtedMagicLinkCode(userInputtedCode string) (*Code, error) {
-	codeModel, err := s.getCode(userInputtedCode)
+	target, err := s.MagicLinkStore.Get(userInputtedCode)
```
carmenlau, line 173: If it is not found, do we need to return `ErrInvalidMagicLink` instead? Please ensure the UI shows the current error if the code is invalid.

pkg/lib/authn/otp/magic_link_store_redis.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (s *MagicLinkStoreRedis) Create(token string, target string, expireAt time.Time) error {
+	return s.set(token, target, expireAt)
+}
+
+func (s *MagicLinkStoreRedis) Get(token string) (string, error) {
+	ctx := context.Background()
+	key := redisMagicLinkKey(s.AppID, token)
+	var codeModel string
```
carmenlau, line 52: This is not `codeModel`

carmenlau (commented):
🙇🏻‍♀️

pkg/lib/authn/otp/service.go
```diff
@@ -41,7 +50,11 @@ func (s *Service) createCode(target string, otpMode OTPMode, codeModel *Code) (*
 	switch otpMode {
 	case OTPModeMagicLink:
 		codeModel.Code = secretcode.MagicLinkOTPSecretCode.Generate()
-		err := s.CodeStore.Create(codeModel.Code, codeModel)
+		err := s.CodeStore.Create(target, codeModel)
+		if err != nil {
+			return nil, err
+		}
+		err = s.MagicLinkStore.Create(codeModel.Code, codeModel.Target, codeModel.ExpireAt)
```
carmenlau, line 57: Given that `MagicLinkStore.Create` may failed due to duplicate code (although it is rare), we should call `MagicLinkStore.Create` first. And call `CodeStore.Create` if `MagicLinkStore.Create` succeed.

IniZio: Updated

carmenlau (approved):
LGTM

I merge to the feature branch first, please rebase the feature branch later. Not urgent!

Thank you Newman!

PR #2814, opened 2023-01-13
[WIP] Add SMS send limit

### LINKED ISSUE #2784: [OTP] SMS send limit
## Requirements

- There will be a daily send limit for each phone number. If the send limit is exceed the enduser cannot receive a new OTP until a day later.

- Effectively this will block the user from logging in.

- Need to check if send limit is hit before checking OTP cooldown. Otherwise the enduser  will need to wait the cooldown period be over and then noticed the limit is exceeded.

- Default value: 30 per day

### AuthUI

When the limit is met, show a new page with button back to login/signup screen

> ### SMS send limit exceeded
> You have exceeded the daily limit of sending one-time passwords. Please try again later.
> 
> [Back]

### Portal

Image

Show a integer field for "Daily SMS sending limit per phone number", max: 100

Rubikium:
fix #2784
ref #2780 
after #2794 (config structure overlap)

TODO:
- [ ] Add `ErrTooManyRequestsFromBucket` error to attach bucket detail to error
- [ ] Emit above error when send limit is hit
- [ ] Check send limit bucket before OTP cooldown bucket
- [ ] Add message bar to show send limit exceeded error


portal/src/graphql/portal/LoginMethodConfigurationScreen.tsx
```diff
@@ -2165,6 +2166,19 @@ function VerificationSettings(props: VerificationSettingsProps) {
...
+      setState((prev) =>
+        produce(prev, (prev) => {
+          prev.smsRatelimitConfig.per_phone ??= {};
+          prev.smsRatelimitConfig.per_phone.size =
+            parseIntegerAllowLeadingZeros(value);
+        })
+      );
+    },
+    [setState]
+  );
```
louischan-oursky, line 2180: I suggest we change the shape of the state to match the UI. That is, I will use the following shape as the state.

```typescript
interface State {
  // The valid values are empty string and integers in string form.
  dailySMSSendLimit: string;
}
```

And then I will convert the state in `constructConfig`

```typescript
if (dailySMSSendLimit === "") {
  per_phone = {
    enabled: false
  };
} else {
  per_phone = {
    enabled: true,
    size: parseInt(dailySMSSendLimit, 10),
    reset_period: "24h",
  }
}
```

pkg/lib/ratelimit/error.go
```diff
@@ -1,9 +1,22 @@
...
+import (
+	"github.com/authgear/authgear-server/pkg/api/apierrors"
+)
 
-var ErrTooManyRequests = apierrors.TooManyRequest.WithReason("RateLimited").
-	New("request rate limited")
+var errTooManyRequestsMsg = "request rate limited"
+var RateLimited = apierrors.TooManyRequest.WithReason("RateLimited")
 
+var ErrTooManyRequests = RateLimited.New(errTooManyRequestsMsg)
```
louischan-oursky, line 10: Delete this so that everyone has to use ErrTooManyRequestsFrom

pkg/lib/ratelimit/error.go
```diff
@@ -1,9 +1,22 @@
 package ratelimit
 
-import "github.com/authgear/authgear-server/pkg/api/apierrors"
+import (
+	"github.com/authgear/authgear-server/pkg/api/apierrors"
+)
 
-var ErrTooManyRequests = apierrors.TooManyRequest.WithReason("RateLimited").
-	New("request rate limited")
+var errTooManyRequestsMsg = "request rate limited"
```
louischan-oursky, line 7: This can simply be a variable in ErrTooManyRequestsFrom

pkg/auth/handler/webapp/enter_oob_otp.go
```diff
@@ -82,6 +83,17 @@ func (h *EnterOOBOTPHandler) GetData(r *http.Request, rw http.ResponseWriter, se
...
+		if channel == model.AuthenticatorOOBChannelSMS {
+			bucket := h.AntiSpamSMSBucket.MakeBucket(target)
+			pass, _, err := h.RateLimiter.CheckToken(bucket)
+			if err != nil {
+				return nil, err
+			}
+			if !pass && baseViewModel.Error == nil {
+				baseViewModel.SetError(bucket.BucketError())
+			}
+		}
```
louischan-oursky, line 95: I thought the end-user should not be in this page if the limit is exceeded. Instead, they should see the error in the previous page in the error message bar. We should not need to inject the bucket maker and query the bucket limit.

louischan-oursky (changes requested):

pkg/lib/infra/sms/ratelimit.go
```diff
@@ -16,10 +18,20 @@ type AntiSpamSMSBucketMaker struct {
 }
 
 func (m *AntiSpamSMSBucketMaker) MakeBucket(phone string) ratelimit.Bucket {
-	return ratelimit.Bucket{
-		Key:         fmt.Sprintf("sms-message:%s", phone),
-		Name:        "AntiSpamSMSBucket",
-		Size:        m.Config.Ratelimit.PerPhone.Size,
-		ResetPeriod: m.Config.Ratelimit.PerPhone.ResetPeriod.Duration(),
+	if m.Config.Ratelimit.PerPhone.Enabled {
```
louischan-oursky, line 21: Do not make an infinite bucket. Check the enabled flag and determine whether we should take a token from the bucket.

PR #2826, opened 2023-01-17
Draft sms gateway spec

tung2744:
ref #2829

### DIFF
--- a/docs/specs/sms_gateway.md
+++ b/docs/specs/sms_gateway.md
@@ -1,0 +1,81 @@
+# SMS Gateway
+
+This document describes configuration of sms gateway.
+
+- [Using Custom SMS Gateway](#using-custom-sms-gateway)
+  - [Configuration](#configuration)
+  - [Webhook Signature](#webhook-signature)
+  - [Webhook Body](#webhook-body)
+  - [Deno Script](#deno-script)
+
+## Using Custom SMS Gateway
+
+When using custom sms gateway, `messaging.sms_provider` must be set to value `custom` in configuration. A secret with key `sms.custom` must also exist in secret configuration.
+
+### Configuration
+
+The following configs must be specified in `authgear.yaml`:
+
+```yaml
+messaging:
+  sms_provider: custom
+```
+
+- `messaging.sms_provider`
+  - Must be `custom` when using custom sms gateway
+
+And the following secrets must be specified in `authgear.secrets.yaml`:
+
+```yaml
+- data:
+    url: authgeardeno:///deno/sms.ts
+    timeout: 10
+  key: sms.custom
+```
+
+- `data.url`
+  - If it is an http / https URL, a webhook will be sent to the URL.
+  - If it is an URL with scheme `authgeardeno://`, the referenced deno module will be executed.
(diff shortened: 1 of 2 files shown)


docs/specs/sms_gateway.md
```diff
@@ -0,0 +1,73 @@
...
+The signature is calculated as the hex encoded value of HMAC-SHA256 of the request body.
+
+The signature is included in the header `x-authgear-body-signature:`.
+
+### Webhook Body
+
+The body is in json format with the following fields:
+
+- `to`: The recipient of the sms.
+- `from`: The sender of the sms.
```
louischan-oursky, line 56: I think we do not need to tell them `from`. It is their choice to select a suitable sender. The only we need to tell them is what the message is and whom the message is for.

docs/specs/sms_gateway.md
```diff
@@ -0,0 +1,73 @@
...
+{
+  "to": "+85298765432",
+  "from": "+85298765432",
+  "body": "You otp is 123456"
+}
+```
+
+### Deno Script
+
+When `url` in config is an URL with scheme `authgeardeno://`, the referenced deno module will be executed instead of sending a webhook request.
```
louischan-oursky, line 71: Just put a link here to reference to the definition of a deno script, instead of repeating the definition here.

docs/specs/sms_gateway.md
```diff
@@ -0,0 +1,73 @@
...
+  - data:
+      url: "authgeardeno:///deno/randomstring.ts"
+    key: sms.custom
+```
+
+### Webhook Signature
+
+Each request is signed with the secret key provided in the `sms.custom` section with `secret` key. The developer must validate the signature and reject requests with invalid signature to ensure the request originates from Authgear.
+
+The signature is calculated as the hex encoded value of HMAC-SHA256 of the request body.
```
louischan-oursky, line 47: Reference the existing definition of webhook signature, instead of repeating it here.

tung2744: @louischan-oursky 🤔 But I think they are actually different things, and actually we are using a different secret to sign the request of sms gateway requests. I think in future the interface can become different too.

docs/specs/sms_gateway.md
tung2744: Maybe I create one more file to explain deno script module and referernce it in both document.

docs/specs/sms_gateway.md
```diff
@@ -0,0 +1,84 @@
...
+  - [Configuration](#configuration)
+  - [Webhook Signature](#webhook-signature)
+  - [Webhook Body](#webhook-body)
+  - [Deno Script](#deno-script)
+
+## Using Custom SMS Gateway
+
+When using custom sms gateway, `messaging.sms_provider` must be set to value `custom` in configuration. A secret with key `sms.custom` must also exist in secret configuration.
+
+### Configuration
```
carmenlau, line 15: For the timeout, does it configurable?

tung2744: I missed it, just added thanks!

PR #2832, opened 2023-01-22
Decouple webapp and oauth

### LINKED ISSUE #2816: [Custom UI] Make /oauth2/authorize less coupled with Auth UI
/oauth2/authorize is now tightly coupled with Auth UI. The goal is to remove this coupling entirely so that Custom UI can replace Auth UI.

To remove the coupling, we no longer create web session in /oauth2/authorize.
Instead, we rely on URL query or cookie to let Auth UI or Custom UI to access the information originally received by /oauth2/authorize.

louischan-oursky:
fix #2816


louischan-oursky:
Updated.

Issue #2748, opened 2023-01-02
[Magic link] Login with email login links


fungc-io:
## Problem

To authenticate a user by possession of a mailbox, OTP is currently supported.
It's also common to send a link to the user which will magically log in the user.
While this method should works for both primary and secondary authenticators.
Support for secondary is priority as a requirement from the [client](https://3.basecamp.com/3096882/buckets/11477446/documents/5653990958)

## Solution

1. User receives an email with a login button
2. Redirected to the login screen hosted by Authgear, and logged in
3. User clicks on the "Continue" button
4. User is redirected via universal link to either the mobile app or a website.

(See Figma inside client project for details)

## Todo

- [ ] Update config file
- [ ] Backend support
- [ ] Email template
- [ ] SDK support
- [ ] Allow configuration in the portal 

---

- [x] #2764
- [x] #2773
- [x] #2765
    - Can refer to setup oob email
    - [x] Add menu item in /settings/mfa
- [ ] ~~#2766~~
    - Can refer to oob email
- [x] #2788
- [x] #2767
    - [x] Loading screen
	    - Can refer to wechat screen
	    - [x] Redirect on websocket event
- [x] #2768
    - [ ] Redirect on websocket event
    - [ ] Handle universal link
        Can refer to wechat's handling of universal link [docs.authgear.com/strategies/how-to-setup-sso-integrations/wechat](https://docs.authgear.com/strategies/how-to-setup-sso-integrations/wechat)
- [ ] ~~#2769~~
    - `authentication.primary.oob_otp_email_magic_link.failed`
    - `authentication.secondary.oob_otp_email_magic_link.failed`
- [x] #2770
- [x] #2771
    - [ ] Refactor to allow using unverfied authenticator
    - [ ] Implement createAuthenticator admin API mutation
- [ ] #2772

fungc-io:
Update the original OOB-OTP for email to use magic link as alternative to code inside the email message

fungc-io:
@IniZio 
Here are the updates on the design (re: [your comment](https://github.com/authgear/authgear-server/issues/2767#issuecomment-1383573202)) 
The UI is mostly good, please update the copywriting.


## Design

**Setup authenticator**

```markdown
## Set up login link 2-step verification

Input placeholder: Email address for receiving login links

[Next]
```

**Waiting for email verification**

```markdown
## Login link

We have sent a login link to xxx****@example.com
Please click on the link to continue

Didn't receive the link? [Resend (XXs)]
[...]

PR #2848, opened 2023-01-31
[TBC] Use OTP code to verify magic link setup

IniZio:
ref: #2748

based on: #2828


IniZio:
As per discussion, will also include OTP code in magic link email instead.

Issue #2807, opened 2023-01-13
Email in 2FA to be unique


fungc-io:
In use cases that 2FA is doubled as user attributes, 
Developers may want their emails to be unique.
Constraint in the table

PR #2849, opened 2023-01-31
Specifiy unique verifiable claims

kiootic:
TODO: interaction with identity uniqueness

> oauth & phone identity, primary phone OOB-OTP, required secondary unique email OOB-OTP; ac A phone + secondary email OOB-OTP; ac B signup oauth same email; conflict resolution?

ref #2807


louischan-oursky:
This should be irrelevant.

Issue #2829, opened 2023-01-18
Connect to custom SMS gateway


fungc-io:
To support an alternative SMS gateway with event hook

TODO:

- [x] Update app config: Add `custom` to `SMSProvider`
- [x] Update app secret: Add `sms.custom` key
- [x] Update send sms task: Support sending sms with custom provider in `sms.Client`
    - [x] Support webhook
      - [x] Make timeout configurable
    - [x] Support denohook
      - [x] Make timeout configurable
      - [x] Update authgear_deno_hook package

fungc-io:
est: 31 Jan

PR #2850, opened 2023-01-31
Implement custom sms gateway

tung2744:
ref #2829


pkg/lib/config/config.go
```diff
@@ -55,9 +56,10 @@ type AppConfig struct {
 	HTTP *HTTPConfig `json:"http"`
 	Hook *HookConfig `json:"hook,omitempty"`
 
-	UI           *UIConfig           `json:"ui,omitempty"`
-	Localization *LocalizationConfig `json:"localization,omitempty"`
-	Messaging    *MessagingConfig    `json:"messaging,omitempty"`
+	UI                       *UIConfig                 `json:"ui,omitempty"`
+	Localization             *LocalizationConfig       `json:"localization,omitempty"`
+	Messaging                *MessagingConfig          `json:"messaging,omitempty"`
+	CustomSMSProviderConfigs *CustomSMSProviderConfigs `json:"custom_sms_provider,omitempty"`
```
carmenlau: One quick comment first, we further review the PR later 🙏🏻

We have to put the custom sms provider config to the `authgear.secrets.yaml`. The reasons:

1. To be consistent with the other providers (twilio and nexmo)
2. We have `build-in`, `custom`, and `app` resources. `custom` is the resource defined per cluster, `app` is the resource defined per app. Different resources (`authgear.yaml`, `authgear.secrets.yaml`...etc) have their own way to load the resource. For `authgear.secrets.yaml`, we support combining `custom` and `app` resources. That means we can provide a custom SMS provider config in `authgear.secrets.yaml` of the custom resource,  then the whole cluster will be able to use the custom SMS provider. This is how we config the production cluster to use Twilio for all projects.

Let me know if further explanation is needed.

tung2744: @carmenlau
I thought we are going to have a custom `authgear.yaml` defined for a certain app.
See if I understand it correctly: In production we want to share the same `custom_sms_provider` of `authgear.secrets.yaml` config, but we will configure the sms provider used (twilio / custom) per app? So I should put `custom_sms_provider` into secrets we don't have to add this config to each app?

carmenlau: @tung2744 Actually we don’t have this use case now (share custom sms provider in production). Putting this to authgear.secrets.yaml is just for future flexibility. Maybe consistency is the stronger reason.

Any reason you want to put it in authgear.yaml instead?

tung2744: @carmenlau The reason is there is no secrets in the config, but I think I can still move it to `authgear.secrets.yaml`.

pkg/lib/hook/deno_client.go
```diff
@@ -15,37 +15,53 @@ import (
...
-	return &DenoClient{
+	return &DenoClientImpl{
 		Endpoint:   string(endpoint),
 		HTTPClient: httputil.NewExternalClient(60 * time.Second),
 		Logger:     logger,
 	}
 }
 
-type DenoClient struct {
+type DenoClientFactory func(timeout time.Duration) DenoClient
```
louischan-oursky, line 46: This approach of supporting timeout complicates the code a lot. As mention in my previous comment, we now have 

```golang
func (h *WebHookImpl) PerformNoResponse(client *http.Client, request *http.Request) (error)
```

We can easily pass in a HTTP client configured with suitable timeout. The HTTP client should be dependency injected into CustomClient, like what we did for hook.SyncHTTPClient and hook.AsyncHTTPClient.

---

So the question becomes how can we do the same for Deno.

It turns out that CustomClient only needs `loadScript` of DenoHookImpl. CustomClient can create its own DenoClient to execute the script. So we only need to expose LoadScript and dependency-inject a DenoClient with suitable timeout.

```golang
type DenoScriptLoader interface {
  LoadScript(u *url.URL) ([]byte, error)
}

type DenoClient interface {
  	Run(ctx context.Context, script string, input interface{}) (out interface{}, err error)
}

func NewDenoClient() DenoClient {
  return &hook.DenoClient{} // Use the suitable timeout here.
}

type CustomClient struct {
  DenoScriptLoader DenoScriptLoader
  DenoClient DenoClient
}
```

louischan-oursky (changes requested):

pkg/lib/hook/hook_web.go
```diff
@@ -100,6 +115,14 @@ func (h *WebHookImpl) performRequest(client *http.Client, request *http.Request,
 		return
 	}
 
+	return resp, nil
+}
+
+func (h *WebHookImpl) performEventRequest(client *http.Client, request *http.Request, withResponse bool) (hookResp *event.HookResponse, err error) {
+	resp, err := h.performRequest(client, request, withResponse)
+
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
tung2744: If the goal is to ensure the response body closed, what do you think if we make something like this?
```golang
func (h *WebHookImpl) performRequest(
	client *http.Client,
	request *http.Request,
	responseHandler *func(response *http.Response) (interface{}, error)) (interface{}, error) {
	resp, err := client.Do(request)
	// ... Check status code or timeout error

	defer resp.Body.Close()

	if responseHandler != nil {
		result, err := (*responseHandler)(resp)
		return result, err
	}

	return nil, err
}
```

So we don't have to close body in the caller.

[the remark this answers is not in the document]
louischan-oursky: I prefer we keep PerformWithResponse and PerformNoResponse. Passing a responseHandler complicates the function signature. performRequest actually does not depend on `h`, we can even move them to module-level functions.

pkg/lib/hook/deno_client.go
tung2744: Instead of creating `DenoScriptLoader`, I make the client used on run a parameter.
```golang
func (h *DenoHook) Run(client DenoClient, u *url.URL, input interface{}) (out interface{}, err error)
```
You may see the change in 44cdd6b

pkg/lib/infra/sms/hook.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/authgear/authgear-server/pkg/lib/config"
+	"github.com/authgear/authgear-server/pkg/util/httputil"
+)
+
+type HookHTTPClient struct {
+	*http.Client
+}
+
+func NewHookHTTPClient(hookCfg *config.HookConfig, smsCfg *config.CustomSMSProviderConfig) HookHTTPClient {
```
louischan-oursky, line 16: In what case smsCfg is nil? We have `func (c *SecretConfig) Validate(appConfig *AppConfig) error` in pkg/lib/config/secret.go. We can enforce when sms provider is custom, the config is present.

I think we should not use the timeout of hook config as fallback. Use a static value as fallback.

tung2744: > In what case smsCfg is nil? We have func (c *SecretConfig) Validate(appConfig *AppConfig) error in pkg/lib/config/secret.go. We can enforce when sms provider is custom, the config is present.

When provider is not `custom`, `smsCfg` config will be nil, but in `NewCustomClient` if this config is nil, `CustomClient` will become nil therefore I found that actually I don't have to check it in `NewHookHTTPClient`.

tung2744:
TODO:
- [ ] Add feature flag to support using deno unstable api in deno hook

pkg/lib/infra/sms/hook.go
```diff
@@ -9,37 +9,37 @@ import (
...
+	} else {
+		return SMSHookTimeout{Timeout: 60 * time.Second}
 	}
+}
+
+type HookHTTPClient struct {
+	*http.Client
+}
+
+func NewHookHTTPClient(hookCfg *config.HookConfig, timeout SMSHookTimeout) HookHTTPClient {
```
louischan-oursky, line 28: Remove hookCfg from arguments.

tung2744:
@louischan-oursky Updated and rebased. The new changes are after [6e47fff](https://github.com/authgear/authgear-server/pull/2850/commits/6e47fff88131772895575b64aecbc5f83591dd7f), thanks!

louischan-oursky:
@tung2744 Please drop the commits related to feature config.

tung2744:
@louischan-oursky Reverted commits related to feature flags & unstable deno api. Also rebased.

louischan-oursky:
Conflicted :(

tung2744:
@louischan-oursky Rebased again, thanks!

PR #2812, opened 2023-01-13
Add magic link config to admin portal 

IniZio:
ref: #2748 #2770

based on: #2802

### Screens

Screen Shot 2023-01-31 at 11 51 39

PR #2857, opened 2023-02-01
Allow magic link as primary authenticator

IniZio:
ref: #2748 #2856

based on: #2812

Screen Shot 2023-02-01 at 16 46 50

### DIFF
--- a/pkg/auth/handler/webapp/alternatives.go
+++ b/pkg/auth/handler/webapp/alternatives.go
@@ -101,14 +101,10 @@ func handleAlternativeSteps(ctrl *Controller) {
 				// expected there is CreateAuthenticatorBeginNode before the steps
 				return webapp.ErrSessionStepMismatch
 			}
-			switch node.GetCreateAuthenticatorStage() {
-			case authn.AuthenticationStagePrimary:
-				panic("Magic link as primary authenticator is not yet supported")
-			case authn.AuthenticationStageSecondary:
-				choiceStep = webapp.SessionStepCreateAuthenticator
-			default:
+			if node.GetCreateAuthenticatorStage() != authn.AuthenticationStagePrimary && node.GetCreateAuthenticatorStage() != authn.AuthenticationStageSecondary {
 				panic(fmt.Sprintf("webapp: unexpected authentication stage: %s", node.GetCreateAuthenticatorStage()))
 			}
+			choiceStep = webapp.SessionStepCreateAuthenticator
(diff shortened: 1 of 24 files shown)


pkg/lib/interaction/nodes/create_authenticator_magic_link_otp_setup.go
```diff
@@ -37,29 +39,74 @@ func (e *EdgeCreateAuthenticatorMagicLinkOTPSetup) AuthenticatorType() model.Aut
...
+		// Normalize the target.
+		var err error
+		target, err = ctx.LoginIDNormalizerFactory.NormalizerWithLoginIDType(model.LoginIDKeyTypeEmail).Normalize(target)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	spec.OOBOTP.Email = target
+	spec.MagicLinkOTP.Email = target
```
carmenlau, line 108: Seems we don't need `MagicLinkOTP` in the authenticator spec, no one is reading it? Now when we call `NewWithAuthenticatorID` only `OOBOTP` is read. If it is not needed, remove it first to avoid confusion.

pkg/auth/handler/webapp/alternatives.go
```diff
@@ -103,7 +103,7 @@ func handleAlternativeSteps(ctrl *Controller) {
 			}
 			switch node.GetCreateAuthenticatorStage() {
 			case authn.AuthenticationStagePrimary:
-				panic("Magic link as primary authenticator is not yet supported")
+				choiceStep = webapp.SessionStepCreateAuthenticator
```
carmenlau, line 106: Both primary and secondary stage are having the same `choiceStep`, don't need the switch case?

IniZio: Updated

pkg/lib/interaction/nodes/create_authenticator_magic_link_otp_setup.go
IniZio: Originally this struct was in case of magic link specific field, removed since not needed atm.

IniZio:
Seems getting timeout when linting in CI from time to time https://github.com/authgear/authgear-server/actions/runs/4063488765/jobs/6995779799, maybe can increase the timeout for  `golangci-lint`?

carmenlau (commented):
If the project enables both password and login link primary authenticator. During signup, the user can switch between password and login link. When the user selects login link, the UI asks the user to input the email again which is unexpected.

Issue #2846, opened 2023-01-31
Check invalid code before entering password in reset password flow


fungc-io:
In reset password flow, a link for users to reset the password will be sent to the user.

The link will be expired after a period or used.
In the current behaviour, the end-user can only know that after they have entered the password. 
A better UX will be knowing the link is invalid immediately after opening it.

- [ ] Reset Primary password
- [ ] ~Reset Secondary Password (when this flow is supported authgear/authgear-server#2845)~

Copy writing:

similar to `"error-password-reset-failed": "This reset password link is invalid, used or expired. Please request a new one."`

```markdown

## Invalid link

This reset password link is invalid, used or expired. Please request a new one.

```

PR #2865, opened 2023-02-01
Display error early for invalid reset code

kiootic:
ref #2846

### DIFF
--- a/pkg/lib/feature/forgotpassword/provider.go
+++ b/pkg/lib/feature/forgotpassword/provider.go
@@ -51,6 +51,7 @@ type TranslationService interface {
 
 type RateLimiter interface {
 	TakeToken(bucket ratelimit.Bucket) error
+	RequireToken(bucket ratelimit.Bucket) error
 }
 
 type ProviderLogger struct{ *log.Logger }
@@ -260,11 +261,6 @@ func (p *Provider) sendSMS(phone string, code string) (err error) {
 // If the code is found but expired, ErrExpiredCode is returned.
 // if the code is found but used, ErrUsedCode is returned.
 func (p *Provider) checkResetPasswordCode(codeStr string) (code *Code, err error) {
-	err = p.RateLimiter.TakeToken(AntiBruteForceVerifyBucket(string(p.RemoteIP)))
-	if err != nil {
-		return
-	}
-
 	codeHash := HashCode(codeStr)
 	code, err = p.Store.Get(codeHash)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/lib/feature/forgotpassword/provider.go
```diff
@@ -247,21 +247,18 @@ func (p *Provider) sendSMS(phone string, code string) (err error) {
...
+// CheckResetPasswordCode check whether the code is valid.
 // If the code is invalid, ErrInvalidCode is returned.
 // If the code is found but expired, ErrExpiredCode is returned.
 // if the code is found but used, ErrUsedCode is returned.
-// Otherwise, the password is reset to newPassword.
-// newPassword is checked against the password policy so
-// password policy error may also be returned.
-func (p *Provider) ResetPasswordByCode(codeStr string, newPassword string) (err error) {
+func (p *Provider) checkResetPasswordCode(codeStr string) (code *Code, err error) {
 	err = p.RateLimiter.TakeToken(AntiBruteForceVerifyBucket(string(p.RemoteIP)))
```
louischan-oursky, line 255: This TakeToken step should only be done for failed verify attempt. So merely checking a correct code does not take a token. And this implies keep refreshing a page with correct code will not encounter rate limit error.

kiootic: In that case, the rate limit would be useless, since potential attacker may check correctness of code by GETting the page instead of POSTing the page.

louischan-oursky: We count failed attempts so the attacker can only try a few times

louischan-oursky: btw there are other places having similar issue. https://github.com/authgear/authgear-server/issues/2869 If you have no tasks on hand, please take it.

louischan-oursky:
I have changed the base branch, please rebase.

pkg/lib/feature/forgotpassword/provider.go
```diff
@@ -284,17 +277,49 @@ func (p *Provider) ResetPasswordByCode(codeStr string, newPassword string) (err
 		return
 	}
 
+	return
+}
+
+func (p *Provider) CheckResetPasswordCode(codeStr string) error {
```
louischan-oursky, line 283: I can now refresh as many times as I want to guess the reset password code.

Instead of rate limiting the attempt, we want to limit failed attempt. Rename `AntiBruteForceVerifyBucket` to `TrackFailedAttemptBucket`. Take the token when the code is invalid. No need to call RequireToken or CheckToken before verifying the code.

kiootic: If we don't check/require token before verifying the code, the result as seen by attacker would be either rate limit error or valid code? i.e. attacker could just ignore rate limit error and continue, since we don't check rate limit for correct code.

kiootic: Updated.

PR #2867, opened 2023-02-01
Workflow HTTP API

### LINKED ISSUE #2818: [Custom UI] Implement the HTTP API for workflow
Some design notes

- Support multiple inputs. A screen may collect information that advance the workflow more than once. It should be easy for the client to handle this as one HTTP call.

louischan-oursky:
fix #2818


carmenlau (commented):
The PR looks good with only one minor problem. When I call the API with incorrect intent or input kind, the error is not handled correctly.

louischan-oursky:
- [ ] Add GetCookies and collect them on finish

louischan-oursky:
Added 5 commits to address the issues.

PR #2871, opened 2023-02-02
Rate limit bucket returns same error instance

### LINKED ISSUE #2868: [OTP] Resend cooldown should not be shared across interactions
Steps to reproduce

1. Sign in with a phone number
2. See the enter OTP page.
3. Click back
4. Sign in with the same phone number
5. See rate limit error

Expected

1. Sign in with a phone number
2. See the enter OTP page and receive a OTP
3. Click back
4. Sign in with the same phone number
5. See the enter OTP page and receive a **different** OTP
6. Be able to repeat the above steps until https://github.com/authgear/authgear-server/issues/2784 is hit

Rubikium:
fix #2868

### DIFF
--- a/pkg/lib/ratelimit/error.go
+++ b/pkg/lib/ratelimit/error.go
@@ -7,8 +7,9 @@ import (
 var ErrUsageLimitExceeded = apierrors.ServiceUnavailable.WithReason("UsageLimitExceeded").
 	New("usage limit exceeded")
 
+var RateLimited = apierrors.TooManyRequest.WithReason("RateLimited")
+
 func ErrTooManyRequestsFrom(bucket Bucket) error {
-	RateLimited := apierrors.TooManyRequest.WithReason("RateLimited")
 	errMsg := "request rate limited"
 	if bucket.Name == "" {
 		return RateLimited.New(errMsg)
--- a/pkg/lib/interaction/nodes/utils_authn.go
+++ b/pkg/lib/interaction/nodes/utils_authn.go
@@ -1,14 +1,14 @@
 package nodes
 
 import (
-	"errors"
-
+	"github.com/authgear/authgear-server/pkg/api/apierrors"
 	"github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/lib/authn"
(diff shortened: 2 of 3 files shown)


pkg/lib/ratelimit/error.go
```diff
@@ -2,18 +2,19 @@ package ratelimit
...
+
 func ErrTooManyRequestsFrom(bucket Bucket) error {
-	RateLimited := apierrors.TooManyRequest.WithReason("RateLimited")
-	errMsg := "request rate limited"
 	if bucket.Name == "" {
-		return RateLimited.New(errMsg)
+		return errTooManyRequests
 	} else {
-		return RateLimited.NewWithInfo(errMsg, apierrors.Details{
+		return apierrors.AddInfo(errTooManyRequests, errorutil.Details{
```
louischan-oursky, line 17: First of all, we cannot modify a global variable in a request goroutine. This means there could be multiple request goroutines race to modify the global variable with different value.

Why do we need the same instance in the first place?

Rubikium: Currently we compare errors using `errors.Is()` during error handling. This function will return false if the two arguments are different error instances: `errors.Is(errors.New("test"), errors.New("test")) == false`, which is the case when comparing errors from `BucketError()`

louischan-oursky: https://pkg.go.dev/errors#Is An error can implement Is() to define equality. @carmenlau I suggest we implement Is() as follows. Thoughts?

```golang
type skyerr struct {
  kind Kind
}

func (e *skyerr) Is(err error) bool {
  apiError := AsAPIError(err)
  if apiError == nil {
    return false
  }
  return e.Kind == apiError.Kind
}

func (e *APIError) Is(err error) bool {
  apiError := AsAPIError(err)
  if apiError == nil {
    return false
  }
  return e.Kind == apiError.Kind
}
```

louischan-oursky: 1. Move `RateLimited := apierrors.TooManyRequest.WithReason("RateLimited")` to a module variable
2. Use apierrors.IsKind instead errors.Is in pkg/lib/interaction/nodes/utils_authn.go:87 and pkg/lib/interaction/nodes/verify_identity.go:124

Issue #2858, opened 2023-02-01
Trigger email verification from SDK


fungc-io:
- [x] Update server to support app session action via authz endpoint
- [ ] Update SDKs to support performing actions
    - [ ] iOS native
    - [ ] Android native
    - [ ] JS web and reactnative
    - [ ] Flutter
    - [ ] Xamarin

carmenlau:
We had a discussion on how to perform setting actions (verify email, change password) in SDK. The following flowcharts explain how it works.

We will work on the simplified version first.

## Simplified version

[...]

PR #2873, opened 2023-02-03
Support start verify email settings action via authz endpoint (without settings action code)

carmenlau:
ref #2858 

This is a simplified version (without settings action code)


carmenlau:
We are going to do it in the client middleware

Issue #2883, opened 2023-02-06
[Magic Link] Implement auto approval through universal link


IniZio:
- Add API for approval request through SDK's universal link handler
- Implement login link handler in SDK's

IniZio:
Will not implement official auto-approval support yet, closing.

PR #2885, opened 2023-02-07
[TBC] Add API handler for verify magic link

IniZio:
ref: #2883

#### Not covered

- Different universal link per client

### DIFF
--- a/pkg/auth/handler/api/magic_link_verification.go
+++ b/pkg/auth/handler/api/magic_link_verification.go
@@ -1,0 +1,92 @@
+package api
+
+import (
+	"net/http"
+
+	"github.com/authgear/authgear-server/pkg/api"
+	"github.com/authgear/authgear-server/pkg/api/apierrors"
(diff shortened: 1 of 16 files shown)


pkg/auth/handler/api/magic_link_verification.go
```diff
@@ -0,0 +1,92 @@
...
+		h.JSON.WriteResponse(resp, &api.Response{Error: err})
+		return
+	}
+
+	_, err = h.MagicLinkOTPCodeService.SetUserInputtedMagicLinkCode(payload.Code)
+	if err != nil {
+		if !apierrors.IsAPIError(err) {
+			h.Logger.WithError(err).Error("magic link verification handler failed")
+		}
+		h.JSON.WriteResponse(resp, &api.Response{Error: err})
```
carmenlau, line 66: We should return here?

IniZio:
Will adopt implementation according to #2898 in another PR, closing.

Issue #2882, opened 2023-02-06
[Magic Link] Specify universal link


IniZio:
- Different cases (https://www.notion.so/oursky/Magic-Link-Cases-5fba0465a39949d797323a112052e331)
- Will host on authgear (https://www.notion.so/oursky/Magic-Link-Universal-link-35a8a6decd38406b8b4bce1da8d2d253)

PR #2887, opened 2023-02-07
Specify universal link

IniZio:
ref: #2882


IniZio:
- [ ] Specify more on exact configure yaml and public API

IniZio:
@carmenlau Not sure if we will close this as well?

Issue #2884, opened 2023-02-06
[Magic Link] Universal link per application


IniZio:
- Support universal link path per client
- Add universal config in admin portal

IniZio:
Will not implement official universal link yet, closing.

Issue #2898, opened 2023-02-08
2023C01 umbrella issue for 2023-02-24


louischan-oursky:
## 2023-02-10
- [x] https://github.com/authgear/authgear-server/issues/2780
- [x] https://github.com/authgear/authgear-server/issues/2891 (Block login after X failed attempt)
- [x] https://github.com/authgear/authgear-server/issues/2829
- [x] https://github.com/authgear/authgear-server/issues/2775

## 2023-02-17
- [ ] https://github.com/authgear/authgear-server/issues/2934
- [ ] https://github.com/authgear/authgear-server/issues/2930
- [ ] https://github.com/authgear/authgear-server/issues/2929
- [ ] https://github.com/authgear/authgear-server/issues/2906
- [x] https://github.com/authgear/authgear-server/issues/2920
- [x] https://github.com/authgear/authgear-server/issues/2820
  - [ ] https://github.com/authgear/authgear-server/issues/2906
  - [x] https://github.com/authgear/authgear-server/issues/2920
- [x] https://github.com/authgear/authgear-server/issues/2819
- [x] https://github.com/authgear/authgear-server/issues/2894
  - [ ] https://github.com/authgear/authgear-server/issues/2899
- [x] https://github.com/authgear/authgear-server/issues/2905
- [x] https://github.com/authgear/authgear-server/issues/2921
- [x] Show email in userinfo: Solved by email identity
- [ ] Show email in JWT: To be done in the middleware
- [ ] Custom email checking rules: To de done in the middleware
- [x] https://github.com/authgear/authgear-server/issues/2748
- [ ] Magic link with universal link
  - [x] https://github.com/authgear/authgear-server/issues/2901
  - [x] ~~https://github.com/authgear/authgear-server/issues/2902~~ Abandoned
  - [ ] Give an example on how to integrate the whole thing. Use the demo app for testing.

## 2023-02-24
- [ ] https://github.com/authgear/authgear-server/issues/2892
- [ ] https://github.com/authgear/authgear-server/issues/2893
- [ ] https://github.com/authgear/authgear-server/issues/2895
- [ ] https://github.com/authgear/authgear-server/issues/2935
- [x] https://github.com/authgear/authgear-server/issues/2921
- [ ] Document instructions on how to create a user without password in the client's portal
- [x] https://github.com/authgear/authgear-server/issues/2896
- [x] https://github.com/authgear/authgear-server/issues/2897
- [ ] Update email templates for the client
  - [ ] Login link email
[...]

PR #2900, opened 2023-02-08
Implement universal link for login link

IniZio:
ref: #2884

based on #2885


IniZio:
Will adopt implementation according to #2898 in another PR, closing.

Issue #2901, opened 2023-02-08
Include .ClientID to all messages render context


louischan-oursky:
So the template can render something different depending on the client ID.

PR #2903, opened 2023-02-08
Include client id in email and sms template data

IniZio:
ref: #2901

### DIFF
--- a/pkg/lib/translation/service.go
+++ b/pkg/lib/translation/service.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/authgear/authgear-server/pkg/lib/clientid"
 	"github.com/authgear/authgear-server/pkg/util/intl"
 	"github.com/authgear/authgear-server/pkg/util/template"
 )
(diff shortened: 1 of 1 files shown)


pkg/lib/translation/service.go
```diff
@@ -119,17 +120,20 @@ func (s *Service) emailMessageHeader(name string, args interface{}) (sender, rep
 }
 
 func (s *Service) EmailMessageData(msg *MessageSpec, args interface{}) (*EmailMessageData, error) {
-	sender, replyTo, subject, err := s.emailMessageHeader(msg.Name, args)
+	data := map[string]interface{}{"ClientID": clientid.GetClientID(s.Context)}
```
carmenlau, line 123: The translation service should handle constructing the message data only, it should not have context about client id. The client id should be provided by the caller. Thought?

carmenlau: Any updates?

IniZio: @carmenlau If so then I think `ClientID` should be only included for magic link generate code email? Kind of weird to enforce client ID passing for messages that should not need client ID at all.

IniZio: Updated thanks!

louischan-oursky: We need to have ClientID for all outgoing messages. I think EmailMessageData is the safest place to handle this. ClientID will never be missed. Please revert back to the original implementation.

IniZio: Updated :pray:

PR #2907, opened 2023-02-08
Rename login link query param and message templates

IniZio:


carmenlau (commented):
@IniZio I found that the copywriting of `setup_primary_login_link` template is incorrect. It says it is Two-Step Verification.

Can you help to double-check the copywriting and make sure the correct templates are loaded?

thanks!!

Issue #2905, opened 2023-02-08
[Custom UI] Disable authentication and settings in auth UI


louischan-oursky:
- Keep public_signup_disabled. It is being used.
- Add `ui.authentication_disabled`. By default it is false. When it is true, /login, /signup, and /promote shows a UI similar to public_signup_disabled, saying that authentication is disabled.
- Add `ui.settings_disabled`. By default it is false, When it is true, /settings and /settings/* shows a UI saying that settings is disabled.

IniZio:
Or should we use existing`custom_ui_enabled` instead? Since otherwise this flag will be pretty confusing.

PR #2909, opened 2023-02-08
[WIP] Support disable authentication and settings page

IniZio:
ref: #2905

### DIFF
--- a/pkg/auth/routes.go
+++ b/pkg/auth/routes.go
@@ -254,7 +254,9 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource) *h
 	router.Add(webapphandler.ConfigureForceChangePasswordRoute(webappPageRoute), p.Handler(newWebAppForceChangePasswordHandler))
 	router.Add(webapphandler.ConfigureForceChangeSecondaryPasswordRoute(webappPageRoute), p.Handler(newWebAppForceChangeSecondaryPasswordHandler))
 	router.Add(webapphandler.ConfigureConnectWeb3AccountRoute(webappSIWERoute), p.Handler(newWebAppConnectWeb3AccountHandler))
-	router.Add(webapphandler.ConfigureMissingWeb3WalletRoute(webappPageRoute), p.Handler(newWebAppMissingWeb3WalletHandler))
+	// router.Add(webapphandler.ConfigureMissingWeb3WalletRoute(webappPageRoute), p.Handler(newWebAppMissingWeb3WalletHandler))
+	router.Add(webapphandler.ConfigureMissingWeb3WalletRoute(webappPageRoute), p.Handler(newWebAppFeatureDisabledHandler))
+	router.Add(webapphandler.ConfigureFeatureDisabledRoute(webappPageRoute), p.Handler(newWebAppFeatureDisabledHandler))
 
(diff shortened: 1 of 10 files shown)


resources/authgear/templates/en/web/__watermark.html
```diff
@@ -1,15 +1,15 @@
...
+      </div>
+      {{ else }}
+      <a href="https://www.authgear.com/?utm_campaign=uibrand" target="_blank">
+        <span class="footer-watermark"></span>
+      </a>
+      {{ end }}
     {{ end }}
-  {{ end }}
-</div>
+  </div>
```
carmenlau, line 14: 1. Please correct the indentation.
2. The `</div>` is outside the `if` unexpectedly

carmenlau: After we changed the `__watermark.html` layout, will it affect the layout of the normal page?

pkg/auth/handler/webapp/viewmodels/base.go
```diff
@@ -217,6 +219,8 @@ func (m *BaseViewModeler) ViewModel(r *http.Request, rw http.ResponseWriter) Bas
 		},
 		ForgotPasswordEnabled:       *m.ForgotPassword.Enabled,
 		PublicSignupDisabled:        m.Authentication.PublicSignupDisabled,
+		AuthenticationDisabled:      m.AuthUI.AuthenticationDisabled && clientid.GetClientID(r.Context()) != "portal",
```
carmenlau, line 222: Why do we need `clientid.GetClientID(r.Context()) != "portal"`?

IniZio: Otherwise the portal admin will be locked out as well when the authgear is self-hosted or developed locally?

resources/authgear/templates/en/web/__watermark.html
IniZio: Looks like I removed the commit that fixes other normal pages. Reverted change on watermark template and handled the vertical padding in the notices instead.

pkg/auth/handler/webapp/viewmodels/base.go
louischan-oursky: We can change the configuration depending on situation. Do not add development hacks.

louischan-oursky:
Use middleware to enforce the effect of the 2 new flags. Do not modify the existing templates.

pkg/auth/handler/webapp/viewmodels/base.go
IniZio: Removed

pkg/auth/routes.go
```diff
@@ -168,10 +168,15 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource) *h
 	)
 	webappAuthEntrypointChain := httproute.Chain(
 		webappPageChain,
+		p.Middleware((newRequireAuthenticationEnabledMiddleware)),
```
louischan-oursky, line 171: Do not add the middleware to webappAuthEntrypointChain. Add a new chain for /login and /signup, and add the middleware there.

IniZio: Updated

louischan-oursky:
Build failed

IniZio:
Gave up on CI, closing

Issue #2820, opened 2023-01-16
[Custom UI] Sign in workflow


louischan-oursky:
requires https://github.com/authgear/authgear-server/issues/2906

PR #2912, opened 2023-02-09
Login workflow

carmenlau:
refs #2820 

- [x] Authenticate phone oob otp
- [x] Authenticate password
- [x] Authenticate magic link
- [x] Fire failed authenticated events
- [x] Fire authenticated event
- [x] Update user's last login at 
- [x] Create session
- [x] Mark email as verified when using email login link


pkg/latte/node_authenticate_oob_otp_phone_end.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/authgear/authgear-server/pkg/lib/authn/authenticator"
+	"github.com/authgear/authgear-server/pkg/lib/workflow"
+)
+
+func init() {
+	workflow.RegisterNode(&NodeAuthenticateOOBOTPPhoneEnd{})
+}
+
+type NodeAuthenticateOOBOTPPhoneEnd struct {
```
louischan-oursky, line 14: I think this node can be generalized to

```golang
type NodeVerifiedAuthenticator struct {
  AuthenticatorInfo *authenticator.Info
}
```

carmenlau:
@louischan-oursky Updated. The login flow is ready for review!

Issue #2894, opened 2023-02-08
[Custom UI] Verify email workflow


louischan-oursky:
requires https://github.com/authgear/authgear-server/issues/2899

PR #2914, opened 2023-02-09
Add verify identity workflow

kiootic:
ref #2894

### DIFF
--- a/docs/workflows/verify-identity.http
+++ b/docs/workflows/verify-identity.http
@@ -1,0 +1,76 @@
+# @no-cookie-jar
+
+{{+
+exports.headers = {
+'Content-Type': 'application/json',
+'Cookie': `session=${SESSION_COOKIE}`
+};
+}}
+
+{{+response
+$global.instance_id = response.parsedBody.result?.workflow?.instance_id ?? $global.instance_id;
+}}
+
+
+###
+# @name start
+POST {{HOST}}/api/workflow/v1
+...headers
+
+{
+    "intent": {
+        "kind": "latte.IntentVerifyUser",
+        "data": {}
+    }
+}
+
+
+###
+# @name get
+GET {{HOST}}/api/workflow/v1/{{$global.instance_id}}
+...headers
+
+
+###
+# @name trigger
+POST {{HOST}}/api/workflow/v1/{{$global.instance_id}}
+...headers
+
+{
+    "input": {
+        "kind": "latte.InputTriggerVerification",
+        "data": {
+            "method": "", // "email", "sms"
+            "claim_name": "", // "email", "phone_number"
+            "claim_value": ""
+        }
+    }
+}
+
+###
+# @name resend_code
+POST {{HOST}}/api/workflow/v1/{{$global.instance_id}}
+...headers
+
+{
+    "input": {
+        "kind": "latte.InputOTPVerificationResend",
+        "data": {
+        }
+    }
+}
+
+
+###
+# @name enter_otp
+POST {{HOST}}/api/workflow/v1/{{$global.instance_id}}
+...headers
+
+{
+    "input": {
+        "kind": "latte.InputOTPVerification",
+        "data": {
+            "code": ""
+        }
+    }
+}
--- a/pkg/latte/input_otp_verification_resend.go
+++ b/pkg/latte/input_otp_verification_resend.go
@@ -1,0 +1,36 @@
+package latte
+
+import (
+	"github.com/authgear/authgear-server/pkg/lib/workflow"
+	"github.com/authgear/authgear-server/pkg/util/validation"
+)
+
+func init() {
+	workflow.RegisterPublicInput(&InputOTPVerificationResend{})
+}
+
+var InputOTPVerificationResendSchema = validation.NewSimpleSchema(`
+	{
+		"type": "object",
+		"additionalProperties": false
+	}
+`)
+
+type InputOTPVerificationResend struct {
+}
+
+func (*InputOTPVerificationResend) Kind() string {
+	return "latte.InputOTPVerificationResend"
+}
+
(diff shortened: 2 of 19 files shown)


pkg/latte/intent_verify_identity.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return nil, workflow.ErrEOF
+}
+
+func (*IntentVerifyIdentity) ReactTo(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow, input workflow.Input) (*workflow.Node, error) {
+	var trigger inputTriggerVerification
+
+	switch {
+	case workflow.AsInput(input, &trigger):
+		userID := session.GetUserID(ctx)
```
louischan-oursky, line 60: userID should be read from the intent itself.

pkg/latte/input_otp_verification_resend.go
```diff
@@ -0,0 +1,39 @@
...
+
+var InputOTPVerificationResendSchema = validation.NewSimpleSchema(`
+	{
+		"type": "object",
+		"additionalProperties": false
+	}
+`)
+
+type InputOTPVerificationResend struct {
+	Code string `json:"code"`
```
louischan-oursky, line 20: This input has no fields.

pkg/latte/intent_verify_user.go
```diff
@@ -0,0 +1,94 @@
...
+
+var IntentVerifyUserSchema = validation.NewSimpleSchema(`
+	{
+		"type": "object",
+		"additionalProperties": false
+	}
+`)
+
+type IntentVerifyUser struct {
+	UserID string `json:"user_id"`
```
louischan-oursky, line 26: This intent has no fields.

pkg/latte/node_do_verify_identity.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/authgear/authgear-server/pkg/lib/feature/verification"
+	"github.com/authgear/authgear-server/pkg/lib/workflow"
+)
+
+func init() {
+	workflow.RegisterNode(&NodeDoVerifyIdentity{})
+}
+
+type NodeDoVerifyIdentity struct {
+	IdentityID    string
```
louischan-oursky, line 15: Add json struct tags.

pkg/latte/ratelimit.go
```diff
@@ -0,0 +1,68 @@
+package latte
```
louischan-oursky, line 1: Ratelimit may be changed recently. Use interaction directly to avoid stale code.

pkg/latte/node_verify_email.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/authgear/authgear-server/pkg/lib/ratelimit"
+	"github.com/authgear/authgear-server/pkg/lib/workflow"
+)
+
+func init() {
+	workflow.RegisterNode(&NodeVerifyEmail{})
+}
+
+type NodeVerifyEmail struct {
+	UserID     string
```
louischan-oursky, line 20: Add json struct tags.

pkg/latte/node_verify_phone_sms.go
```diff
@@ -0,0 +1,133 @@
...
+func init() {
+	workflow.RegisterNode(&NodeVerifyPhoneSMS{})
+}
+
+type NodeVerifyPhoneSMS struct {
+	UserID      string
+	IdentityID  string
+	PhoneNumber string
+
+	CodeLength int
```
louischan-oursky, line 25: Add json struct tags.

docs/workflows/verify-identity.http
```diff
@@ -0,0 +1,76 @@
+# @no-cookie-jar
```
IniZio, line 1: @kiootic Just curious how do you use this file with the env file?

kiootic: I use [httpyac](https://httpyac.github.io/) VS Code extension.

IniZio: Thanks!

pkg/lib/workflow/workflow.go
```diff
@@ -399,3 +399,26 @@ func (w *Workflow) ToOutput(ctx context.Context, deps *Dependencies) (*WorkflowO
 
 	return output, nil
 }
+
+func FindNode[T NodeSimple](w *Workflow) (T, bool) {
```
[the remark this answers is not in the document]
kiootic: Not really, I expect to use it to find singleton nodes. Anyways I'll update the function to be `FindSingleNode`.

pkg/latte/ratelimit.go
kiootic: Do we intend to move ratelimit to `latte` eventually? If so I could alias the functions to `interaction` package.

pkg/lib/workflow/workflow.go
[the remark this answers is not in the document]
kiootic: After some consideration, I think the concept here is to obtain the return value of a sub-workflow (e.g. the result of verification sub-workflow). Should we have a helper function for this?

pkg/latte/ratelimit.go
louischan-oursky: We intent to use the same set of ratelimit eventually. For the time being, aliasing is a good idea.

pkg/lib/workflow/workflow.go
```diff
@@ -399,3 +399,35 @@ func (w *Workflow) ToOutput(ctx context.Context, deps *Dependencies) (*WorkflowO
 
 	return output, nil
 }
+
+func FindNode[T NodeSimple](w *Workflow, cb func(T)) {
```
louischan-oursky, line 403: This function seems to me is doing traversing, with a type guard. What about

```golang
func TraverseNodes[T NodeSimple](w *Workflow, cb func(n T, w *Workflow) error) error
```

PR #2915, opened 2023-02-09
Add intent to approve login link

IniZio:
ref: #2906

depends on: #2912

### DIFF
--- a/pkg/latte/intent_approve_login_link.go
+++ b/pkg/latte/intent_approve_login_link.go
@@ -1,0 +1,61 @@
+package latte
+
+import (
+	"context"
+
+	"github.com/authgear/authgear-server/pkg/lib/workflow"
+	"github.com/authgear/authgear-server/pkg/util/validation"
+)
+
+func init() {
+	workflow.RegisterPublicIntent(&IntentApproveLoginLink{})
+}
+
+var IntentApproveLoginLinkSchema = validation.NewSimpleSchema(`{}`)
+
+type IntentApproveLoginLink struct{}
+
+func (i *IntentApproveLoginLink) Kind() string {
+	return "latte.IntentApproveLoginLink"
+}
+
+func (i *IntentApproveLoginLink) JSONSchema() *validation.SimpleSchema {
+	return IntentApproveLoginLinkSchema
+}
+
+func (i *IntentApproveLoginLink) CanReactTo(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow) ([]workflow.Input, error) {
+	if len(w.Nodes) == 0 {
+		return []workflow.Input{
+			&InputTakeLoginLinkCode{},
+		}, nil
+	}
+
+	return nil, workflow.ErrEOF
+}
+
+func (i *IntentApproveLoginLink) ReactTo(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow, input workflow.Input) (*workflow.Node, error) {
+	var inputTakeLoginLinkCode inputTakeLoginLinkCode
+
+	switch {
+	case workflow.AsInput(input, &inputTakeLoginLinkCode):
+		code := inputTakeLoginLinkCode.GetCode()
+		_, err := deps.OTPCodes.VerifyMagicLinkCode(code, false)
+		if err != nil {
+			return nil, err
+		}
+
+		return workflow.NewNodeSimple(
+			&NodeVerifiedLoginLink{Code: code},
+		), nil
+	default:
+		return nil, workflow.ErrIncompatibleInput
+	}
+}
+
+func (i *IntentApproveLoginLink) GetEffects(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow) (effs []workflow.Effect, err error) {
+	return nil, nil
+}
+
+func (i *IntentApproveLoginLink) OutputData(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow) (interface{}, error) {
+	return map[string]interface{}{}, nil
+}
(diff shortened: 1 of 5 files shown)


pkg/latte/intent_verify_login_link.go
```diff
@@ -0,0 +1,79 @@
...
+	webSession, err := webSessionProvider.GetSession(codeModel.WebSessionID)
+	if err != nil {
+		return nil, err
+	}
+	err = webSessionProvider.UpdateSession(webSession)
+	if err != nil {
+		return nil, err
+	}
+
+	return nil, workflow.ErrEOF
```
IniZio, line 70: @louischan-oursky Currently it returns internal error. From https://github.com/authgear/authgear-server/blob/2023C01/docs/specs/workflow.md#canreactto-and-reactto this return should return 200 with no next node?

pkg/latte/intent_verify_login_link.go
```diff
@@ -0,0 +1,79 @@
...
+func (i *IntentVerifyLoginLink) GetCode() string {
+	return i.Code
+}
+
+func (i *IntentVerifyLoginLink) CanReactTo(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow) ([]workflow.Input, error) {
+	return []workflow.Input{}, nil
+}
+
+func (i *IntentVerifyLoginLink) ReactTo(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow, input workflow.Input) (*workflow.Node, error) {
+	codeModel, err := deps.OTPCodeService.SetUserInputtedMagicLinkCode(i.GetCode())
```
IniZio, line 47: I do it in intent because this is not an official feature of authgear. Or would still prefer to do it in node instead?

pkg/latte/intent_approve_login_link.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, err
+	}
+
+	_, err = deps.OTPCodeService.VerifyMagicLinkCode(i.GetCode(), true)
+	if err != nil {
+		return nil, err
+	}
+
+	// Update the web session and trigger the refresh event
+	webSessionProvider := deps.GlobalSessionServiceFactory.NewGlobalSessionService(
```
kiootic, line 58: Not sure if we should use webapp session in workflow. What is the expected method for workflow API consumer to handle magic link approval?

pkg/latte/intent_verify_login_link.go
kiootic: I also encountered a similar error, but the symptom is a finished workflow would return `continue` action. I think we don't expect `workflow.ErrEOF` to be returned by `ReactTo`, only `CanReactTo` should return this error.

pkg/latte/intent_approve_login_link.go
IniZio: This approval is actually more of a ping to the original session, which will actually consume the login link token.

The reason of moving webapp session refresh to workflow was to make it latte-specific. In the future if we have a different implementation of login link auto-approve, it won't affect this feature

pkg/latte/intent_verify_login_link.go
IniZio: I think from spec returning `nil, workflow.ErrEOF` will exit the intent without doing anything though. 

I did check `FindInputReactor()` method and from comments `workflow.ErrEOF` is an accepted output from `ReactTo`.

pkg/latte/intent_approve_login_link.go
kiootic: So this approval would store the code somewhere, and the custom UI may retrieve it through some API. Do we have this API at the moment?

As a developer of custom UI, how would I detect user has approved the login link in browser/app? Do I poll the get workflow API to see any changes, or workflow API would provide real-time notification on workflow changes?

pkg/latte/intent_verify_login_link.go
louischan-oursky: ErrEOF can only be returned by CanReactTo.

pkg/latte/intent_approve_login_link.go
IniZio: Seems current workflow does not plan to support real-time notification yet so will need to use polling for now?

louischan-oursky: We will add a websocket endpoint for workflow, like the websocket endpoint for webapp.

pkg/latte/intent_approve_login_link.go
```diff
@@ -0,0 +1,79 @@
+package latte
```
louischan-oursky, line 1: 1. This PR need not depend on https://github.com/authgear/authgear-server/pull/2912
2. IntentApproveLoginLink takes nothing to create.
3. IntentApproveLoginLink take an input which contains the code.
4. IntentApproveLoginLink ReactTo the input. It checks the code. If the code is correct, set SetUserInputtedMagicLinkCode and produce a new node NodeVerifiedLoginLinkCode.
5. NodeVerifiedLoginLinkCode is a deadend.
6. IntentApproveLoginLink becomes finished if it has 1 node (NodeVerifiedLoginLinkCode).

pkg/lib/authn/otp/service.go
```diff
@@ -166,7 +166,12 @@ func (s *Service) VerifyMagicLinkCode(code string, consume bool) (*Code, error)
 		return nil, err
 	}
 
-	if !secretcode.MagicLinkOTPSecretCode.Compare(codeModel.UserInputtedCode, codeModel.Code) {
+	userInputtedCode := codeModel.UserInputtedCode
```
louischan-oursky, line 169: @carmenlau VerifyMagicLinkCode is currently overloaded with different use cases. Per our discussion, we should have one method for each use case.

Use case 1: Set the user inputted case in the approve login link flow.
Use case 2: Verify the user inputted case in the original sign in flow.

carmenlau:
continue in #2932

Issue #2896, opened 2023-02-08
Include .HasPassword flag in reset password email's context


louischan-oursky:
So the email can have different copywriting depending on whether the user has a password.

PR #2919, opened 2023-02-10
Add flag to reset password email's context

Rubikium:
ref #2896

### DIFF
--- a/pkg/lib/feature/forgotpassword/provider.go
+++ b/pkg/lib/feature/forgotpassword/provider.go
@@ -22,9 +22,10 @@ import (
 )
 
 type TemplateData struct {
-	Email string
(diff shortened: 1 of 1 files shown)


pkg/lib/feature/forgotpassword/provider.go
```diff
@@ -127,7 +128,7 @@ func (p *Provider) SendCode(loginID string) error {
 		}
 
 		p.Logger.Debugf("sending email")
-		if err := p.sendEmail(email, codeStr); err != nil {
+		if err := p.sendEmail(email, codeStr, info.UserID); err != nil {
```
louischan-oursky, line 131: Do this for SMS message as well.

Issue #2897, opened 2023-02-08
Add mutation sendResetPasswordMessage to Admin API


louischan-oursky:
- Take `loginID String!`

PR #2923, opened 2023-02-10
Add sendResetPasswordMessage mutation

Rubikium:
ref #2897

FIXME: need implementation for `ResetPasswordURL()` in [pkg/admin/web_endpoints.go](https://github.com/authgear/authgear-server/blob/ec3c3406b3fd157e4d2e49f0d65eed073c5d0075/pkg/admin/web_endpoints.go#L55)


louischan-oursky:
I added some commits to refactor endpoints. They now live in pkg/lib/endpoints, and are shared by main, admin, and resolver. @carmenlau

Issue #2920, opened 2023-02-10
[Custom UI] Add websocket to workflow


louischan-oursky:
- Change the endpoint organization

```txt
POST /api/workflow/v1/
GET /api/workflow/v1/{workflow-id}/instances/{instance-id}
POST /api/workflow/v1/{workflow-id}/instances/{instance-id}
101 /api/workflow/v1/{workflow-id}/ws
```

- Add websocket endpoint similar to the existing one

PR #2924, opened 2023-02-10
Add WebSocket endpoint for workflow

kiootic:
ref #2920

### DIFF
--- a/pkg/auth/handler/api/workflow_get.go
+++ b/pkg/auth/handler/api/workflow_get.go
@@ -12,11 +12,11 @@ import (
 func ConfigureWorkflowGetRoute(route httproute.Route) httproute.Route {
 	return route.
 		WithMethods("GET").
-		WithPathPattern("/api/workflow/v1/:instanceid")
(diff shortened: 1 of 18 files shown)


kiootic:
@louischan-oursky Adjusted the path a bit different from discussed:
- Moved version path segment to front for common API convention.
- Use plural for `workflow` for consistency.

pkg/auth/handler/api/workflow_get.go
```diff
@@ -12,7 +12,7 @@ import (
 func ConfigureWorkflowGetRoute(route httproute.Route) httproute.Route {
 	return route.
 		WithMethods("GET").
-		WithPathPattern("/api/workflow/v1/:instanceid")
+		WithPathPattern("/api/v1/workflows/:workflowid/instances/:instanceid")
```
louischan-oursky, line 15: We are ignoring the workflowid in the path. If it is an input now, then we should make ensure the instance id and workflow id pair is valid.

PR #2937, opened 2023-02-14
Fix CORS checking in workflow websocket endpoint

### LINKED ISSUE #2934: Workflow WebSocket endpoint does not respect AllowedOrigins config
It implements own origin checking (`OriginPatterns` in `AcceptOptions`), which has incompatible syntax with the origin matcher library we use.

louischan-oursky:
fix #2934


kiootic (commented):
If I checked correctly, the CORS middleware only omit the CORS headers if the check failed (and rely on browser to prevent the request); however, browser does not check CORS headers for Websocket.

louischan-oursky:
OK. Let me revise the approach.

louischan-oursky:
@kiootic Can you help test if this change fixes the issue? Thank you.

PR #2940, opened 2023-02-15
Allow user to select password authenticator even they don't have it

carmenlau:
refs #2820

### DIFF
--- a/pkg/latte/intent_login.go
+++ b/pkg/latte/intent_login.go
@@ -81,15 +81,9 @@ func (i *IntentLogin) ReactTo(ctx context.Context, deps *workflow.Dependencies, 
 					Authenticator: emailAuthenticator,
 				}), nil
 			case model.AuthenticatorTypePassword:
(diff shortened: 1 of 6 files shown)


pkg/latte/intent_login.go
```diff
@@ -85,7 +86,15 @@ func (i *IntentLogin) ReactTo(ctx context.Context, deps *workflow.Dependencies,
...
 					authenticator.KeepType(model.AuthenticatorTypePassword),
 				)
-				if err != nil {
+				if errors.Is(err, api.ErrNoAuthenticator) {
+					// allow the user to start the intent
+					// if the user doesn't have a password authenticator
+					// the node will always return invalid credentials error
+					// when receiving the password
+					return workflow.NewSubWorkflow(&IntentAuthenticatePassword{
+						Authenticator: nil,
```
louischan-oursky, line 95: The workflow is stuck for the following case:

1. Create the user without password
2. Start the sign-in workflow
3. The user resets (or we should set up) the password
4. The intent and the node has stale information

Issue #2936, opened 2023-02-14
Exceeding `failed_attempt` OTP rate limit would not produce rate limit error


kiootic:
When `failed_attempt` OTP rate limit is exceeded, it always return `InvalidCredentials` error, ad `RateLimited` error is never returned.

kiootic:
The problem is in `handleFailedAttempt` of `otp.Service`. When remaining tokens = 1:
1. `TakeToken` passed, remaining tokens = 0
2. `CheckToken` failed, delete OTP code and return `ErrInvalidCode`

Subsequent attempts would fail early due to missing OTP code and rate limit is not checked anymore. So OTP verification would never return `RateLimited` error due to failed attempt rate limit.

louischan-oursky:
- Rename CanVerifyCode to FailedAttemptRateLimitExceeded
- Call CheckToken in CanVerifyCode and VerifyCode in the beginning

PR #2942, opened 2023-02-15
Update OTP nodes output to include masked target and failed attempt limit exceeded 

carmenlau:
ref #2936 
ref #2820

### DIFF
--- a/pkg/auth/handler/webapp/enter_oob_otp.go
+++ b/pkg/auth/handler/webapp/enter_oob_otp.go
@@ -41,11 +41,11 @@ func ConfigureEnterOOBOTPRoute(route httproute.Route) httproute.Route {
 }
 
 type EnterOOBOTPViewModel struct {
-	OOBOTPTarget           string
-	OOBOTPCodeSendCooldown int
-	OOBOTPCodeLength       int
-	OOBOTPChannel          string
-	OOBOTPCanVerifyCode    bool
+	OOBOTPTarget                   string
+	OOBOTPCodeSendCooldown         int
+	OOBOTPCodeLength               int
+	OOBOTPChannel                  string
+	FailedAttemptRateLimitExceeded bool
 }
 
 type EnterOOBOTPHandler struct {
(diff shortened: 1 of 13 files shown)


kiootic:
If developer set rate limit to N failed attempts, the `RateLimited` error is only returned **after** N attempts, i.e. at N+1 attempt. Naively interpreting the error would give user a meaningless (i.e. always fail) attempt before realizing the rate limit.

We want to check `CheckToken` again after `TakeToken` for failed attempts, and return `RateLimited` error in the last available failed attempt.

carmenlau:
Should be something like this? 
I am checking if this change affects the error handling of the existing auth ui

```diff
git diff 
diff --git a/pkg/lib/authn/otp/service.go b/pkg/lib/authn/otp/service.go
index fda981912..c98f3b570 100644
--- a/pkg/lib/authn/otp/service.go
+++ b/pkg/lib/authn/otp/service.go
@@ -101,17 +101,19 @@ func (s *Service) deleteCode(target string) {
 }
 
 func (s *Service) handleFailedAttempt(target string) error {
-       err := s.RateLimiter.TakeToken(s.TrackFailedAttemptBucket(target))
+       bucket := s.TrackFailedAttemptBucket(target)
+       err := s.RateLimiter.TakeToken(bucket)
        if err != nil {
                return err
        }
 
-       pass, _, err := s.RateLimiter.CheckToken(s.TrackFailedAttemptBucket(target))
+       pass, _, err := s.RateLimiter.CheckToken(bucket)
        if err != nil {
                return err
        } else if !pass {
                // Maximum number of failed attempt exceeded
                s.deleteCode(target)
+               return bucket.BucketError()
        }
        return ErrInvalidCode
 }
```

kiootic:
Yes, that's what I would expect.

carmenlau:
After applying the change, the workflow will return the rate limit error in the last attempt. 

I expected I will see the same thing on the auth ui, but I saw the invalid credentials error instead.

It is because of this:
https://github.com/authgear/authgear-server/blob/master/pkg/lib/interaction/nodes/authn_end.go#L82

It is a bit too tricky. Maybe we should open another issue to address the problem, the auth UI should show the real error.

kiootic:
Tested OK!

Issue #2930, opened 2023-02-14
Should fire UserAuthenticatedEvent when SSO enabled is false


carmenlau:
**Describe the bug**
Currently, UserAuthenticatedEvent is fired only when there is idp session. So when the idp session is suppressed (sso disabled), the UserAuthenticatedEvent is not fired.

**Authgear Version**
 - Version: [staging-2023-02-10.0](https://github.com/authgear/authgear-server/releases/tag/staging-2023-02-10.0)

**Additional context**
If we update to fire the event in this case, the session field in the payload will be empty. Not sure if it is expected.

---

The following table summarizes the intended behavior.

|response_type|IsSSOEnabled|SuppressIDPSessionCookie|Continue / Authenticate|Create IDPSession|Create OfflineGrant|Fire in interaction / workflow |Fire in code exchange|
|---|---|---|---|---|---|---|--|
|code|Yes|No|Authenticate|Yes|Yes|Yes|No|
|code|Yes|No|Continue|No|Yes|No|No|
|code|No|Yes|Authenticate|No|Yes|No|Yes|
|none|Yes|No|Authenticate|Yes|No|Yes|No|

Based on the table, we can derive the condition on firing user.authenticated.

- When an IDP session was created, fire in interaction.
- When IsSSOEnabled is true AND an offline grant was issued, fire in code exchange.

PR #2945, opened 2023-02-15
Fire UserAuthenticatedEvent during code exchange if sso disabled

carmenlau:
ref #2930

### DIFF
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -658,7 +658,8 @@ func (h *TokenHandler) handleBiometricAuthenticate(
 	}
 	resp.IDToken(idToken)
 
-	// Dispatch event user.authenticated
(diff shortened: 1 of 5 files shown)


pkg/lib/oauth/handler/handler_token.go
```diff
@@ -833,6 +834,23 @@ func (h *TokenHandler) issueTokensForAuthorizationCode(
...
 		accessTokenSessionKind = oauth.GrantSessionKindOffline
+
+		// ref: https://github.com/authgear/authgear-server/issues/2930
+		if !opts.SSOEnabled {
+			userRef := model.UserRef{
+				Meta: model.Meta{
+					ID: authz.UserID,
+				},
+			}
+			err = h.Events.DispatchEvent(&nonblocking.UserAuthenticatedEventPayload{
```
carmenlau, line 845: Just realize we need to skip the new sign up user here.

PR #2946, opened 2023-02-16
Specify intent migrate login

carmenlau:


docs/specs/workflow.md
```diff
@@ -318,3 +318,99 @@ And then the frontend will receive a response to telling it to redirect.
...
+The Intent for login id migration allows the developer to create a workflow for migration.
+
+[![](https://mermaid.ink/img/pako:eNqdUl1rwkAQ_CvLPdv2PQ9Cq4JS_EArfWhK2SYbc2ju0ru9ioj_vZvGaCVgSwMHuZ2ZnZu926vEpqQi5ekjkEmor3HlsIgNyFeiY53oEg3DfeB8RejaSC94tsVy1EaG1q4BPfTJWLiDZ3rPpVLzGtVNt9u0jqDnCJlAGybDYy0nkd3WunW2sdta15BF17SIYE4cnDkxYdTvSBPPKIlkc8Vx5mxClALnZyPYas7hafo4mLQ8q0iiQu9rArCtY35qhHO6qnRi96wcxYWE4eVVp5JM8-72mG5RUgJoUoFQXCowQbbuAs-cLWA2nE4Gb5Pl-GEwvzA5hzmO4X8urah_GVJV8SL2v1_Otfv4SfahoAunil0t1VEFuQJ1Ku91_11VwisoVpH8ppRh2HCsYnMQquS0i51JVCSjp44KZSpBj89bRRluPB2-AAamCc8?type=png)](https://mermaid.live/edit#pako:eNqdUl1rwkAQ_CvLPdv2PQ9Cq4JS_EArfWhK2SYbc2ju0ru9ioj_vZvGaCVgSwMHuZ2ZnZu926vEpqQi5ekjkEmor3HlsIgNyFeiY53oEg3DfeB8RejaSC94tsVy1EaG1q4BPfTJWLiDZ3rPpVLzGtVNt9u0jqDnCJlAGybDYy0nkd3WunW2sdta15BF17SIYE4cnDkxYdTvSBPPKIlkc8Vx5mxClALnZyPYas7hafo4mLQ8q0iiQu9rArCtY35qhHO6qnRi96wcxYWE4eVVp5JM8-72mG5RUgJoUoFQXCowQbbuAs-cLWA2nE4Gb5Pl-GEwvzA5hzmO4X8urah_GVJV8SL2v1_Otfv4SfahoAunil0t1VEFuQJ1Ku91_11VwisoVpH8ppRh2HCsYnMQquS0i51JVCSjp44KZSpBj89bRRluPB2-AAamCc8)
+
+#### Configuration in `authgear.yaml`
+
+```yaml
+migration_workflow:
+  url: authgeardeno:///deno/migration.ts
+  timeout: 10
```
louischan-oursky, line 333: I tend to not couple account migration with workflow.

```
account_migration:
  hook:
    url: xxx
    timeout: 10
```

docs/specs/workflow.md
```diff
@@ -318,3 +318,99 @@ And then the frontend will receive a response to telling it to redirect.
...
+```
+
+The developer can pass the request body to the workflow. Authgear will send the request body to the hook, and the hook should return the identities' and authenticators' MigrateSpec.
```
louischan-oursky, line 336: For simplicity, I think we just need to support `migration_token`, which is always a string.

docs/specs/workflow.md
```diff
@@ -318,3 +318,99 @@ And then the frontend will receive a response to telling it to redirect.
...
+	Type  model.LoginIDKeyType `json:"type"`
+	Value string               `json:"value"`
+}
+```
+
+### The authenticator spec
+```go
+type Spec struct {
+	Type      model.AuthenticatorType `json:"type,omitempty"`
+	IsDefault bool                    `json:"is_default,omitempty"`
```
louischan-oursky, line 406: I think the developer has no idea when do they set this to false. I guess we should omit it.

docs/specs/workflow.md
```diff
@@ -318,3 +318,99 @@ And then the frontend will receive a response to telling it to redirect.
...
+	Kind      Kind                    `json:"kind,omitempty"`
```
louischan-oursky, line 407: If we expose Kind, does that mean we support migrating 2FA OOB OTP authenticator? I tend not to support that use case for now. It is OK to keep Kind, but the only valid value is `primary`.

docs/specs/workflow.md
```diff
@@ -318,3 +318,99 @@ And then the frontend will receive a response to telling it to redirect.
...
+
+### Intent for login id migration (Experimental)
+
+The Intent for login id migration allows the developer to create a workflow for migration.
+
+[![](https://mermaid.ink/img/pako:eNqdUl1rwkAQ_CvLPdv2PQ9Cq4JS_EArfWhK2SYbc2ju0ru9ioj_vZvGaCVgSwMHuZ2ZnZu926vEpqQi5ekjkEmor3HlsIgNyFeiY53oEg3DfeB8RejaSC94tsVy1EaG1q4BPfTJWLiDZ3rPpVLzGtVNt9u0jqDnCJlAGybDYy0nkd3WunW2sdta15BF17SIYE4cnDkxYdTvSBPPKIlkc8Vx5mxClALnZyPYas7hafo4mLQ8q0iiQu9rArCtY35qhHO6qnRi96wcxYWE4eVVp5JM8-72mG5RUgJoUoFQXCowQbbuAs-cLWA2nE4Gb5Pl-GEwvzA5hzmO4X8urah_GVJV8SL2v1_Otfv4SfahoAunil0t1VEFuQJ1Ku91_11VwisoVpH8ppRh2HCsYnMQquS0i51JVCSjp44KZSpBj89bRRluPB2-AAamCc8?type=png)](https://mermaid.live/edit#pako:eNqdUl1rwkAQ_CvLPdv2PQ9Cq4JS_EArfWhK2SYbc2ju0ru9ioj_vZvGaCVgSwMHuZ2ZnZu926vEpqQi5ekjkEmor3HlsIgNyFeiY53oEg3DfeB8RejaSC94tsVy1EaG1q4BPfTJWLiDZ3rPpVLzGtVNt9u0jqDnCJlAGybDYy0nkd3WunW2sdta15BF17SIYE4cnDkxYdTvSBPPKIlkc8Vx5mxClALnZyPYas7hafo4mLQ8q0iiQu9rArCtY35qhHO6qnRi96wcxYWE4eVVp5JM8-72mG5RUgJoUoFQXCowQbbuAs-cLWA2nE4Gb5Pl-GEwvzA5hzmO4X8urah_GVJV8SL2v1_Otfv4SfahoAunil0t1VEFuQJ1Ku91_11VwisoVpH8ppRh2HCsYnMQquS0i51JVCSjp44KZSpBj89bRRluPB2-AAamCc8)
```
louischan-oursky, line 326: A typo in the diagram. s/PHONE_NUMBER/TOKEN

docs/specs/workflow.md
carmenlau: How about calling it `login_id_migration` instead of `account_migration`? Not sure if we will have another type of migration that is related to accounts as well.

And the intent is named `IntentMigrateLoginID`

docs/specs/workflow.md
carmenlau: Looks good to me. And for the hook request, we will put the migration_token as the request body without further serialization. Thought?

docs/specs/workflow.md
carmenlau: Removing it from now and only `primary` is supported for this stage

docs/specs/workflow.md
carmenlau: Also renamed the intent to `IntentMigrateAccount`

docs/specs/workflow.md
louischan-oursky: You mean `{ "migration_token": "TOKEN" }`?

PR #2960, opened 2023-02-17
Implement reset password workflow

tung2744:
base on #2959 
ref #2892


pkg/lib/workflow/dependencies.go
```diff
@@ -64,6 +64,11 @@ type ForgotPasswordService interface {
 	SendCode(loginID string) error
 }
 
+type ResetPasswordService interface {
+	ResetPassword(userID string, newPassword string) (err error)
```
louischan-oursky, line 68: Remove this method. It is unused.

Issue #2895, opened 2023-02-08
[Custom UI] Change password workflow


louischan-oursky:
requires https://github.com/authgear/authgear-server/issues/2899

PR #2961, opened 2023-02-17
Add change password workflow

Rubikium:
ref #2895


pkg/latte/intent_change_password.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (*IntentChangePassword) JSONSchema() *validation.SimpleSchema {
+	return IntentChangePasswordSchema
+}
+
+func (*IntentChangePassword) CanReactTo(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow) ([]workflow.Input, error) {
+	if len(w.Nodes) == 0 {
+		return []workflow.Input{
+			&InputTakeLoginID{},
```
carmenlau: Changing the password is a feature that requires an authenticated user, `IntentChangePassword` should not take the login id to search the user.

Instead, the user id should be resolved from the user session which can be obtained through the context. e.g. https://github.com/authgear/authgear-server/blob/c81d8eda8da7ceba70ccc853f25a3c692957abe3/pkg/latte/intent_verify_user.go#L44

pkg/latte/intent_change_password.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/authgear/authgear-server/pkg/lib/authn/identity"
+	"github.com/authgear/authgear-server/pkg/lib/workflow"
+	"github.com/authgear/authgear-server/pkg/util/validation"
+)
+
+func init() {
+	workflow.RegisterPublicIntent(&IntentChangePassword{})
+}
+
+var IntentChangePasswordSchema = validation.NewSimpleSchema(`{}`)
```
carmenlau: Public intent schema cannot be empty, see `IntentVerifyUser`.

pkg/latte/input_take_change_password.go
```diff
@@ -0,0 +1,50 @@
...
+		"additionalProperties": false,
+		"properties": {
+			"old_password": { "type": "string" },
+			"new_password": { "type": "string" }
+		},
+		"required": ["old_password", "new_password"]
+	}
+`)
+
+type InputTakeChangePassword struct {
```
carmenlau, line 24: Maybe we can call the input `InputChangePassword`? `InputTakeChangePassword` has 2 verbs.

pkg/latte/node_change_password.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (n *NodeChangePassword) ReactTo(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow, input workflow.Input) (*workflow.Node, error) {
+	var inputTakeChangePassword inputTakeChangePassword
+	switch {
+	case workflow.AsInput(input, &inputTakeChangePassword):
+		info, err := n.getPasswordAuthenticator(deps)
+		// The user doesn't have the password authenticator
+		// always returns invalid credentials error
+		if errors.Is(err, api.ErrNoAuthenticator) {
+			return nil, api.ErrInvalidCredentials
```
louischan-oursky, line 44: 1. Move `ChangePasswordFailed` and `ErrNoPassword` in pkg/lib/interaction/nodes/change_password.go to pkg/api/errors.go
2. Return `ErrNoPassword` when the user has no password

pkg/latte/node_change_password_end.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/authgear/authgear-server/pkg/lib/authn/authenticator"
+	"github.com/authgear/authgear-server/pkg/lib/workflow"
+)
+
+func init() {
+	workflow.RegisterNode(&NodeChangePasswordEnd{})
+}
+
+type NodeChangePasswordEnd struct {
+	Authenticator *authenticator.Info
```
louischan-oursky, line 15: Remove this field.

PR #2972, opened 2023-02-23
Add change email workflow

Rubikium:
ref #2935

### DIFF
--- a/pkg/latte/intent_change_email.go
+++ b/pkg/latte/intent_change_email.go
@@ -2,6 +2,7 @@ package latte
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/authgear/authgear-server/pkg/api"
 	"github.com/authgear/authgear-server/pkg/api/apierrors"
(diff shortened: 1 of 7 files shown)


pkg/latte/intent_change_email.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+var IntentChangeEmailSchema = validation.NewSimpleSchema(`
+	{
+		"type": "object",
+		"additionalProperties": false
+	}
+`)
+
+type IntentChangeEmail struct {
```
louischan-oursky, line 26: This intent merely updates the identity, but it misses a lot of details. Please take a look at pkg/lib/interaction/intents/update_identity.go and pkg/latte/intent_create_login_id.go.

We have a few things to do

1. Call UpdateWithSpec
2. Call CheckDuplicate
3. Call Update
4. Populate standard attributes
5. Verify the identity
6. Create the OOB OTP authenticator
7. Fire the event produced by NewIdentityLoginIDUpdatedEventPayload

intent_create_login_id.go has done 4, 5, 6
update_identity.go has done 1, 2, 3, 7

Your workflow should do all of them

louischan-oursky: Great! Thank you.

PR #2974, opened 2023-02-28
Support ui_locales and state in workflow

tung2744:


louischan-oursky (commented):
Looks good to me so far

tung2744:
@louischan-oursky I found that we may need to enable users to encode `state` and `ui_locales` in the email template, because they are arbitrary data that given by the user so is possible to contain characters that are invalid as query parameter. Therefore I added `EncodableData` in the latest commit. See if any concern?

The expected usage is
```
{{ .State.URLQueryEncoded }} // encoded
{{ .State }} // raw
```

louischan-oursky:
html/template should be smart enough to handle that. See https://go.dev/play/p/VOC-jIDDuV-

tung2744:
@louischan-oursky However we also have text templates? https://go.dev/play/p/DT5MPXBq2IG

louischan-oursky:
I suggest we add a func `queryEscape` which is https://pkg.go.dev/net/url#QueryEscape so the writer can write `{{ .State | queryEscape }}`

tung2744:
@louischan-oursky If using pipes, there is a provided `urlquery` pipe can be used. But I saw some effort was made to disallow pipes in templates.
https://github.com/authgear/authgear-server/blob/2023C01/pkg/util/template/validation.go#L109

louischan-oursky:
where does `urlquery` come from?

tung2744:
@louischan-oursky https://pkg.go.dev/text/template#hdr-Functions Should be a predefined function

louischan-oursky:
@kiootic Do you remember why we disallowed pipe back in that time?

kiootic:
Pipes could be used to consume lots of memory/CPU and DoS the server.

louischan-oursky:
Since we have context on where state and ui_locales will appear. I suggest we do something like https://go.dev/play/p/y-zrEPaCjuR

tung2744:
🤔 If we assume the state must be put back into query of some url, then this can be a solution.

tung2744:

PR #2978, opened 2023-03-02
Expose redirect_uri via postMessage when redirecting

Rubikium:
ref https://github.com/authgear/2023C01-latte-customization/issues/65

### DIFF
--- a/pkg/lib/oauth/response_mode.go
+++ b/pkg/lib/oauth/response_mode.go
@@ -21,7 +21,7 @@ const htmlRedirectTemplateString = `<!DOCTYPE html>
 {{- else }}
 <script>
 {{- end }}
-window.parent.postMessage("{{ .redirect_uri }}", "{{ .custom_ui_origin }}")
+window.parent.postMessage({ redirect_uri: "{{ .redirect_uri }}" }, "{{ .custom_ui_origin }}")
 window.location.href = "{{ .redirect_uri }}"
 </script>
 </body>
--- a/pkg/lib/oauth/response_mode.go
+++ b/pkg/lib/oauth/response_mode.go
@@ -21,7 +21,7 @@ const htmlRedirectTemplateString = `<!DOCTYPE html>
 {{- else }}
 <script>
 {{- end }}
-window.parent.postMessage("{{ .redirect_uri }}", "{{ .custom_ui_uri }}")
+window.parent.postMessage("{{ .redirect_uri }}", "{{ .custom_ui_origin }}")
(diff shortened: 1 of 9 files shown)


louischan-oursky:
According to https://github.com/authgear/2023C01-latte-customization/issues/65#issuecomment-1450984869 we can just update the template to post message.

kiootic (commented):
The custom UI could be at a sub-path of a domain, e.g. `https://example.com/custom-ui/auth`. In that case, developer would set `x_custom_ui_uri: https://example.com/custom-ui/auth`, but the `targetOrigin` of `postMessage` should be `https://example.com`.

Change `CustomUIURI` in the authorization result structs into `CustomUIOrigin` for use of template.

pkg/lib/oauth/response_mode.go
```diff
@@ -21,6 +21,7 @@ const htmlRedirectTemplateString = `<!DOCTYPE html>
 {{- else }}
 <script>
 {{- end }}
+window.parent.postMessage("{{ .redirect_uri }}", "{{ .custom_ui_uri }}")
```
kiootic, line 24: Also, I'd prefer not posting a raw string as a message, but an object for better flexibility. e.g. `{ redirect_uri: "..." }`

pkg/util/urlutil/origin.go
```diff
@@ -0,0 +1,13 @@
...
+	"net/url"
+)
+
+func ExtractOrigin(u *url.URL) *url.URL {
+	newU := *u
+	newU.Path = ""
+	newU.RawQuery = ""
+	newU.Fragment = ""
+	return &newU
+}
```
louischan-oursky, line 13: According to https://pkg.go.dev/net/url#URL and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin, we only need to copy scheme, opaque and host.

```golang
func ExtractOrigin(u *url.URL) *url.URL {
  return &url.URL{
    Scheme: u.Scheme,
    Opaque: u.Opaque,
    Host: u.Host,
  }
}
```

Write a test for this function.

louischan-oursky:
Try run `make check-tidy` and commit changes.

louischan-oursky:
@Rubikium Have you tested if the message can be delivered to the parent if we have the refresh meta tag with a delay of 0?

Rubikium:
> @Rubikium Have you tested if the message can be delivered to the parent if we have the refresh meta tag with a delay of 0?

Yes, and the message can be delivered with a delay of 0

PR #2979, opened 2023-03-03
Add copy user script

louischan-oursky:


scripts/copyuser/main.go
```diff
@@ -0,0 +1,1504 @@
...
+
+	err = f(tx)
+	return
+}
+
+func (c *Copier) ListUsers() ([]string, error) {
+	var userIDs []string
+
+	err := c.WithTx(func(tx *sql.Tx) error {
+		q := c.SQLBuilder.Select("id").From("_auth_user").Where("app_id = ?", c.FromAppID)
```
carmenlau, line 119: Should be `c.SQLBuilder.TableName("_auth_user")` for all tables

scripts/copyuser/main.go
```diff
@@ -0,0 +1,1504 @@
...
+		is_disabled,
+		disable_reason,
+		standard_attributes,
+		custom_attributes,
+		is_deactivated,
+		delete_at,
+		is_anonymized,
+		anonymize_at,
+		anonymized_at,
+	).Suffix(onConflictDoNothing)
```
carmenlau, line 270: To allow re-running of the script, we should skip copying all the other table entries for that user if the user exists. 

For the scenario,

1. Run the script to copy users from app A to app B
2. User in app A delete and add an identity
3. Run the script again
4. The user will have 2 identities

To keep it simple, maybe we can support only copying the newly added user and skip copying all the other table entries?

louischan-oursky:
Turned out this script was not used. I am closing this.

PR #2987, opened 2023-03-07
Support per IP rate limit to send any SMS

carmenlau:
ref #2986 

Remarks:
- Doesn't support updating via the portal yet
- Question: Currently in the workflow API. When a phone number is submitted and the rate limit is reached (including per phone, per IP, and resend cooldown), we skip sending and no error is returned. Should we show the error to the user if it is per phone or per IP limit?


louischan-oursky:
> Should we show the error to the user if it is per phone or per IP limit?

The error should only be skipped if the rate limit is due to cooldown.

louischan-oursky (commented):
> Should we show the error to the user if it is per phone or per IP limit?

The error should only be skipped if the rate limit is due to cooldown.

Issue #2922, opened 2023-02-10
App list page on portal is slow


louischan-oursky:
The graphql query takes more than 30 seconds to return. The reason is that it loads everything of every app.

PR #2995, opened 2023-03-16
Return app list

louischan-oursky:
ref #2922


carmenlau (approved):
The PR looks good, please fix the error

PR #3000, opened 2023-03-17
Specify `x_application_type=confidential`

louischan-oursky:
ref #2999

### DIFF
--- a/docs/specs/oidc.md
+++ b/docs/specs/oidc.md
@@ -16,6 +16,7 @@
     + [login_hint](#login_hint)
     + [acr_values](#acr_values)
     + [code_challenge_method](#code_challenge_method)
+    + [nonce](#nonce)
   * [Token Request](#token-request)
     + [grant_type](#grant_type)
     + [id_token_hint](#id_token_hint-1)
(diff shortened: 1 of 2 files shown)


docs/specs/oidc.md
```diff
@@ -461,44 +470,114 @@ Silent Authentication
...
 
-OAuth protocol allows user to delegate access to clients, so that clients can
-act on behalf of the user.
+The tables summarizes client Party, confidentiality, PII in ID token, and access to privileged user operations of different `x_application_type`.
 
-Usually, OAuth clients are third-party clients, with limited trust:
-- User need to give consent to a client before clients is allowed access.
-- Privileged user operations (e.g. change password) is usually not exposed
-  through APIs accessible through OAuth protocol.
+|`x_application_type`|Client Party|Confidentiality|PII in ID token|Access to privileged user operations|
```
kiootic, line 479: Can we add reason behind the difference of 'PII in ID token' and 'Access to privileged user operations' between different kinds of client?

docs/specs/oidc.md
```diff
@@ -461,44 +470,114 @@ Silent Authentication
...
+|---|---|---|---|---|
+|`spa`|First-party|public|No|Yes|
+|`traditional_webapp`|First-party|public|No|Yes|
+|`native`|First-party|public|No|Yes|
+|`confidential`|First-party|confidential|Yes|No|
+|`third_party_app`|Third-party|confidential|Yes|No|
```
kiootic, line 485: I think `third_party_app` is supposed to be legacy, so we may mark it as such.

docs/specs/oidc.md
louischan-oursky: Actually they are inverse of each other. I listed them separately in the table with the intention to make things clearer. Maybe I should add a paragraph to explain their logical relationship?

docs/specs/oidc.md
louischan-oursky: In the future we may bring it back. So I did not mark it as legacy. In the implementation, I will "hide" it.

docs/specs/oidc.md
kiootic: I think we would need to explain the rationales of the limitations:
- why confidential clients cannot use re-authentication?
- why only public clients have access to privileged user operations?

louischan-oursky: Added a section to explain.

PR #3004, opened 2023-03-20
Add x_state to ui params

tung2744:


pkg/auth/handler/api/workflow_new.go
```diff
@@ -145,6 +146,7 @@ func (h *WorkflowNewHandler) makeSessionOptions(cookie *http.Cookie) (*workflow.
 		RedirectURI:              uiInfo.RedirectURI,
 		SuppressIDPSessionCookie: uiInfo.SuppressIDPSessionCookie,
 		State:                    uiInfo.State,
+		XState:                   req.XState(),
```
louischan-oursky, line 149: Add XState to UIInfo and read uiInfo.XState to set this field, just like what we did for State.

louischan-oursky (commented):
We have something called WebhookState in interaction. WebhookState is ultimately in the webhook event context. Therefore we should also expose x_state to hooks.

This is a bit tricky because the handling of WebhookState has issues.

- WebhookState only works for some interaction intents (via IntentWithWebhookState interface). Eliminate IntentWithWebhookState, and read from UIParam.
- Remove WebhookState from webapp session; read from UIParam.

Can you double check if my proposed approach works?

tung2744:
@louischan-oursky For the webhook state, we should now expose both state and x_state in `OAuthContext`, am I correct?

louischan-oursky:
Yes.

louischan-oursky:
Build failed.

Issue #2926, opened 2023-02-13
[OTP, Portal] Revoke OTP after failed attempt should be configurable in Portal


fungc-io:
In authgear-server#2780 OTP is revoked after 5 failed attempt.
This number is configurable in yaml but not exposed in Portal.

Please add it in this issue

## Design

[pending]

fungc-io:
@Rubikium description updated.

Rubikium:
Currently the default config is revoke after 5 failed attempts, and unlimited retries is not supported. @louischan-oursky Should we add an enabled flag in the config to support unlimited retries? 
```yaml
otp:
  ratelimit:
    failed_attempt:
      enabled: true
      size: 5
      reset_period: 20m
```

louischan-oursky:
OK. By default, this is disabled.

PR #3007, opened 2023-03-20
Update otp failed attempt config 

Rubikium:
ref #2926


portal/src/graphql/portal/LoginMethodConfigurationScreen.tsx
```diff
@@ -2293,6 +2342,29 @@ function VerificationSettings(props: VerificationSettingsProps) {
...
+          )}
+          offText={renderToString(
+            "VerificationConfigurationScreen.otp-failed-attempt.enabled.offText"
+          )}
+          onText={renderToString("Toggle.on")}
+          checked={otpFailedAttemptConfig.enabled ?? false}
+          onChange={onChangeOTPFailedAttemptEnabled}
+        />
+        {otpFailedAttemptConfig.enabled ? (
+          <TextField
```
louischan-oursky, line 2358: The validation error (min or max validation) is shown in the top bar instead of inline. Please use FormTextField and specify a correct JSON pointer so that the error is shown inline.

Issue #3010, opened 2023-03-21
Consolidate and review rate limits


kiootic:
- Upload profile image: `presign-images-upload`
  - hard-coded: 10 per hour, per user
- Password authentication: `auth-secret`
  - configurable default: 10/min/user/authentication type
- Other authentication (TOTP, passkey, OOB-OTP, etc.): `auth-secret`
  - hard-coded: 10 per minute, per user, per authentication type
- MFA recovery code: `auth-recovery-code`
  - hard-coded: 10 per minute, per user
- MFA device token: `auth-device-token`
  - hard-coded: 10 per minute, per user
- General OTP Failed Attempt (verification OTP, OOB-OTP): `otp-failed-attempt`
  - Invalidate existing OTP if rate limit is exceeded
  - configurable default: 5 per 20 minutes, per email/phone; disabled by default
- Forgot password send message: `reset-password-send-code`
  - hard-coded: 5 per 5 minutes, per login ID
- Forgot password validate code: `reset-password-verify-ip`
  - hard-coded: 10 per minute, per IP
- SWIE authentication: `siwe-nonce`
  - hard-coded: 10 per minute, per IP
- Verification validate code: `verification-verify-code`
  - hard-coded: 10 per minute, per IP
- Send email: `email-message`
  - hard-coded: 10 per minute, per email
- Send SMS (phone): `sms-message`
  - configurable default: 10 per 24 hours, per phone; disabled by default
- Send SMS (IP): `sms-message-per-ip`
  - configurable default: 120 per minute, per IP; disabled by default
- Hard SMS limit: `sms-message-hard`
  - feature config
- Signup user: `signup`
  - hard-coded: 10 per minute, per IP
- Legacy interaction request: `request`
  - hard-coded: 200 per minute, per IP
- Legacy interaction signup anonymous user: `signup-anonymous-user`
  - checked if new user is anonymous, after checking `signup` rate limit
  - hard-coded: 60 per hour, per IP
- Legacy interaction account enumeration: `account-enumeration`
  - checked before verifying input login ID exists
  - hard-coded: 10 per minute, per IP
- Send OTP code cooldown: `otp-code`
  - configurable default: 1 per 60 seconds, per email/phone

kiootic:
Suggestions:
- Consolidate authentication rate limits into one (`auth-secret`, `auth-recovery-code`, `auth-device-token`)
- Exclude OTP from authenticate rate limits to avoid duplication.
[...]

PR #3012, opened 2023-03-22
Add new rate limits spec

kiootic:
ref #3010

### DIFF
--- a/docs/specs/rate-limit.md
+++ b/docs/specs/rate-limit.md
@@ -80,7 +80,7 @@ Rate limits without default are hard-coded (non-configurable).
 
 ## Configuration
 
-In general, rate limits are configured using two fields:
+In general, rate limits are configured using 3 fields:
 ```yaml
 verification:
     rate_limits:
--- a/docs/specs/rate-limit.md
+++ b/docs/specs/rate-limit.md
@@ -34,32 +34,43 @@ Some considerations for rate limits design:
 
 Rate limits without default are hard-coded (non-configurable).
 
-| Operation | global | per-tenant | per-IP | per-target | per-user-per-IP | per-user | Rationales |
-| ----------|--------|------------|--------|------------|-----------------|----------|------------|
-| **Authentication**
-| Verify password authenticator | | | default: 60/minute | | default: 10/minute | | Mitigate credential brute-forcing. Per-user rate limit is not used to avoid DoS on actual user login.
-| Verify TOTP authenticator | | | default: 60/minute | | default: 10/minute
-| Verify OOB-OTP authenticator (email/SMS) | | | default: 60/minute | | default: 10/minute
-| Verify MFA recovery code | | | default: 60/minute | | default: 10/minute
-| Verify MFA device token | | | default: 60/minute | | default: 10/minute
-| Verify Passkey authenticator | | | default: 60/minute | | | | Since Authgear uses discoverable credentials, user is derived from passkey and per-user-per-IP rate limit is N/A.
-| SWIE nonce request | | | default: 60/minute | | | | Mitigate credential brute-force. User is not known at this point for Web3 login so only per-IP rate limit is used.
-| Signup new user | | | default: 10/minute | | | | Mitigate resource exhaustion by rapid registration of new user.
-| Signup new anonymous user | | | default: 60/minute | | | | A more generous limit is given for anonymous user signup, since it usually occurs on app startup of new installation.
-| Account enumeration (check login ID existence) | | | default: 10/minute | | | | Mitigate account enumeration.
-| **Features** 
-| Send OOB-OTP message | | | default: 10/hour; disabled | default: 1 minute cooldown | | | Mitigate mass/targeted message spam.
-| Send forgot password message | | | default: 10/hour; disabled | default: 1 minute cooldown
-| Send verification message | | | default: 10/hour; disabled | default: 1 minute cooldown | | default: 10/hour; disabled | Per-IP rate limit is used for verification during signup, per-user rate limit is used otherwise.
-| OOB-OTP failed verify attempts | | | | default: 5 attempts; disabled | | | Mitigate credential brute-force; revoke OTP when set limit exceeded.
-| Verification OTP failed verify attempts | | | | default: 5 attempts; disabled
-| Verify verification OTP | | | default: 60/minute | | | | Mitigate credential brute-force.
-| Verify forgot password OTP | | | default: 60/minute | | |
-| **Messaging**
-| Send SMS | environment variables | feature config | default: 60/minute; disabled | default: 50/day; disabled | | | If tenant would like rate limits for all SMS/email, contact server operator to adjust feature config.
-| Send Email | environment variables | feature config | default: 60/minute; disabled | default: 50/day; disabled | |
-| **Misc.** 
-| Presign upload image request | | | | | fixed: 10/hour | | Configuration not needed for now.
+
+| Operation                               | per-IP    | per-target        | per-user-per-IP | per-user | Rationales                                                                                                            |
+| --------------------------------------- | --------- | ----------------- | --------------- | -------- | --------------------------------------------------------------------------------------------------------------------- |
+| **Authentication**                      |           |                   |                 |          |                                                                                                                       |
+| Verify any credentials                  | 60/minute |                   | 10/minute       |          | Mitigate credential brute-forcing. Per-user rate limit is not used to avoid DoS on actual user login.                 |
+| Verify Password / Additional PW         | NIL       |                   | NIL             |          |                                                                                                                       |
+| Verify Email OTP                        | NIL       |                   | NIL             |          |                                                                                                                       |
+| Verify SMS OTP                          | NIL       |                   | NIL             |          |                                                                                                                       |
+| Verify TOTP                             | NIL       |                   | NIL             |          |                                                                                                                       |
+| Verify MFA recovery code                | NIL       |                   | NIL             |          |                                                                                                                       |
+| Verify MFA device tokens                | NIL       |                   | NIL             |          |                                                                                                                       |
+| Verify Passkey                          | NIL       |                   |                 |          | Since Authgear uses discoverable credentials, user is derived from passkey and per-user-per-IP rate limit is N/A.     |
+| SWIE nonce request                      | NIL       |                   |                 |          | Mitigate credential brute-force. User is not known at this point for Web3 login so only per-IP rate limit is used.    |
+| Signup new user                         | 10/minute |                   |                 |          | Mitigate resource exhaustion by rapid registration of new user.                                                       |
+| Signup new anonymous user               | 60/minute |                   |                 |          | A more generous limit is given for anonymous user signup, since it usually occurs on app startup of new installation. |
+| Check login ID existence                | 10/minute |                   |                 |          | Mitigate account enumeration.                                                                                         |
+| **Features**                            |           |                   |                 |          |                                                                                                                       |
+| Send Login OTP by SMS                   | NIL       | 1 minute cooldown |                 | NIL      | Mitigate mass/targeted message spam.                                                                                  |
+| Send Login OTP by Email                 | NIL       | 1 minute cooldown |                 | NIL      |                                                                                                                       |
+| Send Forgot Password by SMS             | NIL       | 1 minute cooldown |                 |          |                                                                                                                       |
+| Send Forgot Password by Email           | NIL       | 1 minute cooldown |                 |          |                                                                                                                       |
+| Send Verification by SMS                | NIL       | 1 minute cooldown |                 | NIL      | Per-IP rate limit is used for verification during signup, per-user rate limit is used otherwise.                      |
+| Send Verification by Email              | NIL       | 1 minute cooldown |                 | NIL      |                                                                                                                       | 
+| Login OTP failed verify attempts        |           | 5 attempts        |                 |          | Mitigate credential brute-force; revoke OTP when set limit exceeded.                                                  |
+| Verification OTP failed verify attempts |           | 5 attempts        |                 |          |                                                                                                                       |
+| Verify Verification OTP                 | 60/minute |                   |                 |          | Mitigate credential brute-force.                                                                                      |
+| Verify Forgot Password OTP              | 60/minute |                   |                 |          |                                                                                                                       |
+| **Misc.**                               |           |                   |                 |          |                                                                                                                       |
+| Presign upload image request            |           |                   | fixed: 10/hour  |          | Configuration not needed for now.                                                                                     |
+
+| Operation               | per-IP     | per-target |                                                                                                                      |
+| ----------------------- | ---------- | ---------- | -------------------------------------------------------------------------------------------------------------------- |
+| **Messaging**           |            |            |                                                                                                                      |
+| Send SMS (Global)       | disabled   | 50/day     | Configured using environment variables                                                                               |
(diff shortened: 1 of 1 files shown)


chpapa:
BTW, another minor concern / reminder, some rate limits we might need to support... e.g. IP white-listing in future, hope the implementation would consider that as well (no need to have it now I guess, or just nice to have)

chpapa:
Finally, one thing I think if easy to implement is nice to have, but if it add complexity to the code we should ignore it is:

> Reset period: Duration to re-fill the bucket

Another common patten of rate limit would be, refill x tokens by y seconds, so instead of "refill 10 tokens" for a 10/hour rate limit, we can say refill 1 token per hour after it is exhausted.

But if it adds any complexity to the code, I would prefer ignore it for now.

louischan-oursky (approved):

carmenlau (approved):

kiootic:
> I guess we need to have different rate limit for Send OOB-OTP message between Email vs SMS. For SMS, 10/hour is good, for email default could be a little bit higher. (20/hour?) (Same for Send Verification message I think? Any other Email/SMS requests mixed together?)

Personally I'd prefer to keep SMS/email per-feature rate-limit same for simplicity. I would believe during normal use it's rare to reach 10/hour rate limit per-feature.

> Just to confirm since it wasn't clear to me on the table, is that Send SMS and Send Email have a per tenant config as well?

The per-tenant level rate limit for send SMS/email is configured in feature config, set by server operator. As mentioned in rationale, if tenant would want configure per-tenant rate limit (e.g. for budget), they could contact server operator for this.

> We probably will want some rate limits for other things like our admin API, OIDC endpoints etc, but I think that's out of scope for now.

> BTW, another minor concern / reminder, some rate limits we might need to support... e.g. IP white-listing in future, hope the implementation would consider that as well (no need to have it now I guess, or just nice to have)

Noted, let me add a future works section with this.

> Another common patten of rate limit would be, refill x tokens by y seconds, so instead of "refill 10 tokens" for a 10/hour rate limit, we can say refill 1 token per hour after it is exhausted.

This is the vanilla algorithm, and as mentioned in spec, we're using a variant of it (i.e. reset instead of refill periodically) for simplicity. We're storing the bucket state in Redis, and handling concurrent rate limit access would need extra effort.

chpapa:
> Personally I'd prefer to keep SMS/email per-feature rate-limit same for simplicity. I would believe during normal use it's rare to reach 10/hour rate limit per-feature.

However, objectively, most users/use cases have very different limits between SMS and emails (sending SMS is way more expensive than email) so the sending limits of email and SMS have to be separated. The verify one could be kept as one, though, but would need separate SMS/emails rate limits for sending OOB-OTP / login by SMS/signup by SMS; Or what about we change the "Send SMS per IP" rate limit to 10/hour by default? I think it would be more useful.

@fungc-io can you also check if this table fits the K11 one? See if we missed anything?

BTW, reading the table again, I'm a bit confused, what is the different between "Verify OOB-OTP (email/SMS)" at `authentication` vs the one under `features`? Is it the same? Or one is for Sign in using SMS/Email, another one is for 2FA? If yes, can we distinguish between between the two?

chpapa:
> Personally I'd prefer to keep SMS/email per-feature rate-limit same for simplicity. I would believe during normal use it's rare to reach 10/hour rate limit per-feature.

However, objectively, most users/use cases have very different limits between SMS and emails (sending SMS is way more expensive than email) so the sending limits of email and SMS have to be separated. The verify one could be kept as one, though, but would need separate SMS/emails rate limits for sending OOB-OTP / login by SMS/signup by SMS; Or what about we change the "Send SMS per IP" rate limit to 10/hour by default? I think it would be more useful.

@fungc-io can you also check if this table fits the latte one? See if we missed anything?

BTW, reading the table again, I'm a bit confused, what is the different between "Verify OOB-OTP (email/SMS)" at authentication vs the one under features? Is it the same? Or one is for Sign in using SMS/Email, another one is for 2FA? If yes, can we distinguish between between the two?

chpapa:
> This is the vanilla algorithm, and as mentioned in spec, we're using a variant of it (i.e. reset instead of refill periodically) for simplicity. We're storing the bucket state in Redis, and handling concurrent rate limit access would need extra effort.

Just wondering -- if it is the "vanilla algorithm" that should be easier to implement? Does the data structure we use in redis handle concurrent well already? (if it really takes more effort than it is ok... but just curious since usually redis handle these sort of things well and u mentioned it is the "vanilla algo")

kiootic:
> However, objectively, most users/use cases have very different limits between SMS and emails (sending SMS is way more expensive than email) so the sending limits of email and SMS have to be separated. The verify one could be kept as one, though, but would need separate SMS/emails rate limits for sending OOB-OTP / login by SMS/signup by SMS; Or what about we change the "Send SMS per IP" rate limit to 10/hour by default? I think it would be more useful.

For budget purpose, it would be inconvenient for developers to configure SMS/email limit for each features. Instead, we have a general 'send SMS'/'send email' rate limit at Messaging section of the table.

However, as we considered before, for public WiFi case, rate limiting send SMS per IP may be excessive. I tend to increase the per-feature send message rate limit to align with general send SMS/email rate limit.

> Just wondering -- if it is the "vanilla algorithm" that should be easier to implement? Does the data structure we use in redis handle concurrent well already? (if it really takes more effort than it is ok... but just curious since usually redis handle these sort of things well and u mentioned it is the "vanilla algo")

As far as I know, to implement vanilla token bucket algorithm, we either need to actively re-fill the token periodically (infeasible), or passively re-fill the token when rate limit is accessed. Passively refill the tokens (instead of reset) is more complicated to implement, since time elapsed since last refill would need to be taken into account.

louischan-oursky:
I agree with Ben on the fact that sending SMS is more expensive than sending emails. So I second the rate limit of SMS and emails should be separate.

kiootic:
- I think we agree that email & SMS rate limit should be separate due to cost difference. Do we need to have per-feature rate limits for each message channels? (verification, forgot password, OOB-OTP authenticator * email, SMS = 6 rate limit configs)

- For per-IP rate limits, there are contrasting goals of preventing abuse by single IP and avoiding affecting legitimate users using same IP from public WiFi. Any suggestion on this?

chpapa:
@kiootic @louischan-oursky I took some initiative to quickly rewrite the tables as follows. Please advice?

Minor rewrite principle:
* Separate into two tables for readability of columns
* Minor copywriting edits for readability
* Remove something I'm not sure about first, and put them as questions for Terry below
  
Question for @kiootic :
* Sorry, what is ";disabled" meant in your table? I removed them first, but not quite sure.
* What this meant: "Send verification message"? "Per-IP rate limit is used for verification during signup, per-user rate limit is used otherwise." Even for signup, per-user is still applicable, right? (Since even during sign-up, we know the "target"/user for the verification already)
* Could we merge MFA recovery code and tokens? Or some concern to keep them as two times?

Clarification above would be helpful, and please also help me check if I missed any configurations.

Maybe we can have a quick call to clarify the above tomorrow or Thu?

The table below follows these conventions:
- Only rate limits that need to be supported have a cell filled in with the default values or `NIL`
- `NIL` meant default is unlimited.
- `per-target` meant either per SMS or Email address, respectively.
- All configs are expected to be in X per Y seconds (TBC?)

| Operation                               | per-IP    | per-target        | per-user-per-IP | per-user | Rationales                                                                                                            |
| --------------------------------------- | --------- | ----------------- | --------------- | -------- | --------------------------------------------------------------------------------------------------------------------- |
| **Authentication**                      |           |                   |                 |          |                                                                                                                       |
| Verify any login attempts               | 60/minute |                   | 10/minute       |          | Mitigate credential brute-forcing. Per-user rate limit is not used to avoid DoS on actual user login.                 |
| Verify Password / Additional PW         | NIL       |                   | NIL             |          |                                                                                                                       |
| Verify Email OTP                        | NIL       |                   | NIL             |          |                                                                                                                       |
| Verify SMS OTP                          | NIL       |                   | NIL             |          |                                                                                                                       |
| Verify TOTP                             | NIL       |                   | NIL             |          |                                                                                                                       |
| Verify MFA recovery code / tokens       | NIL       |                   | NIL             |          |                                                                                                                       |
| Verify Passkey                          | NIL       |                   |                 |          | Since Authgear uses discoverable credentials, user is derived from passkey and per-user-per-IP rate limit is N/A.     |
| SWIE nonce request                      | NIL       |                   |                 |          | Mitigate credential brute-force. User is not known at this point for Web3 login so only per-IP rate limit is used.    |
| Signup new user                         | 10/minute |                   |                 |          | Mitigate resource exhaustion by rapid registration of new user.                                                       |
| Signup new anonymous user               | 60/minute |                   |                 |          | A more generous limit is given for anonymous user signup, since it usually occurs on app startup of new installation. |
| Check login ID existence                | 10/minute |                   |                 |          | Mitigate account enumeration.                                                                                         |
| **Features**                            |           |                   |                 |          |                                                                                                                       |
| Send Login OTP by SMS                   | NIL       | 1 minute cooldown |                 | NIL      | Mitigate mass/targeted message spam.                                                                                  |
| Send Login OTP by Email                 | NIL       | 1 minute cooldown |                 | NIL      |                                                                                                                       |
| Send Forgot Password by SMS             | NIL       | 1 minute cooldown |                 | NIL      |                                                                                                                       |
| Send Forgot Password by Email           | NIL       | 1 minute cooldown |                 | NIL      |                                                                                                                       |
| Send Verification by SMS                | NIL       | 1 minute cooldown |                 | NIL      |                                                                                                                       |
| Send Verification by Email              | NIL       | 1 minute cooldown |                 | NIL      |                                                                                                                       | 
| Login OTP failed verify attempts        |           | 5 attempts        |                 |          | Mitigate credential brute-force; revoke OTP when set limit exceeded.                                                  |
| Verification OTP failed verify attempts |           | 5 attempts        |                 |          |                                                                                                                       |
| Verify Verification OTP                 | 60/minute |                   |                 |          | Mitigate credential brute-force.                                                                                      |
| Verify Forgot Password OTP              | 60/minute |                   |                 |          |                                                                                                                       |
| **Misc.**                               |           |                   |                 |          |                                                                                                                       |
| Presign upload image request            |           |                   | fixed: 10/hour  |          | Configuration not needed for now.                                                                                     |

| Operation               | per-IP     | per-target |                                                     |
| ----------------------- | ---------- | ---------- | --------------------------------------------------- |
| **Messaging**           |            |            |                                                     |
| Send SMS (Global)       | disabled   | 50/day     | By EnvVar                                           |
| Send Email (Global)     | disabled   | 50/day     | By EnvVar                                           |
| Send SMS (Per Tenant)   | 60/minute  | 10/hour    | Require adjust by server operator instead of Portal |
| Send Email (Per Tenant) | 200/minute | 50/day     | Require adjust by server operator instead of Portal |

louischan-oursky:
### Send Forgot Password by SMS, Send Forgot Password by Email
- The user is not yet identified. Rate limit should be per-IP and per-target.

### Send Verification by SMS, Send Verification by Email
- The user is not yet identified in case of sign up (because the user ID is newly generated). Rate limit should be per-IP and per-target.

### Verify Verification OTP
- Add per-target rate limit to mitigate credential brute-force.

kiootic:
> Sorry, what is ";disabled" meant in your table? I removed them first, but not quite sure.

It means disabled by default; the current config for rate limits has three field: enabled, size, and reset period. If developer just set enabled to true, the default rate limit values are applied.

> What this meant: "Send verification message"? "Per-IP rate limit is used for verification during signup, per-user rate limit is used otherwise." Even for signup, per-user is still applicable, right? (Since even during sign-up, we know the "target"/user for the verification already)

As mentioned by Louis, applying per-user rate limit during signup is meaningless, since the user ID is newly generated for every signup attempt.

> Could we merge MFA recovery code and tokens? Or some concern to keep them as two times?

The two authentications have different purposes; MFA recovery code is used when user needs emergency access (e.g. due to loss of MFA authenticator), and MFA device token is used when user is logging in normally in a known device (i.e. skip MFA on this device checkbox). It is plausible that rate limit for recovery code would be configure lower.

> All configs are expected to be in X per Y seconds (TBC?)

There are some special rate limits that may not conform to this format:
- Cooldowns are always 1 per Y seconds, since it is used for resend button cooldown, and it's weird that it can be used in quick succession.
- Failed attempts (i.e. number of failed attempts before revoking the OTP code associated with the target phone number/email, don't know a good name for config though) should be X per OTP lifetime, since it is reset when the OTP code is newly generated.


> Verify Verification OTP: Add per-target rate limit to mitigate credential brute-force.

I think this is not needed, since verifying verification OTP is gated by sending verification message, which does have per-target rate limit.

kiootic:
> Verify any login attempts

This rate limit seems ambigous to me; when would it be checked? before user entering their login ID or before entering credientals (?), which can be multiple steps considering MFA?

docs/specs/rate-limit.md
```diff
@@ -63,4 +64,140 @@ Rate limits without default are hard-coded (non-configurable).
...
+    rate_limits:
+        # No rate limit is applied
+        # validate_code_per_ip:
+        #   period: 1h  # required
+        #   brust: 1    # default to 1
+---
+verification:
+    rate_limits:
+        # 1 validation attempt allowed per hour.
+        validate_code_per_ip:
```
louischan-oursky, line 95: Do we still have a enabled flag? All pointer to struct in the config will be created recursively. Not sure if we want to give a meaning on the zero value of the rate limit struct.

kiootic: There is no enabled flag in the planned config, there are 3 cases:
- the field is not present (i.e. value is missing): the rate limit config is initialized to zero value, which represents no rate limit is configured.
- the field is set to an object with just `period` field: `brust` is default to 1.
- the field is set to an object with both `period` & `brust` fields.

Setting the field to an object with just `brust` field is an error (required).

kiootic: Looks like we can't do it this way: required field nested in optional field value does not roundtrip correctly.

louischan-oursky: So what is our plan?

kiootic: I suppose we need to keep the enabled flag for simplicity. So the cases are:
```yaml
# rate limit disabled by omitting the field
# validate_code_per_ip:
---
# rate limit disabled by settings enabled
validate_code_per_ip:
    enabled: false
---
# rate limit enabled with default brust (1)
validate_code_per_ip:
    enabled: true
    period: 1h
---
# rate limit enabled
validate_code_per_ip:
    enabled: true
    period: 1h
    brust: 10
```

`period` is required if `enabled` is true.

kiootic:
- Fixed typo: `brust` -> `burst`
- Adjusted config field names
    - `send_message_per_` -> `trigger_per_`: since send message & generate code is independent, we would not want to still generate a valid code if rate limit is exceeded.
    - `code_expiry` -> `code_valid_period`: clearer indication on expected value; also add same field for OOB-OTP codes.

docs/specs/rate-limit.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Future Works
+
+- We may want to apply request-level rate limits (e.g. admin API, OIDC endpoints)
+- We may want to exclude certain users (e.g. by IP) from applying rate limit.
+
+
+## Configuration
+
+In general, rate limits are configured using two fields:
```
louischan-oursky, line 83: It should be 3 now.

louischan-oursky (commented):
The configuration shape looks good to me.

chpapa (approved):
I think on business requirements level it is ok, thanks!

I didn't proof read the details of the config file, I assume that's being cross checked between developers.

PR #3015, opened 2023-03-24
Fix invalid oauth session causing internal server error

tung2744:


louischan-oursky:
I added a commit to check if the error is invalid oauth session.

PR #3026, opened 2023-03-30
Update orphan authenticator when updating Identities

Rubikium:
ref https://github.com/authgear/2023C01-latte-customization/issues/212

### DIFF
--- a/pkg/lib/authn/authenticator/oob/store.go
+++ b/pkg/lib/authn/authenticator/oob/store.go
@@ -215,7 +215,7 @@ func (s *Store) Update(a *authenticator.OOBOTP) (err error) {
 		Update(s.SQLBuilder.TableName("_auth_authenticator_oob")).
 		Set("phone", a.Phone).
 		Set("email", a.Email).
-		Where("id", a.ID)
+		Where("id = ?", a.ID)
 	result, err = s.SQLExecutor.ExecWith(q)
 	if err != nil {
 		return err
--- a/pkg/lib/authn/authenticator/oob/store.go
+++ b/pkg/lib/authn/authenticator/oob/store.go
@@ -3,6 +3,7 @@ package oob
 import (
 	"database/sql"
 	"errors"
+	"fmt"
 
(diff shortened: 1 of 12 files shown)


louischan-oursky (commented):
This PR does not work because we are using the updated identity to find dependent authenticators which will never find any. We need to change the signature of `Update(newInfo *identity.Info)` to `Update(oldInfo *identity.Info, newInfo *identity.Info)`. And then use the old info to find dependent authenticators to update.

pkg/lib/authn/authenticator/service/service.go
```diff
@@ -311,12 +312,13 @@ func (s *Service) WithSpec(ai *authenticator.Info, spec *authenticator.Spec) (bo
 		changed = (newAuth != a)
 		return changed, newAuth.ToInfo(), nil
 	case model.AuthenticatorTypeOOBEmail:
```
louischan-oursky, line 314: Case switch does not fall through implicitly in Golang. Also, if you have tested update email locally, this will result in a crash.

pkg/lib/authn/authenticator/oob/store.go
```diff
@@ -180,3 +181,54 @@ func (s *Store) Create(a *authenticator.OOBOTP) (err error) {
...
+		return authenticator.ErrAuthenticatorNotFound
+	} else if rowsAffected > 1 {
+		panic(fmt.Sprintf("authenticator_oob: want 1 row updated, got %v", rowsAffected))
+	}
+
+	q = s.SQLBuilder.
+		Update(s.SQLBuilder.TableName("_auth_authenticator_oob")).
+		Set("phone", a.Phone).
+		Set("email", a.Email).
+		Where("id", a.ID)
```
louischan-oursky: This results in invalid SQL statement. Have you tested your PR locally? This problem will be revealed if you have tested it.

louischan-oursky (commented):
Do you find it difficult to test changes locally? 

I have added some commits to fix the problems I found.

Issue #3023, opened 2023-03-30
Refactor OTP rate limits config (verification/login link/OOB-OTP)


fungc-io:
- Revamp verification OTP mechanism
- Consolidate login link/OOB-OTP OTP behaviors

fungc-io:
Ready on staging, pending deployment

PR #3022, opened 2023-03-30
Revamp verification rate limits

kiootic:
ref #3023

config snippet for test:
```yaml
verification:
  code_valid_period: 1m
  rate_limits:
    email:
      trigger_per_ip:
        enabled: true
        period: 1m
        burst: 1
      trigger_per_user:
        enabled: true
        period: 1m
        burst: 1
      trigger_cooldown: 42s
      max_failed_attempts_revoke_otp: 3
      validate_per_ip:
        enabled: true
        period: 1s
        burst: 1
```

note that existing `resend_cooldown_seconds` in `messaging` would likely override the settings.

PR #3030, opened 2023-04-06
Revamp OOB-OTP rate limits

kiootic:
ref #3023
requires #3022


pkg/lib/authn/otp/code_store_redis.go
```diff
@@ -19,15 +19,15 @@ type CodeStoreRedis struct {
 	Clock clock.Clock
 }
 
-func (s *CodeStoreRedis) set(target string, code *Code) error {
+func (s *CodeStoreRedis) set(purpose string, target string, code *Code) error {
```
louischan-oursky, line 22: Binding the OTP to purpose would make the generated OTP code via Admin API only usable for login, not verification.

louischan-oursky (commented):
The resend cooldown blocks login in the default UI.

1. Sign in with a phone number
2. See the enter OTP page
3. Click back
4. Sign in with the same phone number again
5. See rate limit error

Expected
1. Sign in with a phone number
2. See the enter OTP page
3. Click back
4. Sign in with the same phone number again
5. See the enter OTP page with cooldown counting down. No OTP code should be sent. The old OTP code should still be valid for use.

pkg/lib/authn/otp/code_store_redis.go
kiootic: The admin API is `generateOOBOTPCode`, which (I think) implies the OTP code is usable for OOB-OTP login.

I'm concerned that the OTP code for different purposes could be mixed up, since the code valid period for different purpose could be different.

How about adding an optional field `purpose` for the mutation, with default value to login?

kiootic:
> The resend cooldown blocks login in the default UI.

Cannot reproduce this.


https://user-images.githubusercontent.com/6682519/231997077-ff2862c8-1101-479d-91c0-f907854c5ca3.mov

In the video I'm using 12s cooldown.

louischan-oursky:
@kiootic I have whatsapp enabled too. Not sure if it is related.

kiootic:
The error is because generating Whatsapp OTP is also subject to cooldown rate limit, however its nature is different from normal OTP (receive from user rather than send to user). Since the whatsapp feature is being reworked, I think I would setup a quick workaround for it first.

louischan-oursky:
The rate limit error is gone, but the SMS message is not sent when I switch to SMS.

kiootic:
Updated to separate SMS & Whatsapp rate limits.

PR #3035, opened 2023-04-17
Add audit context to admin api calls from portal

tung2744:


pkg/portal/service/admin_api.go
```diff
@@ -23,6 +28,12 @@ type AdminAPIService struct {
 	AuthzAdder     AuthzAdder
 }
 
+type AuthzAuditContext struct {
+	Triggerer   string `json:"triggerer"`
```
louischan-oursky, line 32: We have seen different usage of Admin API here. Let's be specific.

Let's replace `triggerer` by `usage`.

When the Admin API is used by portal for internal purpose, like collaborators. Specify `{ "usage": "internal" }`.

When the Admin API is proxied for an authenticated collaborator, specify `{ "usage": "proxy", "actor_user_id": "xxx" }`

Make usage a type alias of string. Add constants.

pkg/portal/service/admin_api.go
```diff
@@ -23,6 +28,12 @@ type AdminAPIService struct {
...
+	ActorUserID string `json:"actor_user_id"`
```
louischan-oursky, line 33: This can be empty. Specify `,omitempty`.

pkg/portal/service/admin_api.go
```diff
@@ -23,6 +28,12 @@ type AdminAPIService struct {
...
+	Referrer    string `json:"refererer"`
```
louischan-oursky, line 34: This can be empty. Specify `,omitempty`.

louischan-oursky: Rename this to ```HTTPReferer string `json:"http_referer"```. Note that we have a typo in the original struct tag.

pkg/lib/admin/authz/audit.go
```diff
@@ -0,0 +1,29 @@
+package authz
+
+import "context"
+
+type auditContextKeyType struct{}
```
louischan-oursky, line 5: Should be auth context, instead of audit context. Rename the JSON key as well.

louischan-oursky (commented):
In my test, the context is always null. Any idea?

tung2744:
@louischan-oursky How is your test procedure? I test using portal in local, and do something in the User Management tab (Such as add identity).

pkg/lib/admin/authz/audit.go
tung2744: @louischan-oursky I think name it as `AuthContext` looks like something which store auth state / session etc, but in fact we only use it as audit. The content inside is also not verified. People can claim himself as any user.

louischan-oursky: Let's stick with `audit_context` then.

pkg/lib/admin/authz/middleware.go
```diff
@@ -81,6 +81,13 @@ func (m *Middleware) Handle(next http.Handler) http.Handler {
 			}
 
 			authorized = true
+
+			requestCtx := r.Context()
+			auditCtx, ok := token.Get("audit_context")
```
louischan-oursky, line 86: Use the constant of the key name here.

louischan-oursky:
I figured out why audit_context is null. I set ADMIN_API_AUTH=NONE.

tung2744:
@louischan-oursky Updated and rebased, thanks

PR #3041, opened 2023-04-19
Log portal mutations

tung2744:
requires https://github.com/authgear/authgear-server/pull/3035


pkg/portal/service/app.go
```diff
@@ -113,10 +113,10 @@ func (s *AppService) Get(id string) (*model.App, error) {
 	}, nil
 }
 
-func (s *AppService) GetAppProvider(appID string) (*deps.AppProvider, error) {
+func (s *AppService) WithAppProvider(appID string, executor func(*deps.AppProvider) error) error {
```
louischan-oursky, line 116: I prefer we introduce a new event service to log events, instead of reusing the original one in pkg/lib/event. The new event service installs the hook in the database handle being used in the portal graphql handler. So the bug will not occur in the first place. The new event service only needs the audit sink, skip other irrelevant sinks.

tung2744: @louischan-oursky Actually we still need another transaction for writing to app audit db because it might be another db?

louischan-oursky: The new event service manages the transaction within itself. Note that in pkg/portal/transport/graphql_handler.go, a read only transaction is opened there. May need to adjust transaction isolation level to avoid deadlock.

tung2744: @louischan-oursky In the original event service, we have a user resolver so we can resolve user refs in the service. To implement that resolver we need a lot of things inside the AppProvider.
If we make a new event service, I think we don't implement a new resolver for only this service. So user ref resolving will not work in this service.
Currently, all portal mutations doesn't contains user ref in the payload so this still works so I can do in this way first. However we need to implement a new resolver if we want to resolve user refs in portal events in the future.

pkg/lib/event/service.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/authgear/authgear-server/pkg/lib/infra/db"
 	"github.com/authgear/authgear-server/pkg/lib/session"
 	"github.com/authgear/authgear-server/pkg/lib/uiparam"
+	portalsession "github.com/authgear/authgear-server/pkg/portal/session"
```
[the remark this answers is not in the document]
tung2744: Fixed in [fb2fccc](https://github.com/authgear/authgear-server/pull/3041/commits/fb2fccc67ec59e8d391c304c65e62d10e2308b3e)

tung2744:

louischan-oursky:
@tung2744 I added a few commits.

pkg/portal/wire_gen.go
```diff
@@ -296,14 +296,19 @@ func newGraphQLHandler(p *deps.RequestProvider) http.Handler {
 	}
 	remoteIP := deps.ProvideRemoteIP(request, trustProxy)
 	userAgentString := deps.ProvideUserAgentString(request)
+	writeHandle := auditdb.NewWriteHandle(context, pool, databaseEnvironmentConfig, auditDatabaseCredentials, logFactory)
+	writeSQLExecutor := auditdb.NewWriteSQLExecutor(context, writeHandle)
```
tung2744, line 300: @louischan-oursky I think it is possible that the app's audit db is not the same as the portal audit db. And now `auditDatabaseCredentials` comes from environment config of portal but not from config of app, therefore we may write to wrong db if the app has another audit db configured?

louischan-oursky: Actually the portal only reads ONE audit database from the environment variable. And it is assumed (for now) the app audit db is the same as the portal audit db.

Issue #3040, opened 2023-04-19
Refactor authentication rate limits config (password/TOTP/passkey/etc.)


kiootic:

fungc-io:
Ready on staging, pending deployment

PR #3045, opened 2023-04-20
Use new rate limits for authentication

kiootic:
ref #3040


pkg/lib/authn/mfa/service.go
```diff
@@ -55,12 +92,23 @@ func (s *Service) CreateDeviceToken(userID string, token string) (*DeviceToken,
...
+	defer s.RateLimiter.Cancel(perIP)
+
 	if err != nil {
 		return err
 	}
 
 	_, err = s.DeviceTokens.Get(userID, token)
+	if errors.Is(err, ErrDeviceTokenNotFound) {
+		perUserPerIP.Consume()
+		perIP.Consume()
```
louischan-oursky, line 110: Both perUserPerIP and perIP is nullable.

kiootic: If `reserveRateLimit` return nil `err`, both `perUserPerIP` and `perIP` is reserved successfully and non-nil.

Issue #3046, opened 2023-04-20
Add rate limits config in portal


kiootic:
- update portal to use new rate limit config
- add the config "Daily limits to trigger email verification from the setting page (leave blank to turn off limit)"
  - below 'verify email by' field

fungc-io:
Ready on staging, pending deployment

PR #3051, opened 2023-04-24
Update portal to use new rate limit configs

kiootic:
ref #3046

Screenshot 2023-04-24 at 5 57 55 PM

### DIFF
--- a/portal/src/graphql/portal/LoginMethodConfigurationScreen.tsx
+++ b/portal/src/graphql/portal/LoginMethodConfigurationScreen.tsx
@@ -732,6 +732,13 @@ function setPrimaryAuthenticator(
   );
 }
 
+function parseOptionalDuration(s: string | undefined) {
+  if (s == null || s === "") {
+    return undefined;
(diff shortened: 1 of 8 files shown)


louischan-oursky (commented):
The otp valid period is empty in my environment. I debugged and found that `config.authenticator?.oob_otp?.sms?.code_valid_period = "1h"` and `config.verification?.code_valid_period = "1h0m0s"`. They are equivalent but getConsistentValue is string-based.

louischan-oursky:
Screenshot 2023-04-26 at 21 38 18
Screenshot 2023-04-26 at 21 38 24
Screenshot 2023-04-26 at 21 38 30


@fungc-io It looks like this now. I changed **Daily 6-digit OTP limit per email address (leave blank to turn off limit)** to **Daily 6-digit OTP limit per user (leave blank to turn off limit)** because the config is per user, not per target.

fungc-io:
Yes it works :+1:

PR #3050, opened 2023-04-24
Add admin tab to audit log screen in portal

tung2744:
requires https://github.com/authgear/authgear-server/pull/3047

PR #3056, opened 2023-04-25
Log admin api secrets viewed event

tung2744:
requires https://github.com/authgear/authgear-server/pull/3050

### DIFF
--- a/pkg/portal/graphql/app.go
+++ b/pkg/portal/graphql/app.go
@@ -111,11 +111,11 @@ var smtpSecret = graphql.NewObject(graphql.ObjectConfig{
 type AppSecretKey string
 
 const (
-	AppSecretKeyOauthSSOProviderClientSecrets AppSecretKey = "oauthSSOProviderClientSecrets"
+	AppSecretKeyOauthSSOProviderClientSecrets AppSecretKey = "oauthSSOProviderClientSecrets" // nolint:gosec
 	AppSecretKeyWebhookSecret                 AppSecretKey = "webhookSecret"
 	AppSecretKeyAdminAPISecrets               AppSecretKey = "adminAPISecrets"
 	AppSecretKeySmtpSecret                    AppSecretKey = "smtpSecret"
-	AppSecretKeyOauthClientSecrets            AppSecretKey = "oauthClientSecrets"
+	AppSecretKeyOauthClientSecrets            AppSecretKey = "oauthClientSecrets" // nolint:gosec
 )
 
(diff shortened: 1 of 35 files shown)


pkg/portal/graphql/app.go
```diff
@@ -198,11 +220,23 @@ var nodeApp = node(
 			},
 			"secretConfig": &graphql.Field{
 				Type: graphql.NewNonNull(secretConfig),
+				Args: graphql.FieldConfigArgument{
+					"unmaskedSecrets": &graphql.ArgumentConfig{
+						Type: graphql.NewNonNull(graphql.NewList(graphql.NewNonNull(secretKey))),
```
louischan-oursky, line 244: Make the array nullable (keep the element non-null) so that existing queries are compatible.

louischan-oursky (commented):
We currently fire the event in a graphql query. So refreshing the page will generate new events. I think we can improve this by a token mechanism. It works like this

1. To view a secret, one must pass a token.
2. To obtain a token, the auth_time must be recent enough (i.e. the original reauthentication mechanism)
3. The token is bound for what secret can be viewed. So the caller must specify which secret they want to view when obtaining the token.
4. The token is valid for 5 minutes. It expires itself.
5. Using the token to view secret does not invalidate the token.
6. If the auth_time is recent enough, obtaining a token does not require reauthentication
7. Log view secret event when we issue a token.

tung2744:
@louischan-oursky 
> We currently fire the event in a graphql query. So refreshing the page will generate new events.
It won't. This is because in portal, we now have to press "reveal" explicitly to reveal the secret.
Once the user pressed reveal, we will put an object into the browser history. When the user finished reauthentication and back to portal, we will pop that object and query the unmasked secrets. When the user refresh the page, because we already cleared the object in the history, he cannot see the secret until he pressed reveal again.

louischan-oursky:
But if someone repeatedly call the query, it will

tung2744:
Per discussion, will use the token approach.

tung2744:
@louischan-oursky Updated to use token approach.
https://github.com/orgs/authgear/projects/64/views/1?pane=issue&itemId=27683819

> If the auth_time is recent enough, obtaining a token does not require reauthentication

For this one, I found it can be quite complicated and require some refactoring in the code on loading state of the screen / forms.
I've created another issue to handle it:
https://github.com/authgear/authgear-server/issues/3080

louischan-oursky:
It seems that confidential client secret is not masked. It always show the copy button.

Issue #2954, opened 2023-02-16
Potential race condition causing resolver to throw error indefinitely


louischan-oursky:
```
time="2023-02-16T13:33:17Z" level=error msg="panic occurred" error="runtime error: invalid memory address or nil pointer dereference" logger=panic-middleware stack="[log.PanicValue at /src/pkg/util/log/panic.go:56 middleware.(*PanicMiddleware).Handle.func1.3 at /src/pkg/lib/infra/middleware/panic.go:48 runtime.gopanic at /usr/local/go/src/runtime/panic.go:884 runtime.panicmem at /usr/local/go/src/runtime/panic.go:260 runtime.sigpanic at /usr/local/go/src/runtime/signal_unix.go:835 deps.(*RootProvider).NewAppProvider at /src/pkg/lib/deps/providers.go:92 deps.(*RequestMiddleware).Handle.func1 at /src/pkg/lib/deps/middleware_request.go:37 http.HandlerFunc.ServeHTTP at /usr/local/go/src/net/http/server.go:2109 httputil.StaticSecurityHeaders.func1 at /src/pkg/util/httputil/static_security_headers.go:10 http.HandlerFunc.ServeHTTP at /usr/local/go/src/net/http/server.go:2109 middleware.(*SentryMiddleware).Handle.func1 at /src/pkg/lib/infra/middleware/sentry.go:22 http.HandlerFunc.ServeHTTP at /usr/local/go/src/net/http/server.go:2109 middleware.(*BodyLimitMiddleware).Handle.func1 at /src/pkg/lib/infra/middleware/body_limit.go:16 http.HandlerFunc.ServeHTTP at /usr/local/go/src/net/http/server.go:2109 middleware.(*PanicMiddleware).Handle.func1 at /src/pkg/lib/infra/middleware/panic.go:63 http.HandlerFunc.ServeHTTP at /usr/local/go/src/net/http/server.go:2109 httprouter.(*Router).Handler.func1 at /go/pkg/mod/github.com/julienschmidt/httprouter@v1.3.0/router.go:275 httprouter.(*Router).ServeHTTP at /go/pkg/mod/github.com/julienschmidt/httprouter@v1.3.0/router.go:387 httproute.(*Router).ServeHTTP at /src/pkg/util/httproute/httproute.go:85 http.serverHandler.ServeHTTP at /usr/local/go/src/net/http/server.go:2947 http.(*conn).serve at /usr/local/go/src/net/http/server.go:1991]"
```

[...]

PR #3065, opened 2023-04-27
Make global db handle never be shared across threads

tung2744:
ref #2954


louischan-oursky:
I added a commit to use wire to write the body of the factory. One advantage is that the dependencies are checked by wire. If we add a new field to SQLBuilder or SQLExecutor, wire will detect missing dependency and fail to generate. If we write the body by hand, we will miss that and result in nil dereference error.

Issue #3066, opened 2023-04-28
Migrate all configs from old to new rate limit


kiootic:

fungc-io:
@louischan-oursky do we need to do this in the next production deployment?

louischan-oursky:
Yes, as part of the deployment.

PR #3067, opened 2023-04-28
Add rate limits config migration script

kiootic:
ref #3066


louischan-oursky:
I added a commit to not always write password rate limit config. Any reason why you always wrote it? @kiootic

kiootic:
The original password rate limit is always used, so I added a default value. I think the general auth rate limit can cover that behavior so can remove it.

Issue #3068, opened 2023-04-28
Use GCRA for rate limits


kiootic:
ref: https://github.com/throttled/throttled

PR #3070, opened 2023-04-28
Use GCRA for rate limits

kiootic:
ref #3068


docker-compose.yaml
```diff
@@ -13,7 +13,7 @@ services:
       - "5432:5432"
 
   redis:
-    image: redis:6.0
+    image: redis:6.2
```
louischan-oursky, line 16: Does this algorithm require features only available >= 6.2?

kiootic: Yes, `PXAT` is added at 6.2

pkg/lib/ratelimit/gcra_test.go
```diff
@@ -0,0 +1,172 @@
...
+			burst:  2,
+			entries: []scheduleEntry{
+				{fn: func() {
+					s.HSet(testKey, "test", "123")
+				}},
+				{time: "0s", n: 0, ok: true, timeToAct: -5},
+				{time: "0s", n: 1, ok: true, timeToAct: 0},
+				{time: "0s", n: -1, ok: true, timeToAct: -5},
+			},
+		})
```
louischan-oursky, line 170: Does the new algorithm handle negative underflow now? Please add a test case to demonstrate it.

kiootic: It is tested in the reservation test case, let me add some comments to describe the entries.

Issue #3072, opened 2023-05-03
Introduce a feature config to generate fixed OTP


fungc-io:
For load test, a fixed OTP is needed
This should be a internal feature config for a project

Fixed:
- OTP
- Verification code

tung2744:
Created another issue to handle totp
https://github.com/authgear/authgear-server/issues/3074

PR #3073, opened 2023-05-03
Add feature config for fixed oob otp and deterministic link otp

tung2744:
ref #3072

### DIFF
--- a/pkg/lib/config/feature.go
+++ b/pkg/lib/config/feature.go
@@ -26,7 +26,8 @@ var _ = FeatureConfigSchema.Add("FeatureConfig", `
 		"messaging": { "$ref": "#/$defs/MessagingFeatureConfig" },
 		"collaborator": { "$ref": "#/$defs/CollaboratorFeatureConfig" },
 		"web3": { "$ref": "#/$defs/Web3FeatureConfig" },
-		"admin_api": { "$ref": "#/$defs/AdminAPIFeatureConfig" }
+		"admin_api": { "$ref": "#/$defs/AdminAPIFeatureConfig" },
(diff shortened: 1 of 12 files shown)


pkg/lib/config/feature.go
```diff
@@ -26,7 +26,8 @@ var _ = FeatureConfigSchema.Add("FeatureConfig", `
 		"messaging": { "$ref": "#/$defs/MessagingFeatureConfig" },
 		"collaborator": { "$ref": "#/$defs/CollaboratorFeatureConfig" },
 		"web3": { "$ref": "#/$defs/Web3FeatureConfig" },
-		"admin_api": { "$ref": "#/$defs/AdminAPIFeatureConfig" }
+		"admin_api": { "$ref": "#/$defs/AdminAPIFeatureConfig" },
+		"feature": { "$ref": "#/$defs/FixtureFeatureConfig" }
```
louischan-oursky, line 30:
```yaml
test_mode:
  fixed_oob_otp:
    enabled: true
    code: "123456"
```

A second thought, I think we cannot have fixed link_otp. It is because we can look up a particular reset password flow by a link_otp. If every link_otp is the same, then reset password of different user will overwrite each other. Seems we need to agree on a deterministic 32-character length code that the client can generate by themselves.

And recovery code is actually not fixed. But we want to allow a magic recovery code that would verified successfully.

louischan-oursky, line 30: It should be `test_mode` and `TestModeConfig`

tung2744:
@louischan-oursky Updated. Also added the actual implementation. Thanks!

Issue #3075, opened 2023-05-05
Support user agent binding in workflow api


tung2744:
- [ ] Define a new cookie `workflow_ua`
- [ ] In new workflow api, add one flag `bind_user_agent: boolean`
- [ ] If `bind_user_agent` is true, set or renew the expiry time of the cookie `workflow_ua_id` in new workflow api
- [ ] Add `UserAgentID string` to workflow session.  If `bind_user_agent` is true, set it to the value of `workflow_ua_id` on creating workflow.
- [ ] Check `UserAgentID` with the request cookie in all workflow api if it is not empty.

fungc-io:
Already supported

PR #3078, opened 2023-05-08
Implement user agent binding in workflow api

tung2744:
ref ref https://github.com/authgear/authgear-server/issues/3075


tung2744:

PR #3088, opened 2023-05-10
Ensure authenticator cache was updated after updated identities

tung2744:


portal/src/graphql/adminapi/mutations/updateIdentityMutation.generated.ts
```diff
@@ -24,13 +24,23 @@ export const UpdateIdentityMutationDocument = gql`
         edges {
           node {
             id
+            type
```
louischan-oursky, line 27: The standard attributes are still stale. Please also fetch the standard attributes in this mutation result.

louischan-oursky: Let me fix this.

Issue #3106, opened 2023-05-19
Portal become unusable if a sso provider config set without an client id


tung2744:
There are two bugs here:

1. Under some conditions, client_id of sso providers can be set to empty string without validation error.
2. When we successfully trigger 1, we cannot update other configs of the app because it always throw validation error saying that client_id is missing.

tung2744:
I found this is a more generic problem of the whole authgear config, let me create a new issue to describe the details.
For this issue, I made a pr to fix the client_id field only.

fungc-io:
Ready on staging, pending deployment
Follow up in #3109

PR #3108, opened 2023-05-22
Ensure user cannot save a config with empty client id in sso providers

tung2744:
ref #3106


louischan-oursky:
Have you ever encounter a situation the portal saves an empty client ID? Do you know how to reproduce it?

louischan-oursky:
Oh, just saving an empty client ID will trigger this validation error

tung2744:
@louischan-oursky When the user modified the field, the value of the field is not undefined. Therefore it will pass the schema validation, even it is an empty string.
If the user didn't modify the field, it is undefined but not empty string, in this case validation error will be thrown.

louischan-oursky:
@louischan-oursky Check if preexisting projects will be affected by this change.

Issue #3103, opened 2023-05-18
Support sending whatsapp otp in server


tung2744:
- [x] Add secrets config `whatsapp.on-premises`
  - username: string
  - password: string
  - namespace: string

- [x] Add authgear config
  ```yaml
  whatsapp:
    api_type: "on-premises" # Support "cloud" later
    api_endpoint: ENDPOINT # Required if type is "on-premises"
    templates:
      otp:
        name: TEMPLATE_NAME
        languages:
          - en
        components:
          header:
            parameters:
              - '{{ template "app.name" }}'
          body:
            parameters:
              - "one-time password"
              - "{{ .Code }}"
              - "logging in"
  ```

- [ ] Implement a whatsapp api client
  - [ ] Cache the access token in redis by app_id + namespace + username

fungc-io:
ready on staging

Issue #3102, opened 2023-05-18
Support whstapp otp flow in auth ui


tung2744:
- [x] Update implementations of the following edges so that it send the code with whatsapp
  - [x] EdgeCreateAuthenticatorWhatsappOTPSetup
  - [x] EdgeAuthenticationWhatsappTrigger
  - [x] EdgeVerifyIdentityViaWhatsapp

- [x] Update implementations of the following nodes so it requires a code input to verify
  - [x] NodeCreateAuthenticatorWhatsappOTP
  - [x] NodeAuthenticationWhatsapp
  - [x] NodeVerifyIdentityViaWhatsapp

- [x] Update /flows/whatsapp_otp so that it can accept code input on POST

- [x] Update whatsapp otp screen in authui to support inputting code

- [x] Implement resend

- [x] Skip sending in debug mode

- [x] Clean up unused code
    - [x] Wati webhook handler
    - [x] Mark wati credential deprecated
    - [x] Review GetWhatsappOTP
    - [x] Review GetServerWhatsappPhone

fungc-io:
@tung2744 Let's reuse the SMS screen with the title changed to WhatsApp

![Image](https://github.com/authgear/authgear-server/assets/15207470/92935f54-8c6d-4806-b135-de6f3799e106)

fungc-io:
Please make sure this works when switching SMS/WhatsApp
1. Sent code in Whatsapp
2. Switch to SMS, receive SMS code
3. Switch back to WhatsApp, cool down is still counting, the old code received in whatsapp should still work

PR #3119, opened 2023-05-24
Implement whatsapp otp

tung2744:
ref #3103 #3102

# How to test

1. In authgear.yaml, add the following section
  ```yaml
  messaging:
    whatsapp:
      api_type: "on-premises"
      templates:
        otp:
          name: authgear_whatsapp_otp
          type: authentication
          namespace: "WATI_NAMESPACE"
          languages:
            - en
            - zh_CN
            - zh_HK
  ```

2. In authgear.secrets.yaml, add the following
  ```yaml
  - data:
      username: "WATI_USERNAME"
      password: "WATI_PASSWORD"
      api_endpoint: "WATI_ENDPOINT"
    key: whatsapp.on-premises
  ```

3. Ensure `authenticator.oob_otp.sms.phone_otp_mode` is `whatsapp_sms` in authgear.yaml.
4. Login / Signup / Add Identity with whatsapp

Find all credentials in 1password.

## Available templates

I've created two templates in our wati whatsapp account

1. authgear_whatsapp_otp
  -  `type` must be `authentication` when using this template.
  - Supported languages are `en` `zh_CN` `zh_HK`
<del>
2. authgear_otp_util
  - `type` must be `util` when using this template.
  - Supported languages is only `en`
  - `components` must be specified to tell authgear server how to use this template. You can use the following config:
    ```
    components:
      header:
        parameters:
          - '{{ template "app.name" }}'
      body:
        parameters:
          - '{{ template "whatsapp.body.otp" }}'
          - "{{ .Code }}"
          - '{{ template "whatsapp.body.action" }}'
    ```
    - You need to add translation strings `whatsapp.body.otp` and `whatsapp.body.action` in `var/templates/<language>/translation.json`. For reference:
      ```json
      {
        "whatsapp.body.otp": "one-time password",
        "whatsapp.body.action": "logging in",
      }
      ```
</del>

TODO:

  - [x] Remove util template support


pkg/lib/config/secret_whatsapp.go
```diff
@@ -0,0 +1,26 @@
...
+		"namespace": { "type": "string", "minLength": 1 }
+	},
+	"required": ["username", "password", "namespace"]
+}
+`)
+
+type WhatsappOnPremisesCredentials struct {
+	Username  string `json:"username"`
+	Password  string `json:"password"`
+	Namespace string `json:"namespace"`
```
louischan-oursky, line 19: If I understand correctly, namespace should be configured in the template.

pkg/lib/config/secret_whatsapp.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func (c *WhatsappOnPremisesCredentials) SensitiveStrings() []string {
+	return []string{
+		c.Password,
```
louischan-oursky, line 24: Include username as well.

pkg/lib/config/messaging_whatsapp.go
```diff
@@ -32,8 +32,23 @@ var _ = Schema.Add("WhatsappTemplatesConfig", `
...
+
+const (
+	WhatsappTemplateTypeAuthentication WhatsappTemplateType = "authentication"
+	WhatsappTemplateTypeUtil           WhatsappTemplateType = "util"
+)
+
+var _ = Schema.Add("WhatsappTemplateType", `
+{
+	"type": "string",
+	"enum": ["authentication", "util"]
```
louischan-oursky, line 45: The way we support `util` is NOT language-agnostic. I suggest we only support authentication template type for now.

tung2744:
Tested we can have different text and copied text in the template. Will update the code to add app name into the code.

pkg/lib/infra/whatsapp/models.go
```diff
@@ -99,16 +97,15 @@ type UserToken struct {
...
 		return err
 	}
 	*j = LoginResponseUserExpiresTime(t)
 	return nil
 }
 
-func (j LoginResponseUserExpiresTime) MarshalJSON() ([]byte, error) {
-	return json.Marshal(time.Time(j))
+func (j LoginResponseUserExpiresTime) MarshalText() ([]byte, error) {
+	return time.Time(j).MarshalText()
```
louischan-oursky, line 110: Add a test for this pair of methods. The result of MarshalText should be able to be UnmarshalText. I think this is not now.

louischan-oursky (commented):
1. See if we can remove pkg/lib/authn/authenticator/whatsapp/provider.go entirely. The rationale is that whatsapp works like SMS now. I think we should be able to support Whatsapp  in pkg/lib/authn/otp.
2. I think we can stop firing `whatsapp.otp.verified` event. The rationale is that we do not have similar event for SMS nor email. @fungc-io Do you have any concern?
3. Do not add `InspectCode` to `otp.Service`. The rationale is that we do not have it now, but SMS and email OTP are still working fine.
4. KindWhatsapp is actually kindOOBOTP. But we have other kinds like KindVerification and KindForgotPassword. For forgot password, we send a reset link. Given we only support authentication template, we may not support sending reset password link via Whatsapp. But it should be possible to send verification code via Whatsapp.

louischan-oursky (commented):
1. See if we can remove pkg/lib/authn/authenticator/whatsapp/provider.go entirely. The rationale is that whatsapp works like SMS now. I think we should be able to support Whatsapp  in pkg/lib/authn/otp.
2. I think we can stop firing `whatsapp.otp.verified` event. The rationale is that we do not have similar event for SMS nor email. @fungc-io Do you have any concern?
3. Do not add `InspectCode` to `otp.Service`. The rationale is that we do not have it now, but SMS and email OTP are still working fine.
4. KindWhatsapp is actually kindOOBOTP. But we have other kinds like KindVerification and KindForgotPassword. For forgot password, we send a reset link. Given we only support authentication template, we may not support sending reset password link via Whatsapp. But it should be possible to send verification code via Whatsapp.

tung2744:
TODO:
- [x] Eliminate pkg/lib/authn/authenticator/whatsapp/provider.go
- [ ] Update whatsapp audit event
- [ ] Check rate limit configs

tung2744:
@louischan-oursky Updated and rebased, thanks!
The changes start from 1bb40c7265b750b8dfcec78edfc49da40f619c23

pkg/lib/messaging/message_sms.go
```diff
@@ -17,8 +17,7 @@ type SMSMessage struct {
 }
 
 func (m *SMSMessage) Send() error {
-	err := m.events.DispatchEvent(&nonblocking.SMSSentEventPayload{
-		Sender:    m.Sender,
+	err := m.events.DispatchEvent(&nonblocking.WhatsappSentEventPayload{
```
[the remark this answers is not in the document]
louischan-oursky: Let me update this.

PR #3124, opened 2023-05-30
Support create_with_batch_input

louischan-oursky:


pkg/auth/handler/api/workflow_v2.go
```diff
@@ -26,7 +26,7 @@ var WorkflowV2RequestSchema = validation.NewSimpleSchema(`
 		"type": "object",
 		"properties": {
 			"action": {
-				"enum": ["create", "input", "batch_input", "get"]
+				"enum": ["create", "input", "batch_input", "create_with_batch_input", "get"]
```
tung2744, line 29: What if we simply allow `batch_input` field in "create" and do not introduce a new action `create_with_batch_input`?

louischan-oursky: Good idea.

louischan-oursky:
Updated.

PR #3126, opened 2023-06-05
Specify device sso

tung2744:


docs/specs/oidc-sso-device.md
```diff
@@ -0,0 +1,84 @@
...
+
+Applications which are able to share authentication session are expected to be signed by the same vendor certificate, and are expected share the device token and id token by a shared storage which only accessible by applications signed by the same vendor certificate.
+
+## Implementation details
+
+### Server-side
+
+- Token Endpoint
+
+  - Support a new scope `device_sso`
```
louischan-oursky, line 21: Add new a paragraph to specify that when `device_sso` is included in scope, then `x_suppress_idp_session_cookie` is implied as true.

Also specify that `x_sso_enabled` CANNOT be used together with `device_sso`, doing so will result in OAuth error in the authorization endpoint.

tung2744:
@louischan-oursky Updated, thx!

PR #3127, opened 2023-06-05
Workflow customization

louischan-oursky:


docs/specs/workflow-customization.md
```diff
@@ -0,0 +1,1570 @@
...
+      }
+    }
+  }
+}
+```
+
+### Alternative design if Expressions were not used
+
+Here we show how would the configuration looks like if expressions were not used.
+We take the [Use case example 5](#use-case-example-5-comprehensive-example) and rewrite it.
```
fungc-io, line 1467: @chpapa This is the alternative design without IFs. What do you think? This maybe harder to express the conditions of which a authentication method is needed

chpapa: @fungc-io @louischan-oursky I think these look prettier and easier to understand. Did I miss anything? 

My syntax is sth like this:

- `steps` is an array with:
   - `one_of` or `any_of`
   - `type`
- `one_of` or `any_of` is an array with `type`
- `type` is a dictionary with `id` and `steps`

```
signup_flows:
- id: default_signup_flow
  steps:
  - one_of:
    - type: identification
      id: email
      steps:
      - type: authentication
        id: primary_email_code
      - type: verification
        id: email
      - type: authentication
        id: primary_password
      - type: authentication
        id: secondary_totp
    - type: identification
      id: oauth
```

(I noticed that in the config, verification doesn't have id, I put sth there, as I don't feel right it could "imply" which verification it's doing...?)

```
login_flows:
- id: default_login_flow
  steps:
  - one_of:
    - type: identification
      id: email
      steps:
      - one_of:
        - type: authentication
          id: primary_password
          steps:
          - one_of:
            - type: authentication
              id: secondary_totp
            - type: authentication
              id: recovery_code
            - type: authentication
              id: device_token
        - type: authentication
          id: primary_email_code
          steps:
          - one_of:
            - type: authentication
              id: secondary_totp
            - type: authentication
              id: recovery_code
            - type: authentication
              id: device_token
        - type: authentication
          id: primary_passkey
    - type: identification
      id: oauth
    - type: identification
      id: passkey
```

Frankly I like it much better XD I think it would also easier to code a Visual Editor for the flow in future, if our config is like this instead of expression literal?

A few other tweak in my mind:
1. Use `auth`, `identify` and `verify` instead of `authentication`, `identification` and `verification` to save a few keystrokes and easier to read?
2. Use `steps`, `steps_one_of`, `steps_all_of` instead of two layer? Would it be easier to read?

louischan-oursky: According to your proposed syntax, your first example should look like this.

```yaml
# Enumerated flows:
# 1. oauth
# 2. email -> verify the email -> setup OTP authenticator (this is skipped because we have verified the email) -> setup TOTP
# 3. email -> verify the email -> setup password -> setup TOTP

signup_flows:
- id: default_signup_flow
  steps:
    one_of:
    - type: identification
      id: oauth
    - type: identification
      id: email
      steps:
        one_of:
        - type: verification
          id: email
          steps:
            one_of:
            - type: authentication
              id: primary_email_code
              steps:
                one_of:
                - type: authentication
                  id: secondary_totp
            - type: authentication
              id: primary_password
              steps:
                one_of:
                - type: authentication
                  id: secondary_totp
```

One issue of this proposed syntax is the existence of `any_of`.
First, assume we agree on the meaning of `one_of`, `all_of`, and `any_of` according to https://json-schema.org/understanding-json-schema/reference/combining.html and https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/.
Second, assume we have the following configuration.

```yaml
signup_flows:
- id: default_signup_flow
  steps:
    one_of:
    - type: identification
      id: email
      steps:
        one_of:
        - type: verification
          id: email
          steps:
            any_of: # any_of
            - type: authentication
              id: primary_password
              steps:
                one_of:
                - type: authentication
                  id: secondary_totp
            - type: authentication
              id: primary_email_code
              steps:
                one_of:
                - type: authentication
                  id: secondary_totp
```

Since `any_of` meaning one or more, we need to ask the end-user if they want to stop after fulfilling one step.
But this kind of asking is uncommon in signup flows.
Good UX should not make the end-user to make a decision, but this syntax allows this to happen.

Another problem is that how should the nested steps be executed if more than 1 steps in `any_of` were taken?
I do not know the answer because the configuration is ambiguous.

Yet another problem is that without conditionals, we cannot express an optional step.
For example, suppose we want to support a new type of step `prompt_setup_passkey`. (We do have this step in our default UI now).
We cannot express this in the proposed syntax, but we can express this with conditionals.

```
login_flows:
- id: default_login_flow
  steps:
  # ...
  - type: prompt_setup_passkey
    if: !user.has_passkey
```

---

> (I noticed that in the config, verification doesn't have id, I put sth there, as I don't feel right it could "imply" which verification it's doing...?)

`type: verification` uses reference to which step the verification is for.

```yaml
- type: identification_method
  id: setup_identity
  one_of:
   identification_method:
     id: email
- type: verification
  step:
    id: setup_identity
```

---

> Use auth, identify and verify instead of authentication, identification and verification to save a few keystrokes and easier to read?

I agree. Let's change to `authenticate`, `identify`, and `verify`. All of them are verbs and are not in short forms.

chpapa: @louischan-oursky ah I didn't understand the original sample 5; I now understand it better with your comments.

However, I was actually kinda assuming `steps` would perform an action on the array's `type` step by step, so the two `one_of` could be skipped in the `signup_flows`?

```yaml
signup_flows:
- id: default_signup_flow
  steps:
  - one_of:
    - type: identify
      id: oauth
    - type: identify
      id: email
      steps:
      - type: verify
        id: email
      - steps:
        - one_of:
          - type: authenticate
            id: primary_email_code
            steps:
            - type: authenticate
              id: secondary_totp
          - type: authenticate
            id: primary_password
            steps:
            - type: authenticate
              id: secondary_totp
```

Um, actually I thought `any_of` could be for those optional steps; Maybe that's too implicit or complicated. Let's remove it; In that case there are only `steps` and `one_of`, I think we can put them into "same level"? So it should further simplify the redundant steps followed by one_of patterns:

```yaml
signup_flows:
- id: default_signup_flow
  one_of:
  - type: identify
    id: oauth
  - type: identify
    id: email
    steps:
    - type: verify
      id: email
      one_of:
      - type: authenticate
        id: primary_email_code
        steps:
        - type: authenticate
          id: secondary_totp
      - type: authenticate
        id: primary_password
        steps:
        - type: authenticate
          id: secondary_totp
```

Besides `prompt_setup_passkey`, do we have any other optional items in our use case now? If not I think we could:

1. Make the `prompt_setup_passkey` type of action optional by default when the user already have passkey (even make it more explicit say `prompt_setup_passkey_if_not_exists`)
2. We could still have a simple expression in the type configuration or in the flow configuration, but at least we could make sure they're only used in certain edge case; Hence most of the flow would still be eumunerated. That should still make the "flow editor" (if it ever happens) very easy to code?

louischan-oursky: > Make the prompt_setup_passkey type of action optional by default when the user already have passkey (even make it more explicit say prompt_setup_passkey_if_not_exists)

I tend not to couple the condition with the action. Maybe the developer wants to prompt setup passkey when the user did not use passkey to sign in.
Or sometimes the developer wants to prompt setup passkey only when the user never did that before. Having conditionals allow us to support ONE type once and for all.

> We could still have a simple expression in the type configuration or in the flow configuration, but at least we could make sure they're only used in certain edge case; Hence most of the flow would still be eumunerated. That should still make the "flow editor" (if it ever happens) very easy to code?

My concern is I want to follow one principle from the Zen of Python.

> There should be one-- and preferably only one --obvious way to do it.

If we have conditionals only, then it is the only way to do it.
If we have nested steps **AND** conditionals, then we will have 2 ways to do it.

For enumerability, in the current design of only one level of steps and conditionals, we can still enumerate all the flows by enumerating the truth table of all conditionals in the steps.

In case of the "flow editor", the configuration written against the current design does not have duplication. This makes modification and deletion simple to implement too.

chpapa: > I tend not to couple the condition with the action. Maybe the developer wants to prompt setup passkey when the user did not use passkey to sign in.
Or sometimes the developer wants to prompt setup passkey only when the user never did that before. Having conditionals allow us to support ONE type once and for all.

But actually I think we have many of these implicit meanings in various actions already, as our config is basically specifying "what needs to check" on each step, the actual implementation of each action depends on the UI, I think it is unavoidable we will have a lot of reasons why the UX of what the users actually do would depends on the frontend implementation.

I think in your current design, it also assume something like that? (e.g. verify a phone no. after authenticate with SMS, I think we will have a logic to skip / auto satisfy the phone verification?)

> There should be one-- and preferably only one --obvious way to do it.

I don't think that's a valid concern, the expression I proposed here is just for skipping an optional step (And we might not even need that if we have no other optional steps we can think of now besides `prompt_setup_passkey`), not for doing conditions.

Any other cons of using nested configuration? Or pros of using such linear one with complicated expressions?

I think the pros are quite obvious:
1. Easier to implement (just a simple tree in array), both for the config and convert from/to flow chart/flow editor (writing a parser to parse the expression and covert them to tree form is not as simple, I'm not sure why you said "simple to implement")
2. Easier to write the config (ask any developers who love having expressions in declarative config instead of imperative config?)

p.s. if we really need "real conditions" in future, I could imagine we can still implement it via something like this?

```yaml
steps:
  - if: <expressions>
    true:
    false:
```

louischan-oursky: > I think in your current design, it also assume something like that? (e.g. verify a phone no. after authenticate with SMS, I think we will have a logic to skip / auto satisfy the phone verification?)

The point of the configuration is to specify the interaction taken by the end-user.
So each step corresponds to an interaction between the UI and the end-user.
Your example is about whether we mark a phone number as verified every time an OTP is verified for that phone number.
Theoretically this behavior can be attached to a step, thus can be turned on / off for a specific step.

> Any other cons of using nested configuration?

Nested configuration inherently contains duplications.
The editor needs to take care of duplications in case of update / deletion.

> Easier to implement (just a simple tree in array), both for the config and convert from/to flow chart/flow editor (writing a parser to parse the expression and covert them to tree form is not as simple, I'm not sure why you said "simple to implement")

I agree writing a parser to parse the expression is not simple, but it is not very hard neither.
I doubt why we need to convert our linear configuration to a tree.
For enumeration purpose, we can just show all linear configurations with conditionals evaluated.
Or we can draw a flow chart like [this](https://mermaid.live/edit#pako:eNplj0EOgyAQRa9CZl0T3bpoUqm9gF1VXEwEq4mAoZCmMd69qKXadAEh89_MG0aoNReQQtPrZ92isUwRUlgxnMrlrkgUHQmNR6oV72ynFfbxNEM09hG5GicWZKazpSertviC_WPL6ZLTKjgCvzqSvSNZHcmfYz9jjTdHrnjpz3d-YH9COIAURmLH_bfHGWVgWyEFg9Q_uWjQ9ZYBU5NH3cDRityvpQ2kzaw6ADqri5eqIbV-tQCdO7wblKE4oLppLT9N0xv-73S_)

> Easier to write the config (ask any developers who love having expressions in declarative config instead of imperative config?)

Maybe we should create a poll to survey this. In my personal experience, I like the design of GitHub Actions (1 level of steps with conditionals), which I borrow the syntax from.

louischan-oursky:
I updated the spec according to the offline discussion.

docs/specs/workflow-customization.md
```diff
@@ -0,0 +1,1486 @@
...
+  steps:
+  - type: identify
+    one_of:
+    - identification: email
+    - identification: phone
+    - identification: username
+  - type: authenticate
+    one_of:
+    - authentication: primary_password
+    - authentication: primary_sms_code
```
chpapa, line 748: should we allow `target` of `oob_opt_sms` (and email counterpart) specify not just to a step, but also to a login ID or a custom attributes?

But I think that's very edge cases... most case would works by allow target = "*"

docs/specs/workflow-customization.md
```diff
@@ -0,0 +1,1486 @@
...
+We define available IdentificationMethods under `identification_methods`.
+The `id` is in the global namespace.
+
+Example:
+
+```yaml
+identification_methods:
+# Identify the User by a Email Login ID Identity
+- id: email
+  type: "login_id"
```
chpapa, line 183: would it make more sense to just have type `email`, `phone` and `username` instead?

docs/specs/workflow-customization.md
```diff
@@ -0,0 +1,1486 @@
...
+  login_id:
+    type: "username"
+# Identify the User by a OAuth Identity
+- id: oauth
+  type: "oauth"
+  oauth:
+    aliases: ["google", "apple"]
+# Identify the User by a Anonymous Identity
+- id: anonymous
```
chpapa, line 202: Is it really technically possible to have any combination of `anonymous`, `biometric` flow? (Passkey maybe...?)

If not, maybe we shouldn't put them into spec and allow configuration?

chpapa (changes requested):
left some comments.

I think the `verify`, `required` comments might need changes, others are optional to address.

docs/specs/workflow-customization.md
```diff
@@ -0,0 +1,1486 @@
...
+      target_step: setup_identity
+  # Set up a primary password.
+  - type: authenticate
+    one_of:
+    - authentication: primary_password
+  # Verify the phone number or the email address
+  # If this step is not specified, the phone number or the email address is unverified.
+  - type: verify
+    if: contains(fromJSON('["phone", "email"]'), steps.setup_identity.identification)
+    target_step: setup_identity
```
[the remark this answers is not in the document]
louischan-oursky: I think I will remove the `phone_otp_mode` and `email_otp_mode` for now. That means we do not support flow-specific configuration at this stage. We can add it in the future.

docs/specs/workflow-customization.md
louischan-oursky: I chose step because we do not know which login ID would be selected by the end-user. What would happen if the config reference a login ID that was not selected?

And we do not support authenticating with custom attributes. We treat all custom attributes opaque.

docs/specs/workflow-customization.md
louischan-oursky: Actually we can. I did not do that because I wanted to the `type` equal to the types of the identity we have.

docs/specs/workflow-customization.md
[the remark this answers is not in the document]
chpapa: @louischan-oursky So I think a bigger questions here is why there are `identification_methods`, `authentication_methods` but not `verification_methods`

Right now...

1. Each `identification_methods` seems just `id`=`type`
2. Each `authentication_methods` are just `id`= `kind`+`type`

If there are no other options possible in those methods, why not just remove them?

If there are possible options in those methods, there should be similar options in verification as well? We should make them consistent or else we will risks a breaking changes of config in future?

docs/specs/workflow-customization.md
chpapa: I know step is needed for reference, I meant it seems we might need more than that...

E.g. would there be use cases where we want to specific only these list of phone no. could be used for `secondary_oob_otp_sms`? If yes I don't see how we can do it with current configurations.

docs/specs/workflow-customization.md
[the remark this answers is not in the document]
louischan-oursky: > If there are no other options possible in those methods, why not just remove them?

I have removed them.

docs/specs/workflow-customization.md
louischan-oursky: If I understand you correctly, you are referring to a feature like this.

```yaml
authentication_methods:
- id: phone1
  kind: secondary
  type: oob_otp_sms
  labels: ["some_special_label"]
```

And then authenticator created with reference to `phone1`WILL have labels `["some_special_label"]` attached to them and persisted in the database. Later on we can reference to `phone1` to select those authenticators.

We do not support attaching any labels to identity / authenticator at the moment. And this feature does not work with settings page. We do not allow customized workflow for the actions in the settings page.

I think this is out of the scope of this PR.

louischan-oursky:
- [ ] Design how to handle account linking flow.
- [ ] Design how to handle optional 2FA step.

docs/specs/workflow-customization.md
```diff
@@ -0,0 +1,1032 @@
...
+  - id: setup_identity
+    type: identify
+    one_of:
+    - identification: phone
+      steps:
+      - type: authenticate
+        one_of:
+        - authentication: primary_oob_otp_sms
+          target_step: setup_identity
+      - step: verify
```
tung2744, line 183: @louischan-oursky The verify step here is missing `type`.

docs/specs/workflow-customization.md
```diff
@@ -0,0 +1,1032 @@
...
+        target_step: setup_identity
+    - identification: email
+      steps:
+      - type: authenticate
+        one_of:
+        - authentication: primary_oob_otp_email
+          target_step: setup_identity
+      - step: verify
```
tung2744, line 191: @louischan-oursky The verify step here is missing `type`.

docs/specs/workflow-customization.md
louischan-oursky: Fixed.

docs/specs/workflow-customization.md
louischan-oursky: Fixed.

louischan-oursky:
When I am working on https://github.com/authgear/authgear-server/issues/3274 I realize we do need `authentication: recovery_code` and `authentication: device_token`. 

We also need a new `type: change_password` in login flow to support the existing feature of forced password change in login.

I will amend the spec.

tung2744:
@louischan-oursky I guess this spec is quite stable already, lets merge this pr and send new prs for future updates?

louischan-oursky:
Sure.

fungc-io:
thanks all

Issue #3129, opened 2023-06-06
Account Lockout policy for repeated login attempts


fungc-io:
## Goal

To enhance account security by locking out account after repeated failed login attempts.
This is different from API rate limit, which throttles the requests into a constant average number.

## Design

After certain failed attempts (X) in a duration (Y), end-user must wait for a lockout duration (T1) before next attempt.
For subsequent failed attempts, the end-user must wait for another lockout duration (T2), etc.
Z grows exponentially by a backoff factor (n) until a maximum defined duration.
The lockout is reset after:
- The end-user authenticated successfully with the correct credentials.
- Duration Y is passed after their last failed attempts.

A developer can define the lockout policy by:
- **X** - Max attempts: Maximum attempts to login before the account was locked
- **Y** - Duration: The duration of the max attempts is counted
- **Tmin** - minimum
- **Tmax** - Maximum
- **n** - Backoff Factor

Let's say the security policy is
- X: 10, Y: 1440m, Tmin: 5m, Tmax: 5m, n:1
  - It means within 24 hour, if the end-user input the wrong password 10 times, they will be locked for 5 minutes. For every next failed attempt, they must wait for another 5 minutes until the entered password is correct.

fungc-io:
backend change is done in staging, frontend to be added in #3139

fungc-io:
@tung2744 


![Image](https://github.com/authgear/authgear-server/assets/15207470/f754a7ac-f9fb-4e0c-80ae-d508051a762a)


When showing the account is locked,
The input and button should be disabled. And can we make the error message a count down?

louischan-oursky:
@tung2744 We have a ResendButtonController. Maybe you can reference to it.

tung2744:
@fungc-io 
> The countdown will show a round down number in the message.
When they need to wait 2 minutes, it always show "1 min" in the first minute. I think round up should be a better UX?
Currently, the relative time was formatted using the [intl api](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat). The behavior was controlled by the api. Also round up will has problem when you have 2 minutes counting down, you will see "2 minutes" -> "59 seconds".

fungc-io:
> Also round up will has problem when you have 2 minutes counting down, you will see "2 minutes" -> "59 seconds".

I think that's better than "1 min" for 1st min?

[...]

PR #3134, opened 2023-06-08
Implement account lockout

tung2744:
ref #3129 

The spec will be updated in another pr.


pkg/lib/interaction/nodes/do_reset_lockout_attempts.go
```diff
@@ -0,0 +1,44 @@
...
+	n := &NodeDoResetLockoutAttempts{}
+	userID := ""
+	authenticators := graph.GetUsedAuthenticators()
+	types := []model.AuthenticatorType{}
+	for _, a := range authenticators {
+		userID = a.UserID
+		types = append(types, a.Type)
+	}
+	if len(types) > 0 && userID != "" {
+		ctx.Lockout.ClearAttempts(userID, types)
```
louischan-oursky: We should perform this in OnCommitEffect

pkg/latte/intent_login.go
```diff
@@ -93,6 +97,14 @@ func (i *IntentLogin) ReactTo(ctx context.Context, deps *workflow.Dependencies,
 			}
 		}
 	case 2:
+		authenticators, err := i.getVerifiedAuthenticators(w)
+		if err != nil {
+			return nil, err
+		}
+		err = deps.Authenticators.ClearLockoutAttempts(authenticators)
```
louischan-oursky, line 104: We should perform this in OnCommitEffect.

pkg/lib/lockout/error.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+const bucketNameKey = "bucket_name"
+const untilKey = "until"
+
+var ErrLocked = apierrors.TooManyRequest.WithReason("Locked")
+
+func NewErrLocked(bucketName BucketName, until time.Time) error {
+	return ErrLocked.NewWithInfo("resource locked", apierrors.Details{
+		bucketNameKey: bucketName,
```
louischan-oursky, line 16: Remove the bucket name.

pkg/lib/lockout/bucket.go
```diff
@@ -0,0 +1,76 @@
...
+	"time"
+
+	"github.com/authgear/authgear-server/pkg/lib/config"
+)
+
+type BucketName string
+
+const (
+	BucketNameAccountAuthentication BucketName = "AccountAuthentication"
+)
```
louischan-oursky, line 14: Remove BucketName because we are unlikely to have another.

pkg/lib/lockout/bucket.go
```diff
@@ -0,0 +1,76 @@
...
+
+type BucketSpec struct {
```
louischan-oursky, line 16: I think we can eliminate this struct and use AuthenticationLockoutConfig directly.

Or we rename this struct to LockoutSpec. Bucket is a term from rate limit. But I do not think we need to inherit this legacy in this module.

louischan-oursky (commented):
1. We need to format the lockout error in default UI. Currently It just shows `resource locked`
```
English
Your account is locked out. Please try again after {timestamp}
zh-HK
你的帳號已被鎖定。請於{timestamp}後再試。
zh-TW
你的帳戶已被鎖定。請於{timestamp}後再試。
```

tung2744:

louischan-oursky:
@tung2744 I added a commit to change how we format the time in the error message. We have a Stimulus Controller to format relative time. We use that to handle time formatting in the client side.

PR #3138, opened 2023-06-13
Specify app2app authorization

tung2744:


tung2744:
Added ability to configure user authentication and mentioned the oidc ciba flow in the spec.

docs/specs/app2app.md
```diff
@@ -0,0 +1,199 @@
...
+  - A new parameter `device_key` is supported. This value can be specified by the client to enable app2app login. If specified, it should be a public key of a key pair, which the private key should be stored in a secure storage in the device. Such as the Android keystore or the iOS keychain.
+    - Multiple key types such as RS256 and NIST P-256 are supported.
+
+- Token Endpoint
+
+  - A new grant_type `urn:authgear:params:oauth:grant-type:app2app` is supported. When using such grant_type, The following must be provided in the request:
+    - `refresh_token`: a valid refresh token.
+    - `client_id`: current client id.
+    - `app2app_client_id`: the authenticating client id.
+    - `challenge`: a signature of a challenge token obtained from the `/oauth2/challenge` api, signed with the private key binded with the provided refresh token (using the `device_key` parameter during authentication).
```
louischan-oursky, line 26: To follow anonymous user / biometric login, this parameter should be named as `jwt`, and the value is a JWT signed with the private key of the `device_key`. It has the `challenge` claim in the JWT payload.

The JWT header allows us to tell the server which kind of the key is, as well as the signature algorithm. (RS256 or ES256)

docs/specs/app2app.md
```diff
@@ -0,0 +1,199 @@
...
+Assume there are two apps, A and B. App A is holding a valid user session, and app B is trying to authenticate through app A.
+
+- App A
+
+  - The client id is `client_a`
+  - App A is able to handle applinks in domain `https://a.example.com/` with a properly setup [.well-known/assetlinks.json](https://developer.android.com/training/app-links/verify-android-applinks) file.
+
+- App B
+
+  - The configured `x_redirect_uris` is `["appb://redirect"]`
```
louischan-oursky, line 74: The `redirect_uris` includes `https://b.example.com/redirect`

docs/specs/app2app.md
```diff
@@ -0,0 +1,199 @@
...
+
+   - For security reason, App B should verify the signature of the app that is going to handle the intent by:
+     - Fetching the `.well-known/assetlinks.json` from the domain of the uri. In this example it will be `https://a.example.com/.well-known/assetlinks.json`.
+     - Obtain the package name and signature of the app which is going to handle the uri.
+     - Find the `"delegate_permission/common.handle_all_urls"` relation, and obtain the related `package_name` and `sha256_cert_fingerprints` from the section. Verify it with the package name and signature obtained in the previous step. If none of the items match, returns error.
+     - After validated the package name and signature, start an activity with an intent with the above uri.
+
+2. App A handles the intent.
+
+   - Returns an error if App A doesn't have an valid session.
```
louischan-oursky, line 95: I think we should support first `authenticate()` in App A, after that, continue the flow.

docs/specs/app2app.md
```diff
@@ -0,0 +1,199 @@
...
+   - Find out the package name of the app which will handle the `redirect_uri`.
+   - Verify the signature of the above app by using the `.well-known/assetlinks.json` file, from the domain of the `redirect_uri`. In this example it is `https://b.example.com/.well-known/assetlinks.json`.
+   - Call the token endpoint with `grant_type=urn:authgear:params:oauth:grant-type:app2app`.
+     - `refresh_token` should be the refresh token of the existing session of App A.
+     - `client_id` should be the client id of this app.
+     - `app2app_client_id` should be obtained from the intent uri, which is the authorizing app.
+     - `challenge` should be generated with a private key stored in the android keystore, created on the app first authenticated with the server. Refer to the Server-side section for details.
```
louischan-oursky, line 102: - `jwt`: Refer to the Server-side section for details.

docs/specs/app2app.md
```diff
@@ -0,0 +1,199 @@
...
+
+3. App B handles the redirect uri intent.
+
+   - Extract `code` from the uri and use it to perform the code exchange.
+
+#### Android SDK
+
+The following methods will be added in android sdk to support the app2app flow:
+
+- `startApp2AppAuthentication(authorizeUri: String, redirectUri: String)`
```
louischan-oursky, line 124: I think the signature of this method should be

`(authorizationEndpoint: String, clientID: string, redirectURI: string)`

Example

`(authorizationEndpoint: "https://a.example.com/authorize", clientID: "client_a", redirectURI: "https://b.example.com/redirect")`

Then the developer do not need to compose the URI for us. In the previous version, we need to parse the partial composed URI and add query parameters.

docs/specs/app2app.md
```diff
@@ -187,13 +195,20 @@ Assume there are two apps, A and B. App A is holding a valid user session, and a
...
 
-#### iOS SDK
+#### iOS
+
+The following parameters will be added to `authenticate()` method:
+
+- `app2appOptions: App2AppOptions?`
+  - app2app options. If `null`, this app cannot authenticate other apps through app2app.
+  - `App2AppOptions` contains the following fields:
+    - `userAuthenticationFlags: SecAccessControlCreateFlags?`: The authentication the user must perform on authenticating another app through app2app flow. Read [this doc](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags) for possible values. Internally, it was used to set the [access control flags](https://developer.apple.com/documentation/security/1394452-secaccesscontrolcreatewithflags) of the device key.
```
louischan-oursky, line 205: We should follow existing options we have for biometric. See BiometricAccessConstraint and BiometricLAPolicy.

docs/specs/app2app.md
```diff
@@ -111,6 +111,14 @@ Assume there are two apps, A and B. App A is holding a valid user session, and a
 
 #### Android SDK
 
+The following field will be added to `AuthenticateOptions`:
+
+- `app2appOptions: App2AppOptions?`
+  - App2app options. If `null`, this app cannot authenticate other apps through app2app.
+  - `App2AppOptions` contains the following fields:
+    - `userAuthenticationRequired: Boolean`: Whether the user has to pass an authentiction process during the app2app flow. Read [this doc](<https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setUserAuthenticationRequired(boolean)>) for details.
+    - `userAuthenticationTypes: Int?`: Type of user authentication can be used. Read [this doc](<https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setUserAuthenticationParameters(int,%20int)>) for details.
```
louischan-oursky, line 120: We should follow the existing option we have for biometric. See BiometricAccessConstraintAndroid and BiometricOptionsAndroid in the react native SDK.

docs/specs/app2app.md
```diff
@@ -187,13 +195,20 @@ Assume there are two apps, A and B. App A is holding a valid user session, and a
...
+  - app2app options. If `null`, this app cannot authenticate other apps through app2app.
+  - `App2AppOptions` contains the following fields:
+    - `userAuthenticationFlags: SecAccessControlCreateFlags?`: The authentication the user must perform on authenticating another app through app2app flow. Read [this doc](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags) for possible values. Internally, it was used to set the [access control flags](https://developer.apple.com/documentation/security/1394452-secaccesscontrolcreatewithflags) of the device key.
 
 The following methods will be added in android sdk to support the app2app flow:
 
 - `startApp2AppAuthentication(authorizeUri: String, redirectUri: String)`
   - This method should be called to trigger a new app2app authentication through another client through `authorizeUri`, and receive the result through `redirectUri`.
-- `handleApp2AppAuthenticationRequest(uri: URL)`
+- `handleApp2AppAuthenticationRequest(uri: URL, )`
```
louischan-oursky, line 211: typo

docs/specs/app2app.md
```diff
@@ -0,0 +1,219 @@
...
+     - `code_challenge`: The PKCE code challenge with the same purpose in the oidc authorize endpoint.
+
+2. App A handles the universal link.
+
+   - Returns an error if App A doesn't have an valid session.
+   - Call the token endpoint with `grant_type=urn:authgear:params:oauth:grant-type:app2app`.
+     - `refresh_token` should be the refresh token of the existing session of App A.
+     - `client_id` should be the client id of this app.
+     - `app2app_client_id` should be obtained from the intent uri, which is the authorizing app.
+     - `challenge` should be generated with a private key stored in the android keystore, created on the app first authenticated with the server. Refer to the Server-side section for details.
```
louischan-oursky, line 187: `jwt`

docs/specs/app2app.md
```diff
@@ -0,0 +1,219 @@
...
+The following parameters will be added to `authenticate()` method:
+
+- `app2appOptions: App2AppOptions?`
+  - app2app options. If `null`, this app cannot authenticate other apps through app2app.
+  - `App2AppOptions` contains the following fields:
+    - `userAuthenticationFlags: SecAccessControlCreateFlags?`: The authentication the user must perform on authenticating another app through app2app flow. Read [this doc](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags) for possible values. Internally, it was used to set the [access control flags](https://developer.apple.com/documentation/security/1394452-secaccesscontrolcreatewithflags) of the device key.
+
+The following methods will be added in android sdk to support the app2app flow:
+
+- `startApp2AppAuthentication(authorizeUri: String, redirectUri: String)`
```
louischan-oursky, line 209: Apply the change on the signature here as well.

docs/specs/app2app.md
tung2744: @louischan-oursky We don't know what is the parameter we should use to call `authenticate()` in AppA in app2app flow. If we want to support that, we may need to support all parameters of `authenticate()` in `startApp2AppAuthentication()`.

louischan-oursky: Ideally, we should let the developer to decide how to call `authenticate()`. What we need to do is to resume the app2app2 flow after authentication.

tung2744: @louischan-oursky Then I think maybe we don't need `startApp2AppAuthentication`, we can just reuse `authenticate` with an endpoint which will be handled by App A.

tung2744: This flow can actually handled by the current api design by calling `authenticate` in app A before calling `handleApp2AppAuthenticationRequest`, will write it down in the spec.

tung2744:
@louischan-oursky Updated, thanks :adore:

PR #3140, opened 2023-06-14
Fix otp resend timestamp always updated on inspected

tung2744:


pkg/lib/ratelimit/reservation.go
```diff
@@ -34,9 +34,6 @@ func (r *Reservation) DelayFrom(t time.Time) time.Duration {
 		return 0
 	}
 	delay := r.timeToAct.Sub(t)
-	if delay < 0 {
-		return 0
-	}
```
louischan-oursky, line 39: I think we should remove DelayFrom instead. Its existence promotes a source of bug. We use DelayFrom to calculate can_resend_at. can_resend_at is equivalent to timeToAct. So we should simply expose timeToAct as `func (r *Reservation) TimeToAct() *time.Time`

louischan-oursky:
```golang
var unixEpoch = time.Unix(0, 0)

func (r *Reservation) TimeToAct() time.Time {
  if r == nil || r.timeToAct == nil {
    return unixEpoch
  }
  return *r.timeToAct
}
```

tung2744: I found that timeToAct is actually updating on each inspect.

tung2744: It seems it is an expected behavior of gcra if the burst is 1, so it seems difficult to change it.

PR #3142, opened 2023-06-14
Improve the readability of GCRA

louischan-oursky:


pkg/lib/ratelimit/gcra_test.go
```diff
@@ -118,17 +118,17 @@ func TestGCRA(t *testing.T) {
 			period: "20s",
 			burst:  4,
 			entries: []scheduleEntry{
-				{time: "0s", n: 5, ok: false, timeToAct: -15}, // allow burst requests within limit
+				{time: "0s", n: 5, ok: false, timeToAct: 5}, // allow burst requests within limit
```
tung2744, line 121: I actually found the definition of timeToAct is still difficult to understand.
In is_conforming == true case, the meaning of `timeToAct` is the timestamp you are able to make the next 1 attempt, no matter what n is.
However in is_conforming == false case, the meaning of `timeToAct` will be the timestamp you are able to make the next n attempt, which seems a bit inconsistent.

louischan-oursky: What about we multiply it by `n` in is_confirming == true case?

tung2744: However that may cause incorrect next otp resend timestamp because we inspect the timestamp by using n=0

PR #3143, opened 2023-06-15
Make app2app spec clearer & support migration to app2app from existing sessions

tung2744:


tung2744:
@louischan-oursky Updated the spec for the recently discussed config for migration of existing session.

docs/specs/app2app.md
```diff
@@ -21,20 +21,38 @@ An app can start the authentication flow by opening a link to another app, inste
...
+  - A app2app section will be added to the client configs.
+
+  ```yaml
+  oauth:
+    clients:
+      - client_id: CLIENT_ID
+        # ...Other configs
+        x_app2app:
+          enabled: true
+          allow_insecure_device_key_binding: false
```
louischan-oursky, line 35: The convention in client metadata is not to use nested object. So I think we should call them `x_app2app_enabled` and `x_app2app_insecure_device_key_binding_enabled`.

docs/specs/app2app.md
```diff
@@ -21,20 +21,38 @@ An app can start the authentication flow by opening a link to another app, inste
...
+  ```
+
+  - `enabled`: boolean. Whether the client is able to handle app2app authentication requests from other clients.
+  - `allow_insecure_device_key_binding`: boolean. Default `false`. If `true`, refresh tokens of this client without a binded `device_key` can be binded to a new `device_key` during refreshing the tokens. This option is for allowing existing logged in users to use app2app without requiring them to re-login. However this also allows sessions without a binded `device_key` to be binded with any key and participate in app2app authentication which might be a security concern.
```
louischan-oursky, line 39: s/binded/bound

docs/specs/app2app.md
```diff
@@ -21,20 +21,38 @@ An app can start the authentication flow by opening a link to another app, inste
...
 
 - Token Endpoint
 
+  - For `grant_type=authorization_code`, a new parameter `device_key` is supported. This value can be specified by the client to enable app2app login. If specified, it should be a base64 encoded jwk of public key of a key pair, which the private key should be stored in a secure storage in the device. Such as the Android keystore or the iOS keychain. `device_key` will be ignored if `x_app2app.enabled` of the client configuration is not `true`.
```
louischan-oursky, line 43: Rename this to `x_app2app_device_key_jwt` to signify it is our own extension.

And we are going to make it a JWT.

header
```
{
  "alg": "ES256",
  "kty": "EC",
  "kid": "randomly-generated-kid",
  "jwk": { ... }
}
```

Payload
```
{
  "iat": 123,
  "exp": 123,
   "challenge": "one-time-use-challenge-obtain-from-server"
}
```

docs/specs/app2app.md
```diff
@@ -21,20 +21,38 @@ An app can start the authentication flow by opening a link to another app, inste
...
+
+    - Multiple key types such as RS256 and NIST P-256 are supported.
+
+  - For `grant_type=refresh_token`, a new parameter `device_key` is supported if `x_app2app.allow_insecure_device_key_binding` is `true`. The provided `device_key` will be binded to the provided refresh token, if the refresh token isn't already binded to a `device_key`. If the refresh token was already binded to a `device_key` which doesn't match the one provided in this parameter, an error will be returned.
```
louischan-oursky, line 47: Apply the same change.

docs/specs/app2app.md
tung2744: @louischan-oursky Why making it a jwt? This is the public key to bind to the refresh token but not to verify the user owns the private key.

louischan-oursky: 1. It is a proof that the caller has the private key. We will verify the signature of the JWT with the JWK inside it.
2. Avoid replay attack. We will check if the challenge is consumed or not.

tung2744:
@louischan-oursky Updated, thanks :adore:

PR #3147, opened 2023-06-20
Update portal subscription screen

tung2744:


portal/src/graphql/portal/SubscriptionScreen.tsx
```diff
@@ -608,25 +608,23 @@ function SubscriptionScreenContent(props: SubscriptionScreenContentProps) {
   const [enterpriseDialogHidden, setEnterpriseDialogHidden] = useState(true);
   const [cancelDialogHidden, setCancelDialogHidden] = useState(true);
 
-  // @ts-expect-error
```
louischan-oursky, line 601: I prefer using ts-expect-error to casting to unknown. In this particular case, the problem is the type definition is too strict on the type of `subText`. In case the type definition is corrected in a future version of fluent UI, the ts-expect-error will become invalid. This will be caught by tsc. If we cast to unknown, then the compiler will not be able to do its job.

.gitignore
```diff
@@ -29,3 +29,6 @@ resources/portal/static/
 
 # generated code
 pkg/lib/webparcel/parcel_gen.go
+
+# OSX
+.DS_Store
```
louischan-oursky, line 34: Nice

louischan-oursky (commented):
Screenshot 2023-06-23 at 11 13 42

I guess we need to update the number `50000` to `10000` to match the new enterprise card?

portal/src/graphql/portal/SubscriptionScreen.tsx
tung2744: Oh I see. I changed this because I think the `@ts-expect-error` here ignored all errors in the whole object, but what we only have type error in `subText` line. Maybe I rewrite it so that we only ignore type errors in `subText`.

tung2744:
@louischan-oursky Updated thx!

PR #3160, opened 2023-06-23
Collect whatsapp usage and upload to stripe

tung2744:
ref #3155

### DIFF
--- a/pkg/lib/usage/count_collector.go
+++ b/pkg/lib/usage/count_collector.go
@@ -37,6 +37,12 @@ type smsCountResult struct {
 	total        int
 }
 
+type whatsappCountResult struct {
+	northAmerica int
+	otherRegions int
+	total        int
+}
+
 type CountCollector struct {
 	GlobalHandle  *globaldb.Handle
 	GlobalDBStore *GlobalDBStore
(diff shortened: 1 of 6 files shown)


pkg/lib/usage/count_collector.go
```diff
@@ -307,7 +312,57 @@ func (c *CountCollector) querySMSCount(appID string, rangeFrom *time.Time, range
 	}
 }
 
-func (c *CountCollector) querySMSSentEvents(appID string, rangeFrom *time.Time, rangeTo *time.Time, first uint64, after model.PageCursor) (events []*event.Event, lastCursor model.PageCursor, err error) {
+func (c *CountCollector) queryWhatsappCount(appID string, rangeFrom *time.Time, rangeTo *time.Time) (*whatsappCountResult, error) {
```
louischan-oursky, line 315: Maybe we can use generics here to avoid duplicating the algorithm.

tung2744: I think we should not assume whatsapp count and sms count are having same regions, therefore duplicated the function to preserve the flexibility of having different regions in sms and whatsapp. I've reused the same code in `queryEvents` because I think they should always have the same logic.

Issue #3139, opened 2023-06-13
Update portal design for Account Lockout policy


fungc-io:
UI for #3129 
So the user can configure the account lockout policy on the Authgear Portal

fungc-io:
Hi @crys-tal ,

thanks for your design in https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?type=design&node-id=9212%3A97671&t=4Ot1CNlr3Ktyu7Rn-1

I've left 1 comment in the figma,
And we've added 2 more settings to the feature

Please also add these

- for which types of authenticators will their attempts be tracked, only the enabled authenticators should be available in the config
    - type:
        - Password
        - Passwordless vis Email/Phone
        - Authenticator apps
        - Recovery code
    - If the user can login with OTP than Password, and only password is participating in the lockout policy,
        - They can try OTP by many times and only locked when password failed
 
- Lockout type
    - Per user
        - The user will be lockout from logging in on any devices
    - Per user per IP 
        - The user will be lockout from logging in on the same IP

crys-tal:
@fungc-io (correct me if I'm wrong...) User should be able to set different lockout threshold/ lockout duration for different authenticator types? Or would it be one 'universal' lockout policy for all the selected authenticator types? Thanks!

fungc-io:
It will be one 'universal' lockout policy for all the selected authenticator types

crys-tal:
@fungc-io I've added the 2 new settings to the feature. Please take a look and see if it's okay. Thanks a lot! https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?type=design&node-id=9212%3A97671&t=sZ8d06oiccoh8IUE-1

fungc-io:
Hi @chpapa 
see if you have any comments on the design for this portal config?
https://www.figma.com/file/msiE4O5imHONAG5EjhZeiZ/Authgear-UI?type=design&node-id=9212%3A97671&t=ogvDhOD5BYceEY5E-1

chpapa:
Left a few minor comments on copywriting. @crys-tal @fungc-io

fungc-io:
Design confirmed! thanks @crys-tal

tung2744:
@fungc-io Some point would like to mention after looking at the design:
1. Passwordless via Phone / Email should be a single config, we didn't support configuring them seperately.
[...]

PR #3171, opened 2023-06-27
Add account lockout config in portal

tung2744:
ref #3139


portal/src/locale-data/en.json
```diff
@@ -1455,12 +1455,21 @@
...
+  "LoginMethodConfigurationScreen.lockout.title": "Account Lockout Policy",
+  "LoginMethodConfigurationScreen.lockout.threshold.title": "Lockout threshold",
+  "LoginMethodConfigurationScreen.lockout.threshold.description": "Configure the number of failed login attempts your users can make within a time limit before their accounts are locked.",
+  "LoginMethodConfigurationScreen.lockout.threshold.failedAttempts.title": "Failed Login Attempts",
+  "LoginMethodConfigurationScreen.lockout.threshold.failedAttempts.description": "This is the maximum failed attempts to login before the account is locked.",
+  "LoginMethodConfigurationScreen.lockout.threshold.resetAfter.title": "Reset failed attempts after (minutes)",
+  "LoginMethodConfigurationScreen.lockout.threshold.resetAfter.description": "The interval to reset failed attempts after the last failure.",
+  "LoginMethodConfigurationScreen.lockout.threshold.overall.description": "A user account will be locked after {attempts} failed login attempts. The failure attempts will be reset after {resetIntervalMins} minutes of the last failure. {resetIntervalHours, plural, =0 {} other {(i.e. {resetIntervalHours} hours)}}",
```
louischan-oursky: It should be `{resetIntervalHours, plural, =0 {} one {(i.e. # hour)} other {(i.e. # hours)}`

And this is what is rendered:

`0`: `""`
`1`: `"(i.e. 1 hour)"`
other: `"(i.e. x hours)"`

The difference between `zero`, `=0` is that `zero` refers to the plural category, and `=0` is a simple numeric comparison. In English, the applicable plural categories are `one` and `other`, meaning that `englishPluralRule(number) => "one" | "other"`. So if we use `zero` here, it has no effect. But we also need to handle `one` here. The `#` sign allows us to refer to the number in question.

portal/src/graphql/portal/LockoutSettings.tsx
```diff
@@ -72,71 +78,212 @@ function useNumberOnChange<T extends State>(
   );
 }
 
+function formatOptionalHour(hour: number): string {
```
louischan-oursky, line 126: To format an optional number, we need to use two values.

```
{xIsPresent, select, true {{x, plural, one{# unit} other{# units}}} other{none}}
```

```
{
  xIsPresent: String(x != null) // a string of "true" or "false", which can be used in select
  x: x,
}
```

The problem is this function is that it returns `none` which is language specific.

portal/src/locale-data/en.json
```diff
@@ -1462,7 +1462,18 @@
...
+  "LoginMethodConfigurationScreen.lockout.duration.description": "Configure the number of minutes an account will remain locked when it has exceeded the lockout threshold.",
+  "LoginMethodConfigurationScreen.lockout.duration.duration.title": "Lockout Duration (minutes)",
+  "LoginMethodConfigurationScreen.lockout.duration.duration.description": "This determines how long the account will remain locked.",
+  "LoginMethodConfigurationScreen.lockout.duration.backoff.title": "Backoff Factor",
+  "LoginMethodConfigurationScreen.lockout.duration.backoff.description": "This value will increase the lockout duration exponentially.",
+  "LoginMethodConfigurationScreen.lockout.duration.max.title": "Maximum Lockout Duration (minutes)",
+  "LoginMethodConfigurationScreen.lockout.duration.max.description": "This will allow you to put a cap on the lockout duration.",
+  "LoginMethodConfigurationScreen.lockout.duration.overall.description.noBackoff": "If a user account has exceeded the lockout threshold, it will be locked for {durationMins} minutes. For every subsequent failed attempt, the user must wait for another {durationMins} minutes until a correct password is entered.",
```
louischan-oursky, line 1474: Use `plural` for `durationMins` here.

portal/src/locale-data/en.json
```diff
@@ -1462,7 +1462,18 @@
...
+  "LoginMethodConfigurationScreen.lockout.duration.overall.description.withBackoff": "If a user account has exceeded the lockout threshold, it will be locked for {durationMins} minutes. For every subsequent failed attempt, lockout duration will increase by {backoffFactor} times (e.g. {durationMins} mins, {durationMinsSecond} mins, {durationMinsThird} mins...) until it reaches the maximum of {maxDurationMins} minutes {maxDurationHours, select, none {} other {(i.e. {maxDurationHours} hours)}}.",
```
louischan-oursky, line 1475: Use `plural` for `durationMins`, `durationMinsSecond`, `durationMinsThird` here. No need to use `plural` for `backoffFactor` because `times` and `time` means different things.

portal/src/graphql/portal/LoginMethodConfigurationScreen.tsx
```diff
@@ -794,6 +799,14 @@ function parseOptionalDuration(s: string | undefined) {
   return parseDuration(s);
 }
 
+function parseOptionalDurationInMinutes(s: string | undefined) {
```
louischan-oursky, line 802: The name is confusing. This function is not parsing an input "in minutes", it is parsing an duration string into a number in minutes. Maybe we should call it `parseOptionalDurationIntoMinutes(s: string | undefined): number | undefined`.

portal/src/util/duration.ts
```diff
@@ -34,12 +36,25 @@ export function parseDuration(s: string): number {
...
     seconds += value;
   }
   return sign * seconds;
 }
 
-export function formatDuration(seconds: number): string {
-  return seconds.toString() + "s";
+export function formatDuration(
+  quantity: number,
+  unit: DurationUnit = "s"
```
louischan-oursky, line 47: I suggest we make `unit` non-optional.

portal/src/util/duration.ts
```diff
@@ -34,12 +36,25 @@ export function parseDuration(s: string): number {
...
+): string {
+  return quantity.toString() + unit;
+}
+
+export function formatOptionalDuration(
+  quantity: number | undefined,
+  unit: DurationUnit = "s"
```
louischan-oursky, line 54: I suggest we make `unit` non-optional.

portal/src/graphql/portal/LockoutSettings.tsx
```diff
@@ -20,7 +20,7 @@ export interface State {
   historyDurationMins?: number;
   minimumDurationMins?: number;
   maximumDurationMins?: number;
-  backoffFactor?: number;
+  backoffFactorRaw?: string;
```
louischan-oursky, line 23: Why do we need to use string to represent this?

louischan-oursky (commented):
Screenshot 2023-06-28 at 11 43 01

Setting max attempts to 0 to disable account locking is non-trivial to the developer. I suggest we add a toggle to turn this on and off. When turn on, set the max attempts to a sensible value like 10. When the toggle is off, we disable all following fields.

---

Screenshot 2023-06-28 at 11 43 09

1. The error message `invalid format` is not very informative. Let's use `Enter a positive integer`.
2. I realized that `minimum_duration` can be greater than `maximum_duration`. We can prevent this in the custom validate function in the backend. In the frontend, we should detect this error and display it.

portal/src/locale-data/en.json
tung2744: @louischan-oursky I've changed this part in the later commit, because what we actually need is <1 but not 1 or 0, which is not what plural for.

portal/src/graphql/portal/LockoutSettings.tsx
tung2744: @louischan-oursky I am just trying to return something which the formatter will understand, it can be 0, none, null or anything other thing, as long as we handled it in the message.

portal/src/locale-data/en.json
tung2744: @louischan-oursky Why we need plural here?

portal/src/graphql/portal/LockoutSettings.tsx
tung2744: @louischan-oursky If we allow the user to type decimal places, the user can type something like "3.", which if we parse it into number it becomes `3`, and if we put it back into the TextField it becomes "3". As a result the user can never type ".". So, we should store the current input value as string.

louischan-oursky: > In fact I always think form text field value states should always be string, even it represents a number.

Agree.

portal/src/locale-data/en.json
louischan-oursky: When it is 1, it shows `1 minutes` which is not grammatically correct.

tung2744: Oh right. Let me add `plural` for all messages if needed.

tung2744:
> The error message invalid format is not very informative. Let's use Enter a positive integer.

Lets use `Please enter a positive value`, because in fact it can be non-integer in server-side validation.

tung2744:
@louischan-oursky Addressed all comment except the one about enable toggle. Thanks!

portal/src/error/parse.ts
```diff
@@ -91,13 +95,53 @@ function isKnownCauseKind(kind: string): boolean {
...
+    cause.details.error?.startsWith("non-positive duration")
+  ) {
+    return {
+      messageID: "errors.validation.format.duration.nonPositive",
+    };
+  }
+  return null;
+}
+
+function getGeneralErrorMessage(
```
louischan-oursky, line 110: Remove.

louischan-oursky (commented):
Per offline discussion, let's do the following:

1. Change server to emit a `maximum` error instead of `general` error
2. Remove specific error message in error/parse.ts
3. Update error/parse.ts to flavor error rules provided in components over builtin ones.
4. Use errorRules in component to specialize error message.

portal/src/error/parse.ts
```diff
@@ -91,13 +95,53 @@ function isKnownCauseKind(kind: string): boolean {
   return kind in errorCauseMessageIDs;
 }
 
+function getFormatErrorMessage(cause: FormatErrorCause): ParsedAPIError | null {
```
tung2744, line 98: @louischan-oursky 

For this error, actually it applies to all duration string in config. So I think can keep here instead of moving to component rules. What do you think?

pkg/lib/config/config.go
```diff
@@ -293,8 +293,10 @@ func (c *AppConfig) validateLockout(ctx *validation.Context) {
 	minDuration, isMinDurationValid := c.Authentication.Lockout.MinimumDuration.MaybeDuration()
 	maxDuration, isMaxDurationValid := c.Authentication.Lockout.MaximumDuration.MaybeDuration()
 	if isMaxDurationValid && isMinDurationValid && minDuration.Seconds() > maxDuration.Seconds() {
-		ctx.Child("authentication", "lockout", "maximum_duration").
-			EmitErrorMessage("maximum_duration must be greater than or equal to minimum_duration")
+		ctx.Child("authentication", "lockout", "minimum_duration").EmitError("maximum", map[string]interface{}{
+			"maximum": c.Authentication.Lockout.MaximumDuration,
```
louischan-oursky, line 297: Since `maximum` is a keyword for JSON number, the information it contains must also be number. Call `.Seconds()`.

portal/src/error/parse.ts
louischan-oursky: OK.

pkg/lib/config/config.go
tung2744: It seems a bit weird that I pass in a string "1m" and it tells me my input is `60` 🤔

louischan-oursky: It does not matter in this case, because the error message we are going to show is something like `Lockout Duration must be less than or equal to Maximum Lockout Duration`. 

The change is just for conforming the protocol of `maximum`.

tung2744: Yes I understand. However it is still an error returned from the api, which the error message doesn't match the input.
Nvm, let me change it in this way first.

tung2744:

louischan-oursky:
This change now looks very good. Thank you!

PR #3173, opened 2023-06-28
Add workflow JSON schema

louischan-oursky:
ref #3144

### DIFF
--- a/pkg/lib/config/config.go
+++ b/pkg/lib/config/config.go
@@ -45,7 +45,8 @@ var _ = Schema.Add("AppConfig", `
 		"web3": { "$ref": "#/$defs/Web3Config" },
 		"google_tag_manager": { "$ref": "#/$defs/GoogleTagManagerConfig" },
 		"account_migration": { "$ref": "#/$defs/AccountMigrationConfig" },
-		"captcha": { "$ref": "#/$defs/CaptchaConfig" }
+		"captcha": { "$ref": "#/$defs/CaptchaConfig" },
+		"workflow": { "$ref": "#/$defs/WorkflowConfig" }
 	},
 	"required": ["id", "http"]
 }
--- a/pkg/lib/config/config.go
+++ b/pkg/lib/config/config.go
@@ -83,6 +83,8 @@ type AppConfig struct {
 	AccountMigration *AccountMigrationConfig `json:"account_migration,omitempty"`
 
 	Captcha *CaptchaConfig `json:"captcha,omitempty"`
+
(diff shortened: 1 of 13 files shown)


pkg/lib/config/workflow.go
```diff
@@ -0,0 +1,597 @@
...
+	WorkflowLoginFlowStepTypeIdentify     WorkflowLoginFlowStepType = "identify"
+	WorkflowLoginFlowStepTypeAuthenticate WorkflowLoginFlowStepType = "authenticate"
+)
+
+type WorkflowLoginFlowStep struct {
+	ID   string                    `json:"id,omitempty"`
+	Type WorkflowLoginFlowStepType `json:"type,omitempty"`
+
+	// Optional is relevant when Type is authenticate.
+	Optional *bool `json:"optional,omitempty"`
```
tung2744, line 529: Maybe not related to this pr, but what actually is the purpose of having an optional authenticate step?

pkg/lib/config/config.go
```diff
@@ -45,7 +45,8 @@ var _ = Schema.Add("AppConfig", `
 		"web3": { "$ref": "#/$defs/Web3Config" },
 		"google_tag_manager": { "$ref": "#/$defs/GoogleTagManagerConfig" },
 		"account_migration": { "$ref": "#/$defs/AccountMigrationConfig" },
-		"captcha": { "$ref": "#/$defs/CaptchaConfig" }
+		"captcha": { "$ref": "#/$defs/CaptchaConfig" },
+		"workflow": { "$ref": "#/$defs/WorkflowConfig" }
```
tung2744, line 49: I am not sure should we have another file for workflow config instead of mixing it with existing configs.

pkg/lib/config/workflow.go
louischan-oursky: When `optional` is true, and the user does not have any applicable authentication method listed in `one_of`, then this authentication step is skipped. This allows us to represent optional 2FA step. When the user has configured 2FA, then the step must be executed.

pkg/lib/config/config.go
louischan-oursky: I have put changes in pkg/lib/config/workflow.go. Or do you prefer something else?

tung2744: I meant the config file. Currently all configs are in authgear.yaml. However maybe workflow configs are complicated enough to have its own file?

louischan-oursky: We have three already. I prefer not adding one more.

Issue #3174, opened 2023-06-28
Execute SignupFlow


louischan-oursky:
- [x] Track workflow progress with json pointer
- [x] Advance workflow progress with json pointer
- [x] Determine the end of workflow with json pointer
- [ ] type: identify
  - [x] login id
  - [ ] oauth
  - [ ] passkey
  - [ ] siwe
- [x] type: verify
  - [x] verify identity
  - [x] verify oob otp authenticator
- [ ] type: authenticate
  - [x] create password
  - [x] create oob otp authenticator
  - [x] create totp authenticator
  - [ ] create passkey authenticator
- [x] generate and view recovery code
- [x] type: user_profile
- [x] Perform necessary checking and side effects at the end of the flow

louischan-oursky:
Remaining items will be continued in https://github.com/authgear/authgear-server/issues/3273

PR #3196, opened 2023-07-07
Implement SignupFlow

louischan-oursky:
ref #3174


tung2744:
@louischan-oursky Is there any suggested ways to test this pr?

louischan-oursky:
1. Add a signup flow config
2. Use postman to drive the flow.
[SignupFlow.postman_collection.json.zip](https://github.com/authgear/authgear-server/files/12077564/SignupFlow.postman_collection.json.zip) Since instance ID keeps changing, I have added a test script in postman to auto-update the variable.

pkg/lib/workflow/workflowconfig/intent_signup_flow_step_authenticate.go
```diff
@@ -0,0 +1,247 @@
...
+			switch authentication {
+			case config.WorkflowAuthenticationMethodPrimaryPassword:
+				fallthrough
+			case config.WorkflowAuthenticationMethodSecondaryPassword:
+				return workflow.NewNodeSimple(&NodeCreateAuthenticatorPassword{
+					UserID:         i.UserID,
+					Authentication: authentication,
+				}), nil
+			case config.WorkflowAuthenticationMethodPrimaryPasskey:
+				// FIXME(workflow): create primary passkey
```
tung2744, line 135: Are we going to implement this later?

pkg/lib/workflow/workflowconfig/intent_signup_flow.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	// The list of nodes looks like
+	// 1 NodeDoCreateUser
+	// N Nodes for each N steps
+	// 1 IntentCreateSession
+	// So at the end of the flow, it will have 2+N nodes.
+	if len(w.Nodes) >= 2+len(f.Steps) {
+		return nil, workflow.ErrEOF
+	}
```
tung2744: It looks like this kind of assumptions can be easily broken, What if we check if the last node is `NodeDoCreateSession`?

louischan-oursky: OK. But checking the type of the last node could also be fragile, for example, if we later on refactor to use IntentCreateSession instead of NodeDoCreateSession. The last node would a subworkflow with the intent IntentCreateSession.

Issue #3098, opened 2023-05-16
Remove old rate limits config after new rate limit is deployed


fungc-io:

louischan-oursky:
We can schedule this task now.

fungc-io:
same as #3066?

louischan-oursky:
No. This issue is about removing the deprecated configuration fields from the configuration file schema. After this change, any configuration file still containing the removed fields will be considered as invalid.

tung2744:
We will run the migration script again after this cooldown, and observe if any invalid configs exist in next cooldown.
- [ ] Staging
- [ ] Production

tung2744:
Checked in staging and there are still deprecated configs, need to check again where it comes from..

PR #3202, opened 2023-07-12
Add checker for config source integrity

kiootic:
`go run ./cmd/portal internal check config-sources`

ref #3098

### DIFF
--- a/cmd/portal/cmd/cmdinternal/check.go
+++ b/cmd/portal/cmd/cmdinternal/check.go
@@ -1,0 +1,36 @@
+package cmdinternal
+
+import (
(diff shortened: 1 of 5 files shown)


louischan-oursky:
@kiootic I remember you told me you have run this command against the staging server, and found some invalid configs. Did you investigate into that? Any findings?

kiootic:
Some of the configs contain the deprecated field `/authenticator/password/ratelimit`, despite we already run the migration before. However, the current codebase does not set/update the deprecated field, so it's not known how it is still present.

louischan-oursky:
I run again and did not find any invalid configs though. Did you run the migration again?

louischan-oursky:
Oh, I did not apply your another PR first. Sorry.

PR #3203, opened 2023-07-12
Remove deprecated config

kiootic:
ref #3098
requires #3202


tung2744:
@louischan-oursky Updated the pr so that it won't remove deprecated fields.

Issue #3220, opened 2023-07-18
Improve Account Lockout countdown display


fungc-io:
re: #3129

The countdown will show a round down number in the message.
When they need to wait 2 minutes, it always show "1 min" in the first minute. 

Either:
- Show the seconds, e.g. `1 minutes 59 seconds`
- Round up to minutes if time >= 1 min

tung2744:
@fungc-io 

https://drive.google.com/file/d/1B2eg1ffcjQ5yTt9LzI7QGCtNea9QlVUP/view?usp=sharing
Pease see if this looks good?

fungc-io:
@tung2744 
looks good, it'd be better if the zh translation format can be improved.
`請於 1 分鐘及25 秒後再試` -> `請於 1 分鐘 25 秒後再試`

- Remove `及`
- Space on both sides of the number `X 25 X`

tung2744:
@fungc-io It looks like this:
<img width="1315" alt="Screenshot 2023-07-18 at 10 00 37 PM" src="https://github.com/authgear/authgear-server/assets/34942210/f11dfd46-8037-4670-8779-087a1aeb3e8d">
<img width="1318" alt="Screenshot 2023-07-18 at 9 58 38 PM" src="https://github.com/authgear/authgear-server/assets/34942210/8a2e91bc-6fa1-42f1-8dd1-f005cc893fb0">

fungc-io:
tested ok

PR #3222, opened 2023-07-18
Use luxon duration to format countdown numbers

tung2744:
ref #3220


louischan-oursky:
@tung2744 I addressed the comment in https://github.com/authgear/authgear-server/issues/3220#issuecomment-1640310329 and fixed the compile error

Issue #3213, opened 2023-07-17
Introduce authgear-apps.com for default project domain


fungc-io:
## Background

The original domain is wrongly classified as dangerous by Google SafeBrowser. To maintain the service, we are moving to `authgear-apps.com`.

## Solution

- Existing projects will keep their domain by default
  - new entry for *.authgear-apps.com will be added
  - Users should activate the new domain manually
- New app will use authgear-apps.com by default
- Register authgear-apps.com in Google Search Console
- Submit the authgear-apps.com to Public Suffix List

**Must register in search console before submitting to public suffix list**

## No-goes

No need multi domain support

louischan-oursky:
> Register authgear-apps.com in Google Search Console

What should I do specifically? I only see authgear.com (no authgearapps.com) on Google Search Console

louischan-oursky:
- [ ] Add a A record for `*.authgear-apps.com`
- [ ] Add a A record for `.authgear-apps.com`
- [ ] TODO

fungc-io:
done 🎉 !

PR #3228, opened 2023-07-19
Support multiple default domains

louischan-oursky:
ref #3213


cmd/portal/internal/create_default_domain.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return
+}
+
+func makeVerificationNonce() string {
+	nonce := make([]byte, 16)
+	corerand.SecureRand.Read(nonce)
+	verificationNonce := hex.EncodeToString(nonce)
+	return verificationNonce
+}
```
tung2744, line 72: Looks like this code was duplicated from `pkg/portal/service/domain.go`. Maybe we should create a util function for this to ensure their logics are same?

tung2744 (approved):

Issue #3236, opened 2023-07-21
Allow phone login in portal for new project


tung2744:

fungc-io:
tested ok

PR #3241, opened 2023-07-21
Allow choosing phone related options in project setup wizard

tung2744:
ref #3236

### DIFF
--- a/portal/src/graphql/portal/ProjectWizardScreen.tsx
+++ b/portal/src/graphql/portal/ProjectWizardScreen.tsx
@@ -214,6 +214,13 @@ function Step1(props: StepProps) {
     };
   }, [oobOTPEmailTooltip.targetElement]);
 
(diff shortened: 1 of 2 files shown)


louischan-oursky:
@fungc-io @tung2744 I have added a few commits trying to make some improvements.

- I re-ordered the options in step 1 so that the 2 password variants come first, the 2 passwordless variants come later.
- I changed the copywriting of the options in step 1 to make them more consistent.
- I make step 3 question 3 aware of the new option in step 1. In particular, when the variant phone_password is selected, we only show the oob_otp_email variant as the second factor. Not sure if we want to support oob_otp_sms as the second factor as well.
- As we already have a tooltip for oob_otp_email in step 1, I added a tooltip for oob_otp_sms in step 1 as well.

Take a look at the video to see what I mean.

https://github.com/authgear/authgear-server/assets/15082564/08f2fc3a-b608-4028-af8c-3f36ec2991cc

Issue #3267, opened 2023-07-28
[Portal] Do not require reauth when the user cannot reauth


louischan-oursky:
For example, if the user has only Google Sign-in, they cannot reauth. Allow them to do sensitive actions with reauth in such case.

fungc-io:
deployed

PR #3271, opened 2023-07-30
Fix reauthentication on portal

louischan-oursky:
ref #3267

### DIFF
--- a/portal/src/InternalRedirect.tsx
+++ b/portal/src/InternalRedirect.tsx
@@ -1,0 +1,23 @@
+import React, { useEffect } from "react";
+import { useNavigate, useLocation } from "react-router-dom";
+
+export interface InternalRedirectState {
+  originalPath: string;
+  state: unknown;
+}
+
+export default function InternalRedirect(): React.ReactElement | null {
+  const location = useLocation();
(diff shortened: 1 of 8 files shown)


tung2744 (commented):
This pr modified frontend only, but I think user is still unable to visit the secret because he is unable to generate the visit token?
https://github.com/authgear/authgear-server/blob/main/pkg/portal/service/app.go#L236

louischan-oursky:
```
	if !(elapsed >= 0 && elapsed < 5*time.Minute || !sessionInfo.UserCanReauthenticate) {
		return nil, ErrReauthRequrired
	}
```

- `elapsed >= 0 && elapsed < 5*time.Minute` means authenticated recently
- `!sessionInfo.UserCanReauthenticate` means cannot re-authenticate
- `!(a || b)` means `!a && !b`
- So overall the if condition means **NOT** authenticated recently **AND** can re-authenticate
- Therefore, those who cannot re-authenticate can always create the token.

tung2744:
Oh you are right, I missed the ! before the two conditions.

louischan-oursky:
That condition is really hard to read. It is not your fault :)

Issue #3231, opened 2023-07-20
Support app2app configuration in portal


tung2744:
![Image](https://github.com/authgear/authgear-server/assets/34942210/c0752ea0-6480-4b6f-8e73-6b48e686193c)

@fungc-io I suggest to update the portal as shown in the image, please see if any thoughts?

fungc-io:
I think App2App login should belong to "Authentication" rather than "Application" when the user finding for it.

<img width="330" alt="image" src="https://github.com/authgear/authgear-server/assets/15207470/882d6350-b3da-444d-979a-e33bb798fe3a">

tung2744:
@fungc-io 

1. It is a per application config
2. It requires other settings in application to work, such as redirect uri
3. It seems "Authentication" controls authentication methods, such as otp / password. In developer perspective app2app is not an authentication method but about how applications share session with each other.

louischan-oursky:
> It seems "Authentication" controls authentication methods, such as otp / password. In developer perspective app2app is not an authentication method but about how applications share session with each other.

If we put "App2App" under Authentication, I guess the developer will find it easier. On that page we can briefly explain what it is, and direct the developer to configure app2app in the Applications screen.

I just realize we lack the terms to refer to "App A" and "App B". In the spec, I see Tung used the term "the authenticating client" to refer to "App B", but we do not have a term for "App B".

louischan-oursky:
> I suggest to put this 2 config components into feature flag and treat it as experimental. Only enable if the client request it or later when the feature is mature.

Agree.

tung2744:
One point to note: I think we only display app2app options for native apps. Because app2app is not supported in other platforms.

tung2744:
Some suggested changes in the copy writings:

Description of app2app enable toggle:

`Allow this application to handle App2App requests from other applications.`

Because it actually doesn't affect requests **from** this application, it only allows handling app2app request **to** this application.

Description of migration:

`App2App login requires a refresh token generated after App2App login enabled. Enable Migration mode will allow old refresh tokens to participate in app2app login. This mode is less secure and not recommended.`

It seems device key is implementation detail and user / developer have no idea about it. So I updated the wordings a bit to hide it from the description.

@fungc-io See if they are ok?

fungc-io:
looks good on staging

PR #3308, opened 2023-08-14
Support configuring app2app in portal

tung2744:
ref #3231

Screenshot 2023-08-14 at 2 56 46 PM
Screenshot 2023-08-14 at 2 56 37 PM

@fungc-io Please see if the ui looks good. I cannot find a suitable color in the current figma for some colors such as the status green and grey colors, so I just randomly pick one.


fungc-io:
@tung2744 please use the same green and blue from the custom domain page
image

otherwise lgtm

pkg/lib/config/feature_oauth.go
```diff
@@ -20,14 +20,16 @@ var _ = FeatureConfigSchema.Add("OAuthClientFeatureConfig", `
 	"additionalProperties": false,
 	"properties": {
 		"maximum": { "type": "integer" },
-		"custom_ui_enabled": { "type": "boolean" }
+		"custom_ui_enabled": { "type": "boolean" },
+		"app2app_enabled": { "type": "boolean" }
```
louischan-oursky, line 24: The server should also respect this config. If app2app_enabled is false, then app2app request should fail with App2AppDisabled error.

louischan-oursky:
The setup link in app2app screen should link to the app2app section directly. It is a bit confusing that it just directs me to the edit client screen. I need to scroll down manually.

tung2744:

@fungc-io This is the updated colors.
Screenshot 2023-08-16 at 12 51 51 PM

PR #3311, opened 2023-08-14
Create a new intent to support forgot password by email by phone number input

tung2744:


pkg/latte/intent_login.go
```diff
@@ -148,7 +149,30 @@ func (i *IntentLogin) GetEffects(ctx context.Context, deps *workflow.Dependencie
...
 func (i *IntentLogin) OutputData(ctx context.Context, deps *workflow.Dependencies, w *workflow.Workflow) (interface{}, error) {
-	return nil, nil
+	verifiedAuthns, err := i.getVerifiedAuthenticators(w)
+	if err != nil {
+		return nil, err
+	}
+	if len(verifiedAuthns) == 0 {
+		return nil, nil
+	}
+	userID := verifiedAuthns[0].UserID
```
louischan-oursky, line 159: Why do we just use `i.Identity.UserID`?

pkg/latte/intent_login.go
```diff
@@ -148,7 +149,30 @@ func (i *IntentLogin) GetEffects(ctx context.Context, deps *workflow.Dependencie
...
+	identities, err := deps.Identities.ListByUser(userID)
```
louischan-oursky, line 160: I think we can just get the standard attribute `email` and mask it, include it in the data as `masked_email`.

pkg/latte/intent_login.go
tung2744: I think they are the same, but we should check `len(verifiedAuthns)` to ensure the user passed at least one step before including the email in response.

pkg/latte/intent_login.go
tung2744: I am not sure if it is safe to assume the email in standard attribute is same as the email identity. As forgot password only works for login ids, I get them from login ids.

tung2744:
Added a commit to expose ui params in sms data.

PR #3322, opened 2023-08-28
Implement reauth workflow in latte

tung2744:


pkg/latte/intent_reauthenticate.go
```diff
@@ -0,0 +1,88 @@
...
+	userID, err := i.userID(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	switch len(w.Nodes) {
+	case 0:
+		if err != nil {
+			return nil, err
+		}
```
louischan-oursky, line 56: This err checking block is redundant.

pkg/latte/intent_reauthenticate.go
```diff
@@ -0,0 +1,88 @@
...
+		return workflow.NewSubWorkflow(&IntentAuthenticatePassword{
+			UserID:            userID,
+			AuthenticatorKind: authenticator.KindPrimary,
+		}), nil
+	case 1:
+		return workflow.NewSubWorkflow(&IntentCreateSession{
```
louischan-oursky, line 62: IntentCreateSession actually does not implement the logic for reauthenticate. Please refer to do_ensure_session.go to add the missing logic.

tung2744: @louischan-oursky I am not sure if I understand it correctly:
In `do_ensure_session`, if mode is `EnsureSessionModeNoop`, basically nothing will be done except updating the login time.
And the condition of using `EnsureSessionModeNoop` is `SuppressIDPSessionCookie`, which means `x_sso_enabled` is not true.
In latte case, `x_sso_enabled` should be always false, therefore actually we only need to update the login time?

tung2744: In fact, it seems `IntentCreateSession` already handles reauth correctly, as it will skip session creation using `SkipCreate` by checking `SuppressIDPSessionCookie`.

louischan-oursky: Oh, you're correct. I am sorry that I overlooked in latte case SuppressIDPSessionCookie is always true.

pkg/latte/intent_reauthenticate.go
tung2744: Handled in [ae3db4b](https://github.com/authgear/authgear-server/pull/3322/commits/ae3db4b3079c7174b8f18f6c7dd42ad2603d9f1d)

pkg/latte/intent_reauthenticate.go
tung2744: I refactored the code in latte so that it looks more similar to the original workflow in 20fee68

louischan-oursky: Thank you

Issue #3353, opened 2023-09-12
Return Client ID in Session Node


fungc-io:
**Problem**

- In the Session node, it only contains the following info
  - `createdAt`
  - `createdByIP`
  - `displayName`
  - `lastAccessedAt`
  - `lastAccessedByIP`
  - `type`: `IDP` or `OFFLINE_GRANT`
  - `updatedAt`

If there are multiple sessions from the same device, say multiple apps, the developer is unable to identify which sessions correspond to which app. So they cannot see which app are logged in (have valid tokens) and it's hard to revoke the specific token related to the app.

**Appetite**

2 weeks

**Solution**

- Suggest to show the **Client ID** of the session so the dev knows which app it is.
- For `type = "IDP"`, client ID should be `null` because the session is for SSO, i.e. it's a device-wide session, it's not specific to a certain client ID.

**Rabbit holes**
> Details about the solution worth calling out to avoid problems

**No-goes**
> Functionality or use cases we intentionally aren’t covering to fit the appetite

louischan-oursky:
Remember the type of clientID should be `*string`.

tung2744:
@fungc-io Do we need to add client id in portal session list too?

tung2744:
- Currently settings screen grouped sso session together
- Portal lists session from user query as is

- Rewrite the query in portal such that it groups the session in the same way in settings

fungc-io:
@tung2744 yes please

fungc-io:
@tung2744 I found a problem


![Image](https://github.com/authgear/authgear-server/assets/15207470/f6b141d7-c50a-485a-872c-12f0162332e1)

The default length of our random generated client ID is truncated in the table, which is missing the last 3 characters, and it looks weird,
Can we adjust the width so it is showing the full client id?

fungc-io:
checked ok

PR #3374, opened 2023-09-20
Include client id in session node

tung2744:
ref #3353 
Screenshot 2023-09-20 at 7 39 17 PM


louischan-oursky:
Looks good to me. See if @fungc-io has any comments.

fungc-io:
yes this looks good 👍

Issue #3275, opened 2023-07-31
Execute SignupLoginFlow


louischan-oursky:
To be filled in

fungc-io:
Move to the next cycle because of other priorities.

PR #3375, opened 2023-09-20
Support SignupLoginFlow

louischan-oursky:
ref #3275


pkg/auth/handler/api/authenticationflow_v1_create.go
```diff
@@ -15,35 +15,31 @@ import (
...
-			"type": "array",
-			"items": {
-				"type": "object"
-			}
-		}
-	}
+var AuthenticationFlowV1NonRestfulCreateRequestSchema *validation.SimpleSchema
+
+func init() {
+	b := validation.SchemaBuilder{}.Type(validation.TypeObject)
```
tung2744, line 29: Why we need to construct the schema in init?

louischan-oursky: Because I need to build the schema, which references authflow.AllFlowTypes.

Issue #3286, opened 2023-08-03
Add tester app for trying out the login box


fungc-io:
## Todo

- Allow trigger "Try it out" from the portal
  - Serve a tester application at `[authgear-endpoint]/tester`
    - The tester app cannot trigger authentication by itself
    - The tester app offers a callback url to perform code exchange
    - It will show:
      - user info of the logged in enduser
      - a button to open the AuthUI Settings page
      - A button to go back to Authgear portal

---

- [ ] Always allow {authgear}/tester as a redirect uri.
- [x] Add a portal mutation, which on called, generates a token in redis `app:{app_id}:test:{token}`.
- [x] Add a "Try it out" out button, on triggers, call the above mutation, and navigate to `{authgear}/tester?token={generated_token}&portal_url={current_location}`
- [ ] Add a handler to handle `/tester`, which
    - [ ] If `token` exist in query, consume the token in redis. Redirect to authorization endpoint. Encode `portal_url` in `state`.
    - [ ] Else, if `code` exist in query, consumes `code` and display user info.
        - [ ] Display settings page button.
        - [ ] Display back to portal button, using `state.portal_url`.
    - [ ] Show an error if the above steps failed
    - [ ] Else, show a not found screen. (@fungc-io See what we want to show if someone enter `/tester` directly?)

louischan-oursky:
1. I suggest we rename `portal_url` to `return_uri`.
2. We only allow `return_uri` to begin with the portal origin (configured in environment variable), and with a path prefix of the same project (i.e. `/project/{encoded_project_id}`). This mitigates open redirection.
3. What client_id is used in this flow? `client_id=` (the empty client id?) Maybe we should specify when `client_id=`, a synthetic client is used, the only redirect_uri is `origin/tester`.

tung2744:
@fungc-io Will there be a design for the "tester" screen?

tung2744:
@louischan-oursky After some thought, I think we can simply store `return_uri` with the test token in redis, so it cannot be changed by the query, and don't have to worry about whitelisting.

fungc-io:
@tung2744 sorry let me draft one today

fungc-io:
@tung2744 Let's use this notion page for the design:

https://www.notion.so/oursky/Login-Success-5c3074df9107422f95f75354ac748015?pvs=4

Like this:
[...]

PR #3380, opened 2023-09-21
Update styles of tester user info section

tung2744:
ref #3286 
@fungc-io Please see this updated ui

Screenshot 2023-09-21 at 5 58 50 PM


fungc-io:
ok that looks good 👍

Issue #3392, opened 2023-10-10
Password strength in passowrd setup UX issue


fungc-io:
The UX of password setup page is strange because it's unclear what strength of password is needed.
Please make the following changes:

1. In the requirement section, Instead of showing the bar indicator, only show the strength requirement in text. Only normal or error state, no need for diff colors for diff strength level.

2. Since the policy is also checked in frontend, If the overall requirements are not met, disable the "Next" button and do not allow the user to proceed.

<img width="491" alt="image" src="https://github.com/authgear/authgear-server/assets/15207470/2fcfa6f8-9260-40b4-9f11-5d4817b80ff3">

fungc-io:
@tung2744 lol
thanks for pointing it out "You new password must be Very unguessable" sounds very absurd too.

I've just quickly check the UX of common design, usually the strength requirement is not shown in the policy list.
Usually just a note saying "Please use a stronger password" below the password input

Let's do this:
<img width="381" alt="image" src="https://github.com/authgear/authgear-server/assets/15207470/2341e406-86f8-4adc-b2d0-bd2e8c8c814d">

louischan-oursky:
I suggest we change the tone of the sentence so that it can always be there. If the tone is like "Please use a strong password", that sounds like an error message, which should only appear when the password is weak. But the strength indicator is real time. When the end-user is typing the password, flickering will occur.



I suggest we change it to something like "Password is strong enough"

https://github.com/authgear/authgear-server/assets/15082564/8366df40-3ce4-4adc-bc65-020eb488b582

fungc-io:
"Please use a strong password" is shown only when the requirement is not meet. And no need to show the message if the pw is strong enough. To avoid flickering, we can keep it in one line, or reserve a margin for the line
- Default don't show text when pw is empty: ` `
- `Strength: Fair - Please use a stronger password` (in red)
- `Strength: Unguessable` (in green)

Reference from the design in Woocommerce for min strength requirement - 

GIF taken from: https://rudrastyh.com/woocommerce/password-strength-meter.html
![password-strength-meter-woocommerce](https://github.com/authgear/authgear-server/assets/15207470/41e574d4-4da1-4fc9-a49a-4e6304b3602f)

fungc-io:
@calumhale can you help create the flow and design in Figma for this change?

fungc-io:
Design:
[...]

PR #3393, opened 2023-10-11
Improve the UX of creating new password

louischan-oursky:
ref #3392 

https://github.com/authgear/authgear-server/assets/15082564/8ce3d992-62ee-4874-9276-922ebffe3700


louischan-oursky:
@fungc-io See if you are satisfied with this improvement.

Issue #3280, opened 2023-07-31
Replace interaction with workflow


louischan-oursky:

fungc-io:
Need to follow up by if no authenticators is available in the middle of steps:
- [ ] https://github.com/authgear/authgear-server/issues/3346
- [ ] No 2FA set but 2FA is mandatory
- [ ] explore if any other cases

PR #3398, opened 2023-10-16
Use authflow in authui

louischan-oursky:
ref #3280


pkg/auth/routes.go
```diff
@@ -248,10 +248,28 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource) *h
 
 	router.Add(webapphandler.ConfigureRootRoute(webappAuthEntrypointRoute), p.Handler(newWebAppRootHandler))
 	router.Add(webapphandler.ConfigureOAuthEntrypointRoute(webappAuthEntrypointRoute), p.Handler(newWebAppOAuthEntrypointHandler))
-	router.Add(webapphandler.ConfigureLoginRoute(webappRequireAuthEnabledAuthEntrypointRoute), p.Handler(newWebAppLoginHandler))
-	router.Add(webapphandler.ConfigureSignupRoute(webappRequireAuthEnabledAuthEntrypointRoute), p.Handler(newWebAppSignupHandler))
+	//router.Add(webapphandler.ConfigureLoginRoute(webappRequireAuthEnabledAuthEntrypointRoute), p.Handler(newWebAppLoginHandler))
+	//router.Add(webapphandler.ConfigureSignupRoute(webappRequireAuthEnabledAuthEntrypointRoute), p.Handler(newWebAppSignupHandler))
```
tung2744, line 252: Is this intended?

louischan-oursky: Yes. I think we will not use the legacy interaction-based signup / login flow, so I did not make it configurable.

tung2744:
@louischan-oursky Although we are having unit tests already, I'm thinking about should we have some tests which execute a complete flow which we expected to support on http level 🤔
For example,
- Signup with email + password
- Signup with phone number + oobotp
- Login with phone number + oobotp
... Any other flows we can think of

What do you think?

tung2744:
I found a case which is a bit weird:
Current Behavior:

https://github.com/authgear/authgear-server/assets/34942210/e9db14dc-9aa9-49f9-83fb-13c39babd72d



Original Behavior:

https://github.com/authgear/authgear-server/assets/34942210/19573f6b-e6d4-4338-92f6-a30e064fd0db

My authentication settings:

```yaml
authentication:
  identities:
  - login_id
  - oauth
  - biometric
  - passkey
  lockout:
    backoff_factor: 2
    history_duration: 1440m
    lockout_type: per_user
    max_attempts: 10
    maximum_duration: 5s
    minimum_duration: 2s
    oob_otp:
      enabled: true
    password:
      enabled: true
  primary_authenticators:
  - oob_otp_sms
  - oob_otp_email
  - password
  - passkey
  rate_limits:
    oob_otp:
      email:
        trigger_cooldown: 60s
      sms:
        trigger_cooldown: 60s
  recovery_code:
    disabled: true
  secondary_authentication_mode: disabled
```

louischan-oursky:
> What do you think?

Agree. Let me add some tests.

> I found a case which is a bit weird

Is the email an existing account? I thought login link should only be used in login, not signup.

tung2744:
It is signup and the email is not used before.

louischan-oursky:
I added 2 commits to fix the case you mentioned.

tung2744:
For the test cases, lets create another issue to handle?

louischan-oursky:
https://github.com/authgear/authgear-server/issues/3401

tung2744:
@louischan-oursky There are still something different:
Screenshot 2023-10-17 at 3 16 18 PM

in original authui, I can choose to setup password, but now the options seems disappeared.

louischan-oursky:
I guess you turned off verification of email?

louischan-oursky:
I guess I found the cause. In SignupFlow, we use `step: verify` to verify the OOB OTP, while in LoginFlow, `step: authenticate` verifies the OTP. When we are at `step: verify`, we lose the branches so we do not see password as a branch.

louischan-oursky:
Added more commits to not use `step: verify` for creating authenticator.

louischan-oursky:
I edited the base branch to `authflow`.

tung2744:
@louischan-oursky There is still a different.

If we enabled `passkey` as primary authenticator in config, in original authui we will see setup passkey option, but now this option disappeared.
Original:
Screenshot 2023-10-18 at 12 03 07 PM
Now:
Screenshot 2023-10-18 at 12 04 06 PM

louischan-oursky:
Actually I intentionally deviate from the original behavior. A new account with only passkey is uncommon.

Issue #3402, opened 2023-10-17
Just-in-time create missing primary OOB OTP authenticator in authflow


louischan-oursky:
- [ ] As a special case, just-in-time create missing primary OOB OTP authenticator in login flow.

fungc-io:
Tested in the latest authflow and it magically works 👍

PR #3406, opened 2023-10-19
Just in time create primary OOB OTP authenticator in authflow

louischan-oursky:
ref #3402

### DIFF
--- a/pkg/lib/authenticationflow/declarative/util.go
+++ b/pkg/lib/authenticationflow/declarative/util.go
@@ -11,7 +11,6 @@ import (
 	authflow "github.com/authgear/authgear-server/pkg/lib/authenticationflow"
 	"github.com/authgear/authgear-server/pkg/lib/authn/authenticator"
 	"github.com/authgear/authgear-server/pkg/lib/authn/identity"
-	"github.com/authgear/authgear-server/pkg/lib/authn/mfa"
 	"github.com/authgear/authgear-server/pkg/lib/authn/otp"
 	"github.com/authgear/authgear-server/pkg/lib/authn/sso"
 	"github.com/authgear/authgear-server/pkg/lib/config"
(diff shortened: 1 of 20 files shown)


pkg/lib/authenticationflow/declarative/util.go
```diff
@@ -143,176 +142,215 @@ func flowRootObjectForSignupLoginFlow(deps *authflow.Dependencies, flowReference
...
-	addPrimaryPassword := func() {
-		count := len(as)
-		allUsable = append(allUsable, NewUseAuthenticationOptionPassword(
-			config.AuthenticationFlowAuthenticationPrimaryPassword,
-			count,
+func useAuthenticationOptionAddRecoveryCodes(options []UseAuthenticationOption, isOptional bool, userHasSomeSecondaryAuthenticators bool) []UseAuthenticationOption {
+	shouldAdd := false
+	switch {
+	case !isOptional:
+		// We always add recovery_code even though the end-user does not actually has any.
```
tung2744, line 230: Is it because we want the frontend to show the option even the user doesn't have recovery code?

tung2744: But seems some other options such as passkey only shows if user has one. Will it be a bit inconsistent?
Actually, is "always show some options" a ui logic which should not affect the api?

louischan-oursky: Let me update the code so that passkey is always shown when it is enabled.

> a ui logic which should not affect the api?

Authui now rely solely on the information of authflow to render. Authui cannot "peak" the configuration to something extra. Since primary_password is always shown, to be consistent with this behavior, I made all other options always shown as well (except passkey, which I will fix it).

louischan-oursky: Added two commits to address this comment.

Issue #3370, opened 2023-09-19
Execute ForgotPasswordFlow and ResetPasswordFlow


louischan-oursky:
- request_account_recovery_flows
    - [x] Modify config
    - [x] Generate config
    - [x] Implement intent

- account_recovery_flows
    - [x] Modify config
    - [x] Generate config
    - [x] Implement intent

tung2744:
```yaml

send_recovery_code_flows:
  # The current forgot password flow
  # 1. Take a email or phone, it won't tell the user if it is correct or not
  # 2. Send the code if the email or phone is valid
  - name: may_recover
    steps:
    - type: identify
      on_failure: ignore
      one_of:
      - identification: email
        steps:
        - type: send_recovery_code
          on_failure: ignore
          channel: email
      - identification: phone
        steps:
        - type: send_recovery_code
          on_failure: ignore
          channel: sms
  # Forgot password for reauth flow
  # 1. Take a id token and identify the user
  # 2. Take a channel (sms/email)
  # 3. Must send the code
  - name: must_recover
    steps:
    - type: identify
      on_failure: error
      one_of:
      - identification: id_token
        steps:
        - type: select_channel
        - type: send_recovery_code

recovery_flows:
  # Reset password flow
  - name: default
    steps:
      - type: input_recovery_code
      - type: reset_password

```

@louischan-oursky Please see the proposed design above. You can add comments here or have a huddle if you want to have a discussion. :adore:

louischan-oursky:
```yaml
request_account_recovery_flows:
- name: default
  steps:
  - type: identify
    # on_failure is moved to the branch.
    one_of:
    # Available identification are "email", "phone", and "id_token".
    - identification: email
      # on_failure is either "ignore" or "error"
      # If it is "ignore", do not report error and proceed.
      # If it is "error", report not found (i.e. allow account enumeration).
      on_failure: ignore
    - identification: phone
      on_failure: ignore
    - identification: id_token
      on_failure: ignore
  # This step outputs an array of options, where the options look like
  # [
  #   {"masked_display_name": "+852*****325", "channel": "sms", "otp_form": "code"},
  #   {"masked_display_name": "+852*****325", "channel": "whatsapp", "otp_form": "code"},
  #   {"masked_display_name": "lou******@oursky.com", "channel": "email", "otp_form": "link"}
  # ]
  - type: select_destination
  # No more steps because we have gathered enough information to deliver
  # the account recovery code.

account_recovery_flows:
[...]

PR #3416, opened 2023-10-24
Execute account recovery flow - request code

tung2744:
ref #3370


tung2744:
@louischan-oursky Ready for review, thanks!

pkg/lib/authenticationflow/declarative/intent_request_account_recovery_flow.go
```diff
@@ -0,0 +1,66 @@
...
+	case len(flows.Nearest.Nodes) == 0:
+		return authflow.NewSubFlow(&IntentRequestAccountRecoveryFlowSteps{
+			JSONPointer: i.JSONPointer,
+		}), nil
+	}
+
+	return nil, authflow.ErrIncompatibleInput
+}
+
+func (i *IntentRequestAccountRecoveryFlow) GetEffects(ctx context.Context, deps *authflow.Dependencies, flows authflow.Flows) ([]authflow.Effect, error) {
```
louischan-oursky: We can skip conforming to EffectGetter if we do not have any effects to perform.

pkg/lib/config/authentication_flow_test.go
```diff
@@ -193,3 +193,54 @@ reauth_flows:
...
+			err = json.Unmarshal([]byte(inputJSON), &input)
+			So(err, ShouldBeNil)
+
+			encodedCfg, err := json.Marshal(cfg)
+			So(err, ShouldBeNil)
+
+			encodedInput, err := json.Marshal(input)
+			So(err, ShouldBeNil)
+
+			// NOTE(tung): Disable the schema test for now because it was removed from config schema
```
louischan-oursky, line 221: Do we need to fix the notes in this file in this PR?

tung2744: @louischan-oursky If we don't make the flow configurable at the moment, I think we don't need to fix it.

Issue #3415, opened 2023-10-24
[Authflow] Remove id and add websocket_url


louischan-oursky:
- [x] Remove `id` from response
- [x] Add `websocket_url` when `otp_form` is `link`

PR #3417, opened 2023-10-25
Remove id and add websocket_url

louischan-oursky:
ref #3415


authui/src/authflow_websocket.ts
```diff
@@ -0,0 +1,95 @@
...
+    refreshPage();
+  };
+
+  connectWebSocket = () => {
+    const url = this.urlValue;
+
+    this.ws = new WebSocket(url);
+
+    this.ws.onopen = (e) => {
+      console.log("authflow_websocket onopen", e);
```
tung2744, line 45: Should we use console.info?

tung2744 (approved):

authui/src/authflow_websocket.ts
louischan-oursky: Let me change to use console.info

Issue #3276, opened 2023-07-31
Execute ReauthFlow


louischan-oursky:
- Re-authentication based on interaction supports only primary authenticator.

fungc-io:
Tested, with this issue created:
- [ ] https://github.com/authgear/authgear-server/issues/3469

PR #3432, opened 2023-10-30
Add ReauthFlow

louischan-oursky:
ref #3276

### DIFF
--- a/pkg/lib/authenticationflow/declarative/input_step_identify.go
+++ b/pkg/lib/authenticationflow/declarative/input_step_identify.go
@@ -41,6 +41,9 @@ func (i *InputSchemaStepIdentify) SchemaBuilder() validation.SchemaBuilder {
 		}
 
 		switch option.Identification {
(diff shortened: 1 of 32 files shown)


pkg/lib/authenticationflow/declarative/input_step_identify.go
```diff
@@ -96,6 +99,8 @@ func (i *InputSchemaStepIdentify) MakeInput(rawMessage json.RawMessage) (authflo
 type InputStepIdentify struct {
 	Identification config.AuthenticationFlowIdentification `json:"identification,omitempty"`
 
+	IDToken string `json:"id_token,omitempty"`
```
tung2744, line 102: It seems in reauth, the id token is from id_token_hint of the oauth session. I am not sure if it could be difficult for the ui to pass in id token as an input.

tung2744: Oh I saw id_token_hint was forwarded to ui in this pr.

Issue #3408, opened 2023-10-20
Integrating ForgotPasswordFlow and ResetPasswordFlow to AuthUI


fungc-io:
Continuing #3370

- [x] Create forgot password page
- [x] Create forgot password success page
- [x] Generate account recovery flow config according to identity config
- [x] Create reset password page
    - [x] Fix websocket auto refresh the page and causing flow not found
    - [x] Handle finish redirect
    - [x] Handle invalid code
- [x] Create reset password success page

fungc-io:
tested ok

PR #3433, opened 2023-10-30
Add authflow forgot password page

tung2744:
requires https://github.com/authgear/authgear-server/pull/3419
ref #3408 
https://github.com/authgear/authgear-server/assets/34942210/44659982-79de-4cdf-a443-f8dbc311cd87

### DIFF
--- a/pkg/auth/handler/webapp/authflow_controller.go
+++ b/pkg/auth/handler/webapp/authflow_controller.go
@@ -780,7 +780,7 @@ func (c *AuthflowController) checkPath(w http.ResponseWriter, r *http.Request, s
 	if u.Path != r.URL.Path {
 		// We do not know what causes the mismatch.
 		// Maybe x_step was tempered.
-		c.Logger.Error("path mismatch")
+		c.Logger.Warningln("path mismatch")
 		return webapp.ErrInvalidSession
 	}
 
--- a/pkg/auth/handler/webapp/authflow_controller.go
+++ b/pkg/auth/handler/webapp/authflow_controller.go
@@ -152,7 +152,6 @@ func (c *AuthflowController) HandleStartOfFlow(w http.ResponseWriter, r *http.Re
 		c.renderError(w, r, err)
 		return
 	}
-
 	handler := c.makeHTTPHandler(s, screen, handlers)
 	handler.ServeHTTP(w, r)
 }
@@ -781,6 +780,7 @@ func (c *AuthflowController) checkPath(w http.ResponseWriter, r *http.Request, s
 	if u.Path != r.URL.Path {
 		// We do not know what causes the mismatch.
 		// Maybe x_step was tempered.
+		c.Logger.Error("path mismatch")
 		return webapp.ErrInvalidSession
 	}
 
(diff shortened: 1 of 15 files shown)


pkg/auth/handler/webapp/authflow_controller.go
```diff
@@ -781,6 +780,7 @@ func (c *AuthflowController) checkPath(w http.ResponseWriter, r *http.Request, s
 	if u.Path != r.URL.Path {
 		// We do not know what causes the mismatch.
 		// Maybe x_step was tempered.
+		c.Logger.Error("path mismatch")
```
louischan-oursky, line 783: Maybe it should be a warning.

pkg/auth/handler/webapp/authflow_forgot_password_success.go
```diff
@@ -0,0 +1,52 @@
...
+	data := make(map[string]interface{})
+
+	baseViewModel := h.BaseViewModel.ViewModel(r, w)
+	viewmodels.Embed(data, baseViewModel)
+
+	return data, nil
+}
+
+func (h *AuthflowForgotPasswordSuccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var handlers AuthflowControllerHandlers
```
louischan-oursky, line 48: We do not actually need data from the screen. Maybe we can skip using AuthflowController here.

louischan-oursky (commented):
What does `x_from_step` actually do?

pkg/auth/handler/webapp/authflow_forgot_password_success.go
tung2744: Using `Controller.HandleStep` here will make sure the user see this page only if his account recovery flow is on the correct step.

tung2744:
> What does x_from_step actually do?

Originally I introduced it for restoring the previous state of the screen after finished the account recovery flow. However I found it is not needed because we always back to the login screen which start a new flow. So let me remove it.

pkg/auth/handler/webapp/authflow_enter_password.go
```diff
@@ -38,6 +38,7 @@ type AuthflowEnterPasswordViewModel struct {
 	PasswordManagerUsername string
 	ForgotPasswordInputType string
 	ForgotPasswordLoginID   string
+	CurrentXStep            string
```
louischan-oursky, line 41: CurrentXStep is unused. Please remove it.

louischan-oursky (commented):
1. The reset password link references `/flows/reset_password` which is interaction-based. I think we need to implement a authflow-based version of it.
2. Have you resolved the issues you mentioned in the last standup meeting?

tung2744:
> The reset password link references /flows/reset_password which is interaction-based. I think we need to implement a authflow-based version of it.

Yes, I will do it in next pr.

> Have you resolved the issues you mentioned in the last standup meeting?

Should be yes.
Websocket problem -> I disable it for all authflow based ui
Redirect URI problem -> Instead of relying on redirect URI, I added a mechanism to navigate to a defined path on flow end.

There will be another pr for reset password flow.

pkg/auth/handler/webapp/authflow_enter_password.go
louischan-oursky: @tung2744 In case you missed this.

tung2744: @louischan-oursky  Removed, thanks!

Issue #3436, opened 2023-10-31
grant_type app2app not allowed even app2app is enabled


fungc-io:

fungc-io:
tested ok

PR #3437, opened 2023-10-31
Fix app2app request grant type is not allowed even app2app enabled

tung2744:
ref #3436

### DIFF
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -54,6 +54,7 @@ const AppSessionTokenDuration = duration.Short
 var whitelistedGrantTypes = []string{
 	AnonymousRequestGrantType,
 	BiometricRequestGrantType,
+	App2AppRequestGrantType,
 	IDTokenGrantType,
 }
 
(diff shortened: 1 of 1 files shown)


pkg/lib/oauth/handler/handler_token.go
```diff
@@ -155,6 +155,9 @@ func (h *TokenHandler) doHandle(
 		allowedGrantTypes = []string{"authorization_code"}
 	}
 	allowedGrantTypes = append(allowedGrantTypes, whitelistedGrantTypes...)
+	if client.App2appEnabled {
+		allowedGrantTypes = append(allowedGrantTypes, App2AppRequestGrantType)
```
louischan-oursky, line 159: What about adding `App2AppRequestGrantType` to `whitelistedGrantTypes`?

tung2744: But for a app didn't enabled app2app, do we want to allow using app2app grant type?

louischan-oursky: We have done this for anonymous user and biometric login. I think we should follow.

Issue #3379, opened 2023-09-21
Allow all known origins of app to be redirect uri


tung2744:
~If the redirect uri is relative, it should be interpreted as `{{currentOrigin}}{{redirect_uri}}`.~

According to [oauth spec](https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2), it must be an absolute URI.

Maybe we could simply allow all known origins to be redirect uri.

Related to https://github.com/authgear/authgear-sdk-js/issues/248

louischan-oursky:
- Add all known origins of a project to `AppContext`
- Make sure database configsource construct and cache `AppContext` with origins correctly.
- May need to extract the redirect_uri checking logic into reusable code. In the code, respect all known origins. Make every place uses this code.

PR #3440, opened 2023-11-01
Allow all verified domains to be redirect uri

tung2744:
ref #3379


pkg/lib/oauth/handler/resolve.go
```diff
@@ -75,6 +75,14 @@ func validateRedirectURI(client *config.OAuthClientConfig, httpOrigin httputil.H
 		}
 	}
 
+	// Implicitly allow URIs for all whitelisted domains in https
+	for _, domain := range domainWhitelist {
+		origin := fmt.Sprintf("https://%s", domain)
```
louischan-oursky, line 80: Use the injectable`httputil.HTTPProto` here, instead of hard coding `https`.

tung2744: Updated, thanks

Issue #3443, opened 2023-11-02
`state` is only available in `user.pre_create` blocking event, but not `user.created` or `user.authenticated`


louischan-oursky:
**Describe the bug**
`state`, if provided by the developer, should appear in `user.created` and `user.authenticated` event.

**Authgear Version**
 - Version: SaaS

**To Reproduce**
1. Register a webhook for `user.pre_create`, `user.created` and `user.authenticated`
2. Provide `state`
3. Observe that only `user.pre_created` has `state` propagated.

**Expected behavior**
All three events have `state` if `state` was provided.

**Screenshots**
N/A

**Client Env (if applicable, please complete the following information):**
N/A

**Additional context**
N/A

fungc-io:
We may want to sanitize the state in the auditlog

fungc-io:
checked ok, i can see the state in user.authenticated

PR #3450, opened 2023-11-03
Include `state` in all events

louischan-oursky:
ref #3443


pkg/auth/webapp/ui_param.go
```diff
@@ -1,19 +1,21 @@
...
 	"github.com/authgear/authgear-server/pkg/util/intl"
 )
 
-// ClientIDCookieDef is a HTTP session cookie.
-var ClientIDCookieDef = &httputil.CookieDef{
-	NameSuffix: "client_id",
-	Path:       "/",
-	SameSite:   http.SameSiteNoneMode,
-}
+// ClientIDCookieDef is deprecated.
```
tung2744, line 13: Why don't we simply remove it? 🤔

louischan-oursky: Leaving the lines here to prevent us from re-adding it in the future. Also for easier `git blame`

tung2744:
@louischan-oursky Conflicted, please rebase :adore:

louischan-oursky:
Rebased.

Issue #3452, opened 2023-11-03
[AuthUI] Show "flow not found" error if proceed after going to forgot password page


joyzoursky:
## Environment
- Server: Staging
- App: qa-authflow
- Build: iOS TestFlight Authgear Demo iOS 1.0 (1698821396)
- Device: iPhone 15 (iOS 17.0.3)

## Description
If user has navigated to the forgot password page during login, then when click Next to proceed the login it will show error page.

## Steps to reproduce
1. Configure email + phone + password (or others)
2. Click "Authenticate" to start authentication
3. Login by email
4. Fill in email and see Enter Password screen
5. **Click "Forgot password"**
6. Go back
7. Fill in password and click "Next"
8. See "flow not found" error

## Expected result
Should be able to proceed normally after accessing forgot password page

## Video

https://github.com/authgear/authgear-server/assets/12539486/27bf429e-7fd8-4ec8-93a6-cbfe41973032

fungc-io:
@louischan-oursky please add me to qa-authflow so i can test it too,
thanks

fungc-io:
tested ok

PR #3457, opened 2023-11-06
Remember all screens from different flows

louischan-oursky:
ref #3452


tung2744 (approved):
LGTM. But one question: Originally the session only contains one flow, therefore `RedirectURI` of the session means the redirect uri after completing the flow.
As now there can be multiple flows in one session, it seems redirect uri is not very useful?

Issue #3445, opened 2023-11-02
[AuthUI] Web view does not anchor to top when error message is displayed


joyzoursky:
## Environment
- Server: Staging
- App: qa-authflow (Phone + Passwordless)
- Tested user agent:
  - iOS Native demo app
- Device: iPhone 15 (iOS 17.0.3)

## Description
When error message is shown on top of the web view, the page scroll position always stays at the current position, and cannot see the error message on top as it is blocked by the web view bar.

Should be an existing issue for a long time but I can't find an opening issue, so created a new one here.

## Steps to reproduce
1. Setup a project
2. Sign up a new user on iOS or Android app
3. Logout
4. Try to sign up with the same phone number or email
5. Looks like no response as there is actually an error message on top hidden behind the web view bar

Same problem occurs on other kinds of errors, e.g. wrong OTP.

## Expected result
If the error message is always on top, it should anchor to top position when there is error displayed, so user can see the error.

## Screenshot

https://github.com/authgear/authgear-server/assets/12539486/b53c079b-c4dc-4484-8055-a1b78ecdfcf2

<img width=47% src="https://github.com/authgear/authgear-server/assets/12539486/b9d5184d-deb4-40cd-958f-1486fc78128a" />

tung2744:
I can fix it in desktop by using sticky position, but seems difficult to be fixed in mobile. The mobile browser such as safari will make the page scrollable, and sticky is not working.

tung2744:
@fungc-io @joyzoursky 

Please see this video for the latest fix:

https://github.com/authgear/authgear-server/assets/34942210/78bf2d52-6ee1-42c3-a042-56cc9a6692f6

joyzoursky:
@tung2744 seems ok. What did you press at 0:07?
And double confirm keyboard will still show and auto-focus when first enter page (no top bar showing) right?

tung2744:
@joyzoursky  I pressed signup again.

And yes, the initial auto focus is still there.

joyzoursky:
I think we can merge first so we could start testing

fungc-io:
@tung2744 
thanks! 🙏  the video looks ok
Please merge it and we can test

fungc-io:
tested ok

PR #3467, opened 2023-11-07
Make error message bar stick to top

tung2744:
ref #3445

### DIFF
--- a/resources/authgear/templates/en/web/__message_bar.html
+++ b/resources/authgear/templates/en/web/__message_bar.html
@@ -1,5 +1,5 @@
 {{ define "__message_bar.html" }}
-<div class="absolute w-full">
+<div class="sticky top-0">
 
 {{ template "__error.html" . }}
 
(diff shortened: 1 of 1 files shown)


louischan-oursky:
We can use `fixed` instead of `sticky`. `fixed` works well on mobile Safari.

tung2744:
https://github.com/authgear/authgear-server/assets/34942210/7acfd6d1-d553-4ccb-a584-a0ead928398d

@louischan-oursky This is the effect of using fixed, it doesn't work too. The problem is the webview is scrolled by "something" outside the body element when keyboard is shown. All things in the screen will be scrolled.

fungc-io:
Seems like Webkit really hate it when we stick the elements on the top... 😞

tung2744:
https://github.com/authgear/authgear-server/assets/34942210/0bd61b60-99e6-46ba-a9eb-19515cb75e74

It seems even using position fixed, the message bar is still not always visible.

fungc-io:
Let's try make it not autofocus when error message is present

louischan-oursky:
I am closing this PR because we will go for another approach.

Issue #3430, opened 2023-10-29
Using HTTP for AuthUI and Authflow integration


fungc-io:
When releasing the new Authentication flow API, 
we have made the default Auth UI to employ the new set of API so we could:
1. unify the backend for custom and default AuthUI and reduce future maintenance effort
2. Test the API out

However the integration was done with internal functions, which is not an good representation of how the API would be used.
This hinder ourselves to accomplish the 2nd goal of this exercise.

Therefore it's suggested to change and implement the integration by actual HTTP calls.

Issue #3058, opened 2023-04-26
Possible deadlock if multiple request handled at the same time due to db connection exhausted in pool


tung2744:
**Describe the bug**
Currently, there are global db, app db, audit db. In a single request, we begin transactions on each db. All of these dbs shares the same connection pool.
In the connection pool, the connection pool cache is using database url as the cache key. Therefore, if global db / audit db / audit db share the same database url, they will share the same pool.
Therefore, it is possible that in a single request we opened multiple connections to the same database in the same database pool. And if the max connection limited of the pool was reached, it is possible to become a deadlock. 

**Authgear Version**
 - Version: fe4ea3092de0ab537dc6bfbc98e6fff3d869d855

**To Reproduce**
Trigger multiple graphql mutations together.

**Expected behavior**
Should not cause deadlock.
Maybe we should separate the database pool of global db, app db and audit db.

**Screenshots**
-

**Client Env (if applicable, please complete the following information):**
-

**Additional context**
-

chpapa:
> if global db / audit db / audit db share the same database url, they will share the same pool.

could we consider to fix this, by showing a warning at startup, if someone use the same DB url for the 3 DBs, and update our help page / setup with different pgsql users (and permissions) for each DB url

(Ironically as someone who suffer from pgsql connection pool cannot share between users in many ocassions, this is the first time I think that’s a feature not a bug lol)

louischan-oursky:
linear

PR #3482, opened 2023-11-11
Use Authflow HTTP API in authui

louischan-oursky:
ref #3430


pkg/lib/authenticationflow/authflowclient/client.go
```diff
@@ -0,0 +1,145 @@
...
+	// Only the port is important, the host is always the loopback address.
+	localEndpointURL.Host = fmt.Sprintf("127.0.0.1:%v", localEndpointURL.Port())
+
+	return &Client{
+		Context:       ctx,
+		HTTPClient:    httpClient,
+		LocalEndpoint: localEndpointURL,
+		HTTPHost:      httpHost,
+	}
+}
```
tung2744, line 47: I think calling self might cause a deadlock in the server, similar to https://github.com/authgear/authgear-server/issues/3058:

Assume the db connection pool have a max of 2. And now there are two incoming webapp requests at the same time.

1. Two requests start at the same time, and run util it calls a authflow api.
2. Now, two new authflow requests received, however, as the previous two webapp requests are not finished, the db connection pool is still exhausted and therefore no connections can be used to handle the authflow requests. As a result, they wait indefinitely.
3. Finally the requests should timeout.

And I am not sure how many threads are used to handle the requests, but I guess the thread pool will have the same problem.

tung2744: I checked the code again and seems we are only using database connections in authenticationflow.Service, and not using it in the authflow webapp 🤔  However we must be careful in the future to avoid such problem. Maybe we should consider deploying authflow api as a separated service?

louischan-oursky: Authflow-based webapp handlers do not acquire database connection. I used webapp.Session to store authflow state because I do not want to maintain yet another state in the webapp. I think those handlers, and authflowclient can be moved to a new package. The new package is a client implementation of the Authflow API, so database connection is forbidden. Making authflow API a separate service will complicate further the deployment, especially in case of non k8s deployment.

tung2744: ok. And I've also checked the concurrent requests golang http server can handle and it seems a large number, so maybe it is not a problem too. Will you move the packages in this pr too?

louischan-oursky: Moved the authflow client implementation to a new package.

tung2744:
@louischan-oursky Conflicted, and should we merge it into cooldown-8 first?

louischan-oursky:
> and should we merge it into cooldown-8 first?

yes

louischan-oursky:
rebased.

Issue #3338, opened 2023-09-04
Add cognitive complexity linting


tung2744:

tung2744:
The followings will not be fixed:
1. Code without tests.
2. Code will be removed in near future, e.g. interaction / old webapp.
3. Scripts.

PR #3486, opened 2023-11-14
Replace linter gocyclo with gocognit

tung2744:
ref #3338 

What have been done:

1. Turn on gocognit liniter
2. Ignore gocognit errors if:
    - The code do not have tests. We don't want to risk introducing bugs before next release.
    - The code will not be updated or maintained, and is going to be removed. e.g. latte, interaction, interaction webapp.
    - The code is one time use, e.g. scripts.
3. Fix remaining gocognit errors.
4. Turn off gocyclo linter.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ test:
 
 .PHONY: lint
 lint:
-	golangci-lint run ./cmd/... ./pkg/... --timeout 7m --max-issues-per-linter 0
+	golangci-lint run ./cmd/... ./pkg/... --timeout 7m
 	go run ./devtools/translationlinter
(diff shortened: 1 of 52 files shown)


Makefile
```diff
@@ -39,7 +39,7 @@ test:
 
 .PHONY: lint
 lint:
-	golangci-lint run ./cmd/... ./pkg/... --timeout 7m
+	golangci-lint run ./cmd/... ./pkg/... --timeout 7m --max-issues-per-linter 0
```
louischan-oursky, line 42: This flag should be set in the configuration file. https://golangci-lint.run/usage/configuration/

tung2744: Updated, thanks!

PR #3489, opened 2023-11-14
Describe login_hint in app2app spec

tung2744:
@louischan-oursky Please see if any comments on the design.

### DIFF
--- a/docs/specs/app2app.md
+++ b/docs/specs/app2app.md
@@ -162,6 +162,7 @@ The following methods will be added in android sdk to support the app2app flow:
   - This method should be called to trigger a new app2app authorization request through another client.
   - `App2AppAuthenticateOptions.authorizationEndpoint` should be an url of an applink pointing to another client which handles the app2app authentication request.
   - `App2AppAuthenticateOptions.redirectUri` should be an uri for the another client to return the authentication result. It must be an applink which opens the current app.
+  - `App2AppAuthenticateOptions.state`: Opaque value used to maintain state between the request and the callback. Same purpose as [state in oidc](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
 - `parseApp2AppAuthenticationRequest(uri: Uri): App2AppAuthenticateRequest?`
   - Parse an url into an `App2AppAuthenticateRequest`, or `null` if the uri is not a valid app2app request.
 - `approveApp2AppAuthenticationRequest(request: App2AppAuthenticateRequest)`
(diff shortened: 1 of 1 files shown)


docs/specs/app2app.md
```diff
@@ -56,12 +56,16 @@ An app can start the authentication flow by opening a link to another app, inste
...
+    - `refresh_token`: Required. A valid refresh token.
+    - `client_id`: Required. Current client id.
+    - `app2app_client_id`: Required. The authenticating client id.
+    - `jwt`: Required. A jwt with a challenge token obtained from the `/oauth2/challenge` api, signed with the private key of the device key bound to the refresh token. If `x_app2app_insecure_device_key_binding_enabled` is `true`, a jwt with a new device key will be allowed if there were no device key bound in the current offline grant. In this case the new key will be bound to the offline grant after the request.
+    - `app2app_redirect_uri`: Required. The redirect uri used to return the result used by the app. The server is reponsible to check the provided uri is in the whitelist of at least one of the configured client redirect uri.
+    - `login_hint`: Optional. When provided, it must be in the format described in [here](./oidc.md#login_hint). The server will check the corresponding user identity referenced by the `refresh_token`, and if it does not match the provided `login_hint`, a `unmatched_user` error will be returned. Only the following combinations of parameters are supported.
```
louischan-oursky, line 65: In the client's use case, App B actually can perform the checking locally by itself.

When the end-user has authenticated in App B
- App B is opened by the authorization URL. It see login_hint in the the query parameters. It then call the userinfo endpoint to fetch the standard attributes.
- If the phone_number does not match, App B return directly to App A.
- If the phone_number match, App B proceeds.

When the end-user has NOT authenticated in App B
- App B construct an authorization URL to include the login_hint.
- The end-user authenticated.
- App B calls the userinfo endpoint to check the login_hint.

The reason I am proposing this is because login_hint is originally a hint to the authorization server. In our original spec, we explicitly states that the login_hint will be ignored when it cannot be fulfilled. But in here, we are changing its meaning to be mandatory.

tung2744: I am trying to support login_hint in server is to ensure the checking logic is same in all sdks. But i agree that we should not change the meaning of login_hint at server. Let me  remove the parameter from token endpoint.

louischan-oursky: Per our offline discussion, instead of introducing login_hint, we will support `state` in app2app request. Its meaning is the same as in the authorization code flow. It will be included in the redirect URI to App A.

tung2744: Updated to add state only in the spec, thanks!

docs/specs/app2app.md
```diff
@@ -155,6 +155,7 @@ The following parameter will be added to constructor of `Authgear`:
   - `App2AppOptions` contains the following fields:
     - `isEnabled`: If true, new sessions will be prepared for participating in app2app authentication.
     - `isInsecureDeviceKeyBindingEnabled`: If true, the sdk will try to create and bind device key during `approveApp2AppAuthenticationRequest` to "migrate" to app2app sessions.
+    - `state`: Opaque value used to maintain state between the request and the callback. Same purpose as [state in oidc](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
```
louischan-oursky, line 158: state should be provided in App2AppAuthenticateOptions, not when Authgear is configured.

tung2744: Sorry I put it at a wrong place, fixed thanks!

PR #3490, opened 2023-11-17
Enable pprof and fix an issue

louischan-oursky:


pkg/util/httproute/httproute.go
```diff
@@ -85,6 +86,29 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	r.router.ServeHTTP(w, req)
 }
 
+func (r *Router) Pprof() {
```
tung2744, line 89: Should there be a config to disable it in production?

louischan-oursky: https://go.dev/doc/diagnostics

> Can I profile my production services?

> Yes. It is safe to profile programs in production, but enabling some profiles (e.g. the CPU profile) adds cost. You should expect to see performance downgrade. The performance penalty can be estimated by measuring the overhead of the profiler before turning it on in production.

The CPU profile is active only when we call /pprof/debug/profile https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/net/http/pprof/pprof.go;l=133 So it is OK to leave it enabled in production. But we need to make sure /pprof/debug/* is not reachable from the internet. Let me use another port for pprof.

louischan-oursky: Added original port + 10000 for internal use. Our helm chart is not aware of these new ports so they are not exposed to the internet. These ports are intended to be accessed with `kubectl port-forward` or something similar.

Issue #3488, opened 2023-11-14
Capture unknown error to Sentry and do not show in UI


fungc-io:
**Describe the bug**

From https://github.com/authgear/authgear-server/issues/3487
We see that the unknown error message are shown directly in the UI, which is not helpful for the users and may create huge frustration.

We should capture the error to Sentry and show a generic error message.

> ## Error
>
> An unknown error occurred. If the problem persists, please seek assistance in our  [Community Forum](https://github.com/orgs/authgear/discussions/categories/get-help) or contact us at [help@authgear.com](mailto:help@authgear.com)

PR #3493, opened 2023-11-20
Improve error reporting

louischan-oursky:
ref #3488

### DIFF
--- a/pkg/auth/handler/webapp/panic_middleware.go
+++ b/pkg/auth/handler/webapp/panic_middleware.go
@@ -47,7 +47,8 @@ func (m *PanicMiddleware) Handle(next http.Handler) http.Handler {
 		defer func() {
 			if e := recover(); e != nil {
 				err := panicutil.MakeError(e)
-				log.PanicValue(m.Logger.Logger, err)
+				m.Logger.WithError(err).Error("panic occurred")
+
 				apiError := apierrors.AsAPIError(err)
 
 				if !written {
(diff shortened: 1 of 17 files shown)


pkg/auth/handler/webapp/panic_middleware.go
```diff
@@ -51,13 +48,16 @@ func (m *PanicMiddleware) Handle(next http.Handler) http.Handler {
...
-					baseViewModel.SetError(err)
-					viewmodels.Embed(data, baseViewModel)
-					m.Renderer.RenderHTML(w, r, TemplateWebFatalErrorHTML, data)
+					result := &webapp.Result{
+						RedirectURI:      "/errors/error",
+						NavigationAction: "replace",
+					}
+					// This call cannot fail.
+					_ = m.ErrorCookie.SetNonRecoverableError(result, apiError)
+					result.WriteResponse(w, r)
```
tung2744, line 61: I think this could be bad in ux, usually user will try to refresh the same page if he saw an unexpected error.

louischan-oursky: Are you suggesting we should redirect to the same URL, and use recoverable error?

tung2744: 🤔 It seems originally we don't redirect and render `fatal_error.html` directly, why we need to change it?

tung2744: Ok, so what we assume here is panic will only occur on non-recoverable error such as programming error, and is impossible to recover by refreshing the page? If it is the case I think it is ok.

louischan-oursky: I/O error could also lead to panic. I changed back to show the error in the original path, the error is also stored in cookie instead of pinned to the query.

PR #3499, opened 2023-11-22
Add scripts for dump and restoring auth, portal, audit databases

tung2744:


louischan-oursky:
I remember you mentioned that restoring audit log is not supported. Is it still true?

tung2744:
Maybe I will create a issue to follow up the id conflict problem.

Issue #3501, opened 2023-11-23
Support reset password by sms otp - Config and Authflow


tung2744:
From: https://github.com/authgear/authgear-server/issues/3500

- [x] Add otp valid period config
- [x] Add otp templates
- [x] Support sending otp messages
- [x] Add forgot password channel & form configs
- [x] Update authflow configs
- [x] Update authflow api

PR #3506, opened 2023-11-27
Support reset password by otp in authflow

tung2744:
ref #3501


tung2744:
https://github.com/authgear/authgear-server/pull/3515 also fixed some bug in this pr, so lets review the whole feature in there.

Issue #3502, opened 2023-11-23
Support reset password by sms otp - Auth ui


tung2744:
From https://github.com/authgear/authgear-server/issues/3500

tung2744:
@fungc-io 

I found a problem at the flow end:



![Image](https://github.com/authgear/authgear-server/assets/34942210/f1ed0448-94ed-4e21-8e95-2297f90456f9)


The original reset password success page has no action button, the user will stuck in this page after resetting the password with otp.

Maybe we add a Sign In button, just like forgot password success page?



![Image](https://github.com/authgear/authgear-server/assets/34942210/92d7322b-9d8a-4a2b-95fe-358ee4b5b64e)

fungc-io:
yes, can we write "Back to sign in" on the button?

tung2744:
I found it could be confusing if the "Back to sign in" button always exist, because if user opened the reset password from link in other device, he shouldn't continue sign in in that device. So I think we should only show the button if we are resetting password in the same browser session.

fungc-io:
> So I think we should only show the button if we are resetting password in the same browser session.

👍

tung2744:
@fungc-io Fixed and deployed to staging. Please try again.

PR #3515, opened 2023-11-28
Support reset password by otp in auth ui & authflow

tung2744:
ref #3501 #3502

https://github.com/authgear/authgear-server/assets/34942210/d4c69bbd-3124-4d21-9857-f6379af0ff3d

https://github.com/authgear/authgear-server/assets/34942210/a444d577-5306-45f9-ad7b-a7e2361025fd


resources/authgear/templates/zh-HK/translation.json
```diff
@@ -298,6 +298,7 @@
 
   "reset-password-success-page-title": "成功重設密碼",
   "reset-password-success-description": "你已成功重設密碼，請以新密碼登入。",
+  "login-button-label--reset-password-success-page": "Back to 登入",
```
louischan-oursky, line 306: 返回登入

pkg/auth/handler/webapp/authflow_forgot_password.go
```diff
@@ -96,16 +113,35 @@ func (h *AuthflowForgotPasswordHandler) GetData(w http.ResponseWriter, r *http.R
...
+
+		var screen2 *webapp.AuthflowScreenWithFlowResponse
+
+		identification := r.Form.Get("q_login_id_input_type")
+		loginID := r.Form.Get("q_login_id")
+
+		if identification != "" && loginID != "" {
+			var err error
+			// Advance one step to get the options
+			_, screen2, err = h.Controller.AdvanceWithInput(r, s, screen1, map[string]interface{}{
```
louischan-oursky, line 129: I think this should be handled by

1. Not to use HandleStartOfFlow, use something else that support creating a flow AND pass an input at the same time. Then we do not need to call AdvanceWithInput in GET. We also need not to change the signature of AdvanceWithInput.

tung2744: @louischan-oursky Updated to add a optional input to `HandleStartOfFlow`, see if it is better?

resources/authgear/templates/en/web/authflow_reset_password_success.html
```diff
@@ -7,6 +7,10 @@ <h1 class="m-0 primary-txt text-center text-xl font-bold">{{ template "reset-pas
 
 <div class="text-sm break-words primary-txt text-center">{{ template "reset-password-success-description" }}</div>
 
+{{ if .CanBackToSignIn }}
+<a class="as-btn not-a btn primary-btn mt-5" href="{{ call $.MakeURL "/login" "x_step" "" }}">{{ template "login-button-label--reset-password-success-page" }}</a>
```
louischan-oursky, line 11: The generate URL should remove `x_can_back_to_login`, otherwise, this query is persisted. (due to our convention of persisting query of name starting with `x_`)

louischan-oursky (commented):
The description on the forgot password is still incorrect. It shows "... link ..." when I am requesting reset password with my phone number and OTP.

tung2744:
@louischan-oursky Updated again, thanks!

PR #3518, opened 2023-11-29
Support editing new forgot password configs in portal

tung2744:
requires #3515 

Screenshot 2023-11-29 at 3 13 13 PM

Screenshot 2023-11-29 at 3 16 14 PM


tung2744:
@louischan-oursky Updated according to our discussion.
This issue is created to follow up: https://github.com/authgear/authgear-server/issues/3524

I also found that we are not constructing the new config using effective config, so I modified it and now we always construct the config according to effective config.

tung2744:
@louischan-oursky I added [18b1312](https://github.com/authgear/authgear-server/pull/3518/commits/18b1312d5527bb451df15ce9d507e532b2add298) to use effective config to construct configs in portal, but it cause the default values always being saved into the app's config, therefore the config become unnecessarily detailed.

I suggest instead of `SetDefault`, we introduce a new method called `Sanitize`. It should be called after `SetDefault` was called in effective config. 
And in `LoadRawAppConfig`, we call `Sanitize` but do not call `SetDefault`, so that deprecated fields are also removed in raw config.
What do you think?

Issue #3504, opened 2023-11-24
Support reset password by whatsapp otp


tung2744:
From https://github.com/authgear/authgear-server/issues/3500

- [x] Add whatsapp to account recovery channel
- [x] Add config validation to disallow (channel: whatsapp, otp_form: link)
- [x] Support sending whatsapp in forgot password service
- [x] Support whatsapp otp in auth ui forgot password
- [x] Update portal to allow changing otp channel of phone
- [x] Update docs

fungc-io:
@tung2744 this project is using Authflow:
When I press "Send one-time password via SMS instead", It shows an error message saying "Please wait for a moment"


![Image](https://github.com/authgear/authgear-server/assets/15207470/e234bdaf-428d-488c-9009-2241b10edda5)

tung2744:
@fungc-io which project are you using? I guess it might be related to rate limits.

fungc-io:
tested working

PR #3522, opened 2023-11-30
Support whatsapp otp in forgot password flow

tung2744:
requires https://github.com/authgear/authgear-server/pull/3518
ref #3504


tung2744:

Issue #3528, opened 2023-12-05
Detect invalid whatsapp numbers and fallback to SMS


fungc-io:
## Problem 

When sending OOB-OTP via "Whatsapp or SMS" to passwordless login/verify the numbers
Authgear will first send the OTP using whatsapp, and the user needs to manually select "Send via SMS" if they don't have a whatsapp account.

## Solution

- Fallback to SMS automatically if the number is not registered with a whatsapp account
- In our testing, if the number is a vali number but not registered on whatsapp, the API will return "user is not valid".

fungc-io:
@tung2744 
In this project https://demo-enterprise.authgear-staging.com/ , seems it's always sending SMS even when "WhatsApp or SMS" is enabled and the phone number i'm using is registered on whatsapp. Can you check if any issue?

tung2744:
Reproducible with the same project config in local, investigating.

fungc-io:
checked ok

PR #3541, opened 2023-12-20
Fallback to sms if target is not a valid whatsapp user

tung2744:
ref #3528

### DIFF
--- a/pkg/lib/authn/otp/errors.go
+++ b/pkg/lib/authn/otp/errors.go
@@ -12,7 +12,7 @@ var ErrCodeNotFound = InvalidOTPCode.NewWithCause("otp code is expired or invali
 var ErrInvalidCode = InvalidOTPCode.NewWithCause("invalid otp code", apierrors.StringCause("InvalidCode"))
 var ErrConsumedCode = InvalidOTPCode.NewWithCause("used otp code", apierrors.StringCause("UsedCode"))
 
-var ErrInvalidWhatsappUser = InvalidWhatsappUser.NewWithCause("invalid whatsapp user", apierrors.StringCause("InvalidWhatsappUser"))
+var ErrInvalidWhatsappUser = InvalidWhatsappUser.New("invalid whatsapp user")
 
 // FIXME: backward compat; should not use RateLimited
 var ErrTooManyAttempts = ratelimit.RateLimited.NewWithInfo("too many verify OTP attempts", apierrors.Details{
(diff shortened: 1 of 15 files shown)


tung2744:
Updated to return an api error when whatsapp user is not available.

louischan-oursky (commented):
1. When `phone_otp_mode` is `whatsapp`, we cannot fallback, the error message is unknown error. We should recognize the error and tell the number is not a Whatsapp number. 

https://github.com/authgear/authgear-server/assets/15082564/accca584-2cd2-48da-a09a-4cb289a172fb

2. When `phone_otp_mode` is `whatsapp_sms`, we can fallback. After fallback, we can see the branch of switching back to whatsapp (I think it is fine to show this). Clicking the branch will reload the page and make the branch disappear. I think we show error in this case. That is, when the branch is explicitly taken by the end-user, we do not fallback and show error instead.


https://github.com/authgear/authgear-server/assets/15082564/d6f23046-2d7c-4c0c-81a0-fb930bd3a691

tung2744:
@louischan-oursky Updated:

1. When user explicitly selected whatsapp, an error will be shown instead of fallback to sms.
https://github.com/authgear/authgear-server/assets/34942210/d29733b5-e687-4955-b27f-9384f0deedf2

2. When the system selected whatsapp automatically and found it is not a valid whatsapp user, fallback to sms and the whatsapp branch will be hidden.
https://github.com/authgear/authgear-server/assets/34942210/06a8b8b9-72a1-4d01-a193-aac563e6b7c1

tung2744:
@louischan-oursky Updated again to always show all options

pkg/lib/authn/otp/errors.go
```diff
@@ -12,7 +12,7 @@ var ErrCodeNotFound = InvalidOTPCode.NewWithCause("otp code is expired or invali
 var ErrInvalidCode = InvalidOTPCode.NewWithCause("invalid otp code", apierrors.StringCause("InvalidCode"))
 var ErrConsumedCode = InvalidOTPCode.NewWithCause("used otp code", apierrors.StringCause("UsedCode"))
 
-var ErrInvalidWhatsappUser = InvalidOTPCode.NewWithCause("invalid whatsapp user", apierrors.StringCause("InvalidWhatsappUser"))
+var ErrInvalidWhatsappUser = InvalidWhatsappUser.NewWithCause("invalid whatsapp user", apierrors.StringCause("InvalidWhatsappUser"))
```
louischan-oursky, line 15: No need to specify cause here. Just call `New("invalid whatsapp user")`

tung2744: @louischan-oursky Updated, thanks!

Issue #2371, opened 2022-08-25
Address bar and tool bar do not hide on scroll


andychow326:
## Environment
- Browser: Safari on IPhone SE (2019) Version 15.6.1 
- Server: Staging

## Description
When scrolling horizontally in any screen, the top address bar and the bottom tool bar do not hide automatically.

## Steps to reproduce
1. Login to the portal on Safari
2. Select any project
3. Scroll on any screen
4. See the address bar and tool bar wont hide

## Expected result
The address bar and tool bar will hide on scroll. 

## Video

https://user-images.githubusercontent.com/70891443/186574803-dab79757-381c-4872-9b4e-f0f02f254205.MOV

Issue #3544, opened 2023-12-21
Portal cannot render when width <=640px


fungc-io:
Env: staging
The portal project pages became blank when viewport width <= 640px

fungc-io:
tested ok

PR #3545, opened 2023-12-22
Fix portal content is not showing in mobile

tung2744:
ref #3544

### DIFF
--- a/portal/src/ScreenLayoutScrollView.module.css
+++ b/portal/src/ScreenLayoutScrollView.module.css
@@ -5,5 +5,5 @@
   overflow: auto;
 
   /* Let body to scroll in mobile */
-  @apply mobile:overflow-visible mobile:flex-grow-only mobile:min-h-min;
+  @apply mobile:overflow-visible mobile:flex-1-0-auto mobile:min-h-min;
 }
--- a/portal/src/ScreenLayoutScrollView.module.css
+++ b/portal/src/ScreenLayoutScrollView.module.css
@@ -3,4 +3,7 @@
   min-width: 0;
   min-height: 0;
   overflow: auto;
+
+  /* Let body to scroll in mobile */
+  @apply mobile:overflow-visible mobile:flex-grow-only mobile:min-h-min;
 }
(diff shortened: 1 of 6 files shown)


louischan-oursky:
This change will revert the changes made in https://github.com/authgear/authgear-server/issues/2371. Is this acceptable?

louischan-oursky:
cc @fungc-io @tung2744

fungc-io:
😱 is that a bug in Fluent

tung2744:
I think it is not a bug of fluent ui. Maybe it is related to style changes of fluent ui, but it is not a bug. The original css we have is already buggy.

tung2744:
Let me see if there could be a better fix which also fix #2371

tung2744:
@louischan-oursky I tried to use flex to relayout the portal so that the url bar can hide during scroll in mobile layout. I've also tried to scroll screens in desktop layout and it seems working as usual. Please have a look.


https://github.com/authgear/authgear-server/assets/34942210/9135c54e-4696-49bc-a9db-9311ed9e8e17

portal/src/ScreenLayoutScrollView.module.css
```diff
@@ -3,4 +3,7 @@
   min-width: 0;
   min-height: 0;
   overflow: auto;
+
+  /* Let body to scroll in mobile */
+  @apply mobile:overflow-visible mobile:flex-grow-only mobile:min-h-min;
```
louischan-oursky, line 8: Instead of introducing `flex-grow-only`, we can rewrite this block into

```
flex-grow: 1;
flex-shrink: 0;
flex-basis: 0;

@apply mobile:grow mobile:shrink-0 mobile:basis-auto
```

tung2744: Yes, but seems `flex: 1 0 auto` is something we use in a group frequently so I declared it in the tailwind config.

louischan-oursky: Fair enough. What about renaming it to `flex-1-0-auto`?

tung2744: @louischan-oursky Added a commit to rename it to `flex-1-0-auto`

PR #3550, opened 2024-01-04
Fix portal is not showing otp mode correctly

tung2744:
ref #3528 

There is another issue which related to secondary oob otp authenticator, follow up in https://github.com/authgear/authgear-server/issues/3549

### DIFF
--- a/pkg/auth/handler/webapp/authflow_controller.go
+++ b/pkg/auth/handler/webapp/authflow_controller.go
@@ -691,6 +691,9 @@ func (c *AuthflowController) takeBranchRecursively(s *webapp.Session, screen *we
 		// This taken branch require an input to select.
 		case webapp.TakeBranchResultInput:
 			output, screen, err = c.handleTakeBranchResultInput(s, screen, takeBranchResult)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/auth/handler/webapp/authflow_controller.go
```diff
@@ -691,6 +691,9 @@ func (c *AuthflowController) takeBranchRecursively(s *webapp.Session, screen *we
 		// This taken branch require an input to select.
 		case webapp.TakeBranchResultInput:
 			output, screen, err = c.handleTakeBranchResultInput(s, screen, takeBranchResult)
+			if err != nil {
+				return
+			}
```
louischan-oursky, line 696: What bug does this change fix?

tung2744: The bug here is even taking branch has error, the webapp will still continue the loop and finally panic with nil pointer exception.
So I fixed it and ensure an error will be returned.

However, the real error is actually something more difficult to fix, which I opened an issue to fix later: https://github.com/authgear/authgear-server/issues/3549

Issue #3551, opened 2024-01-04
Setup for new auth ui development


tung2744:
- [x] Add config to switch to new authui
- [x] Add new handlers
- [x] Decouple navigation logic with authflow controller

PR #3554, opened 2024-01-05
Setup authflowv2 UI

tung2744:
ref #3551


authui/src/authflowv2/base.css
```diff
@@ -0,0 +1,35 @@
+@layer base {
+  html,
+  body {
+    @apply m-0 p-0 h-full min-h-min flex flex-col;
```
louischan-oursky, line 6: I suggest we move these to page_frame.html, on the element classname.

`m-0` and `p-0` are redundant because Tailwind's preflight.css has done this for us.

authui/src/authflowv2/base.css
```diff
@@ -32,4 +39,20 @@
     --c-grey-600: #303135;
     --c-grey-700: #131315;
   }
+
+  /* Border Radius */
+  :root {
+    --border-radius-full: 9999px;
+    --border-radius-sqaure: 0px;
+    --border-radius-rounded-square: 14px;
```
louischan-oursky, line 47: According to our offline discussion, we do not need these 3 variables. The border radius of primary button should be derived from its font-size.

authui/src/authflowv2/base.css
```diff
@@ -32,4 +39,20 @@
...
+  }
+
+  /* Typography */
+  :root {
+    --font-family-label-large: inherit;
```
louischan-oursky, line 64: If we follow our convention of `primary-btn`, `label-large` should come first, making it `label-large-font-family`.

authui/src/authflowv2/base.css
```diff
@@ -67,4 +67,10 @@
     --letter-spacing-label-large: 0.00625rem;
     --font-weight-label-large: 600;
   }
+
+  /* Alignment */
+  :root {
+    --alignment-branding: center;
```
louischan-oursky, line 73: I suggest we keep the CSS property in the variable name, making it `branding-text-align: center`.

authui/src/authflowv2/components/primary-button.css
```diff
@@ -1,45 +1,78 @@
...
+    --primary-btn-font-weight: var(--font-weight-label-large);
+    --primary-btn-ring-width: 0px;
+    --primary-btn-ring-color: transparent;
+    --primary-btn-py: var(--primary-btn-font-size);
+
+    --primary-btn-px: calc(
+      (var(--primary-btn-py) * 2 + var(--primary-btn-line-height)) / 2
+    ); /* = Height / 2 */
+
+    --primary-btn-bg-color-hover: var(--c-primary-500);
```
louischan-oursky, line 19: I suggest we borrow the idea of BEM https://getbem.com/naming/ in naming our variables.

Examples

`--primary-btn__bg-color--hover`, `--primary-btn__px`

Block in BEM is our component name.
Element in BEM is the CSS property name of the component.
Modifier in BEM is the same.

louischan-oursky (commented):
Visiting /signup will crash. Can we fix it first?

authui/src/authflowv2/base.css
tung2744: @louischan-oursky I think we still need these 3 variables. Only `rounded-square` should be calculated from the font-size.

authui/src/authflowv2/base.css
tung2744: @louischan-oursky What if naming it as `--t-label-large-font-family`, t for typography?

authui/src/authflowv2/base.css
tung2744: @louischan-oursky 
The consideration here is it may not be only used in `text-align`, but also `align-items` `align-self` or other alignment related rules, depends on our implementation.

louischan-oursky: 1. Rename --alignment-branding to --alignment-logo.
2. Document --alignment-logo as a enum. Its valid values are `start`, `center`, `end`, and `hidden`.
3. Document --alignment-content as a enum. Its valid values are `start`, `center` and `end`.

authui/src/authflowv2/base.css
tung2744: I suggest we leave it in the css file so that it can be overridden if needed, let me remove the redundant classes

tung2744:
Updated, thanks!

louischan-oursky:
Build failed.

tung2744:
Fixed, thanks!

Issue #3537, opened 2023-12-12
Roles and Groups


fungc-io:
## Problem

Role-based access control is a common approach to restrict access to certain functions in an applications to authorized users.

In the same project, the applications may have different defined roles. For example:
- In the company bulletin board (A): `admin`, `writer`, `reader`
- In the roster management app (B): `admin`, `manager`, `senior_staff`, `junior_staff` 
- A store manager is always assigned to `writer` in “A” and `manager` in “B”. 

It will be great if the store managers can be assigned to a `store_manager` group and automatically get the corresponding application roles.

## Appetite

6 weeks

## Solution

Introducing Roles and Groups in Authgear:

- Roles are associated with an application. A user can be assign to roles.
- Roles can be associated to a group. A user assigned to a group will automatically receive the entitlement of the corresponding application roles.
- A user can be assigned to many groups.

### Role Creation and Editing

In the portal, developers can define the roles for the applications. They can also assign the users to the roles.

- Name: Display name in the portal
- Key: the string that will be returned in the user info
- Description: Optional description
- Default?: A boolean, the default role to this application if the user sign up without explicitly defining it in the API request

### Groups creation and Editing

In the portal, developers can define the user groups for the applications. They can also assign users into groups

- Group ID: Unique group ID to be used in API
- Name: Display name
- Description: Optional description
- Roles: The application roles which the group members are automatically assigned to.

### Search users by group

In the portal, developers can filter the users by groups. e.g. See all users in the `Manager` group.

### Using roles in the application

Return roles in the JWT and user info as an Array of String. Only the roles in the application are returned, this is determined by the client ID. The user info will not return the roles for other applications of another client id.

The roles can also be obtained from the resolver

### Group and roles assignment on signup

- When creating the users by API, developers can specify the groups and roles the users belong
- In the user pre-create hook, developers can specify the groups and roles the users belong

## Discussion

- Should a role be specific to an application? App B cannot see role in app A? 
- How should the application-specific role returned in Admin API or when there is no client id
  - Will that lead to different representation in userInfo for specific app and in Admin API?
- How should the access token be revoked/refreshed after the roles of the user is updated

louischan-oursky:
> Default?: A boolean, the default role to this application if the user sign up without explicitly defining it in the API request

If we have "Group and roles assignment on signup", I think this is not necessary.

[...]

PR #3556, opened 2024-01-08
Specify roles and groups

louischan-oursky:
ref #3537

### DIFF
--- a/docs/specs/roles-groups.md
+++ b/docs/specs/roles-groups.md
@@ -440,13 +440,13 @@ and then we can add support for roles / groups assignment.
 
 # Changes in JWT access token
 
-The effective role keys of a user will appear as `https://authgear.com/claims/user/effective_roles` in the JWT access token.
+The effective role keys of a user will appear as `https://authgear.com/claims/user/roles` in the JWT access token.
 It is an array of strings, for example, `["store_manager", "salesperson"]`.
 The order is unspecified.
 
 # Changes in User Info
 
-The effective role keys of a user will appear as `https://authgear.com/claims/user/effective_roles` in the User Info.
+The effective role keys of a user will appear as `https://authgear.com/claims/user/roles` in the User Info.
 It is an array of strings, for example, `["store_manager", "salesperson"]`.
 The order is unspecified.
 
--- a/docs/specs/roles-groups.md
+++ b/docs/specs/roles-groups.md
@@ -237,7 +237,7 @@ type DeleteGroupPayload {
 }
 
 input AddRoleToUsersInput {
-  roleID: ID!
+  roleKey: String!
   userIDs: [ID!]!
 }
 
@@ -246,7 +246,7 @@ type AddRoleToUsersPayload {
 }
 
 input RemoveRoleFromUsersInput {
-  roleID: ID!
+  roleKey: String!
   userIDs: [ID!]!
 }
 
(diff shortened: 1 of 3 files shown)


docs/specs/roles-groups.md
```diff
@@ -0,0 +1,472 @@
...
+# Changes in the headers of resolver
+
+See [./api-resolver.md#x-authgear-user-roles](./api-resolver.md#x-authgear-user-roles).
+
+# Changes in blocking event hooks
+
+Some blocking event hooks can mutate the `roles` and `groups` of a user.
+See [./hook.md#blocking-event-mutations](./hook.md#blocking-event-mutations).
+
+# Changes in user listing
```
louischan-oursky, line 458: Clarify that we will take a compromise first.

louischan-oursky:
I have a suggestion. Rename `roles` to `computed_roles` when roles are computed, and `roles` always stand for non-computed roles. Thoughts? @tung2744 @fungc-io @chpapa 

It is because we want to allow the developer to mutate `roles` and `groups` in blocking hooks. Then we should give them `roles` and `groups` in the event payload. In this context, `roles` is non-computed. So `roles` sometimes is computed and sometimes is not, depending what context the developer finds it. I think this will be confusing.

fungc-io:
In our offline discussion,
we plan to use:
- `effective_roles` to represent all roles, including the directly assigned roles and the derived roles
- And `roles` for those directly assigned roles, which can be edited by hooks

louischan-oursky:
@fungc-io The key appearing in JWT and userinfo will be `https://authgear.com/claims/user/effective_roles` instead of `effective_roles`, to be consistent of other Authgear-defined claims.

chpapa:
@louischan-oursky @fungc-io sorry I am confused, so what's the difference between effective_role and role? Why we need two keys?

chpapa:
If the definition of effective_roles is just a union of roles + roles of the group user in.

I think we should either:

- just have roles in user info endpoint and tokens.
- if there are some reason that's confusing (eg when used with admin api), maybe less confusing if we define each users can only belong to groups, and roles are always computed from groups

Definitely not introduce another terms that need a page to explain in doc again 🙃

fungc-io:
Agree to better keep one `roles` only.

@louischan-oursky can you elaborate more on how the user maybe confused by cannot tell "directly assigned roles" and the "derived roles" apart. Perhaps in actual usecase they don't need to tell them apart

louischan-oursky:
We had implemented JWT access token in [PR #750 in Nov 2020](https://github.com/authgear/authgear-server/pull/750), predating [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.txt), which was published in Oct 2021.

In Section 2.2.3.1. of RFC 9068, it says that `roles` and `groups` should be referred to [RFC 7643](https://www.rfc-editor.org/rfc/rfc7643.txt). In RFC 7643, `roles` is an array of objects. This is different from our definition.

Therefore, we want to avoid using the name `roles` in JWT access token and User Info.

Ben suggested that we simply use the name `https://authgear.com/claims/user/roles` in JWT access token and User Info, to represent effective roles. In event payload and Admin API, `roles` means roles directly associated with the user.

fungc-io:
thanks all

louischan-oursky:
@Ngkaokis Beware of the spec update.

Issue #3555, opened 2024-01-08
AuthUI V2 - Implement common component styles


tung2744:
- [x] Screen Title 

![Image](https://github.com/authgear/authgear-server/assets/34942210/818c714d-3ea3-4504-b12d-6be0e0d87e36)

- [x] Screen Description

![Image](https://github.com/authgear/authgear-server/assets/34942210/6642010b-a1d4-4918-9f46-baab05b76254)

- [x] Primary Button 

![Image](https://github.com/authgear/authgear-server/assets/34942210/16a5e027-71f2-40b8-a094-af95f698cd44)

- [x] Secondary Button 
    - [ ] Secondary Button Icon

![Image](https://github.com/authgear/authgear-server/assets/34942210/e3ea504a-0b4c-4e3b-a596-512c6fbc0929)

- [x] Tertiary Button 

![Image](https://github.com/authgear/authgear-server/assets/34942210/67cc5505-99f7-46ff-9dd9-a20a68d31909)

- [x] Text Button

![Image](https://github.com/authgear/authgear-server/assets/34942210/419b6019-ba5a-43b9-953c-b10941f9057b)

PR #3563, opened 2024-01-09
Implement common component styles

Ngkaokis:
ref #3555 

require #3554

https://github.com/authgear/authgear-server/assets/63200569/7f5c808e-1af4-45e4-8897-ee63fd52bfb6


authui/src/authflowv2/components/secondary-button.css
```diff
@@ -0,0 +1,64 @@
+@layer components {
+  :root {
+    --secondary-btn-bg-color: var(--c-white);
+    --secondary-btn-text-color: var(--c-grey-700);
+    --secondary-btn-border-radius: var(--border-radius-rounded-square);
+    --secondary-btn-font-family: var(--font-family-label-large);
```
tung2744, line 6: @Ngkaokis We've renamed some css variable names yesterday, can you please rebase and use the latest name? :adore:

authui/src/authflowv2/base.css
```diff
@@ -47,6 +47,12 @@
     --border-radius-rounded-square: 14px;
   }
 
+  /* Ring Width */
+  :root {
+    --ring-width-0: 0px;
+    --ring-width-1\.5: 1.5px;
```
tung2744, line 53: 1. I  think we don't want to name a css variable using the actual value. Maybe we should call it something like `--ring-width-common`.
2. Maybe ring width is a special case that we can directly write `1.5px` as the default value in the component, and don't need a variable in base styles, because ring is unlikely to be changed together with other components. @louischan-oursky Also see how louis think

authui/src/authflowv2/components/secondary-button.css
```diff
@@ -0,0 +1,64 @@
...
+    --secondary-btn-text-color-active: var(--secondary-btn-text-color);
+    --secondary-btn-ring-width-active: var(--secondary-btn-ring-width);
+    --secondary-btn-ring-color-active: var(--secondary-btn-ring-color);
+  }
+
+  .secondary-btn {
+    @apply text-center;
+    @apply flex justify-center items-center;
+
+    fill: currentColor;
```
tung2744, line 34: fill should be a style applied to the svg class.

resources/authgear/templates/en/web/authflowv2/enter_password.html
```diff
@@ -6,5 +6,32 @@ <h1 class="screen-title">Enter Password</h1>
 <div class="w-[400px] p-6 space-y-2">
   <button class="primary-btn w-full">Primary</button>
   <button class="primary-btn w-full" disabled>Primary Disabled</button>
+  <button class="secondary-btn w-full">
+    <svg
```
tung2744, line 15: Can we put this svg into a css class? (Maybe check how tabler write the css?)

authui/src/authflowv2/components/screen-description.css
```diff
@@ -0,0 +1,24 @@
+@layer components {
+  :root {
+    --screen-description-font-family: var(--font-family-title-medium);
+    --screen-description-font-size: var(--font-size-title-medium);
+    --screen-description-line-height: var(--line-height-title-medium);
+    --screen-description-letter-spacing: var(--letter-spacing-title-medium);
+    --screen-description-font-weight: var(--font-weight-title-medium);
+    --screen-description-text-align: var(--alignment-content);
```
tung2744, line 8: Please check `authui/src/authflowv2/components/screen-title.css` for the text-align style of this component. They should be similar.

authui/src/authflowv2/base.css
```diff
@@ -72,6 +72,18 @@
...
+    --font-size-label-medium: 0.875rem;
+    --line-height-label-medium: 1.25em;
+    --letter-spacing-label-medium: 0.03125em;
+    --font-weight-label-medium: 600;
+  }
+
+  /* Padding */
+  :root {
+    --padding-2\.5: 0.625rem;
+    --padding-6: 1.5rem;
```
tung2744: Same as above, do not name the variable with the actual value.

authui/src/authflowv2/components/tertiary-button.css
```diff
@@ -0,0 +1,78 @@
...
+    --tertiary-btn-font-family: var(--font-family-label-medium);
+    --tertiary-btn-font-size: var(--font-size-label-medium);
+    --tertiary-btn-line-height: var(--line-height-label-medium);
+    --tertiary-btn-letter-spacing: var(--letter-spacing-label-medium);
+    --tertiary-btn-font-weight: var(--font-weight-label-medium);
+    --tertiary-btn-ring-width: 0px;
+    --tertiary-btn-ring-color: transparent;
+    --tertiary-btn-py: var(--padding-2\.5);
+
+    --tertiary-btn-px: var(--padding-6);
```
tung2744, line 15: Let's calculate the paddings from the font size, just like primary-btn.

According to design, the y padding is 10, and font size is 14. So y padding can be `0.7143em` (10 / 14 = 0.71428571428). Document how we calculated it.

And x padding is half of the button height, just like primary button. This is because the maximum border radius is half of the height.

So `--padding-2\.5` and `--padding-6` could be removed.

authui/src/authflowv2/base.css
louischan-oursky: Let's write `ring-width-[1.5px]` directly on the component HTML.

tung2744: @louischan-oursky  The border width of the component should be adjustable, therefore there should be a component level css variable. I meant the value of component level css variable can be `1.5px`, i.e.
```css
:root {
  --secondary-button-ring-width: 1.5px;
}
```

So the base variables `--ring-width-x` can be removed.

louischan-oursky: `--secondary-btn__ring-width`, to be exact.

tung2744:
Sorry there is one more change in the css variable naming.
We decided to follow the [BEM naming convention](https://getbem.com/naming/) for css variables.
I've pushed a new commit https://github.com/authgear/authgear-server/pull/3554/commits/b0ece56e67eb9f2f5b86190fb5c871f9c704a774 to rename the variables in primary button and screen title.
Please also rename the variables add in this PR.

resources/authgear/templates/en/web/authflowv2/enter_password.html
louischan-oursky: We used tabler icon font. So we just use `color` to change the tint color.

tung2744: Opened another issue to implement the icons.
https://github.com/authgear/authgear-server/issues/3568

authui/src/authflowv2/components/label-button.css
```diff
@@ -0,0 +1,82 @@
...
+    --label-btn__ring-width--hover: var(--label-btn__ring-width);
+    --label-btn__ring-color--hover: var(--label-btn__ring-color);
+
+    --label-btn__bg-color--active: transparent;
+    --label-btn__text-color--active: var(--color-primary-500);
+    --label-btn__ring-width--active: var(--label-btn__ring-width);
+    --label-btn__ring-color--active: var(--label-btn__ring-color);
+
+    --label-btn__bg-color--disabled: transparent;
+    /* FIXME(tung): It is grey-600 with opacity in figma, confirm with designer */
```
louischan-oursky, line 32: Are this FIXME and the FIXMEs in primary-button.css still relevant? Please remove the FIXMEs if they are resolved.

Ngkaokis: Yes, asked designer in the meeting. Let me update it later.

Issue #3568, opened 2024-01-09
AuthUI V2 - Sign in Method icons


tung2744:
- SSO icons: Two set of icons. One for light mode, one for dark mode. Hardcoded as background image.
- Other sign in methods (Passkey / Email / Phone..): Material Icon

PR #3599, opened 2024-01-10
Setup authflow v2 icons

Ngkaokis:
ref #3568

require #3563 

reference site
https://developers.google.com/fonts/docs/material_icons#icon_font_for_the_web

image


authui/src/authflowv2/icons/material-symbols.css
```diff
@@ -0,0 +1,37 @@
...
+  text-rendering: optimizeLegibility;
+
+  /* Support for Firefox. */
+  -moz-osx-font-smoothing: grayscale;
+
+  /* Support for IE. */
+  font-feature-settings: "liga";
+}
+
+.material-icons__passkey::before {
```
louischan-oursky: Per offline discussion, let's use the ligatures feature documented in https://developers.google.com/fonts/docs/material_icons#using_the_icons_in_html then we can save our time declaring every icon we want to use here.

resources/authgear/templates/en/web/authflowv2/enter_password.html
```diff
@@ -18,5 +18,9 @@ <h1 class="screen-title">Enter Password</h1>
   <p class="screen-description">
     We've just sent a 6-digit verification code to Marianne20@gmail.com.
   </p>
+  <button class="secondary-btn w-full">
+    <div class="sso-icon google-icon mr-2"></div>
```
louischan-oursky, line 24: Use span instead of div.

authui/src/authflowv2/icons/authgear-icon.css
```diff
@@ -0,0 +1,20 @@
...
+    @apply mr-auto ml-auto;
+  }
+
+  :root[alignment-logo="end"] & {
+    @apply mr-0 ml-auto;
+  }
+
+  :root[alignment-logo="hidden"] & {
+    @apply hidden;
+  }
```
louischan-oursky, line 19: Alignment depends on the parent container. Here we assume the parent container is a block. If it is a flex container, we need different sets of style to position to logo. Let's keep this for now. We need to revisit this when we actually implement the logo header.

authui/src/authflowv2/icons/authgear-icon.css
```diff
@@ -0,0 +1,20 @@
+.authgear-icon {
+  @apply w-16 h-16 bg-no-repeat bg-center;
+  background-image: url(./icons/raw/icon-authgear.svg);
```
tung2744, line 3: @Ngkaokis Sorry I forgot to mention in the issue:
The logo is not hardcoded and can be configured per project.
Please see `resources/authgear/templates/en/web/__header.html` for the current implementation.
Also, the logo size is adjustable according to the spec, so please make it a css variable.

Ngkaokis: why display two logos if dark theme enabled :derpthink:
```
{{ if .darkThemeEnabled }}
<img class="banner light" src="{{ .light }}">
<img class="banner dark" src="{{ .dark }}">
{{ else }}
<img class="banner" src="{{ .light }}">
{{ end }}

{{ end }}

```

tung2744: @Ngkaokis If dark theme disabled, no matter your device is dark or light theme, it shows the light icon.
If dark theme enabled, depends on your device's theme, it show one icon and hide another one.

Ngkaokis: Do you expect to duplicate the header implementation from v1 to v2?

louischan-oursky: See authui/src/tailwind.css:308

tung2744:
Added 5774cd4 to remove unused authgear icon.

PR #3602, opened 2024-01-10
Implement common component styles

Ngkaokis:
ref #3555 

require #3599 

https://github.com/authgear/authgear-server/assets/63200569/966f949d-9e1d-47ab-b31c-78cc012634bc

### DIFF
--- a/authui/src/authflowv2/base.css
+++ b/authui/src/authflowv2/base.css
@@ -79,9 +79,15 @@
 
     --typography-body-small__font-family: inherit;
     --typography-body-small__font-size: 0.75rem;
-    --typography-body-small__line-height: 1rem;
-    --typography-body-small__letter-spacing: 0.025rem;
+    --typography-body-small__line-height: 1em;
+    --typography-body-small__letter-spacing: 0.025em;
     --typography-body-small__font-weight: 500;
+
+    --typography-body-large__font-family: inherit;
+    --typography-body-large__font-size: 1rem;
+    --typography-body-large__line-height: 1.5em;
+    --typography-body-large__letter-spacing: 0.03125em;
+    --typography-body-large__font-weight: 500;
   }
 
   /* Alignment */
--- a/authui/src/authflowv2/base.css
+++ b/authui/src/authflowv2/base.css
@@ -55,7 +55,7 @@
   :root {
     --typography-headline-medium__font-family: inherit;
     --typography-headline-medium__font-size: 1.75rem;
-    --typography-headline-medium__line-height: 2.125em;
+    --typography-headline-medium__line-height: 1.21428571429em; /* 34 / 28 = 1.21428571429 */
     --typography-headline-medium__letter-spacing: 0px;
     --typography-headline-medium__font-weight: 700;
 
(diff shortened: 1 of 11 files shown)


authui/src/authflowv2/base.css
```diff
@@ -76,6 +76,12 @@
     --typography-label-medium__line-height: 1.25em;
     --typography-label-medium__letter-spacing: 0.03125em;
     --typography-label-medium__font-weight: 600;
+
+    --typography-body-small__font-family: inherit;
+    --typography-body-small__font-size: 0.75rem;
+    --typography-body-small__line-height: 1rem;
+    --typography-body-small__letter-spacing: 0.025rem;
```
tung2744, line 83: Use `em` as unit because it should be related to the font size of the current element not the root element.

resources/authgear/templates/en/web/authflowv2/__divider.html
```diff
@@ -0,0 +1,14 @@
+{{ define "authflowv2/__divider.html" }}
+
+<div class="divider__container {{$.Classname}}">
```
tung2744, line 3: Lets use [hr](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr) as the  container element.

authui/src/authflowv2/components/divider.css
```diff
@@ -0,0 +1,36 @@
...
+
+    background-color: var(--divider__bg-color);
+  }
+
+  .divider__container {
+    @apply flex justify-center items-center relative;
+  }
+
+  .divider__text-container {
+    @apply bg-white text-center w-11;
```
tung2744, line 27: 1. Instead of using fixed width, give it a padding so that it won't broken when the text length changed.

2. The background should be transparent because the background color might not be white. It is customizable. The element structure can be `<divider><OR><divider>` so we don't need to specify a background color.

resources/authgear/templates/en/web/authflowv2/__divider.html
tung2744: @Ngkaokis I think `<hr>` should be used as the outer most element, which contains the whole divider.
It is because the all elements inside `divider__container` is actually used as one horizontal rule.
And the elements inside `hr` can use non-semantic elements like div or span, because the outermost `<hr>` already tell us they are part of the horizontal rule.

resources/authgear/templates/en/web/authflowv2/__divider.html
```diff
@@ -0,0 +1,15 @@
...
+    {{/* TODO(authflowv2): Template don't support passing dynamic variable so
+    hardcoding "or" here. Need to implement a function like helm include
+    function
+    https://helm.sh/docs/howto/charts_tips_and_tricks/#know-your-template-functions
+    */}}
+  <hr class="divider"></hr>
+  <p class="divider_text">{{template "or-label"}}</p>
```
tung2744, line 10: I think we should use new translation strings for v2 ui.
Lets prefix the translation ids with `v2-`. So we should add a `v2-or-label`.
@louischan-oursky See if any thoughts?

resources/authgear/templates/en/web/authflowv2/__divider.html
Ngkaokis: Oh, I misunderstood your meaning. Let me update it.

louischan-oursky: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr

hr should not have children.

authui/src/authflowv2/components/code-block.css
```diff
@@ -0,0 +1,34 @@
...
+
+  .code-block {
+    @apply flex flex-col justify-center items-center p-4;
+
+    border-radius: var(--code-block__border-radius);
+    background-color: var(--code-block__bg-color);
+  }
+
+  .code-block__text {
+    @apply whitespace-pre-line;
```
louischan-oursky, line 25: whitespace-pre

resources/authgear/templates/en/web/authflowv2/__divider.html
louischan-oursky: OK. Let's use `v2-` prefix.

tung2744:
The line-height and letter spacing of some typography is wrong, pushed 034391f to fix it.

Issue #3557, opened 2024-01-08
AuthUI V2 - Implement Text Field


tung2744:
![Image](https://github.com/authgear/authgear-server/assets/34942210/aa7f77f9-b25d-4576-99fb-79092fe7fdfe)

- Error state should be supported.

PR #3610, opened 2024-01-11
Add text input component

Ngkaokis:
ref #3557

### DIFF
--- a/authui/src/authflowv2/base.css
+++ b/authui/src/authflowv2/base.css
@@ -15,6 +15,10 @@
     cursor: not-allowed;
   }
 
+  *:focus {
+    @apply outline-none;
+  }
+
   /* Colors */
   :root {
     --color-white: #ffffff;
@@ -51,6 +55,12 @@
     --border-radius-rounded-square: 0.875em;
   }
 
+  /* Border Width */
+  :root {
+    --border-width-regular: 1.5px;
+    --border-width-regular--active: 2px;
+  }
+
   /* Typography */
   :root {
     --typography-headline-medium__font-family: inherit;
@@ -74,7 +84,7 @@
     --typography-label-medium__font-family: inherit;
     --typography-label-medium__font-size: 0.875rem;
     --typography-label-medium__line-height: 1.42857142857em; /* 20 / 14 = 1.42857142857 */
-    --typography-label-medium__letter-spacing: 0.025em; /* 0.5 / 20 = 0.025 */
+    --typography-label-medium__letter-spacing: 0.03571428571em; /* 0.5 / 14 = 0.03571428571 */
     --typography-label-medium__font-weight: 600;
 
     --typography-body-small__font-family: inherit;
@@ -82,6 +92,12 @@
     --typography-body-small__line-height: 1.33333333333em; /* 16 / 12 = 1.33333333333 */
     --typography-body-small__letter-spacing: 0.03333333333em; /* 0.4 / 12 = 0.03333333333 */
     --typography-body-small__font-weight: 500;
+
+    --typography-body-medium__font-family: inherit;
+    --typography-body-medium__font-size: 0.875rem;
+    --typography-body-medium__line-height: 1.42857142857em; /* 20 / 14 = 1.42857142857 */
+    --typography-body-medium__letter-spacing: 0.03571428571em; /* 0.5 / 14 = 0.03571428571 */
+    --typography-body-medium__font-weight: 500;
 
     --typography-body-large__font-family: inherit;
     --typography-body-large__font-size: 1rem;
--- a/authui/src/authflowv2/components/input.css
+++ b/authui/src/authflowv2/components/input.css
@@ -1,0 +1,74 @@
+@layer components {
+  :root {
+    --input__ring-color: var(--color-grey-100);
+    --input__ring-color--focus: var(--color-primary-base);
+    --input__ring-color--error: var(--color-error);
+    --input__ring-width: var(--border-width-regular);
+    --input__ring-width--focus: var(--border-width-regular--active);
+    --input__border-radius: var(--border-radius-rounded-square);
+    --input__placeholder-color: var(--color-grey-200);
+    --input__text-color: var(--color-grey-700);
+    --input__font-family: var(--typography-body-large__font-family);
+    --input__font-size: var(--typography-body-large__font-size);
+    --input__line-height: var(--typography-body-large__line-height);
+    --input__letter-spacing: var(--typography-body-large__letter-spacing);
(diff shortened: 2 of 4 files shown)


authui/src/authflowv2/base.css
```diff
@@ -51,6 +51,12 @@
     --border-radius-rounded-square: 0.875em;
   }
 
+  /* Border Weight */
+  :root {
+    --border-radius-regular: 1.5px;
```
louischan-oursky, line 56: I think you mean `--border-width-regular: 1.5px`.

@tung2744 Any thoughts on `regular=1.5px & medium=2px`?

authui/src/authflowv2/base.css
```diff
@@ -51,6 +51,12 @@
     --border-radius-rounded-square: 0.875em;
   }
 
+  /* Border Weight */
```
louischan-oursky, line 54: Weight -> Width

authui/src/authflowv2/components/input.css
```diff
@@ -0,0 +1,87 @@
...
+    --input__border-radius: var(--border-radius-rounded-square);
+    --input__placeholder-color: var(--color-grey-200);
+    --input__text-color: var(--color-grey-700);
+    --input__font-family: var(--typography-body-large__font-family);
+    --input__font-size: var(--typography-body-large__font-size);
+    --input__line-height: var(--typography-body-large__line-height);
+    --input__letter-spacing: var(--typography-body-large__letter-spacing);
+    --input__font-weight: var(--typography-body-large__font-weight);
+    --input__y-padding: 1em;
+    --input__x-padding: 1.09em;
```
louischan-oursky, line 18: 1.09 seems very arbitrary to me. I think it should be 1em

authui/src/authflowv2/components/input.css
```diff
@@ -0,0 +1,87 @@
...
+
+    --input-error-message__text-color: var(--color-error);
```
louischan-oursky, line 20: It should be `--input__error-message-text-color`. `input` is the BEM block. `error-message-text-color` is the BEM element.

authui/src/authflowv2/components/input.css
```diff
@@ -0,0 +1,87 @@
...
+}
+
+.input::placeholder {
+  color: var(--input__placeholder-color);
+  opacity: 1; /* Firefox */
+}
+
+.input::-ms-input-placeholder {
+  /* Edge 12 -18 */
+  color: var(--input__placeholder-color);
```
louischan-oursky, line 77: Can we write this? This saves us from getting styling placeholder correctly. Tailwind should have done this for us.

```
.input {
  @apply placeholder:[var(--input\_\_placeholder-color)]
}
```

authui/src/authflowv2/base.css
tung2744: Prefer calling `medium` as `active` or `highlight`, because the purpose of this border width is to indicate something is active / highlighted.

authui/src/authflowv2/components/input.css
tung2744: Should be 18px (4.5 spacing unit). Ideally should be multiple of 4. We can confirm with designer.
BTW, name it `px` and `py` which is same as other components.

authui/src/authflowv2/components/input.css
```diff
@@ -0,0 +1,87 @@
...
+}
+
+:root.dark {
+  /* TODO(davis): Dark mode styles */
+}
+
+.input {
+  padding: var(--input__y-padding) var(--input__x-padding);
+  border-radius: var(--input__border-radius);
+  border: var(--input__border-width) solid var(--input__border-color);
```
tung2744, line 44: Simply name the variable `--input__border` and write all 3 elements (width, style, color) in it?

tung2744: I saw you used ring already, do we need border too?

authui/src/authflowv2/components/input.css
```diff
@@ -0,0 +1,87 @@
...
+  line-height: var(--input__line-height);
+  letter-spacing: var(--input__letter-spacing);
+  font-weight: var(--input__font-weight);
+  caret-color: var(--input__ring-color--focus);
+  @apply ring-inset	
+      ring-[length:var(--input\_\_ring-width)]
+      ring-[color:var(--input\_\_ring-color)];
+
+  &:focus {
+    @apply outline-none;
```
tung2744, line 57: Should this be a style exist in base stylesheet which apply to all input?

authui/src/authflowv2/components/input.css
```diff
@@ -0,0 +1,87 @@
...
+    @apply ring-inset	
+      ring-[length:var(--input\_\_ring-width--focus)]
+      ring-[color:var(--input\_\_ring-color--focus)];
+  }
+
+  &.error {
```
tung2744, line 63: Prefer .input--error to avoid classname collision.

authui/src/authflowv2/base.css
louischan-oursky: `--border-width-regular: 1.5px`
`--border-width-regular--active: 2px`

Something like this?

authui/src/authflowv2/components/input.css
louischan-oursky: Should we use this https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid ?

authui/src/authflowv2/base.css
Ngkaokis: `--border-width-regular--active` this naming seems not so suitable to put in `base.css`? 
State related naming should put back to `component.css`?

authui/src/authflowv2/components/input.css
Ngkaokis:
```
  *:focus {
    @apply outline-none;
  }

```
Actually I thought it's set globally in `tailwind.css`. But don't know why it's not functioning on my own testing :derpthink:

authui/src/authflowv2/components/input.css
Ngkaokis: Then we will need to use `Constraint API` to do the validation.

https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation#complex_constraints_using_the_constraint_validation_api

authui/src/authflowv2/base.css
louischan-oursky: If the width is shared by multiple components, it can be placed in base.css

authui/src/authflowv2/components/input.css
tung2744: That is for validation error only, but we need to display error from server so it seems not suitable for us.

Issue #3562, opened 2024-01-08
AuthUI V2 - Implement OTP Field


tung2744:
![Image](https://github.com/authgear/authgear-server/assets/34942210/4350a853-ac5b-494a-ba92-ec54a68ae0cc)

- Implement as a stimulus controller attached to input element
- Auto submit the form after the 6th digit was entered.
- Error state should be supported.

IniZio:
@tung2744 What's the recommended way for adding variant to components? Currently `.label-btn` only has fixed font size `1rem` and padding 🙏 .

tung2744:
@IniZio It seems the resend button is not same as the label button. I think we might want another class for the resend button.

louischan-oursky:
> It seems the resend button is not same as the label button. I think we might want another class for the resend button.

It seems like we need to confirm with the designer on this. @IniZio Please leave a comment in the Figma file for Arvin.

IniZio:
Tracking in https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE?node-id=769:49233&mode=dev#665880997

PR #3611, opened 2024-01-11
AuthUI V2 - Signup

tung2744:
requires #3567 
ref #3569 

TODO:
- [ ] Display error
- [ ] Integrate text field
- [ ] Integrate phone input field

Screenshot 2024-01-11 at 7 18 57 PM
Screenshot 2024-01-11 at 7 18 49 PM

PR #3612, opened 2024-01-12
Add OTP input component

IniZio:
ref #3562

image


resources/authgear/templates/en/web/authflowv2/enter_password.html
```diff
@@ -53,5 +53,94 @@ <h1 class="screen-title">Enter Password</h1>
...
+      <button
+        class="hidden"
+        type="submit"
+        name="x_action"
+        value="submit"
+        data-otp-input-target="submit"
+      ></button>
+    </form>
+    <form method="post" novalidate>
+      <button
```
IniZio, line 83: After to confirming with designer, this button behaves more line an inline element. Should I just use span with role button?

resources/authgear/templates/en/web/authflowv2/enter_password.html
```diff
@@ -53,5 +53,94 @@ <h1 class="screen-title">Enter Password</h1>
     By registering, you agree to the <a href="#">Terms of Service</a> and
     <a href="#">Privacy Policy</a>.
   </p>
+
+  <!-- OTP Input -->
+  <div data-controller="otp-input">
```
tung2744, line 58: Can we make it a reusable template? The template should:
  - generate all necessary ui elements.
  - The name / value of resend button, submit button, text input should be configurable
  - The error message should be shown depends on a provided context value (.Error)

authui/src/otpInput.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export class OtpInputController extends Controller {
+  static targets = ["input", "submit", "textContainer"];
+
+  declare readonly inputTarget: HTMLInputElement;
+  declare readonly submitTarget: HTMLButtonElement;
+  declare readonly textContainerTarget: HTMLElement;
+
+  value: string = "";
+  maxLength: number = 6;
```
IniZio, line 11: Currently use input maxlength to init, or should I be explicit and use values instead?

pkg/auth/handler/webapp/authflowv2/enter_password.go
```diff
@@ -135,5 +135,33 @@ func (h *AuthflowV2EnterPasswordHandler) ServeHTTP(w http.ResponseWriter, r *htt
...
+		input := map[string]interface{}{}
+
+		result, err := h.Controller.UpdateWithInput(r, s, screen, input)
+		if err != nil {
+			return err
+		}
+
+		result.WriteResponse(w, r)
+		return nil
+	})
```
tung2744, line 164: I guess enter password screen don't have these code. If it is for testing only I think we can add a FIXME comment to remove them later.

authui/src/authflowv2/components/otp-input.css
```diff
@@ -0,0 +1,124 @@
...
+    input {
+      @apply w-full;
+      @apply h-full;
+      @apply outline-none;
+      @apply selection:bg-white/0;
+      @apply caret-white/0;
+      @apply opacity-0;
+    }
+
+    .text-container {
```
tung2744, line 54: Lets use BEM naming, i.e. otp-input__text-container. Same for other elements.

resources/authgear/templates/en/web/authflowv2/enter_password.html
louischan-oursky: Use `button` and style it with a different style.

authui/src/authflowv2/components.css
```diff
@@ -6,6 +6,7 @@
 @import "./components/secondary-button.css";
 @import "./components/tertiary-button.css";
 @import "./components/label-button.css";
+@import "./components/inline-button.css";
```
louischan-oursky, line 9: I think link-button is a better name. @tung2744 Thoughts?

authui/src/otpInput.ts
louischan-oursky: using maxlength looks good to me.

authui/src/authflowv2/components.css
tung2744: Maybe simply call it resend button, because it is actually only for resending and not a generic component in figma. 🤔

louischan-oursky: Arvin said it is used in other places too.

authui/src/authflowv2/components/otp-input.css
```diff
@@ -0,0 +1,124 @@
...
+
+    --otp-input-text-digit__ring-color: var(--color-grey-200);
+    --otp-input-text-digit__ring-color--focus: var(--color-primary-base);
+    --otp-input-text-digit__ring-color--error: var(--color-error);
+    --otp-input-text-digit__ring-width: 1px;
+    --otp-input-text-digit__ring-width--focus: 1px;
+    --otp-input-text-digit__ring-width--error: 1px;
+    --otp-input-text-digit__border-radius: var(--border-radius-rounded-large);
+    --otp-input-text-digit__border-width: 1px;
+    --otp-input-text-digit__border-color: var(--color-grey-200);
```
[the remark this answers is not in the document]
tung2744: Or do we actually need border if we are using ring already?

resources/authgear/templates/en/web/authflowv2/enter_password.html
louischan-oursky: Ideally in the template it should look like

```
<input data-controller="otp-input" type="text" inputmode="numeric" pattern="\d*" maxlength="6">
```

When the controller `connect()`, it transform the element into a OTP component.

I think the OTP component only deals the OTP input. It does not include the resend button, submit button, etc.

authui/src/authflowv2/components.css
tung2744: It is not a component in figma.
I saw the same style here (The login button): https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=762-37040&mode=design&t=HdwuDJWdD4Lf2u1u-4 
But they are having different purpose, this one is really inline link / button inside a body text.
I've implemented it in https://github.com/authgear/authgear-server/pull/3611

authui/src/authflowv2/components/otp-input.css
```diff
@@ -0,0 +1,124 @@
+@layer components {
+  :root {
+    --otp-input__spacing: 21px;
+
+    --otp-input-text-digit__ring-color: var(--color-grey-200);
```
louischan-oursky, line 5: `--otp-input` is the BEM block. `text-digit-ring-color` is the BEM element. So it should be `--otp-input__text-digit-ring-color`.

resources/authgear/templates/en/web/authflowv2/enter_password.html
tung2744: I think your comment is about the controller at the input, but I mean on top of it, we want a reusable template which generates all neccesary elements at once so we can reuse it in different screen.
Because these things are always used together and never be separated:
- OTP input
- Resend button
- Error message

authui/src/authflowv2/components.css
louischan-oursky: In translation.json, we have some translation containing `<a class="link">`. By using `link`, the element should look like a styled link in the design. Ideally, if we write `<button class="link">`, then the button should look like a link.

resources/authgear/templates/en/web/authflowv2/enter_password.html
IniZio: > Can we make it a reusable template? The template should:

I think generating ui elements with connect should avoid those that will cause layout shift 🤔.

louischan-oursky:
```
<!-- in OTP page -->
{{ template "__otp_widget.html" (props) }}

<!-- in __otp_widget.html -->
{{ template "__otp_input.html (props) }}
{{ template "__resend_button.html (props) }}
...

<!-- in __otp_input.html -->
In here you can write out all necessary elements that is required for the input.
But only <input> is visible initially. The interactive part is activated by the connected controller 
```

tung2744: By "generate", it is not neccesary to be generated on client side, maybe like:

```html
{{ define "authflowv2/__otp_input.html" }}
    <div data-controller="otp-input">
      <div class="otp-input h-11">
        <input
          data-otp-input-target="input"
          /* ...other things */
        />
        <div data-otp-input-target="textContainer" class="text-container"></div>
      </div>
      <...error>
      <...resend>
    </div>
{{ end }}
```

tung2744:
@IniZio Please rebase too :adore:

louischan-oursky (commented):
https://github.com/authgear/authgear-server/assets/15082564/b099d309-3962-4931-b50d-9db452698e0e

The input still have some weird behaviors

## When the input is focused, it should allow me to append digit only.

Focus the input, type 3 digits. Blur. Focus again by tapping the first box. The first box is highlighted. I think the input should always highlight the last box that has digit. Also when the input is focused, it should always append digit, never inserting digit in the middle.

## When the input is focused, it should only allow me to remove digits from right to left

I think this is related to the above problem. It seems that it is now possible to place the cursor in the middle of the value. Thus pressing backspace sometimes do not delete digits from the right.

authui/src/authflowv2/components.css
tung2744: I've implemented link in body text by styling `<a>` inside body-text in the signup screen pr.

louischan-oursky:
@IniZio I added a commit to simplify the implementation.

1. Remove unnecessary instance property. Use getter instead.
2. Achieve the behavior we want by manipulate the cursor position. No need to use keydown event.

Issue #3560, opened 2024-01-08
AuthUI V2 - Implement Password Field


tung2744:
Based on:
- https://github.com/authgear/authgear-server/issues/3557

![Image](https://github.com/authgear/authgear-server/assets/34942210/1a8c1b59-168a-4768-9290-10ff5458701a)

- Error state should be supported
- Password strength meter should be implemented, with on / off config
- Can toggle visibility using the "eye" button

Ngkaokis:
Any detail requirement on the Password strength meter ? @tung2744

tung2744:
@Ngkaokis The design is here:
https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=1109-21880&mode=design&t=BCEU1x8kYuE0BIdk-4

But I found it only have 4 level
- Strong
- Medium
- Weak
- Very Weak
where our existing meter (See __password_strength_meter.html) has 5 level.

I think the missing level is "Very Strong" and the color can be the same as Strong.

@arvinhoux @fungc-io See if any opinion?

Ngkaokis:
Do you expect to reuse `password-policy.ts` / `passwordVisibility.ts`? No need to have a v2 version right?

tung2744:
@Ngkaokis 
- `password-policy.ts`: It seems we don't need it in v2, because I didn't see we any password policy checklist in figma.
- `passwordVisibility.ts`: You can reuse it in password field if you find it useful.

Ngkaokis:
> password-policy.ts: It seems we don't need it in v2, because I didn't see we any password policy checklist in figma.

Oh right, it check the password point by point. But what I want to ask is that is there a reference on determining the strength of password? Like the password doesn't have symbol so it's a weak password etc.

@tung2744

PR #3614, opened 2024-01-12
Add password input component 

Ngkaokis:
ref #3560 

https://github.com/authgear/authgear-server/assets/63200569/0df11876-61fa-487a-a013-3decef5f139c


resources/authgear/templates/en/translation.json
```diff
@@ -270,6 +273,10 @@
   "password-policy-password-strength-meter-3": "Fair",
   "password-policy-password-strength-meter-4": "Very unguessable",
   "password-policy-password-strength-meter-5": "Extremely unguessable",
+  "v2-password-policy-password-strength-meter-1": "Very Weak",
+  "v2-password-policy-password-strength-meter-2": "Weak",
+  "v2-password-policy-password-strength-meter-3": "Medium",
+  "v2-password-policy-password-strength-meter-4": "Strong",
```
tung2744, line 279: - As mentioned in slack, we need 5 level.
- Please put all v2-.. translation strings at the bottom.

authui/src/authflowv2/password-policy.ts
```diff
@@ -0,0 +1,64 @@
...
+  const score = Math.min(4, Math.max(1, result.score + 1));
+  let strengthClass = "";
+  switch (score) {
+    case 1:
+      strengthClass = "password-strength-meter--very-weak";
+      input.classList.add("input--error");
+      break;
+    case 2:
+      strengthClass = "password-strength-meter--weak";
+      input.classList.add("input--error");
```
tung2744, line 33: I think error state shouldn't be related to password strength.
Some project can be configured to accept weak password.

resources/authgear/templates/en/web/authflowv2/__password_input.html
```diff
@@ -0,0 +1,55 @@
...
+		class="w-5 absolute inset-y-0 right-4 hidden"
+		type="button"
+		tabindex="-1"
+		title="{{ template "hide-password-label" }}"
+		data-password-visibility-toggle-target="hideButton"
+		data-action="password-visibility-toggle#hide">
+      <span class="material-icons align-middle text-xl">
+        visibility
+      </span>
+	</button>
```
tung2744, line 53: The indentation here is a bit weird. (tab / space?)
Maybe convert all to tabs.

louischan-oursky: I think this should be done by a tool, and be enforced on CI (we currently do not have it yet).

louischan-oursky: This website uses Go template https://github.com/golang/pkgsite I did not find any configuration related to formatting the go templates :(

authui/src/authflowv2/components/input.css
```diff
@@ -72,3 +74,7 @@
   letter-spacing: var(--input__error-message-letter-spacing);
   font-weight: var(--input__error-message-font-weight);
 }
+
+.input.input__password {
+  @apply pr-[length:var(--input\_\_password-pr)];
```
louischan-oursky, line 79: Use padding-inline-end instead of padding-right.

resources/authgear/templates/en/web/authflowv2/__password_input.html
```diff
@@ -0,0 +1,34 @@
...
+	<input
+		tabindex="0"
+		data-password-policy-target="input"
+		class="input input__password w-full"
+		type="password"
+		placeholder="Password"
+		data-password-visibility-toggle-target="input"
+	>
+	<button
+		class="w-5 absolute inset-y-0 right-4"
```
louischan-oursky: Use logical properties. https://tailwindcss.com/docs/top-right-bottom-left#using-logical-properties

authui/src/authflowv2/components/password-strength-meter.css
```diff
@@ -0,0 +1,65 @@
...
+    --password-strength-meter__line-height: var(
+      --typography-body-medium__line-height
+    );
+    --password-strength-meter__letter-spacing: var(
+      --typography-body-medium__letter-spacing
+    );
+    --password-strength-meter__font-weight: var(
+      --typography-body-medium__font-weight
+    );
+    --password-strength-meter__color--very-weak: var(--color-error);
```
louischan-oursky, line 18: Name these variables by zxcvbn score. `--password-strength-meter__color--zxcvbn-score-{0,1,2,3,4}`

resources/authgear/templates/en/web/authflowv2/__password_strength_meter.html
```diff
@@ -0,0 +1,13 @@
+{{ define "authflowv2/__password_strength_meter.html" }}
+<div class="{{ $.Classname }} password-strength-meter" data-password-policy-target="currentMeter">
```
louischan-oursky, line 2: In the original implementation, we use `<meter>` element. Any specific reason we did not follow?

Make use of `<meter value>` in the implementation. The job of the controller is to assign correct `value`. The style should be written against `<meter value>`. I mean, you can write selector that looks like `password-strength-meter[value=0] .password-strength-meter__someelement`

tung2744: But the new design is actually not a meter, ui wise 🤔

louischan-oursky: Visually it does not look like a meter, but semantically it is. We can use `appearance: none` to make it look like a div, and apply our design on it.

resources/authgear/templates/en/web/authflowv2/__password_strength_meter.html
```diff
@@ -1,14 +1,16 @@
...
-			class="password-strength-meter-description"
-			data-password-policy-target="currentMeterDescription"
-			data-desc-1="{{ template "v2-password-policy-password-strength-meter-1" }}"
-			data-desc-2="{{ template "v2-password-policy-password-strength-meter-2" }}"
-			data-desc-3="{{ template "v2-password-policy-password-strength-meter-3" }}"
-			data-desc-4="{{ template "v2-password-policy-password-strength-meter-4" }}"
-			data-desc-5="{{ template "v2-password-policy-password-strength-meter-5" }}"
-		></span>
-</div>
+<meter
```
louischan-oursky, line 2: Let's define a new controller responsible for the look of this meter. Let's call it PasswordStrengthMeterController. The interface of this Controller is

- Assume `element` is HTMLMeterElement.
- Assume `value` is in range `[1,5]`
- It changes the appearance according to `value`.

And PasswordPolicyController is responsible for setting the `value` of the meter only.

resources/authgear/templates/en/web/authflowv2/__password_strength_meter.html
```diff
@@ -1,16 +1,19 @@
 {{ define "authflowv2/__password_strength_meter.html" }}
 <meter
-  class="{{ $.Classname }} password-strength-meter w-full"
+  class="{{ $.Classname }} password-strength-meter"
   min="1"
   max="5"
+  data-controller="password-strength-meter"
   data-password-policy-target="currentMeter"
+  data-password-strength-meter-target="currentMeter"
```
louischan-oursky, line 8: We can just use `this.element` instead of adding a duplicate target. See `mirrorbutton.ts` as an example.

resources/authgear/templates/en/web/authflowv2/__password_strength_meter.html
```diff
@@ -1,16 +1,19 @@
...
+  data-action="password-strength-updated@window->password-strength-meter#display"
```
louischan-oursky, line 9: The original `<meter>` automatically renders when `value` changes. We should keep this capability. So instead of requiring another component to dispatch a CustomEvent, set up a mutation observer on `value` and update the rendering in PasswordStrengthMeterController. See `mirrorbutton.ts` as an example.

Issue #3561, opened 2024-01-08
AuthUI V2 - Implement Phone Input Field


tung2744:
![Image](https://github.com/authgear/authgear-server/assets/34942210/7f6de684-17a0-45f2-a24a-dca5f20920de)

- Responsive
    - Mobile: Full screen modal opened on click
    - Desktop: Dropdown opened on click
- Search (Frontend search only)
- Error state should be supportted.

fungc-io:
Design: https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=890-8744&mode=design&t=87GCqRocri2Fd5th-4

fungc-io:
Reference for flags:
- https://github.com/jackocnr/intl-tel-input?tab=readme-ov-file
- https://github.com/fonttools/region-flags

fungc-io:
- Other emoji font for reference: https://fonts.google.com/noto/specimen/Noto+Color+Emoji

IniZio:
Don't think Noto Color Emoji has variant that provides flag without wave.

Found Twitter emoji compiled as font https://github.com/13rac1/twemoji-color-font, tried the font and pretty close to current design

Preview provided on their repo:
![Image](https://github.com/authgear/authgear-server/assets/3119725/07ec8be4-e622-453b-93c6-e1c9bd6de2a3)

PR #3623, opened 2024-01-16
Implement phone input component

IniZio:
ref #3561

image
image


authui/package.json
```diff
@@ -41,6 +41,7 @@
     "@sentry/tracing": "7.80.0",
     "@tabler/icons": "1.107.0",
     "axios": "1.6.1",
+    "countries-list": "^3.0.6",
```
louischan-oursky, line 45: Use libphonenumber-js and cldr-localenames-full instead. It is based on Google's libphonenumber.
https://gitlab.com/catamphetamine/libphonenumber-js
https://github.com/unicode-org/cldr-json/tree/main/cldr-json/cldr-localenames-full

In particular, please

1. `import "libphonenumber-js/max"` to ensure we are using the full data set
2. Use `getCountries()` and `getCountryCallingCode()` to generate the country list.
3. Use `cldr-localenames-full` to get the localized locale names in the **CURRENT** language. The current language is `<html lang>`
4. Sort the list according to localized locale names.

And then follow this algorithm to get the emoji flag code point https://github.com/annexare/Countries/blob/main/packages/countries/src/getEmojiFlag.ts

Example

```
String.fromCodePoint(127462 - 'A'.charCodeAt(0) + 'H'.charCodeAt(0), 127462 - 'A'.charCodeAt(0) + 'K'.charCodeAt(0)) // 🇭🇰
```

Finally we can get rid of `countries-list`.

authui/src/authflowv2/components/phone-input.css
```diff
@@ -0,0 +1,134 @@
...
+  :root {
+    --phone-input__space-x: 0.5em;
+    --phone-input__text-color: var(--color-grey-700);
+    --phone-input__font-size: var(--typography-body-large__font-size);
+    --phone-input__line-height: var(--typography-body-large__line-height);
+    --phone-input__letter-spacing: var(--typography-body-large__letter-spacing);
+    --phone-input__font-weight: var(--typography-body-large__font-weight);
+    --phone-input__font-family: twemoji-countries, Inter;
+
+    --phone-input__trigger__ring-color: var(--color-grey-100);
```
louischan-oursky, line 11: We follow BEM and BEM does not have two elements. So it should be named as `--phone-input__trigger-ring-color`.

authui/src/authflowv2/components/phone-input.css
```diff
@@ -0,0 +1,134 @@
...
+
+  .phone-input__options {
+    @apply flex;
+    @apply flex-col;
+    @apply p-4;
+
+    max-height: 288px;
+    overflow-y: auto;
+
+    @apply space-y-[var(--phone-input\_\_options\_\_space-y)];
```
louischan-oursky: use gap.

authui/src/customSelect.ts
```diff
@@ -1,38 +1,160 @@
...
+    }).then(({ x, y }) => {
+      Object.assign(this.dropdownTarget.style, {
+        left: `${x}px`,
+        top: `${y}px`,
+      });
+    });
+
+    this.triggerTarget.addEventListener("keydown", this.handleKeyDown);
+    this.dropdownTarget.addEventListener("keydown", this.handleKeyDown);
+    document.addEventListener("click", this.handleClickOutside);
```
louischan-oursky, line 71: Have you tried https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog ? Does it work?

authui/src/authflowv2/components/phone-input.css
```diff
@@ -103,6 +103,42 @@
...
+  .phone-input__search {
+    @apply relative;
+    @apply flex;
+    @apply flex-row;
+    @apply items-center;
+
+    border-bottom: 1px solid var(--color-grey-100);
+  }
+
+  .phone-input__search-input {
```
louischan-oursky: Can we reuse the input field component here?

authui/src/authflowv2/components/phone-input.css
```diff
@@ -20,30 +20,11 @@
...
-    --phone-input__options-height: 100%;
-    --phone-input__options__space-y: 0.5em;
-
-    --phone-input__option__border-radius: calc(
-        var(--phone-input__font-size) * 0.75
-      )
-      /* 12 / 16 ratio to font size */;
-    --phone-input__option__background-color--active: var(--color-grey-100);
-
+    --phone-input__error-message-margin-top: 0.5em;
```
louischan-oursky, line 27: `error-message-spacing`

authui/src/customSelect.ts
```diff
@@ -50,6 +50,7 @@ export class CustomSelectController extends Controller {
 
   keyword = "";
   value?: string;
+  focusedValue?: string;
```
louischan-oursky, line 53: Do not introduce redundant instance properties when the source of truth is the attributes of some elements in our control.

louischan-oursky (commented):
Just left some preliminary comments.

authui/src/customSelect.ts
IniZio: Couldn't get focus to work when using dialog, will try again after addressing other comments.

authui/src/authflowv2/components/phone-input.css
IniZio: The search input's style is pretty different from the input component, or you meant adding a variant to input?

authui/src/customSelect.ts
IniZio: Able to make dialog work with current country select. 

The dialog features used: `::backdrop`, `.show` should have ~95% on browser support.

Issue #3622, opened 2024-01-15
AuthUI V2 - Signup (Integrate Text Field & Phone Input)


tung2744:
- [x] Integrate text input
- [ ] Integrate phone input
- [x] Handle errors

PR #3625, opened 2024-01-16
Integrate input field into signup screen with error

tung2744:
ref #3622 

https://github.com/authgear/authgear-server/assets/34942210/78d18531-a2b6-4987-a0aa-3196bb13c8d3


resources/authgear/templates/en/web/authflowv2/signup.html
```diff
@@ -50,38 +55,47 @@ <h2 class="screen-description">
...
+              inputmode="tel"
+              autocomplete="tel-national"
+              autocapitalize="none"
+              name="q_login_id"
+              placeholder="{{ template "v2-placeholder.phone" }}"
+              data-retain-form-form-target="input"
+              data-retain-form-form-name-param="phone"
+            >
+          {{ else }}
+            <!-- TODO(tung): Update to use text input (https://github.com/authgear/authgear-server/issues/3557) -->
```
Ngkaokis, line 76: Remove comment if already using the component

resources/authgear/templates/en/web/authflowv2/signup.html
```diff
@@ -50,38 +55,47 @@ <h2 class="screen-description">
...
+              data-controller="retain-form-input"
+              data-action="input->retain-form-input#input retain-form-input:input->retain-form-form#input"
+              data-retain-form-form-target="input"
+              data-retain-form-form-name-param="text"
+            >
+          {{ end }}
+          {{ if $display_error }}
+            <ul class="input__error-message mt-2" data-text-field-target="errorMessage">
+              {{ template "authflowv2/__error.html" $ }}
+            </ul>
```
Ngkaokis, line 96: Not sure if we should use `ul` here. Because ultimately we will only display one message?

resources/authgear/templates/en/web/authflowv2/signup.html
```diff
@@ -82,6 +89,11 @@ <h2 class="screen-description">
             data-retain-form-form-name-param="text"
           >
         {{ end }}
+        {{ if $display_error }}
+          <ul class="input__error-message mt-2">
```
louischan-oursky: 1. We should use gap instead of `mt-2`. There are some other places using margin-top as well. We should correct those (Maybe in another issue)
2. By using gap, the usage should become

```html
<div class="input__container">
  <input class="input">
  <ul class="input__error-message">{{ template "authflowv2/__error.html" $ctx }}</ul>
</div>
```

authui/src/authflowv2/components/text-field.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.inputTarget.removeEventListener("input", this.onInput);
+  }
+
+  onInput = () => {
+    if (this.inputTarget.classList.contains(this.inputErrorClassValue)) {
+      this.inputTarget.classList.remove(this.inputErrorClassValue);
+    }
+
+    if (this.errorMessageTarget != null) {
+      this.errorMessageTarget.style.display = "none";
```
louischan-oursky, line 27: Set the tailwind class `hidden` to hide the element.

resources/authgear/templates/en/web/authflowv2/signup.html
tung2744: @Ngkaokis @IniZio @louischan-oursky 
I think it depends on what __error.html do.
In this pr, it creates <li> elements, so I used `<ul>` here. The advantage is it will be able to handle multiple error messages. The disadvantage is there is an assumption that it must be used as html element.
We can change __error.html to render texts, so it can be used in attributes too. See if any opinion on it?

Ngkaokis: How do you expect to handle form with multiple fields? For example, in create password screen, there are enter password field and confirm password field and both can display corresponding error message. Do you expect to have two `__error.html` under each field and conditionally render each one?

authui/src/authflowv2/components/text-field.ts
tung2744: ok. But if we use tailwind css class in typescript files, I think we need to include the file paths in authui/tailwind.config.js so that tailwind won't remove the classes used in script.

resources/authgear/templates/en/web/authflowv2/signup.html
Ngkaokis: > I think it depends on what __error.html do.
> In this pr, it creates `<li>` elements, so I used `<ul>` here. 

But it will be something look like the following eventually in the dom?
```
<ul>
<li>Error Message</li>
</ul>
```

The `<ul>` and `<li>` seems not necessary?
Seem just `<p>ErrorMessage</p>` is enough?

tung2744: @Ngkaokis I think there should only be one place to display the error message. For example, in create password screen, the error message should always be under the first text input.

However, if there are really cases we need two error messages, we need to conditionally render the template with a correct context.

Ngkaokis: > I think there should only be one place to display the error message. For example, in create password screen, the error message should always be under the first text input.

Oh, I thought error like confirm password not the same as the above one would be displayed under the second one

resources/authgear/templates/en/web/authflowv2/signup.html
tung2744: I think it is we cannot replace all margin-top with gap, because some elements are not having equal spacing, but for text field we can use gap here.

authui/src/authflowv2/components/text-field.ts
louischan-oursky: I meant using the literal `"hidden"` only. I guess it is available by default. No need to mess with typescript.

tung2744: Tailwind removes all classes that are unused from the generated stylesheet, so you need to tell it where to find those classes:
https://tailwindcss.com/docs/content-configuration#class-detection-in-depth

I think I just add `"./src/**/*.{css,scss,ts}"` to the tailwind config.

louischan-oursky: I see. Though `hidden` is used in many places already.

tung2744:
@louischan-oursky @Ngkaokis  Updated for all comments above, thanks!

resources/authgear/templates/en/web/authflowv2/signup.html
tung2744: For the actual error position, I think we can discuss it on implementing create password screen.
Maybe let me change `__error.html` to render text only first which seems more flexible.

louischan-oursky:
And this conflicts again :(

tung2744:
Rebased to latest authflowv2

Issue #3565, opened 2024-01-09
AuthUI V2 - Countdown Controller


tung2744:
![Image](https://github.com/authgear/authgear-server/assets/34942210/4b9eb237-5247-4bd3-a7ea-3b43d68b37ad)

![Image](https://github.com/authgear/authgear-server/assets/34942210/4943b820-019a-45c0-9530-496d3f2e044d)

- Basically same as `ResendButtonController`, but not only for resend.
- Support disabling a button before count down finish, and reenable it after finish.

IniZio:
@tung2744 What is different beside renaming, want to make it extendable as well? 🙏

tung2744:
@IniZio Basically the same, but the original implementation is only for resend button. I think we might want a more generic implementation which:
- Do not assume the host element is a button element, use target to specify the button to disable instead.
- Except `cooldown` which is the original relative time, also accept using a timestamp as a absolute time. <- Can be used in account lockout

PR #3636, opened 2024-01-17
Add countdown stimulus controller

IniZio:
ref #3565


louischan-oursky (approved):
LGTM

authui/src/countdown.ts
```diff
@@ -0,0 +1,87 @@
...
+      const formatted = duration.toFormat(format);
+
+      if (seconds === 0) {
+        button.disabled = false;
+        button.textContent = label;
+        this.animHandle = null;
+      } else {
+        button.disabled = true;
+        button.textContent = labelUnit
+          .replace("%d", String(seconds))
```
tung2744, line 55: I think we don't need to display the raw seconds. Always display a formatted duration, and give a default format to the format value which only displays "mm:ss"

PR #3646, opened 2024-01-18
Add "include" for executing template as a string

tung2744:
ref #3644 

Usage

```
{{ $errorMessage := include "authflowv2/__error.html" . }}
```

### DIFF
--- a/pkg/util/template/engine.go
+++ b/pkg/util/template/engine.go
@@ -2,13 +2,14 @@ package template
 
 import (
 	"fmt"
-	"golang.org/x/text/language"
 	htmltemplate "html/template"
 	"net/http"
 	"strconv"
 	"strings"
 	texttemplate "text/template"
 	"text/template/parse"
+
+	"golang.org/x/text/language"
 
 	"github.com/authgear/authgear-server/pkg/util/intl"
 	"github.com/authgear/authgear-server/pkg/util/messageformat"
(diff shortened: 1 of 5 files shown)


IniZio:
Not sure if this is intended to generate html content or text node? Currently if run on `authflowv2/__error.html`, it will return as text node `<span>....</span>`

tung2744:
It should return html content. Let me check how to allow html content in go template.

louischan-oursky:
See if this is what you are looking for? https://pkg.go.dev/html/template#HTML

IniZio:
Seems works :adore2x:

```go
func makeInclude(t tpl) func(tplName string, data any) (template.HTML, error) {
	return func(
		tplName string,
		data any,
	) (template.HTML, error) {
		buf := &bytes.Buffer{}
		err := t.ExecuteTemplate(buf, tplName, data)
		return template.HTML(buf.String()), err
	}
}
```

tung2744:
Updated to return HTML :adore:

louischan-oursky:
Can we add some tests?

1. To show that it supports variable template name.
```
{{ $var := "temp2" }}
{{ template "temp1" (dict "Prop1" (include $var (dict "Prop2" "world"))) }}
```

2. To show that it supports HTML

tung2744:
Added two commits to fix linting issues and add tests

louischan-oursky:
Added one more test to showcase nesting.

Issue #3578, opened 2024-01-09
AuthUI V2 - Login: Enter Password


tung2744:
https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=717-22404&mode=design&t=t2ocn0LIQuJP59g8-4

PR #3654, opened 2024-01-19
Add enter password screen

Ngkaokis:
ref #3578 

require #3645 

https://github.com/authgear/authgear-server/assets/63200569/114ef0ed-09ce-412d-ac49-fdf54a674ca0


authui/src/authflowv2/components/primary-button.css
```diff
@@ -38,6 +38,19 @@
     /* TODO(tung): Dark mode styles */
   }
 
+  .primary-btn__content {
+    @apply block;
+  }
+
+  .primary-btn__loading {
+    @apply hidden;
+    height: calc(1.5 * var(--primary-btn__font-size)) /* 24 / 16 = 1.5 */;
```
tung2744, line 47: Use em to calculate something relative to the font-size of the element.

pkg/auth/handler/webapp/authflowv2/enter_password.go
```diff
@@ -22,9 +22,9 @@ var AuthflowEnterPasswordSchema = validation.NewSimpleSchema(`
 	{
 		"type": "object",
 		"properties": {
-			"x_password": { "type": "string" }
+			"q_password": { "type": "string" }
```
tung2744, line 25: Is there any reason we use `q_password` instead of `x_password`?
In my understanding `q_` should only be used for query paramenters.

resources/authgear/templates/en/web/authflowv2/__authflow_branch.html
```diff
@@ -0,0 +1,152 @@
...
+<input type="hidden" name="x_channel" value="{{ .Channel }}">
+<input type="hidden" name="x_action" value="take_branch">
+{{ if ($.Translations.HasKey "customer-support-link") }}
+    {{ template "enter-recovery-code-instead-with-customer-support-v1" (dict "customerSupportLink" ($.Translations.RenderText "customer-support-link" nil)) }}
+{{ else }}
+    {{ template "enter-recovery-code-instead-v1" }}
+{{ end }}
+</form>
+{{- end }}
+{{- end }}
```
tung2744, line 25: Increase one indentation level for each if / for block if they are not on the same line, or else it could be difficult to read.

resources/authgear/templates/en/web/authflowv2/enter_password.html
```diff
@@ -1,122 +1,68 @@
...
+    {{ else }}
+      {{ template "v2-enter-password-page-title" }}
+    {{ end }}
+	{{ end }}
+</h1>
+  <form method="POST" novalidate class="flex flex-col gap-y-4 items-center">
+  {{ $.CSRFField }}
+  <!-- This field is for Chrome and Safari to correctly associate the username with the password -->
+  {{ if $.PasswordManagerUsername }}
+  <input style="display: none" aria-hidden="true" type="text" autocomplete="username" name="" value="{{ $.PasswordManagerUsername }}">
```
tung2744, line 23: We can use `class="hidden"`

resources/authgear/templates/en/web/authflowv2/enter_password.html
```diff
@@ -1,122 +1,68 @@
...
+    data-action-button
+  >
+  <span class="primary-btn__content">
+    {{ template "v2-button.continue" }}
+  </span>
+  <span class="primary-btn__loading">
+    <span class="primary-btn__loading-icon material-icons animate-spin">
+      progress_activity
+    </span>
+  </span>
```
tung2744, line 54: We should create a controller to handle the loading state of the button.
We should modify `XHRSubmitFormController` so that we can distinguish a normal disable and a loading disable of the button, probably by setting some additional attributes.
And the new controller should read that attribute, and display the spinner when needed.
We can created another issue for this because it affects many pages.

pkg/auth/handler/webapp/authflowv2/enter_password.go
louischan-oursky: There are pages not following this rule, for example authflow_signup.go
Actually, using `q_` or `x_` does not matter in a form. The distinction only matters in the query parameter.

Ngkaokis: Talked with Louis before and I misunderstood that I should use `q_`. Let me stick with `x_` then.

tung2744:
Screenshot 2024-01-23 at 1 29 13 PM

The spacing around forgot password button seems incorrect. I think we can use margins for these arbitrary spacing.

tung2744:
Added 5db4415 to fix the above problem

Issue #3596, opened 2024-01-09
AuthUI V2 - Signup: Signup Disallowed


tung2744:

tung2744:
@fungc-io @arvinhoux 

There is a feature to disable public signup. We may want a ui to display when someone access the signup page when public signup is disabled.

Or I guess we can simply reuse the generic error screen? https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=738-25095&mode=design&t=t2ocn0LIQuJP59g8-4


![Image](https://github.com/authgear/authgear-server/assets/34942210/b26da48f-a67f-496a-ac03-a38fa4e00783)

fungc-io:
Use the generic error page with the following copy:

"Unexpected Issues"
"Signup is disabled. Please contact system admin."
No button, the user can only cancel or close this page.

louischan-oursky:
@IniZio I believe this flag is not currently honored in authflow. We need to block public signup at API level.

IniZio:
Yes am trying to add, thanks 🙏

louischan-oursky:
Note that we have `signup` and `signup_login` flow. If public signup is disabled, then the creation of `signup` and `signup_login` flow should be disallowed.

PR #3665, opened 2024-01-22
Add fatal error message for public signup disabled

IniZio:
ref #3596

Screenshot 2024-01-22 at 16-09-39 Authgear

### DIFF
--- a/pkg/auth/handler/webapp/authflow_controller.go
+++ b/pkg/auth/handler/webapp/authflow_controller.go
@@ -13,6 +13,7 @@ import (
 	"github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/auth/webapp"
 	authflow "github.com/authgear/authgear-server/pkg/lib/authenticationflow"
+	"github.com/authgear/authgear-server/pkg/lib/authenticationflow/declarative"
 	"github.com/authgear/authgear-server/pkg/lib/authn/user"
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/lib/oauth/oauthsession"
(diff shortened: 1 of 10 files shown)


tung2744:
Please also update resources/authgear/templates/en/web/authflowv2/signup.html

IniZio:
Updated to remove the condition in signup page

resources/authgear/templates/en/web/authflowv2/__error.html
```diff
@@ -5,6 +5,7 @@
   {{ end }}
   {{- if $display_error -}}
     {{ $info := .Error.info }}
+    {{ .Error }}???
```
louischan-oursky, line 8: Typo?

louischan-oursky:
Now we will see this error `ERRO[0066] failed to create screen                       app=accounts error="public signup is disabled"` and it will become Sentry log. It is not desirable.

louischan-oursky:
This manual exclusion is not very suitable. But let's accept this for now. I have created https://github.com/authgear/authgear-server/issues/3684

pkg/auth/handler/webapp/authflow_controller.go
```diff
@@ -152,7 +153,10 @@ func (c *AuthflowController) HandleStartOfFlow(
 	if err != nil {
 		if errors.Is(err, authflow.ErrFlowNotFound) {
 			screen, err := c.createScreen(r, s, flowReference, input)
-			if err != nil {
+			if errors.Is(err, declarative.ErrNoPublicSignup) {
+				c.renderError(w, r, err)
```
tung2744, line 157: Should update makeErrorResult to explicitly mark ErrNoPublicSignup as non recoverable error?

Issue #3630, opened 2024-01-16
Auth UI V2 - Make existing components or layout writing-mode friendly


louischan-oursky:
Require https://github.com/authgear/authgear-server/issues/3628

PR #3669, opened 2024-01-22
RTL support

louischan-oursky:
ref #3630

### DIFF
--- a/authui/src/authflowv2/components/select.css
+++ b/authui/src/authflowv2/components/select.css
@@ -18,10 +18,9 @@
 
     --select__search-border: 1px solid var(--color-grey-100);
 
-    --select__search-input-padding: 0.875em 0.625em;
-    @media (max-width: theme("screens.tablet")) {
-      --select__search-input-padding: 0.875em 0.125em;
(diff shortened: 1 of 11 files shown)


authui/src/authflowv2/components/select.css
```diff
@@ -18,14 +18,45 @@
 
     --select__search-border: 1px solid var(--color-grey-100);
 
-    --select__input-padding: 0.75em 1em 0.75em 3.25em;
+    --select__search-input-padding: 0.875em 0.625em;
+    @media (max-width: theme("screens.tablet")) {
+      --select__search-input-padding: 0.875em 0.125em;
+    }
```
tung2744, line 24: According to our discussion yesterday, it should be two variables instead of media query?
```
--select__search-input-padding--mobile
--select__search-input-padding--desktop
```

louischan-oursky: Addressed in https://github.com/authgear/authgear-server/pull/3669/commits/2488054563bfea4ce7410ecace52c1618ca65498

Issue #3593, opened 2024-01-09
Auth UI V2 - Signup/Login: SSO login


tung2744:
![Image](https://github.com/authgear/authgear-server/assets/34942210/3bd50940-40a7-4b09-9d29-ea0ae88b6fc0)

- Implement sso in signup
- Implement sso in login
- Ensure SSOCallbackHandler navigate to correct route according to ui implementation

Non-goal:
  - Wechat login is handled in another issue

PR #3671, opened 2024-01-22
Auth ui v2 - Add SSO login

tung2744:
ref #3593 

Screenshot 2024-01-22 at 6 27 05 PM

### DIFF
--- a/authui/src/authflowv2/icons/sso-icons.css
+++ b/authui/src/authflowv2/icons/sso-icons.css
@@ -17,7 +17,9 @@
 .sso-icon.whatsapp-icon {
   background-image: url(./icons/raw/icon-whatsapp.svg);
 }
-.sso-icon.microsoft-icon {
+.sso-icon.azureadv2-icon,
+.sso-icon.azureadb2c-icon,
+.sso-icon.adfs-icon {
   background-image: url(./icons/raw/icon-microsoft.svg);
 }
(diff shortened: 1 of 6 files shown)


pkg/auth/handler/webapp/authflow_controller.go
```diff
@@ -187,7 +187,7 @@ func (c *AuthflowController) HandleOAuthCallback(w http.ResponseWriter, r *http.
 		return
 	}
 
-	s, err := c.getWebSession(r)
+	s, err := c.Sessions.Get(state.WebSessionID)
```
tung2744, line 190: This is a bug which I think also affects authflow v1. If we are not going to deploy v2 onto staging shortly, maybe we need to pick this fix to main.

louischan-oursky: I could not reproduce this bug on latest of authflowv2

louischan-oursky:
Other parts look good to me.

pkg/auth/handler/webapp/authflow_controller.go
tung2744: I tested again but cannot find how to reproduce the bug. Maybe I will drop this commit first.

louischan-oursky:
Rebase.

tung2744:
Rebased, thanks!

PR #3675, opened 2024-01-23
Add editorconfig config

IniZio:
Noticed quite some files e.g. `resources/authgear/templates/en/web/authflowv2/setup_oob_otp.html`, `resources/authgear/templates/en/web/authflowv2/login.html` got tabs accidentally added.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = false
+
+[Makefile]
+indent_style = tab


.editorconfig
```diff
@@ -0,0 +1,15 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
```
louischan-oursky, line 6: Not familiar with editorconfig. Does this line mean applying to all files in this repository?

IniZio: Yes, see if will want to be more conservative and only cover authui v2 code first?

louischan-oursky:
If the intention is to align the format of those html and css files, then I suggest we set up prettier instead. But prettier does not know how to format Go template :(

IniZio:
Yes I am adding prettier and other lint stuff in https://github.com/authgear/authgear-server/issues/3666. Don't think it will help with the gotcha I mentioned in this pr though :/

louischan-oursky:
Let's put our effort in getting auth flow v2 done first. Formatting is easy when we have a right tool.

Ngkaokis:
Found this https://github.com/NiklasPor/prettier-plugin-go-template

Does it help?

IniZio:
Will close this pr first.

Issue #3676, opened 2024-01-23
Handle loading state of button globally


Ngkaokis:
>We should create a controller to handle the loading state of the button.
We should modify XHRSubmitFormController so that we can distinguish a normal disable and a loading disable of the button, probably by setting some additional attributes.
And the new controller should read that attribute, and display the spinner when needed.
We can created another issue for this because it affects many pages.

https://github.com/authgear/authgear-server/pull/3654#discussion_r1461778200

figma 
https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=1118-26749&mode=design&t=MsEmFWFXKDuv5l6c-4

PR #3678, opened 2024-01-23
Add loading state to button for enter password page

Ngkaokis:
ref #3676

https://github.com/authgear/authgear-server/assets/63200569/dc0d4952-b855-44d1-b5d8-b81755c39a4d

### DIFF
--- a/authui/src/authflowv2/components/primary-button.css
+++ b/authui/src/authflowv2/components/primary-button.css
@@ -79,4 +79,10 @@
         ring-[color:var(--primary-btn\_\_ring-color--disabled)];
     }
   }
+
+  .primary-btn__loading-icon {
+    font-size: calc(1.5 * var(--primary-btn__font-size)) /* 24 / 16 = 1.5 */;
+    height: 1em;
+    display: block !important;
+  }
 }
--- a/authui/src/authflowv2/components/primary-button.css
+++ b/authui/src/authflowv2/components/primary-button.css
@@ -38,6 +38,16 @@
     /* TODO(tung): Dark mode styles */
   }
 
+  /* NOTE: @apply animate-spin raises error @parcel/transformer-css: Unknown at rule: @keyframes. So define it here */
+  @keyframes spin {
+    from {
+      transform: rotate(0deg);
+    }
+    to {
+      transform: rotate(360deg);
+    }
+  }
+
   .primary-btn {
     @apply block;
     @apply text-center;
@@ -78,11 +88,21 @@
         ring-[length:var(--primary-btn\_\_ring-width--disabled)]
         ring-[color:var(--primary-btn\_\_ring-color--disabled)];
     }
-  }
 
-  .primary-btn__loading-icon {
-    font-size: calc(1.5 * var(--primary-btn__font-size)) /* 24 / 16 = 1.5 */;
-    height: 1em;
-    display: block !important;
+    &:after {
+      @apply material-icons align-middle absolute inset-x-0;
+      animation: spin 1s linear infinite;
+
+      font-size: calc(1.5 * var(--primary-btn__font-size)) /* 24 / 16 = 1.5 */;
+      color: var(--primary-btn__text-color--disabled);
+      height: 1em;
+      content: "progress_activity";
+    }
+    &.primary-btn--loading {
+      color: transparent;
+    }
+    &:not(.primary-btn--loading)::after {
+      @apply hidden;
+    }
   }
 }
--- a/authui/src/authflowv2/components/primary-button.css
+++ b/authui/src/authflowv2/components/primary-button.css
@@ -38,16 +38,6 @@
     /* TODO(tung): Dark mode styles */
   }
 
-  /* NOTE: @apply animate-spin raises error @parcel/transformer-css: Unknown at rule: @keyframes. So define it here */
-  @keyframes spin {
-    from {
-      transform: rotate(0deg);
-    }
-    to {
-      transform: rotate(360deg);
-    }
-  }
-
   .primary-btn {
     @apply block;
     @apply text-center;
@@ -88,21 +78,22 @@
         ring-[length:var(--primary-btn\_\_ring-width--disabled)]
         ring-[color:var(--primary-btn\_\_ring-color--disabled)];
     }
+  }
 
-    &:after {
-      @apply material-icons align-middle absolute inset-x-0;
-      animation: spin 1s linear infinite;
+  .primary-btn::after {
+    @apply material-icons align-middle absolute inset-x-0 animate-spin;
 
-      font-size: calc(1.5 * var(--primary-btn__font-size)) /* 24 / 16 = 1.5 */;
-      color: var(--primary-btn__text-color--disabled);
-      height: 1em;
-      content: "progress_activity";
-    }
-    &.primary-btn--loading {
-      color: transparent;
-    }
-    &:not(.primary-btn--loading)::after {
-      @apply hidden;
-    }
+    font-size: calc(1.5 * var(--primary-btn__font-size)) /* 24 / 16 = 1.5 */;
+    color: var(--primary-btn__text-color--disabled);
+    height: 1em;
+    content: "progress_activity";
+  }
+
+  .primary-btn[data-loading="true"] {
+    color: transparent;
+  }
+
+  .primary-btn:not([data-loading="true"])::after {
+    @apply hidden;
   }
 }
--- a/authui/src/authflowv2/form.ts
+++ b/authui/src/authflowv2/form.ts
@@ -1,13 +1,9 @@
 import { visit, clearCache } from "@hotwired/turbo";
 import { Controller } from "@hotwired/stimulus";
 import axios, { Method } from "axios";
-import {
-  disableAllButtons,
-  hideProgressBar,
-  showProgressBar,
-  progressEventHandler,
-} from "../loading";
+import { progressEventHandler } from "../loading";
 import { handleAxiosError } from "../messageBar";
+import { LoadingController } from "./loading";
 
 // Turbo has builtin support for form submission.
 // We once migrated to use it.
@@ -28,7 +24,6 @@ import { handleAxiosError } from "../messageBar";
 //
 // See https://github.com/authgear/authgear-server/issues/2333
 export class XHRSubmitFormController extends Controller {
-  revertDisabledButtons: { (): void } | null = null;
   forms: HTMLFormElement[] = [];
 
   onSubmitCapture = (e: Event) => {
@@ -55,18 +50,14 @@ export class XHRSubmitFormController extends Controller {
     formData.forEach((value, name) => {
       params.set(name, value as string);
     });
-    const loadingEvent = new CustomEvent<{ button?: HTMLButtonElement }>(
-      "onLoading",
-      {
-        detail: {},
-      }
-    );
+    let loadingButton: HTMLButtonElement | null = null;
     // FormData does not include any submit button's data:
     // include them manually, since we have at most one submit button per form.
     const submitButtons = form.querySelectorAll('button[type="submit"]');
     for (let i = 0; i < submitButtons.length; i++) {
       const button = submitButtons[i] as HTMLButtonElement;
       params.set(button.name, button.value);
+      loadingButton = button;
     }
     if (form.id) {
       const el = document.querySelector(
@@ -75,11 +66,16 @@ export class XHRSubmitFormController extends Controller {
       if (el) {
         const button = el as HTMLButtonElement;
         params.set(button.name, button.value);
-        loadingEvent.detail.button = button;
+        loadingButton = button;
       }
     }
-
-    this.element.dispatchEvent(loadingEvent);
+    const loadingController: LoadingController | null =
+      this.application.getControllerForElementAndIdentifier(
+        this.element,
+        "loading"
+      ) as LoadingController | null;
+    const { onError: onLoadingError, onFinally: onLoadingFinally } =
+      loadingController?.startLoading(loadingButton) ?? {};
     try {
       const resp = await axios(form.action, {
         method: form.method as Method,
@@ -109,15 +105,9 @@ export class XHRSubmitFormController extends Controller {
       }
     } catch (e: unknown) {
       handleAxiosError(e);
-      this.element.dispatchEvent(
-        new CustomEvent("onLoadingError", {
-          detail: {
-            error: e,
-          },
-        })
-      );
+      onLoadingError?.();
     } finally {
-      this.element.dispatchEvent(new CustomEvent("onLoadingEnd"));
+      onLoadingFinally?.();
     }
   }
 
--- a/authui/src/authflowv2/form.ts
+++ b/authui/src/authflowv2/form.ts
@@ -1,0 +1,144 @@
+import { visit, clearCache } from "@hotwired/turbo";
+import { Controller } from "@hotwired/stimulus";
+import axios, { Method } from "axios";
+import {
+  disableAllButtons,
+  hideProgressBar,
+  showProgressBar,
+  progressEventHandler,
+} from "../loading";
+import { handleAxiosError } from "../messageBar";
+
+// Turbo has builtin support for form submission.
+// We once migrated to use it.
+// However, redirect to external is broken because
+// the redirect is made with fetch, which is subject to CORS.
+// A typical problem is the support for the post login redirect URI.
+// The post login redirect URI is usually an external link that
+// the origin of the link does not list our origin as allowed origin.
+// If we rely on Turbo to handle form submission,
+// the post login redirect URI will be broken.
+// Therefore, we reverted to handle form submission ourselves.
+// To disable the builtin form submission of Turbo,
+// I studied its source code and discovered that
+// Turbo checked in the bubble event listener to see if
+// the event is prevented.
+// So I added a capture event listener to call preventDefault()
+// to stop Turbo from submitting forms.
+//
+// See https://github.com/authgear/authgear-server/issues/2333
+export class XHRSubmitFormController extends Controller {
+  revertDisabledButtons: { (): void } | null = null;
+  forms: HTMLFormElement[] = [];
+
+  onSubmitCapture = (e: Event) => {
+    e.preventDefault();
+  };
+
+  onSubmit = (e: Event) => {
+    this.submitForm(e);
+  };
+
+  async submitForm(e: Event) {
+    const form = e.currentTarget as HTMLFormElement;
+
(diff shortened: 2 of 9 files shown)


authui/src/loading.ts
```diff
@@ -21,6 +21,18 @@ export function disableAllButtons(): () => void {
   };
 }
 
+// displayButtonLoadingState remembers the content of button.
+// It show loading state on button and returns a function to revert to original content.
+export function displayButtonLoadingState(el: HTMLButtonElement): () => void {
+  const original: string = el.innerHTML;
+  el.innerHTML =
+    '<span class="primary-btn__loading-icon material-icons animate-spin">progress_activity</span>';
```
tung2744, line 29: 1. I think we should decouple this with XHRSubmitFormController.
2. Replacing the original content with the loading spinner could cause layout bouncing because their content size is different.

My proposed solution:

- Define a new controller named `loadable-button`.
  - Usage: `data-controller="loadable-button"`
  - On connect(): Inject the `<span class="loading-spinner material-icons">` element immediately.
  - `loading-spinner` should be a class which by default hide the element if the parent is not loading.
  ```css
  :not(.loading) > .loading-spinner {
    display:  none;
  }
  ```
- In `displayButtonLoadingState`, you should only add a class `loading` to the button element. Do not mutate the children. The returned function (dismissButtonLoadingState) should only remove the class from classList of the same element.
- Overlay the spinner on top of the content, instead of replacement.
  ```css
  .loading {
    display: relative;
  }
  .loading :not(.loading-spinner) {
    visibility: hidden; /* Hide all things except the spinner */
  }
  .loading-spinner {
    @apply absolute inset-0; /* Covers the whole button and center align */
  }
  ```

Ngkaokis: Create a controller solely for injecting the loading spinner but not responsible for controlling the loading state display looks quite odd to me. 

Seems like we don't actually need a connector. Can just do something like the follow? Which is similar to what I do at the very first place? https://github.com/authgear/authgear-server/pull/3654#discussion_r1461778200

Just we delegate the loading state visibility control to XHRSubmitFormController
```html
<button class="button">
  this is a button
  <span class="loading-spinner"></span>
</button>
```

tung2744: My thought is we don't want the user of the primary button to declare the loading state ui each time they use the button. i.e. `<span class="loading-spinner material-icons">` should be reused in all reloadable button, because we expect the loading ui is same.

Ngkaokis: Then I think define a template would be a more proper solution?

```html
{{ define "loading_spinner" }}
 <span class="loading-spinner"></span>
{{ end }}
```

tung2744: I feel like it is more complicated.

In my understanding, `loadable-button` is one component, and the css `loadable-button.css` handles the display and hide logic of the spinner.

And using controller has an advantage that you can query the element by selector `data-controller~="loadable-button"`. So you don't need the additional attribute `data-button-loading-state`.

IniZio: Should we just move primary button be moved to html template to explicitly declare the loading spinner. `loadable` controller will solely control the class list 🤔

tung2744: I think we can do it with css pseudo element.

1. Make the class `.material-icons` a utiliy.
  ```css
  @layer utilities {
    .material-icons { ... }
  }
  ```
2. Define a class with the pseudo element using the above utility.
  ```css
  .loadable {
    position: relative;

    &::after {
      @apply material-icons animate-spin;
      content: "progress_activity";
    }

    &:not(.loading)::after {
      @apply hidden;
    }
  }
  ```

In this way we don't need template nor controller, what do you think?

louischan-oursky: In portal, we decouple the state of loading and the presentation of loading. We have src/hook/loading to allow components at any levels to declare there is something loading. We also allow components at any levels to read the current loading state. Would it be too overkill if we introduce a similar abstraction here? 

In my mind, We can have a LoadingController and XHRSubmitFormController mounted at very top level in the HTML document. XHRSubmitFormController can ask LoadingController to become loading, either via [event](https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events) or [direct invocation](https://stimulus.hotwired.dev/reference/controllers#directly-invoking-other-controllers). The missing piece is how do we assign an element responsible for the presentation of loading. Personally, I think the API is the most intuitive if I can simply declare which element is responsible for the presentation of loading in our templates. For example

```
<button class="primary-btn" data-loading-presentation-target>Continue</button>
```

louischan-oursky:
Rebase first.

authui/src/authflowv2/components/primary-button.css
```diff
@@ -78,5 +88,21 @@
         ring-[length:var(--primary-btn\_\_ring-width--disabled)]
         ring-[color:var(--primary-btn\_\_ring-color--disabled)];
     }
+
+    &:after {
```
louischan-oursky, line 92: ::after

authui/src/authflowv2/form.ts
```diff
@@ -0,0 +1,144 @@
...
+// To disable the builtin form submission of Turbo,
+// I studied its source code and discovered that
+// Turbo checked in the bubble event listener to see if
+// the event is prevented.
+// So I added a capture event listener to call preventDefault()
+// to stop Turbo from submitting forms.
+//
+// See https://github.com/authgear/authgear-server/issues/2333
+export class XHRSubmitFormController extends Controller {
+  revertDisabledButtons: { (): void } | null = null;
```
louischan-oursky, line 31: This is unused.

authui/src/authflowv2/loading.ts
```diff
@@ -0,0 +1,63 @@
...
+    // The success branch also loads a new page.
+    // Keeping the buttons in disabled state reduce flickering in the UI.
+    if (this.revertDisabledButtons) {
+      this.revertDisabledButtons();
+      this.revertDisabledButtons = null;
+    }
+    this.button?.classList.remove("primary-btn--loading");
+  }
+  onLoadingEnd(e: LoadingEvent) {
+    hideProgressBar();
```
louischan-oursky, line 46: We do not use progress bar anymore.

authui/src/authflowv2/components/primary-button.css
```diff
@@ -38,6 +38,16 @@
...
 
+  /* NOTE: @apply animate-spin raises error @parcel/transformer-css: Unknown at rule: @keyframes. So define it here */
+  @keyframes spin {
+    from {
+      transform: rotate(0deg);
+    }
+    to {
+      transform: rotate(360deg);
+    }
+  }
```
tung2744, line 49: Unnest the rule should fix the problem.
```
.class::after { /* not &::after */
  @apply animate-spin;
}
```

authui/src/authflowv2/form.ts
```diff
@@ -0,0 +1,144 @@
...
+    // e.stopPropagation();
+
+    const formData = new FormData(form);
+
+    const params = new URLSearchParams();
+    formData.forEach((value, name) => {
+      params.set(name, value as string);
+    });
+    const loadingEvent = new CustomEvent<{ button?: HTMLButtonElement }>(
+      "onLoading",
```
tung2744, line 59: Usually event name won't be onXX.
Maybe we name it "authgear:loading"?

authui/src/authflowv2/form.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+    );
+    // FormData does not include any submit button's data:
+    // include them manually, since we have at most one submit button per form.
+    const submitButtons = form.querySelectorAll('button[type="submit"]');
+    for (let i = 0; i < submitButtons.length; i++) {
+      const button = submitButtons[i] as HTMLButtonElement;
+      params.set(button.name, button.value);
+    }
+    if (form.id) {
```
tung2744, line 71: It should work even the form doesn't have an id. Usually it is unneccesary to define form id in the button.

authui/src/authflowv2/form.ts
```diff
@@ -0,0 +1,144 @@
...
+      {
+        detail: {},
+      }
+    );
+    // FormData does not include any submit button's data:
+    // include them manually, since we have at most one submit button per form.
+    const submitButtons = form.querySelectorAll('button[type="submit"]');
+    for (let i = 0; i < submitButtons.length; i++) {
+      const button = submitButtons[i] as HTMLButtonElement;
+      params.set(button.name, button.value);
```
tung2744, line 69: Set this button element to loadingEvent.detail.button too, if found.

authui/src/authflowv2/form.ts
```diff
@@ -0,0 +1,144 @@
...
+      handleAxiosError(e);
+      this.element.dispatchEvent(
+        new CustomEvent("onLoadingError", {
+          detail: {
+            error: e,
+          },
+        })
+      );
+    } finally {
+      this.element.dispatchEvent(new CustomEvent("onLoadingEnd"));
```
tung2744, line 120: suggest to  rename to event to authgear:loadingend

louischan-oursky:
```
diff --git a/authui/src/authflowv2/components/primary-button.css b/authui/src/authflowv2/components/primary-button.css
index bb1c23fcb..01f5696fd 100644
--- a/authui/src/authflowv2/components/primary-button.css
+++ b/authui/src/authflowv2/components/primary-button.css
@@ -89,7 +89,7 @@
         ring-[color:var(--primary-btn\_\_ring-color--disabled)];
     }
 
-    &:after {
+    &::after {
       @apply material-icons align-middle absolute inset-x-0;
       animation: spin 1s linear infinite;
 
@@ -98,10 +98,10 @@
       height: 1em;
       content: "progress_activity";
     }
-    &.primary-btn--loading {
+    &[data-loading] {
       color: transparent;
     }
-    &:not(.primary-btn--loading)::after {
+    &:not([data-loading])::after {
       @apply hidden;
     }
   }
diff --git a/authui/src/authflowv2/form.ts b/authui/src/authflowv2/form.ts
index 1fc762508..044173ac0 100644
--- a/authui/src/authflowv2/form.ts
+++ b/authui/src/authflowv2/form.ts
@@ -1,12 +1,7 @@
 import { visit, clearCache } from "@hotwired/turbo";
 import axios, { Method } from "axios";
-import {
-  disableAllButtons,
-  hideProgressBar,
-  showProgressBar,
-  progressEventHandler,
-} from "../loading";
+import { progressEventHandler } from "../loading";
 import { handleAxiosError } from "../messageBar";
 
 // Turbo has builtin support for form submission.
@@ -55,12 +50,9 @@ export class XHRSubmitFormController extends Controller {
     formData.forEach((value, name) => {
       params.set(name, value as string);
     });
-    const loadingEvent = new CustomEvent<{ button?: HTMLButtonElement }>(
-      "onLoading",
-      {
-        detail: {},
-      }
-    );
+
+    let loadingButton: HTMLButtonElement | null = null;
+
     // FormData does not include any submit button's data:
     // include them manually, since we have at most one submit button per form.
     const submitButtons = form.querySelectorAll('button[type="submit"]');
@@ -75,11 +67,12 @@ export class XHRSubmitFormController extends Controller {
       if (el) {
         const button = el as HTMLButtonElement;
         params.set(button.name, button.value);
-        loadingEvent.detail.button = button;
+        loadingButton = button;
       }
     }
 
-    this.element.dispatchEvent(loadingEvent);
+    const { onError, onFinally } =
+      loadingController.startLoading(loadingButton);
     try {
       const resp = await axios(form.action, {
         method: form.method as Method,
@@ -109,15 +102,9 @@ export class XHRSubmitFormController extends Controller {
       }
     } catch (e: unknown) {
       handleAxiosError(e);
-      this.element.dispatchEvent(
-        new CustomEvent("onLoadingError", {
-          detail: {
-            error: e,
-          },
-        })
-      );
+      onError();
     } finally {
-      this.element.dispatchEvent(new CustomEvent("onLoadingEnd"));
+      onFinally();
     }
   }
 
diff --git a/authui/src/authflowv2/loading.ts b/authui/src/authflowv2/loading.ts
index b010ba69a..021d6828d 100644
--- a/authui/src/authflowv2/loading.ts
+++ b/authui/src/authflowv2/loading.ts
@@ -1,63 +1,41 @@
-import {
-  disableAllButtons,
-  hideProgressBar,
-  showProgressBar,
-} from "../loading";
+import { disableAllButtons } from "../loading";
 
-interface LoadingEvent extends Event {
-  detail?: {
-    button?: HTMLButtonElement;
-    error?: unknown;
-  };
+export interface LoadingHandle {
+  onError(): void;
+  onFinally(): void;
 }
 
 export class LoadingController extends Controller {
-  revertDisabledButtons: { (): void } | null = null;
-  button: HTMLButtonElement | null = null;
+  revert: (() => void) | null = null;
 
   // Revert disabled buttons before Turbo caches the page
   // To avoid flickering in the UI
   beforeCache = () => {
-    if (this.revertDisabledButtons) {
-      this.revertDisabledButtons();
-    }
+    this.revert?.();
+    this.revert = null;
   };
 
-  onLoading(e: LoadingEvent) {
-    this.revertDisabledButtons = disableAllButtons();
-    showProgressBar();
-    if (e.detail?.button?.getAttribute("data-loading-state") != null) {
-      e.detail.button.classList.add("primary-btn--loading");
-      this.button = e.detail.button;
-    }
-  }
-  onLoadingError(e: LoadingEvent) {
-    // revert is only called for error branch because
-    // The success branch also loads a new page.
-    // Keeping the buttons in disabled state reduce flickering in the UI.
-    if (this.revertDisabledButtons) {
-      this.revertDisabledButtons();
-      this.revertDisabledButtons = null;
-    }
-    this.button?.classList.remove("primary-btn--loading");
-  }
-  onLoadingEnd(e: LoadingEvent) {
-    hideProgressBar();
-    this.button?.classList.remove("primary-btn--loading");
+  startLoading(e: HTMLElement | null): LoadingHandle {
+    const revert = disableAllButtons();
+    e?.setAttribute("data-loading", "true");
+
+    return {
+      onError: () => {
+        revert();
+        e?.removeAttribute("data-loading");
+      },
+      onFinally: () => {
+        e?.removeAttribute("data-loading");
+      },
+    };
   }
 
   connect() {
-    this.element.addEventListener("onLoading", this.onLoading);
-    this.element.addEventListener("onLoadingError", this.onLoadingError);
-    this.element.addEventListener("onLoadingEnd", this.onLoadingEnd);
     document.addEventListener("turbo:before-cache", this.beforeCache);
   }
 
   disconnect() {
-    this.element.removeEventListener("onLoading", this.onLoading);
-    this.element.removeEventListener("onLoadingError", this.onLoadingError);
-    this.element.removeEventListener("onLoadingEnd", this.onLoadingEnd);
     document.removeEventListener("turbo:before-cache", this.beforeCache);
   }
 }
```

authui/src/authflowv2/form.ts
```diff
@@ -0,0 +1,144 @@
...
+        case "replace":
+        case "advance":
+          // We assume the URL returned by the server have query preserved.
+          visit(redirect_uri, { action });
+          break;
+      }
+    } catch (e: unknown) {
+      handleAxiosError(e);
+      this.element.dispatchEvent(
+        new CustomEvent("onLoadingError", {
```
tung2744, line 113: suggest to rename the event to authgear:loading-error

authui/src/authflowv2/form.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (elems[i].querySelector('[data-turbo="false"]')) {
+        continue;
+      }
+      this.forms.push(elems[i] as HTMLFormElement);
+    }
+    for (const form of this.forms) {
+      form.addEventListener("submit", this.onSubmitCapture, true);
+      form.addEventListener("submit", this.onSubmit);
+    }
+  }
```
tung2744, line 136: The original implementation listened to events of turbo, I guess they are still needed?

tung2744: Oh I found them in loading controller.

authui/src/authflowv2/loading.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (this.revertDisabledButtons) {
+      this.revertDisabledButtons();
+    }
+  };
+
+  onLoading(e: LoadingEvent) {
+    this.revertDisabledButtons = disableAllButtons();
+    showProgressBar();
+    if (e.detail?.button?.getAttribute("data-loading-state") != null) {
+      e.detail.button.classList.add("primary-btn--loading");
```
tung2744, line 31: Do not write `primary-btn--loading` here, because this coupled this controller with primary button.
I expect it could be an attribute `loading` or `data-loading` added to the button, which doesn't care about the button type.

The component (primary button) should handle this attribute change in his own code (We can do it in the stylesheet)

authui/src/authflowv2/loading.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+  onLoadingError(e: LoadingEvent) {
+    // revert is only called for error branch because
+    // The success branch also loads a new page.
+    // Keeping the buttons in disabled state reduce flickering in the UI.
+    if (this.revertDisabledButtons) {
+      this.revertDisabledButtons();
+      this.revertDisabledButtons = null;
+    }
+    this.button?.classList.remove("primary-btn--loading");
```
tung2744, line 43: Same as above, use a generic attribute instead of class coupled with primary button.

authui/src/form.ts
```diff
@@ -29,6 +30,7 @@ import { handleAxiosError } from "./messageBar";
 // See https://github.com/authgear/authgear-server/issues/2333
 export class XHRSubmitFormController extends Controller {
```
tung2744, line 31: It seems we can revert all changes in this file, as we duplicated one new controller for v2.

resources/authgear/templates/en/web/authflowv2/enter_password.html
```diff
@@ -39,12 +39,14 @@ <h1 class="screen-title">
     {{ end }}
     <!-- TODO: Handle button loading state in https://github.com/authgear/authgear-server/issues/3676 -->
     <button
+      form="main-form"
```
tung2744, line 42: This button is inside the form already, so we don't need the form attribute.

resources/authgear/templates/en/web/authflowv2/enter_password.html
```diff
@@ -39,12 +39,14 @@ <h1 class="screen-title">
...
       class="primary-btn w-full"
       type="submit"
       name="x_action"
       value=""
       data-authgear-event="authgear.button.enter_password"
       data-action-button
+      data-loading-state
```
tung2744, line 49: Lets name this attribute `data-loadable`, which seems clearer it configures the button as a loadable button.

Issue #3667, opened 2024-01-22
OTP sent via WhatsApp while "SMS" is shown


fungc-io:
@tung2744 I've found another problem in the Whatsapp/SMS verification 😢 

It's the fung-authflow project again in the staging server
- The project is Phone+PW, require phone verification via whatsapp.
- When I sign up, the screen says "SMS", but i got the OTP from Whatsapp.
- I can verify myself by entering the OTP from whatsapp in this page.

![Image](https://github.com/authgear/authgear-server/assets/15207470/90ea4523-f010-4a96-a16e-8909f96c230e)

![Image](https://github.com/authgear/authgear-server/assets/15207470/5e87fc2d-3f7b-4df6-a27e-e03b1afaa030)

PR #3680, opened 2024-01-23
Fix verify step always show sms screen if no branches in v2

tung2744:
ref #3667


louischan-oursky:
Rebased authflow2 and applied in this commit https://github.com/authgear/authgear-server/commit/632860668f8e0a7f4261c0b337d65c52638b042f

PR #3689, opened 2024-01-24
Auth UI v2 - Clear otp input on submit

IniZio:
ref #3688

### DIFF
--- a/authui/src/form.ts
+++ b/authui/src/form.ts
@@ -180,6 +180,9 @@ export class RestoreFormController extends Controller {
     if (form == null) {
       return;
     }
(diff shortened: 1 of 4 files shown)


IniZio:
@tung2744 Thanks, I was finding how other pages achieved forgetting on submit. Updated to add prop for opting out from from restore 🙏 .

louischan-oursky:
Would it be simpler if we clear the otp input on submit in the otp input controller?

tung2744:
I'm not sure which one will run first (The iniatial clear of otp input, or the restoration from restore controller). I think we better disable the restore if we don't need it.

PR #3699, opened 2024-01-25
Render error html based on UI implementation in panic middleware

Ngkaokis:
ref #3692

https://github.com/authgear/authgear-server/assets/63200569/a721a0a7-9df2-40dd-b612-2e82d1433ae4


pkg/lib/web/csp.go
```diff
@@ -122,13 +122,25 @@ type dynamicCSPContextKeyType struct{}
 
 var dynamicCSPContextKey = dynamicCSPContextKeyType{}
 
-func WithCSPNonce(ctx context.Context, nonce string) context.Context {
-	return context.WithValue(ctx, dynamicCSPContextKey, nonce)
+type CSPNonceContextValue struct {
```
louischan-oursky, line 125: CSPNonceContextValue should be private.

pkg/lib/web/csp.go
```diff
@@ -122,13 +122,25 @@ type dynamicCSPContextKeyType struct{}
...
+	Nonce string
+}
+
+func WithCSPNonce(ctx context.Context, ctxValue *CSPNonceContextValue) context.Context {
```
louischan-oursky, line 129: func WithCSPNonce(ctx context.Context, nonce string) context.Context

The caller need not to know we are using holder pattern.

pkg/auth/routes.go
```diff
@@ -58,6 +58,7 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource) *h
 		httproute.MiddlewareFunc(httputil.StaticSecurityHeaders),
 		RequestMiddleware(p, configSource, newRequestMiddleware),
 		p.Middleware(newUIParamMiddleware),
+		p.Middleware(newContextHolderMiddleware),
```
louischan-oursky, line 60: This should be put above newPanicMiddleware

Ngkaokis: Per offline discussion, only need to above `newPanicWebAppMiddleware` but not `newPanicMiddleware `

pkg/auth/webapp/context_holder_middleware.go
```diff
@@ -20,8 +22,10 @@ func (m *ContextHolderMiddleware) Handle(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// This middleware only creates the holder of context.
 		// This enables the holder to be mutated later in other places.
-		var empty web.CSPNonceContextValue
-		ctx := web.WithCSPNonce(r.Context(), &empty)
+		var emptyCSPNonceContext web.CSPNonceContextValue
+		var emptyUIParamContext uiparam.T
```
tung2744: Just curios, why we need to modify something related to ui param and csp nonce in this pr?

louischan-oursky:
4ec47e486a3968154651c6aa25e813245cc3b604

pkg/auth/webapp/context_holder_middleware.go
Ngkaokis: In v1, we don't need js to make the error page works fine. But in v2, we need js to set the alignment attribute so discovered this bug :sosad:

PR #3706, opened 2024-01-26
Add go template prettier for v2 pages

IniZio:
ref #3666

### Caveats

- According to https://github.com/NiklasPor/prettier-plugin-go-template/issues/61#issuecomment-1047013795, prettier treats nested double quotes as error. Need to use backtick instead since single quote cannot be handled by golang.
- It puts the first conditional at same line with `{{ define "[template name]" }}, e.g.

	```html
	{{ template "authflowv2/__page_frame.html" . }}
	{{ define "page-content" }}
	
	  {{ $display_otp_input_error := false }}
	  {{ if $.Error }}
	    {{ if eq $.Error.reason "InvalidVerificationCode" }}
	      {{ $display_otp_input_error = true }}
	    {{ end }}
	    {{ if eq $.Error.reason "InvalidCredentials" }}
	      {{ $display_otp_input_error = true }}
	    {{ end }}
	  {{ end }}
	```
	becomes
	```html
	{{ template "authflowv2/__page_frame.html" . }} {{ define "page-content" }} {{
	$display_otp_input_error := false }} {{ if $.Error }} {{ if eq $.Error.reason
	"InvalidVerificationCode" }} {{ $display_otp_input_error = true }} {{ end }} {{
	if eq $.Error.reason "InvalidCredentials" }} {{ $display_otp_input_error = true
	}} {{ end }} {{ end }}
	```

### Comparison

- djlint: https://github.com/IniZio/authgear-server/tree/add-v2-go-template-djlint
	- Caveat: Does not support golang block (bug?) https://github.com/djlint/djLint/issues/734


scripts/npm/package.json
```diff
@@ -4,13 +4,17 @@
   "version": "0.1.0",
   "scripts": {
     "export-graphql-schema": "node export-graphql-schema.js",
-    "generate-go-timezones": "node generate-go-timezones.js"
+    "generate-go-timezones": "node generate-go-timezones.js",
+    "format": "prettier --write --list-different ../../resources/authgear/templates/en/web/authflowv2/view_recovery_code.html",
```
louischan-oursky, line 8: The whole directory?

louischan-oursky:
Per our offline discussion, we will introduce a new devtools/gotemplatelinter

```
$ go run devtools/gotemplatelinter ./resources/authgear/templates/en/web/authflowv2
./resources/authgear/templates/en/web/authflowv2/a.html:2: mixed indentation
./resources/authgear/templates/en/web/authflowv2/a.html:2: missing EOL character
$ echo $?
1
```

It exits with 0 if no problems are found. It exits with 1 otherwise. It output to stderr.

IniZio:
Will implement custom go template linter in another pr. Closing first.

Issue #3691, opened 2024-01-24
Create Spec for Native SSO


fungc-io:

fungc-io:
Let's put this to backlog and revisit later. Priorities changed

PR #3713, opened 2024-01-26
Specify Native SSO

louischan-oursky:
ref #3691 

https://github.com/louischan-oursky/authgear-server/blob/3691-spec-native-sso/docs/specs/oidc-native-sso.md

@fungc-io @chpapa


docs/specs/oidc-native-sso.md
```diff
@@ -0,0 +1,145 @@
+- [OIDC Native SSO](#oidc-native-sso)
+  * [Key points to note](#key-points-to-note)
+  * [Changes on Client](#changes-on-client)
+  * [Changes on OfflineGrant](#changes-on-offlinegrant)
+  * [Changes on Token Endpoint](#changes-on-token-endpoint)
+    + [`grant_type=authorization_code` and `scope=device_sso`](#grant_typeauthorization_code-and-scopedevice_sso)
+    + [`grant_type=biometric` and `scope=device_sso`](#grant_typebiometric-and-scopedevice_sso)
+    + [`grant_type=refresh_token` and `scope=device_sso`](#grant_typerefresh_token-and-scopedevice_sso)
+    + [`grant_type=app2app` and `scope=device_sso`](#grant_typeapp2app-and-scopedevice_sso)
```
tung2744, line 9: grant_type=urn:authgear:params:oauth:grant-type:app2app

docs/specs/oidc-native-sso.md
```diff
@@ -0,0 +1,145 @@
+- [OIDC Native SSO](#oidc-native-sso)
+  * [Key points to note](#key-points-to-note)
+  * [Changes on Client](#changes-on-client)
+  * [Changes on OfflineGrant](#changes-on-offlinegrant)
+  * [Changes on Token Endpoint](#changes-on-token-endpoint)
+    + [`grant_type=authorization_code` and `scope=device_sso`](#grant_typeauthorization_code-and-scopedevice_sso)
+    + [`grant_type=biometric` and `scope=device_sso`](#grant_typebiometric-and-scopedevice_sso)
```
tung2744, line 7: grant_type=urn:authgear:params:oauth:grant-type:biometric-request

chpapa:
Quick look, not sure if I omitted anything:

1. Shall we specify when and how the `device_secret` will be rotated when `/token` endpoint is called? 
2. Any idea how the spec could bind `device_secret` to the device?

tung2744:
Added a commit to mention how `device_secret` will be rotated.

> Any idea how the spec could bind device_secret to the device?

Looks like the spec didn't mention it.
In here: https://openid.net/specs/openid-connect-native-sso-1_0.html#name-cross-device-sso

It only says 

> An implementation of this specification SHOULD use best efforts to bind the device_secret to the device instance to prevent exfiltration of the device_secret and id_token from the device.

louischan-oursky:
We do have binding to the device in one of our feature, that is biometric login. The private key is stored in the keychain of the device, with proper configuration so that it is not backed up, or extracted from the hardware. However, in Native SSO, the device secret is supposed to be generated by the AS. Not sure if we can come up with a design that relies on key-pair.

louischan-oursky:
RFC 8473 requires TLS to work. So it is not a suitable choice. [RFC 9449 OAuth 2.0 Demonstrating Proof of Possession (DPoP)](https://datatracker.ietf.org/doc/html/rfc9449) is the spec that specify token binding on the application level (i.e. OAuth 2.0).

docs/specs/oidc-native-sso.md
```diff
@@ -0,0 +1,155 @@
...
+- If `refresh_token.scope=device_sso` and `device_secret` is absent or it is invalid, a new `device_secret` is generated. `DeviceSecretHash` is updated.
+
+### `grant_type=urn:authgear:params:oauth:grant-type:biometric-request`
+
+- Support a new parameter `scope`.
+- Allow `scope=device_sso` if the client has `x_device_sso_enabled=true`.
+- If `scope=device_sso` and `device_secret` is present and it is valid, a new `refresh_token` is added to Native SSO offline grant.
+- If `scope=device_sso` and `device_secret` is absent or it is invalid, a new Native SSO offline grant is created.
+
+### `grant_type=urn:authgear:params:oauth:grant-type:app2app`
```
tung2744, line 80: I suggest we do not support native sso with app2app. So we don't allow specifying scope in this grant_type.

Or else, we need to ensure the provided scope is a subset of the scope of the original offline grant.

louischan-oursky: > we need to ensure the provided scope is a subset of the scope of the original offline grant

Let's do this.

louischan-oursky: I added a commit to clarify this.

docs/specs/oidc-native-sso.md
```diff
@@ -0,0 +1,153 @@
...
+- The following fields become client-specific
+  - `ClientID`
+  - `AuthorizationID`
+  - `IdentityID`
+  - `Scopes`
+  - `TokenHash`
+- Add a new field `DeviceSSORefreshTokens`. It has the above client-specific fields.
+  - `ClientID`
+  - `AuthorizationID`
+  - `IdentityID`
```
tung2744, line 50: It seems `AuthorizationID` and `IdentityID` does not need to be token specific:

- AuthorizationID points to an Authorization. And Authorization actually refers to a "group" of session displayed in the settings page. When you terminate a session, you are actually deleting an Authorization. In native SSO, the new refresh tokens are actually sharing a same session with the original session, thus the Authorization should be the same. It represents the "parent" offline grant.
- IdentityID is for biometric login. As biometric login always creates new offline grant, it won't exist in the child refresh tokens?

louischan-oursky: Authorization associates a Client with a User. The refresh token is now client-specific. Actually a session can now be shared by different clients.

---

Your reasoning on IdentityID seems correct. Would there be any problem if we still make IdentityID token specific?

tung2744: > Authorization associates a Client with a User. The refresh token is now client-specific. Actually a session can now be shared by different clients.

Yes, now session can be shared by different client, but that is already represented by the `ClientID` of each token. There should be a "initial" client, which seems represented by "Authorization"? Conceptually, what actually is "Authorization"? 

> Your reasoning on IdentityID seems correct. Would there be any problem if we still make IdentityID token specific?

No problem, but IdentityID will always be empty.

docs/specs/oidc-native-sso.md
```diff
@@ -0,0 +1,153 @@
...
+  - `TokenHash`
+- Add a new field `DeviceSecretHash`. It is the hex of SHA256 of `device_secret`.
+- If `DeviceSSORefreshTokens` is non-empty, then the offline grant is a Native SSO offline grant.
+- Otherwise, the offline grant is an ordinary offline grant.
+
+## Changes on Authorization Endpoint
+
+- Allow `scope=device_sso` if the client has `x_device_sso_enabled=true`.
+
+## Changes on Token Endpoint
```
tung2744, line 61: 1. First party and third party can still share user authentication if the scopes are compatible. When using SDK, they are never compatible.
2. Apps can share user authentication when device secret can be shared.
    - If two apps are written by same vendor, they can share user authentication by sharing the device secret in the shared keychain.
    - If two apps are not written by the same vendor, they cannot share the authentication because they don't have a shared keychain.

Give some examples on 1+2.

docs/specs/oidc-native-sso.md
tung2744: - Each token will have different Authorization which is unique per client id.
- Update scope of existing Authorization if already exist and if needed.

docs/specs/oidc-native-sso.md
tung2744, line 61: It seems that we need to support `device_sso` on anonymous user if we plan to use it for settings page in future

louischan-oursky: I think anonymous user cannot open settings page.

tung2744:
Included in https://github.com/authgear/authgear-server/pull/4298

Issue #3574, opened 2024-01-09
AuthUI V2 - Signup: Setup 2FA TOTP


tung2744:
- QR Code Screen
  - https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=742-32281&mode=design&t=t2ocn0LIQuJP59g8-4
- Verify Code Screen
  - https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=742-32622&mode=design&t=t2ocn0LIQuJP59g8-4

These two screens share the same authflow step, the "Continue" and "Rescan QR code" button can be a link that simply navigate to the corresponding url with the same `x_step` set.

PR #3717, opened 2024-01-26
Auth UI v2 - Add setup totp page

IniZio:
ref #3574 

Used query param ~~`x_action`~~ `q_setup_totp_step=verify` instead of different path because will need to override `checkPath`

### Screens

Setup|Verify
-|-
Screen Shot 2024-01-26 at 17 55 45|Screen Shot 2024-01-26 at 18 01 25

### DIFF
--- a/resources/authgear/templates/en/translation.json
+++ b/resources/authgear/templates/en/translation.json
@@ -1066,5 +1066,7 @@
   "v2-reset-password-success-page-title": "Password updated",
   "v2-reset-password-success-page-description": "Your password successfully updated! \nYou can now log in using the new password.",
   "v2-error-password-reset-failed-title": "Invalid Link",
-  "v2-error-password-reset-failed-description": "This reset password link is invalid, used or expired. Please request a new one."
+  "v2-error-password-reset-failed-description": "This reset password link is invalid, used or expired. Please request a new one.",
+  "v2-setup-totp-title": "Set up authenticator",
+  "v2-setup-totp-description": "First, download Google Authenticator App from the <a class=\"link-btn\" target=\"_blank\" href=\"https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2\">Google Play Store</a> or <a class=\"link-btn\" target=\"_blank\" href=\"https://apps.apple.com/app/google-authenticator/id388497605\">Apple App Store</a>.\n\nScan the QR code or enter the key with your authenticator device / app."
 }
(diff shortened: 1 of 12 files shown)


resources/authgear/templates/en/web/authflowv2/setup_totp.html
```diff
@@ -31,7 +31,7 @@ <h2 class="screen-description whitespace-pre-wrap">{{ template "v2-setup-totp-de
 
     <a
       class="primary-btn w-full mt-10"
-      href="{{ call $.MakeURL "setup_totp_verify" }}"
+      href="{{ call $.MakeURL "setup_totp" "x_action" "submit"}}"
```
louischan-oursky, line 34: Instead of using `x_action` (which is already a name for submit button, let's use `q_setup_totp_step=verify`. Since it starts with `q_`, it is eliminated automatically.

If `q_setup_totp_step=verify`, you render the verify screen. Otherwise you render the qr code screen.

IniZio: Fixed in ad6c8d2

resources/authgear/templates/en/translation.json
```diff
@@ -1066,5 +1066,10 @@
   "v2-reset-password-success-page-title": "Password updated",
   "v2-reset-password-success-page-description": "Your password successfully updated! \nYou can now log in using the new password.",
   "v2-error-password-reset-failed-title": "Invalid Link",
-  "v2-error-password-reset-failed-description": "This reset password link is invalid, used or expired. Please request a new one."
+  "v2-error-password-reset-failed-description": "This reset password link is invalid, used or expired. Please request a new one.",
+  "v2-setup-totp-title": "Set up authenticator",
+  "v2-setup-totp-description": "First, download Google Authenticator App from the <a class=\"link-btn\" target=\"_blank\" href=\"https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2\">Google Play Store</a> or <a class=\"link-btn\" target=\"_blank\" href=\"https://apps.apple.com/app/google-authenticator/id388497605\">Apple App Store</a>.\n\nScan the QR code or enter the key with your authenticator device / app.",
```
tung2744, line 1071: I think we can use `<br>` to make line breaks as it is used as an html.

tung2744: Discuss in https://github.com/authgear/authgear-server/issues/3722

Issue #3572, opened 2024-01-09
AuthUI V2 - Signup/Login: Login Link


tung2744:
- Sent Screen
  - https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=191-19113&mode=design&t=t2ocn0LIQuJP59g8-4

- Use polling to check the approved state

- Success Screen
  - https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=757-35718&mode=design&t=t2ocn0LIQuJP59g8-4

PR #3739, opened 2024-01-30
Implement authflow v2 login link

Ngkaokis:
ref #3572

### DIFF
--- a/pkg/auth/handler/webapp/authflowv2/oob_otp_link.go
+++ b/pkg/auth/handler/webapp/authflowv2/oob_otp_link.go
@@ -1,0 +1,139 @@
+package authflowv2
+
+import (
+	htmltemplate "html/template"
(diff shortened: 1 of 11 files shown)


pkg/auth/handler/webapp/authflowv2/oob_otp_link.go
```diff
@@ -0,0 +1,141 @@
...
+		input := map[string]interface{}{
+			"resend": true,
+		}
+
+		result, err := h.Controller.UpdateWithInput(r, s, screen, input)
+		if err != nil {
+			return err
+		}
+
+		h.FlashMessage.Flash(w, string(webapp.FlashMessageTypeResendLoginLinkSuccess))
```
tung2744, line 120: There should be no flash message in v2

Issue #3732, opened 2024-01-29
Setup OOB OTP 2FA via Phone should not have alternative


fungc-io:
When 2FA is OOB OTP and verification is whatsapp or SMS, this screen is shown when setting up.

Phone number input in upper part and a "Set up Phone" button below

User should not be allowed to select alternative when entering the phone number.

![Image](https://github.com/authgear/authgear-server/assets/15207470/8746578f-9739-43d7-a276-5f498e7d47d4)

tung2744:
I think the expected behavior should be hide all phone otp related alternatives in this screen, but keep other alternatives such as totp.

Issue #3731, opened 2024-01-29
Setup OOB OTP via phone weird copywritings


fungc-io:
The message should clarify it's on whatsapp, and the button below is too long

![Image](https://github.com/authgear/authgear-server/assets/15207470/0d3547ac-aa18-446c-873e-7499408e89e1)

Todo:

- Change upper copywriting: "We've just sent a 6-digit verification code to {number} on WhatsApp"
- Change button copywriting: "Send via SMS instead"

Issue #3730, opened 2024-01-29
Wrong error msg for invalid OTP code in Reset password


fungc-io:
In account recovery, when the OTP entered is incorrect,
The error message is now showing "link invalid..."

![Image](https://github.com/authgear/authgear-server/assets/15207470/4da321d4-7403-491f-9371-5fc7b830bfe1)

Todo:
The error message should match e.g.
"Invalid code entered"

PR #3744, opened 2024-01-30
Various auth ui v2 fixes

tung2744:
ref #3730 #3731 #3732


pkg/auth/handler/webapp/authflowv2/forgot_password_otp.go
```diff
@@ -191,7 +193,11 @@ func (h *AuthflowV2ForgotPasswordOTPHandler) ServeHTTP(w http.ResponseWriter, r
 		}
 
 		result, err := h.Controller.AdvanceWithInput(r, s, screen, input, nil)
-		if err != nil {
+		if apierrors.IsKind(err, forgotpassword.PasswordResetFailed) {
```
louischan-oursky, line 196: Instead of adding info here, I think we can add info in forgotpassword/service.go
In particular we add `otp_form: code` or `otp_form: link`.

tung2744:
Updated, thanks!

louischan-oursky:
I added a commit to improved the addition of the error details.

Issue #3729, opened 2024-01-29
Unclear error message when contain disallowed keywords


fungc-io:
1. Disallow "Hi" in password
2. Create password containing "Hi"
3. Unclear error message

![Image](https://github.com/authgear/authgear-server/assets/15207470/af6c7ea6-25a3-4d76-9480-b7094939d106)

Todo:
- Show error message: "Password contains disallowed keywords"
- Handle other errors such as password history
- If multiple error, show the first one.

fungc-io:
@Ngkaokis please help update the copywriting to "Password contains disallowed keywords" instead of "No banned words"

fungc-io:
tested

PR #3759, opened 2024-01-31
Display explicit error for excluded keyword and reused pw policy

Ngkaokis:
ref #3729

https://github.com/authgear/authgear-server/assets/63200569/a7229d56-653f-4eb5-855c-a9eb236bb23f

### DIFF
--- a/resources/authgear/templates/en/translation.json
+++ b/resources/authgear/templates/en/translation.json
@@ -967,6 +967,8 @@
   "v2-error-server": "Server encountered an internal error.",
   "v2-error-screen-title": "Unexpected Issues",
   "v2-error.error-password-policy-violated": "Invalid password format.",
+  "v2-error.error-password-policy-banned-words": "No banned words",
+  "v2-error.error-password-policy-reuse": "No reuse of {size, plural, one{# previous password} other{# previous passwords}} / previous password within {day, plural, one{# day} other{# days}}",
   "v2-error-phone-number-format": "Incorrect phone number format for the selected country.",
   "v2-create-password-page-title": "Create Password",
   "v2-password-policy-strength": "Password strength:",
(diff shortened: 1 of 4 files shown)


louischan-oursky:
Not related to this issue, but in the video the password strength is NOT the first item in the list.

pkg/auth/handler/webapp/authflowv2/change_password.go
```diff
@@ -57,6 +58,15 @@ func (h *AuthflowV2ChangePasswordHandler) GetData(w http.ResponseWriter, r *http
...
 		},
 	)
+
+	// put policy PasswordBelowGuessableLevel to front
+	for p, x := range passwordPolicyViewModel.PasswordPolicies {
+		if x.Name == password.PasswordBelowGuessableLevel {
+			passwordPolicyViewModel.PasswordPolicies = append([]password.Policy{x}, append((passwordPolicyViewModel.PasswordPolicies)[:p], (passwordPolicyViewModel.PasswordPolicies)[p+1:]...)...)
+			break
+		}
+	}
```
tung2744, line 68: It looks like a ui logic.
What if simply use `order-1` in the corresponding element?

Ngkaokis: You mean access the array directly by index? e.g. `policy = policies[order-1]`

Seems it's will be easily broken if we assume the order of policies :derpthink:
And I think the policies length is not consistent?

tung2744: @Ngkaokis Sorry I just rechecked the doc and it should be `order-first`. It is a tailwind util.
https://tailwindcss.com/docs/order

I mean in the corresponding frontend element, add this class to make it always be the first item.

Issue #3734, opened 2024-01-29
Password Strength indicator is not shown on mobile


fungc-io:
it's not showing the password strength on my iPhone

![Image](https://github.com/authgear/authgear-server/assets/15207470/c50c8bc9-9e22-4a9a-be33-8100039a104a)

PR #3761, opened 2024-02-01
Update password strength meter to use role meter

Ngkaokis:
ref #3734

Seems safari won't display element inside `<meter>` element, so update to use `role="meter"` instead
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/meter_role

https://github.com/authgear/authgear-server/assets/63200569/bd06f367-b96c-4b98-a748-43ed20c6e39a

### DIFF
--- a/authui/src/authflowv2/components/password-strength-meter.css
+++ b/authui/src/authflowv2/components/password-strength-meter.css
@@ -31,35 +31,35 @@
     & > .password-strength-meter__icon::after {
       content: "sentiment_very_dissatisfied";
     }
-    &[value="1"] {
+    &[aria-valuenow="1"] {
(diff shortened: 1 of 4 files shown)


tung2744:
However the v1 meter is also using <meter> element, does that show up in safari?

Ngkaokis:
> However the v1 meter is also using element, does that show up in safari?

I think in v1 we didn't have children element in meter?

tung2744:
Yes I misread your message, I've updated the comment.

tung2744:
But `<span>` is also phrasing content 🤔

tung2744:
It seems safari doesn't expect meter to have any children, just ignore my comment.

Issue #3742, opened 2024-01-30
[AuthUI V2, Performance] Take long time to load first contentful paint


joyzoursky:
## Background
- Env: Staging
- Project: qa-ui-vtwo (V2), projectqa (V1)
- Browser: Chrome on Mac

## Description
When runnning lighthouse performance test, V2 takes much longer time than V1 to load the first contentful paint (FCP) and largest contentful paint (LCP).

## Lighthouse Result (AuthUI V2)

![Image](https://github.com/authgear/authgear-server/assets/12539486/2eec61fd-9305-4dd2-8c4d-c405c908622a)

## Lighthouse Result (AuthUI V1)

![Image](https://github.com/authgear/authgear-server/assets/12539486/dda15e96-569c-4fd2-92e6-8956d1027e59)

fungc-io:
@IniZio :
- Need to improve the loading time for icon fonts and emoji flags used in phone input

IniZio:
Current bundle size: 
<img width="1501" alt="image" src="https://github.com/authgear/authgear-server/assets/3119725/601b7b21-49ed-4900-96e7-32ef0db3c8c0">

IniZio:
@louischan-oursky @tung2744 For the country list, is it ok is we only include the official locales (en,zh-TW,zh-HK)? Seems the 3MB from cldr library is already minimal if want to include all locales.

Otherwise will try lazy import / pass from backend instead.

louischan-oursky:
We only use country names. So let's build a custom JSON that contains country names in all supported locales in cldr.

IniZio:
Noticed that currently static assets do not have brotoli / gzip, would it worth adding compression to reduce download time?

louischan-oursky:
We have util/httputil/file_server.go. Right now, FileServer has no constructor. Let's introduce a constructor for it. The constructor should return a FileServer wrapped by https://github.com/nytimes/gziphandler Then we should have gzip enabled.

louischan-oursky:
Opened https://github.com/authgear/authgear-server/issues/3790

fungc-io:
After optimization the FCP is similar to AuthUI v1,
will close this issue

![Image](https://github.com/authgear/authgear-server/assets/15207470/a6ba1535-fe2f-4f35-ae2b-007b37fafac1)

Issue #3728, opened 2024-01-29
Phone input UI issues


Ngkaokis:
Seems the phone input still not working fine in safari
- Desktop (macos 14.1.1 )
  - Can't see flag 
- mobile (simulator ios 15)
  - The search bar extends beyond the viewport
  - The menu can't not be collapsed after picking an option

UX improvements:
- If not searching, selected option is the original option. Else, should by default select first row.
- Should not blur the input when pressing arrows.

https://github.com/authgear/authgear-server/assets/63200569/1914cd5d-c9b1-473f-9514-df295da7b354

IniZio:
For emoji display issue, there were 2 reasons:
- The font used https://github.com/mozilla/twemoji-colr cannot render properly on Safari. Using https://github.com/13rac1/twemoji-color-font can render well on Safari and other platforms fine.
- Previously there is post-processing done that subsets only flag emojis from the font to file size. https://fonttools.readthedocs.io/en/latest/subset/index.html was used but also makes font not render on Safari `pyftsubset $TTF \
  --no-subset-tables+=FFTM \
  --unicodes=U+1F1E6-1F1FF,U+1F3F4,U+E0061-E007A,U+E007F \
  --output-file=./twemoji-countries.woff2 \
  --flavor=woff2`

PR #3762, opened 2024-02-01
Auth UI v2 - Fix phone input issues

IniZio:
ref #3728

~~Haven't resolved flag display issue~~ Used raw font file without applying subset. Will handle subset in #3742 instead.

### DIFF
--- a/authui/src/phoneInput.ts
+++ b/authui/src/phoneInput.ts
@@ -55,7 +55,7 @@ function compileCountryList(): PhoneInputCountry[] {
     const countryFlag = getEmojiFlag(countryCode);
     const countryCallingCode = getCountryCallingCode(countryCode);
     return {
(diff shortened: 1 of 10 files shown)


authui/src/phoneInput.ts
```diff
@@ -55,7 +55,7 @@ function compileCountryList(): PhoneInputCountry[] {
     const countryFlag = getEmojiFlag(countryCode);
     const countryCallingCode = getCountryCallingCode(countryCode);
     return {
-      flagEmoji: `<span class="country-flag-icon phone-input__country-flag">${countryFlag}</span>`,
+      flagEmoji: `<span class="country-flag-icon phone-input__country-flag twemoji-countries">${countryFlag}</span>`,
```
tung2744, line 58: What actually is the cause?

IniZio: This one is not related, just think better to apply icon default settings. The cause lies on the font files.

Issue #3735, opened 2024-01-29
Cannot see the OTP entered


fungc-io:
I think we should not mask the OTP input at all or at least when editing that specific digit field
I cannot see the OTP i've entered at all

https://github.com/authgear/authgear-server/assets/15207470/f042f4b7-d453-41e3-b36c-a0d9ed8f9527

fungc-io:
@arvinhoux Ok to not masking the OTP input?

arvinhoux:
Yes, the user should see the number they are currently entering. Once they move to the next input box after entering, the previous number should be hidden.

Additionally, when the user re-enters, the previous error state should be reset (meaning no error state should be displayed while entering, covering all input boxes).

louischan-oursky:
Implementation note: Should TextFieldController be applicable to OTP Input?

fungc-io:
When focus, unmask the last input digit.
When loss focus, mask all digits

PR #3764, opened 2024-02-01
Auth UI v2 - Unmask otp last digit when focused

IniZio:
ref #3735

### DIFF
--- a/authui/src/otpInput.ts
+++ b/authui/src/otpInput.ts
@@ -1,5 +1,4 @@
 import { Controller } from "@hotwired/stimulus";
-import { TextFieldController } from "./authflowv2/text-field";
 
 export class OtpInputController extends Controller {
   static targets = ["input", "submit", "digitsContainer"];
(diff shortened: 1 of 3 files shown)


tung2744:
Seems format check failed. Maybe you can install the vscode prettier plugin?

authui/src/otpInput.ts
```diff
@@ -1,4 +1,5 @@
 import { Controller } from "@hotwired/stimulus";
+import { TextFieldController } from './authflowv2/text-field';
```
louischan-oursky: Can we apply `text-field` controller in `__otp_input.html`?

authui/src/otpInput.ts
```diff
@@ -9,6 +10,14 @@ export class OtpInputController extends Controller {
...
   spans: HTMLElement[] = [];
 
+  get textField(): TextFieldController | null {
+    const ctr = this.application.getControllerForElementAndIdentifier(
+      this.inputTarget,
+      "text-field"
+    );
+    return ctr as TextFieldController | null;
+  }
+
```
louischan-oursky, line 20: I think this is unnecessary. If text-field is applied correctly, it should do its job correctly. If not, there should be a bug we need to fix.

authui/src/otpInput.ts
```diff
@@ -64,6 +73,7 @@ export class OtpInputController extends Controller {
       this.submitTarget.click();
     }
 
+    this.textField?.onInput();
```
louischan-oursky, line 76: This is unnecessary.

authui/src/otpInput.ts
IniZio: The issue is the hidden input is not the one with error state. Maybe I add an input container target.

Issue #3587, opened 2024-01-09
AuthUI V2 - Promote


tung2744:
Reference pkg/auth/handler/webapp/authflow_promote.go

Should have same design as signup screen.

PR #3772, opened 2024-02-02
Auth UI v2: Add promote screen

IniZio:
ref #3587

> Same as signup page except no login button

image


resources/authgear/templates/en/web/authflowv2/promote.html
```diff
@@ -0,0 +1,167 @@
+{{ template "authflowv2/__page_frame.html" . }}
```
tung2744, line 1: Is it possible to share the same template with signup instead of duplicating it?

Issue #3672, opened 2024-01-22
Add Authgear Logo Watermark to AuthUI v2


fungc-io:
A dim logo of Authgear should be present in the AuthUI in all screens for free projects

arvinhoux:
The design has been updated here.
https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=1346-18993&mode=design&t=jcQhKdczld30fx6K-4

tung2744:
@fungc-io I think this is ready to develop?

fungc-io:
yes 👍

PR #3776, opened 2024-02-02
Add watermark to widget

Ngkaokis:
ref #3672

https://github.com/authgear/authgear-server/assets/63200569/b388cb46-21a8-405a-87a2-41c36c5c91ca


resources/authgear/templates/en/web/authflowv2/__page_frame.html
```diff
@@ -37,5 +37,6 @@
   <div class="widget">
     {{ template "page-content" . }}
     {{ template "authflowv2/__lockout.html" . }}
+    {{ template "authflowv2/__watermark.html" . }}
```
tung2744, line 40: __lockout.html should still be the last element to ensure it is the last element, so that it is on top by default.

resources/authgear/templates/en/web/authflowv2/__watermark.html
```diff
@@ -0,0 +1,15 @@
+{{ define "authflowv2/__watermark.html" }}
+<div class="watermark__container">
```
tung2744, line 2: watermark__container should not exist as well if not $.WatermarkEnabled ?

arvinhoux:
I updated the rules.
https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=1371-19046&mode=design&t=TyERarSQmrq8JXEF-4

Issue #3581, opened 2024-01-09
AuthUI V2 - Login: Using Passkey


tung2744:
![Image](https://github.com/authgear/authgear-server/assets/34942210/42043252-8351-43e8-92ea-e736f1c972bf)

- Use passkey at login screen
- Use passkey as alternatives in other screens
- Use passkey screen @fungc-io The design of use passkey screen is still missing

fungc-io:
@arvinhoux 
The use passkey screen is missing
Please refer to this comment: https://github.com/authgear/authgear-server/issues/3553#issuecomment-1877279865

fungc-io:
@arvinhoux similar screen on android


![Image](https://github.com/authgear/authgear-server/assets/15207470/f7cad33c-27b4-45f3-88cd-2df756a1d3ba)
Reference: https://developer.android.com/design/ui/mobile/guides/patterns/passkeys

fungc-io:
@arvinhoux pointed out that the "Use Passkey" is a bit weird UX-wise, 
Let's talk and decide if an alternative UX can be implemented

arvinhoux:
Design has been updated
https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=805-63257&mode=design&t=3ZEbz4t2gQfotR9a-4

tung2744:
@fungc-io @arvinhoux It seems the use passkey screen doesn't exist.
It is basically just a screen required by the server to display to the user, and the native passkey dialog will pop-up automatically. What is the conern in ux?

fungc-io:
@tung2744 When they dismiss the Passkey dialog, showing the alternative method seems a better UX.

<img width="609" alt="image" src="https://github.com/authgear/authgear-server/assets/15207470/b086f1b3-450d-4654-a3d8-980850dde9d9">

Can we avoid the Use passkey screen?

tung2744:
@fungc-io It is difficult.
The problem is if the project configured passkey as the first authenticator, and when a user login in the login page (Not pressing "login with passkey", but type username and press enter), the server returns with a response which redirects the user to the use passkey screen.
And in this case, we cannot show the passkey dialog in the login screen directly, because we don't know the selected authenticator is passkey until server returned the redirect response.
It is possible to refactor the flow and eliminate the screen, but I suggest we simply add a screen for it. If your concern is user should be able to select other alternatives, user passkey screen can also show alternatives.

[...]

PR #3778, opened 2024-02-02
Update v2 use passkey text

tung2744:
ref #3581


tung2744:
Merge directly because this pr only changes some text in translations

PR #3780, opened 2024-02-05
Auth UI v2 - Only fetch required country select options

IniZio:
ref #3742


pkg/util/intl/resolve.go
```diff
@@ -27,20 +27,18 @@ func Resolve(preferred []string, fallback string, supported []string) (int, lang
...
-func ResolveLocaleCode(resolved string, fallback string, supported []string) string {
-	var matcher = language.NewMatcher(SupportedLanguageTags(supported))
-	var locale language.Tag
-
-	locale, _, confidence := matcher.Match(language.MustParse(resolved))
-	if confidence == language.No {
-		locale, _ = language.Parse(fallback)
+// ResolveUnicodeCldr resolves language tag to Unicode CLDR language tag.
+func ResolveUnicodeCldr(lang language.Tag, fallback language.Tag) string {
+	var clrdTags []language.Tag
```
louischan-oursky, line 32: The matcher is actually static, since the CldrLanguages are static. Let use init() to initialize a module-level private matcher.

language.NewMatcher is actually a very expensive operation. In 6cbdcbf8f1 we cached matcher instance because of this.

pkg/auth/handler/webapp/viewmodels/base.go
```diff
@@ -58,6 +59,7 @@ type BaseViewModel struct {
 	IsNativePlatform      bool
 	FlashMessageType      string
 	ResolvedLanguageTag   string
+	ResolvedLocale        string
```
louischan-oursky, line 62: Name this `ResolvedCLDRLocale`

pkg/util/intl/language.go
```diff
@@ -42,3 +44,13 @@ func Supported(supportedLanguageTags []string, fallbackLanguage FallbackLanguage
 	s = append(s, supportedLanguageTags[fallbackIdx+1:]...)
 	return s
 }
+
+func SupportedLanguageTags(languages []string) []language.Tag {
```
louischan-oursky, line 48: This function seem unused.

authui/src/phoneInput.ts
```diff
@@ -71,7 +92,7 @@ async function compileCountryList(): Promise<PhoneInputCountry[]> {
   const preferredCountries = preferredCountryCodes.map(countryCodeToCountry);
 
   onlyCountries.sort((a, b) => {
-    return a.localizedName.localeCompare(b.localizedName);
+    return a.name.localeCompare(b.name);
```
louischan-oursky, line 95: Why do you change this?

pkg/util/intl/resolve.go
```diff
@@ -26,3 +26,19 @@ func Resolve(preferred []string, fallback string, supported []string) (int, lang
 
 	return -1, tag
 }
+
+// ResolveUnicodeCldr resolves language tag to Unicode CLDR language tag.
+func ResolveUnicodeCldr(lang language.Tag, fallback language.Tag) string {
+	var cldrTags []language.Tag
+	for _, cldr := range CldrLanguages {
+		cldrTags = append(cldrTags, language.MustParse(cldr))
+	}
```
tung2744, line 35: Parse all cldr languages and put them in memory so that we don't have to parse it in each request?
I remember there was a performance issue saying that constructing the matcher is expensive.

IniZio: Yes, I updated to use `matcher := GetMatcher(cldrTags)` in aad899c54f1c24cd4e7641987957ce044b380c53

tung2744: My thought is to construct the matcher in `init()` and store it to a variable, because the list of clrd locales never changes.
Now for each request we looped all items and called `language.MustParse` for each item, which might be expensive?

IniZio: Sure updated

Issue #3775, opened 2024-02-02
Weird flags in chrome


tung2744:
![Image](https://github.com/authgear/authgear-server/assets/34942210/00abb43e-372f-443a-a851-9b161d92e14e)

PR #3784, opened 2024-02-06
Auth UI v2 - Apply font subset on material icons and twemoji icons

IniZio:
ref #3742 #3775

dependencies:
- [x] #3780


authui/src/authflowv2/password-policy.ts
```diff
@@ -108,13 +107,14 @@ function checkPasswordSymbol(value: string, el: HTMLElement) {
...
 
-function checkPasswordStrength(
+async function checkPasswordStrength(
   value: string,
   el: HTMLElement,
   currentMeter: HTMLElement
 ) {
   let isViolated = false;
   const minLevel = Number(el.getAttribute("data-min-level"));
+  const { default: zxcvbn } = await import("zxcvbn");
```
tung2744, line 117: Maybe we should load zxcvbn once the controller is connected?

IniZio: connect already calls `this.check()` once, you prefer moving it there explicitly?

tung2744: Oh I think it is ok too.

IniZio: Sorry I deleted the comment because the check is already exited when value is empty, maybe I still add `await import("zxcvbn");` in `connect` to try downloading the bundle earlier?

IniZio: Amended in cbda0cb, not awaited since dynamic import de-duplicates bundle download

scripts/python/subset_material_icons.py
```diff
@@ -0,0 +1,95 @@
...
+import re
+import requests
+from bs4 import BeautifulSoup
+
+from lib import subset_font
+
+def parse_html_file(file_path):
+  with open(file_path, 'r') as file:
+    soup = BeautifulSoup(file.read(), 'html.parser')
+    icons = soup.find_all(class_='material-icons')
```
tung2744, line 11: material-icons is a tailwind utility, which can be used with `@apply`. So this doesn't work. Maybe consider one of the following?
1. Maybe we can simply search for all existence of defined icon names in the files (html, css, js, json), not limited to class names. This could cause false matches (For example, "phone" is too common that it always appear, which may not be used as icon). But I believe even with the false matches, the subset is small enough.
2. We can simply define a list of icon name. And we manually add icons there.

IniZio: Yes that's why I also parse_css_file as well to cover this case. 

Nevertheless, maybe I added a custom list if icon names that developers can add if they found auto-parse failed to include the glyph.

IniZio: Added in 41a8e5d

Makefile
```diff
@@ -152,6 +152,14 @@ generate-timezones:
 generate-rtl:
 	go run ./scripts/characterorder/main.go | gofmt > pkg/util/intl/rtl_map.go
 
+.PHONT: generate-material-icons
+generate-material-icons:
+	pushd ./scripts/python && pip install -r requirements.txt && python ./subset_material_icons.py && popd
```
louischan-oursky: Should use venv. Otherwise running this make target will pollute the running machine.

authui/src/authflowv2/icons/twemoji-countries.css
```diff
@@ -34,6 +34,13 @@
 }
 
 @supports (-webkit-appearance: none) and (stroke-color: transparent) {
+  @font-face {
+    font-family: "twemoji-countries";
+    src:
+      url("./icons/twemoji-countries-subset.woff2") format("woff2"),
+      url("./icons/twemoji-countries-subset.ttf") format("ttf");
```
louischan-oursky, line 41: Then here should also be changed to `truetype`. https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src#font_formats

scripts/python/subset_material_icons.py
louischan-oursky: Since we are using ligature to apply the icon, if we can get the full list of ligature keywords. Then we only need to scan resources/**/*.{html,css,ts} for those keywords. This algorithm is much simpler. (I remember tailwind is doing something similar, instead of really parsing the HTML, CSS, JS files)

louischan-oursky: I guess this is the list? https://github.com/google/material-design-icons/blob/master/variablefont/MaterialSymbolsOutlined%5BFILL%2CGRAD%2Copsz%2Cwght%5D.codepoints

louischan-oursky:
1. Add material-symbols.gitcommit to store the git commit
2. Add instruction in CONTRIBUTING.md about how to update material-symbols https://github.com/google/material-design-icons/tree/master/variablefont
3. From https://github.com/google/material-design-icons/tree/master/variablefont, update the fonts we are using this in PR. It because I saw the fonts were updated 5 days ago in the upstream

authui/src/authflowv2/icons/twemoji-countries.css
IniZio: Fixed in 8c39acd

scripts/python/subset_material_icons.py
IniZio: Updated in 2d5f8a1

louischan-oursky:
I added a commit to fix material-icons.txt does not end with a newline.

louischan-oursky (approved):

Issue #3695, opened 2024-01-25
AuthUI V2 - Force Update Password Success Screen


Ngkaokis:
ref #3585

It requires changes on authflow controller

https://oursky.slack.com/archives/C03AH1ACM/p1706098152824109?thread_ts=1706097103.235049&cid=C03AH1ACM

Ngkaokis:
- Tried Approach:
Persist the `Result` in `AuthflowScreen` and hijack the redirect URI of `Result.RedirectURL`. Then redirect it to ResetPasswordSuccessScreen and then restore the previous persisted `Result` in order to continue the original flow.

- Problem Found:
As the session is deleted after the flow is finished, it's not possible to restore the session to get the persisted `Result`.

- Next Approach:
Every flows is required to be redirected to `/authflow/finsh` to manually finish the flow instead of bind it with `AdvanceInput`. And hence we can inject intermediate screen before the screen is finished.

fungc-io:
tested ok

PR #3792, opened 2024-02-07
Add force update password success page

Ngkaokis:
ref #3695 

require #3807 

https://github.com/authgear/authgear-server/assets/63200569/be7a1d73-1064-40aa-9f58-36a54056c04c

### DIFF
--- a/pkg/auth/handler/webapp/authflowv2/change_password.go
+++ b/pkg/auth/handler/webapp/authflowv2/change_password.go
@@ -103,9 +103,14 @@ func (h *AuthflowV2ChangePasswordHandler) ServeHTTP(w http.ResponseWriter, r *ht
 		if err != nil {
 			return err
 		}
+		injectedScreen, err := h.Controller.InjectNewAuthflowScreen(r, s, screen.Screen, result.RedirectURI)
+		if err != nil {
+			return err
+		}
+		newResult := &webapp.Result{}
 
-		h.Navigator.NavigateChangePasswordSuccessPage(screen.Screen, r, s.ID, result)
-		result.WriteResponse(w, r)
+		h.Navigator.NavigateChangePasswordSuccessPage(injectedScreen, r, s.ID, newResult)
+		newResult.WriteResponse(w, r)
 		return nil
 	})
 	h.Controller.HandleStep(w, r, &handlers)
--- a/pkg/auth/handler/webapp/authflowv2/change_password.go
+++ b/pkg/auth/handler/webapp/authflowv2/change_password.go
@@ -103,7 +103,7 @@ func (h *AuthflowV2ChangePasswordHandler) ServeHTTP(w http.ResponseWriter, r *ht
 		if err != nil {
 			return err
 		}
-		injectedScreen, err := h.Controller.InjectNewAuthflowScreen(r, s, screen.Screen, result.RedirectURI)
+		injectedScreen, err := h.Controller.InjectNewAuthflowScreen(r, s, screen.Screen, result)
 		if err != nil {
 			return err
 		}
--- a/pkg/auth/handler/webapp/authflowv2/change_password.go
+++ b/pkg/auth/handler/webapp/authflowv2/change_password.go
@@ -103,12 +103,12 @@ func (h *AuthflowV2ChangePasswordHandler) ServeHTTP(w http.ResponseWriter, r *ht
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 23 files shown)


resources/authgear/templates/en/web/authflowv2/change_password_success.html
```diff
@@ -0,0 +1,18 @@
...
+    <h1 class="screen-title mt-2">
+      {{ template "v2-change-password-success-page-title" }}
+    </h1>
+    <p class="screen-description">{{ template "v2-change-password-success-page-description" }}</p>
+  </header>
+  <div></div>
+    <footer>
+      <a class="primary-btn w-full" href="{{ call $.MakeURL "/settings" "x_step" "" "x_can_back_to_login" "" }}" data-turbo="false">
+        {{ template "v2-button.continue" }}
+      </a>
```
tung2744, line 15: I'm not sure should we hardcode `/settings` here. Referencing the interaction webapp it seems we didn't hardcode the redirect uri here.

pkg/auth/handler/webapp/authflowv2/change_password.go
```diff
@@ -99,6 +99,7 @@ func (h *AuthflowV2ChangePasswordHandler) ServeHTTP(w http.ResponseWriter, r *ht
 			return err
 		}
 
+		h.Controller.Navigator.NavigateChangePasswordSuccessPage(screen, r, s.ID, result)
```
tung2744, line 102: Make navigator a direct dependency of this handler, instead of using it through the controller.

pkg/auth/webapp/session_authflow.go
```diff
@@ -105,6 +105,9 @@ func newXStep() string {
 // Some steps in an authflow can have branches.
 // In order to be able to switch between branches, we need to remember the state that has branches.
 type AuthflowScreen struct {
+	// Store FlowType and FinishRedirectURI when the flow is finish
+	FlowType          authflow.FlowType `json:"flow_type,omitempty"`
+	FinishRedirectURI string            `json:"finish_redirect_uri,omitempty"`
```
tung2744, line 110: Why we don't simply derive finish_redirect_uri in `Finish`?

Ngkaokis: As `finish_redirect_uri` is derived from `*authflow.FlowResponse` but `*authflow.FlowResponse` can't be stringify because `*authflow.FlowResponse.Action.Data` is a interface instead of struct. So we need to derive and save `finish_redirect_uri` from previous screen.

resources/authgear/templates/en/web/authflowv2/change_password_success.html
Ngkaokis: Should go to finish route but forgot updating it.

How about hardcoding finish route / send post request to navigate to finish route?

pkg/auth/handler/webapp/authflowv2/change_password.go
Ngkaokis: Any reason we shouldn't do that?

resources/authgear/templates/en/web/authflowv2/change_password_success.html
tung2744: It might not be finish route if the flow is not finished, so we cannot hardcode it here. We need to send a post request to the server, and let the server to decide where to go, which is what `AdvanceWithoutInput` should do.

pkg/auth/handler/webapp/authflowv2/change_password.go
tung2744: Some reasons we don't want to do it:
1. If we use `Controller.Navigator`, we assumes `Controller` is a struct with a property `Navigator`. We might not want to make such assumption (Maybe we want to change it into an interface in the future)
2. In my opinion, If you call `Navigator.NavigateChangePasswordSuccessPage` in the handler, it means the thing that "needs" a `Navigator` is the handler. So it should be a direct dependency of the handler.

pkg/auth/webapp/session_authflow.go
tung2744: The problem here is:
1. The flow will be deleted on finish, so it is not gettable in the service.
1. If we try to store the data in session, we are unable to recover the data from the serialized json, because we don't know the type of data.

tung2744 (approved):

louischan-oursky:
The password success screen should show error alert when continue results in an error.

Ngkaokis:


https://github.com/authgear/authgear-server/assets/63200569/7287265a-7d69-41f2-8244-10e79734c695

Issue #3548, opened 2024-01-03
[AuthUI] Find algorithm to calculate the color palette given a base color


fungc-io:
The hover color is derived from the primary color. After the user set the primary color, the button will be filled with it.

One idea is to derive the state color by adjusting the L component in HSL.
For example, 
given a button_color = hsl(H_primary, S_primary, L_primary) 
if L_primary <= threshold
    button_color_hover = hsl(H_primary, S_primary, L_primary-10%) 
else 
   button_color_hover = hsl(H_primary, S_primary, L_primary+10%) 

We should test the algorithm by applying it to the new AuthUI

louischan-oursky:
> Limit the selectable base colors. i.e. Only colors on 400 is selectable.

Need a POC to see if this works nice.

fungc-io:
I believe our clients will have a wide range brightness in their brand colors depending on the brand image and position. This limitation may be too restricting to them

tung2744:
How other design systems works:

1. Material UI
    - For any color, convert it into HCT (hue, chroma, tone).
    - Generate a tonal palette for the color (0-100), and always use the same tone value on the same place.
    - Cons: The selected color will be first converted into tonal palette, and components such as primary button will only use a specific tone (For example, always 40). So the selected color is not exactly the color you see on primary button.
2. Fluent UI
    - The theme generator: https://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/7.0/theming-designer/index.html
    - No docs mentioning how the colors are generated, but we can find the source code of theme generator: https://github.com/microsoft/fluentui/blob/master/packages/react/src/components/ThemeGenerator/ThemeGenerator.ts
    - All colors (slots) are "shades" of another color, which finally must be generatable from the selected "base" color.
    - Every slots has defined his own logic to generate the color from the "parent" color.
    - Cons: Basically we need to define the generation logic of each color token.

tung2744:
Mapping of original color palette to fluent ui color palette
- Primary/base or Primary/400 -> Primary themePrimary
- Primary/500 -> Primary themeDark

- Grey/base or Grey/400 -> Foreground neutralPrimary
- Grey/100 -> Foreground neutralTertiary
- Grey/200 -> Foreground neutralSecondary
- Grey/300 -> Foreground neutralPrimaryAlt
- Grey/600 -> Foreground neutralDark
- Grey/700 -> Foreground black

And I also changed some colors to use fluent ui palette:
[...]

PR #3794, opened 2024-02-07
Switch to fluent ui palette

tung2744:
ref #3548 

TODO:
- [x] Switch back to tonal palatte for text color


authui/src/authflowv2/base.css
```diff
@@ -30,31 +30,40 @@
...
+    --color-primary-theme-secondary: #317bf4;
+    --color-primary-theme-tertiary: #70a4f7;
+    --color-primary-theme-light: #b7d1fb;
+    --color-primary-theme-lighter: #d8e6fd;
+    --color-primary-theme-lighter-alt: #f5f9fe;
+
+    /* Text colors is tonal palette */
+    /* https://m3.material.io/styles/color/system/how-the-system-works#395813c8-b314-48d1-bb55-266f421eb3a4 */
+    /* text */
+    --color-text-100: #e7e7e9;
```
tung2744, line 57: I've renamed grey to text color in this pr, because I think the role of the color is to display texts.

authui/src/authflowv2/base.css
```diff
@@ -36,15 +36,21 @@
...
-    --color-primary-600: #052e6f;
-    --color-primary-700: #011234;
+    /* Primary colors is generated using fluent ui theme generator */
+    /* https://developer.microsoft.com/en-us/fluentui?fabricVer=5#/styles/themegenerator */
+    /* Parameters:
+      Primary color: #176df3
+    */
+
+    --color-primary-theme-darker: #0c3c88;
+    --color-primary-theme-dark: #1151b8;
```
louischan-oursky, line 45: I tested a bit with the theme generator. When the input color is light enough, then `light` is NOT actually light, but visually darker than the input color. So the name `light`, `lighter`, `lighter-alt` may be confusing.

If we use `-primary` and `-dark` only, maybe we should only include these two in our palatte?

authui/src/authflowv2/base.css
tung2744: @louischan-oursky Yes, or maybe I should simply name it --color-secondary?

authui/src/authflowv2/base.css
tung2744: The name of the color token is defined by fluent ui palatte.
I think I can remove the other colors first.

authui/src/authflowv2/base.css
louischan-oursky: https://m3.material.io/styles/color/system/how-the-system-works

Design system often names the color "Primary", "Secondary", and "Tertiary". But "Secondary" is not for displaying text. Material design calls it "Neutral". 

https://spectrum.adobe.com/page/color-system/ Spectrum calls it grey, but it also mentions it is neutral.

tung2744: Actually the problem here is foreground elements and background elements are using the same palette, which are usually separated.
Anyway, lets rename it to neutral first.

tung2744:

louischan-oursky:
Now we have 3 different styles of naming for colors

1. `-dark`, `-primary`
2. `-100`, `-200`, ...
3. `-base`

We need to settle down with one. Let's discuss with Arvin?

tung2744:
@louischan-oursky According to discussion before the new year holiday,
1. Primary color uses fluent ui palette 
2. Text color uses original tonal palatte (100..700)
3. Base should not exist.

PR #3797, opened 2024-02-07
Adjust watermark styles

Ngkaokis:
ref #3793


authui/src/authflowv2/components/screen-icon-layout.css
```diff
@@ -31,7 +31,7 @@
           <buttons.../>
         </footer>
      */
-    grid-template-rows: auto auto auto;
+    grid-template-rows: max-content max-content max-content;
```
tung2744, line 34: It seems what you want is align-content: start

Issue #3796, opened 2024-02-07
Password Strength Meter Broken


Ngkaokis:
![Image](https://github.com/authgear/authgear-server/assets/63200569/d10ea266-1674-43c3-824b-3ce626df86b9)

IniZio:
Some notes before reverting lazy loading zxcvbn:

- Lazy load works for some lib but fails for others (tested stimulus and `axios` works, `zxcvbn` and `@floating-ui/dom` always fails).
- `parcel@2.8.3` is the last version that lazy load works for zxcvbn. `2.9.0` introduced new transform and default to swc optimizer. Suspect it changed behaviour of lazy loading cjs, thus causing bundle not found error.

louischan-oursky:
parcel seems very unstable. Switching to vite is also a choice worth exploring.

PR #3798, opened 2024-02-07
Auth UI v2 - Fix dynamic import type for zxcvbn

IniZio:
ref #3796

### DIFF
--- a/authui/src/authflowv2/password-policy.ts
+++ b/authui/src/authflowv2/password-policy.ts
@@ -1,4 +1,5 @@
 import { Controller } from "@hotwired/stimulus";
+import { runZxcvbn } from "./zxcvbn";
 
 enum PasswordPolicyName {
   Strength = "strength",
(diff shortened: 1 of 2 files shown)


authui/src/authflowv2/password-policy.ts
```diff
@@ -114,8 +117,8 @@ async function checkPasswordStrength(
 ) {
   let isViolated = false;
   const minLevel = Number(el.getAttribute("data-min-level"));
-  const { default: zxcvbn } = await import("zxcvbn");
-  const result = zxcvbn(value);
+  const z = (await import("zxcvbn")) as unknown as Zxcvbn;
```
tung2744, line 120: Or instead of directly importing zxcvbn dynamically, we should isolate the import to a new file which defined functions to be called, and dynamically import that file instead?

In this way we can also add zxcvbn to banned imports.

IniZio: Seems like lazy loading controller then, or another way would be have a deps provider declared per controller? 🤔

tung2744: I think the controller must be loaded in authflowv2.ts, so we cannot lazy load it.

My thought was to simply having a module to "isolate" some large dependencies, and ban direct import of those dependencies using eslint.

authui/src/authflowv2/zxcvbn.ts
```diff
@@ -0,0 +1,10 @@
+// This file lazily imports zxcvbn, so that it's not included in the main bundle
+
+import type zxcvbn from "zxcvbn";
```
tung2744, line 3: I wanted to use eslint to ban import from `zxcvbn`, but I found that we did not lint authui 🥲 
Maybe I create a new issue for adding eslint.

Issue #3802, opened 2024-02-19
Add linting to authui


tung2744:
- [ ] Setup eslint

PR #3805, opened 2024-02-19
Auth UI v2 - Setup eslint and stylelint for authui lib

IniZio:
ref #3802

dependencies:

- [x] #3809


.github/workflows/ci.yaml
```diff
@@ -13,6 +13,31 @@ on:
     - '!gh-pages'
 
 jobs:
+  authui-test:
```
louischan-oursky, line 16: Make authgear-image depends on this new job.

authui/src/authflowv2/customSelect.ts
```diff
@@ -29,7 +29,7 @@ export class CustomSelectController extends Controller {
   declare readonly inputTarget: HTMLInputElement;
   declare readonly triggerTarget: HTMLButtonElement;
   declare readonly dropdownTarget: HTMLElement;
-  declare readonly searchTarget: HTMLInputElement;
+  declare readonly searchTarget: HTMLInputElement | undefined;
```
tung2744, line 32: target cannot be undefined. If you try to access a target which doesn't exist, is is an error.
We need to use hasXTarget to check for existence.
https://stimulus.hotwired.dev/reference/targets#optional-targets

Issue #3748, opened 2024-01-30
Test the AuthUI v2 is working on password managers on popular browsers and platforms


fungc-io:
Make sure these can be filled using password managers:
- Login ID
- Password
- OTP
- TOTP

tung2744:
- Desktop Browser
  - Create / enter password: Username cannot be recongnized, fixing in new pr
  - totp: working
- Android Custom tabs
  - Create password: One password on android (9) always failed to fill the first password field
  - Login enter password: working
  - totp: working
- Android webview
  - We need asset link for password manager to work
- iOS
  - To be tested

tung2744:
@fungc-io Updated the result in table

fungc-io:
thanks

fungc-io:
I cannot fill the email address on iOS during login neither,
but it also doesn't work on Auth UI v1,
created another issue to follow up: https://github.com/authgear/authgear-server/issues/3843

PR #3808, opened 2024-02-19
Fix username is missing in password manager

tung2744:
ref #3748


IniZio (approved):
Tested can save username after the change, need to update icon font

Noticed that phone number cannot be autofilled with the saved username in 1pass though. It works if I update `autocomplete="tel-national"` to `autocomplete="username"`

I tried to set an identity with phone number e.g. `+85223382338` and autofill with existing code but it does not have `+` so the phone input fails to parse phone code from it.

louischan-oursky:
# When `autocomplete="tel-national"` and I have set `phone` in my 1Password entry

I need to manually open the extension and press Autofill to fill in the phone number.

https://github.com/authgear/authgear-server/assets/15082564/3b8c08c4-63c7-4bc9-9393-618460cfe1d7

# When `autocomplete="username"`

The usual 1Password icon appears in the text field, and focus the text field will offer the option to select.
I think this results in better UX. So I think we need to change to `autocomplete="username"`

tung2744:
Updated to use autocomplete="username" in phone input

Issue #3813, opened 2024-02-20
Open and close “Change Password” page from SDK


fungc-io:
## Problem

Some users wish to use our pre-built "Change Password" page but not the full "Settings" page.

## Solution

- Provide `authgear.open(Page.ChangePassword)` in the SDKs
- When the change password page is opened by this approach, the user cannot navigate to the `/settings` page or other pages from the UI. After changing the password, in the success screen, the primary button will close the webview.

IniZio:
Currently we use interaction to change password, with redirectURI hard-coded to `/settings`, can we pass custom redirectURI as follow?

```swift
authgear.open(.changePassword, escOnSuccess: true)
```

It concats `?redirect_uri=%2F%3Ferror%3Dsuccess` which replaces server side's redirectURI https://github.com/authgear/authgear-server/blob/c536713e7fc73b5dc585501090ccd6cc52ea7693/pkg/auth/handler/webapp/settings_change_password.go#L78

fungc-io:
Need to discuss how should web be handled

IniZio:
After discussion:

- If page is opened in new tab, will close on success similar to other SDK's behaviours
- If page is opened in same tab, will try to pop history til original page

fungc-io:
If pop history is not feasible, we may need to provide a field in the portal for "Post-change-password URL"..

IniZio:
After testing implementing with web sdk, found original double redirect_uri approach hard to accommodate it.

Will do following changes:
- Separate web SDK from prs since it will need special handling on history after success 
- Pick up https://github.com/authgear/authgear-server/pull/2886 again, treating settings action as an intermediate before redirect URI
- Update SDK to have explicit `changePassword()` method that resolves on redirectURI and fails on cancel

IniZio:
@tung2744 @louischan-oursky Since we have settings action behave like authenticate now, don't think we will have `openInSameTab` option now? Will need to try blocking user going back to change password page after complete instead.

louischan-oursky:
Let's remove the `openInSameTab` option in our new methods.

fungc-io:
@newman @louischan,

I found that the change password function is not on API reference yet.

[https://authgear.github.io/authgear-sdk-js/docs/capacitor/classes/CapacitorContainer/](https://authgear.github.io/authgear-sdk-js/docs/capacitor/classes/CapacitorContainer/)

Is it available in the v2.7.0? If so, should we update the references?

[...]

PR #3827, opened 2024-02-22
Allow sdk redirect on settings actions

IniZio:
ref #3813

Initially wanted to apply this derive function on all settings sub-pages. However it will be odd to, for example close the webview on remove an identity in `/settings/identity` page

### DIFF
--- a/pkg/auth/webapp/redirect.go
+++ b/pkg/auth/webapp/redirect.go
@@ -7,6 +7,10 @@ import (
 	"github.com/authgear/authgear-server/pkg/util/httputil"
 )
 
+var reservedRedirectURIs = []string{
+	"authgearsdk://host/path", // For Authgear SDK only, used for closing the webview
+}
+
 func GetRedirectURI(r *http.Request, trustProxy bool, defaultURI string) string {
 	redirectURI, err := httputil.GetRedirectURI(r, trustProxy)
 	if err != nil {
@@ -17,6 +21,42 @@ func GetRedirectURI(r *http.Request, trustProxy bool, defaultURI string) string 
 
 type OAuthClientResolver interface {
 	ResolveClient(clientID string) *config.OAuthClientConfig
+}
+
+func DeriveSettingsRedirectURIFromRequest(r *http.Request, defaultURI string) string {
+	// 1. Redirect URL in query param (must be whitelisted)
+	// 2. Default redirect URL
+	// 3. `/settings`
+	redirectURIFromQuery := func() string {
+		redirectURI := r.URL.Query().Get("redirect_uri")
+		allowed := false
+
+		for _, u := range reservedRedirectURIs {
+			if u == redirectURI {
+				allowed = true
+				break
+			}
+		}
+
+		// 1. Redirect URL in query param (must be whitelisted)
+		if allowed && redirectURI != "" {
+			return redirectURI
+		}
+
+		return ""
+	}()
+
+	if redirectURIFromQuery != "" {
+		return redirectURIFromQuery
+	}
+
+	// 2. Default redirect URL
+	if defaultURI != "" {
+		return defaultURI
+	}
+
+	// 3. `/settings`
+	return "/settings"
 }
 
 func DerivePostLoginRedirectURIFromRequest(r *http.Request, clientResolver OAuthClientResolver, uiConfig *config.UIConfig) string {
@@ -38,7 +78,7 @@ func DerivePostLoginRedirectURIFromRequest(r *http.Request, clientResolver OAuth
 		allowedURIs := client.RedirectURIs
 		allowed := false
 
-		for _, u := range allowedURIs {
+		for _, u := range append(reservedRedirectURIs, allowedURIs...) {
 			if u == redirectURI {
 				allowed = true
 				break
--- a/pkg/auth/webapp/redirect.go
+++ b/pkg/auth/webapp/redirect.go
@@ -5,11 +5,8 @@ import (
 
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/util/httputil"
+	"github.com/iawaknahc/originmatcher"
 )
-
-var reservedRedirectURIs = []string{
-	"authgearsdk://host/path", // For Authgear SDK only, used for closing the webview
-}
 
 func GetRedirectURI(r *http.Request, trustProxy bool, defaultURI string) string {
 	redirectURI, err := httputil.GetRedirectURI(r, trustProxy)
@@ -23,19 +20,29 @@ type OAuthClientResolver interface {
 	ResolveClient(clientID string) *config.OAuthClientConfig
 }
 
-func DeriveSettingsRedirectURIFromRequest(r *http.Request, defaultURI string) string {
+func DeriveSettingsRedirectURIFromRequest(r *http.Request, clientResolver OAuthClientResolver, defaultURI string) string {
 	// 1. Redirect URL in query param (must be whitelisted)
 	// 2. Default redirect URL
 	// 3. `/settings`
 	redirectURIFromQuery := func() string {
+		clientID := r.URL.Query().Get("client_id")
 		redirectURI := r.URL.Query().Get("redirect_uri")
-		allowed := false
+		if clientID == "" {
+			return ""
+		}
+		client := clientResolver.ResolveClient(clientID)
+		if client == nil {
+			return ""
+		}
 
-		for _, u := range reservedRedirectURIs {
-			if u == redirectURI {
-				allowed = true
-				break
-			}
+		allowed := true
+		matcher, err := originmatcher.New(client.SettingsRedirectURIOrigins)
+		if err != nil {
+			return ""
+		}
+
+		if matcher.MatchOrigin(redirectURI) {
+			allowed = true
 		}
 
 		// 1. Redirect URL in query param (must be whitelisted)
@@ -75,10 +82,8 @@ func DerivePostLoginRedirectURIFromRequest(r *http.Request, clientResolver OAuth
 			return ""
 		}
 
-		allowedURIs := client.RedirectURIs
 		allowed := false
-
-		for _, u := range append(reservedRedirectURIs, allowedURIs...) {
+		for _, u := range client.RedirectURIs {
 			if u == redirectURI {
 				allowed = true
 				break
--- a/pkg/auth/webapp/redirect.go
+++ b/pkg/auth/webapp/redirect.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/util/httputil"
-	"github.com/iawaknahc/originmatcher"
 )
 
 func GetRedirectURI(r *http.Request, trustProxy bool, defaultURI string) string {
@@ -36,15 +35,6 @@ func DeriveSettingsRedirectURIFromRequest(r *http.Request, clientResolver OAuthC
 		}
 
 		allowed := true
-		matcher, err := originmatcher.New(client.SettingsRedirectURIOrigins)
-		if err != nil {
-			return ""
-		}
-
-		if matcher.MatchOrigin(redirectURI) {
-			allowed = true
-		}
-
 		// 1. Redirect URL in query param (must be whitelisted)
 		if allowed && redirectURI != "" {
 			return redirectURI
--- a/pkg/auth/webapp/redirect.go
+++ b/pkg/auth/webapp/redirect.go
@@ -19,43 +19,6 @@ type OAuthClientResolver interface {
 	ResolveClient(clientID string) *config.OAuthClientConfig
 }
 
-func DeriveSettingsRedirectURIFromRequest(r *http.Request, clientResolver OAuthClientResolver, defaultURI string) string {
-	// 1. Redirect URL in query param (must be whitelisted)
-	// 2. Default redirect URL
-	// 3. `/settings`
-	redirectURIFromQuery := func() string {
-		clientID := r.URL.Query().Get("client_id")
-		redirectURI := r.URL.Query().Get("redirect_uri")
-		if clientID == "" {
-			return ""
-		}
-		client := clientResolver.ResolveClient(clientID)
-		if client == nil {
-			return ""
-		}
-
-		allowed := true
-		// 1. Redirect URL in query param (must be whitelisted)
-		if allowed && redirectURI != "" {
-			return redirectURI
-		}
-
-		return ""
-	}()
-
-	if redirectURIFromQuery != "" {
-		return redirectURIFromQuery
-	}
-
-	// 2. Default redirect URL
-	if defaultURI != "" {
-		return defaultURI
-	}
-
-	// 3. `/settings`
-	return "/settings"
-}
-
 func DerivePostLoginRedirectURIFromRequest(r *http.Request, clientResolver OAuthClientResolver, uiConfig *config.UIConfig) string {
 	// 1. Redirect URL in query param (must be whitelisted)
 	// 2. Default redirect URL of the client
@@ -72,8 +35,10 @@ func DerivePostLoginRedirectURIFromRequest(r *http.Request, clientResolver OAuth
 			return ""
 		}
 
+		allowedURIs := client.RedirectURIs
 		allowed := false
-		for _, u := range client.RedirectURIs {
+
+		for _, u := range allowedURIs {
 			if u == redirectURI {
 				allowed = true
 				break
--- a/pkg/auth/webapp/service2.go
+++ b/pkg/auth/webapp/service2.go
@@ -504,7 +504,7 @@ func (s *Service2) afterPost(
 			result.NavigationAction = "replace"
 		}
 	}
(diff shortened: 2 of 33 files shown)


pkg/auth/webapp/redirect.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/authgear/authgear-server/pkg/util/httputil"
 )
 
+var reservedRedirectURIs = []string{
+	"authgearsdk://host/path", // For Authgear SDK only, used for closing the webview
+}
```
tung2744, line 12: I think we can simply allow all uri with scheme `authgearsdk://`?

IniZio: Wanted to be specific since this is the only use case, updated to compare schema instead 🙏

resources/authgear/templates/en/web/settings_close.html
```diff
@@ -0,0 +1,38 @@
...
+</div>
+
+<script nonce="{{ $.CSPNonce }}">
+(function() {
+  var authgear = window.authgear || {};
+  authgear.close = function() {
+    const hasParent = window.parent !== window;
+    if (!hasParent) {
+      // Remove close page from history
+      window.history.replaceState(null, '', window.location.href);
```
louischan-oursky, line 20: Does this work if the end-user navigates back?

resources/authgear/templates/en/web/settings_close.html
```diff
@@ -0,0 +1,38 @@
...
+      location.href = document.referrer;
+      return;
+    }
+
+    window.close();
```
louischan-oursky, line 25: If the Web SDK supports two modes: 1) open in the same tab, 2) open in a new tab, then we should

1. If open in the same tab, then we should call `history.go(-n)` where `n` is the correct number that brings the end-user back to the app.
2. If open in the new tab, then the redirectURI is supposed to be `javascript:window.close()`. And if it is `href` of a `<a>`, then visiting the link will close the tab.

pkg/auth/handler/webapp/enter_login_id.go
```diff
@@ -83,6 +83,7 @@ type EnterLoginIDHandler struct {
 	AuthenticationViewModel *viewmodels.AuthenticationViewModeler
 	Renderer                Renderer
 	Identities              EnterLoginIDService
+	OAuthClientResolver     WebappOAuthClientResolver
```
louischan-oursky, line 86: Do we need this line of change?

pkg/lib/config/oauth.go
```diff
@@ -119,6 +119,10 @@ var _ = Schema.Add("OAuthClientConfig", `
 		"client_uri": { "type": "string", "format": "uri" },
 		"client_name": { "type": "string", "minLength": 1 },
 		"name": { "type": "string" },
+		"x_settings_redirect_uri_origins": {
+			"type": "array",
+			"items": { "type": "string", "format": "uri" }
```
louischan-oursky, line 124: We have a format `http_origin` for an origin. Use that instead.

pkg/lib/oauth/handler/resolve.go
```diff
@@ -101,3 +102,64 @@ func validateRedirectURI(
...
+		queryItems.Set("client_id", r.ClientID())
+
+		settingsRedirectURI.RawQuery = queryItems.Encode()
+		return settingsRedirectURI, nil
+	default:
+		return nil, protocol.NewErrorResponse("invalid_request", "invalid settings action")
+	}
+}
+
+func validatePostSettingsRedirectURI(
```
louischan-oursky, line 141: Rename it to `validateSettingsRedirectURI`

pkg/lib/oauth/handler/resolve.go
```diff
@@ -101,3 +102,64 @@ func validateRedirectURI(
...
+	if matcher.MatchOrigin(redirectURIString) {
+		return nil
+	}
+
+	err = validateRedirectURI(client, httpProto, httpOrigin, domainWhitelist, redirectURI)
+	if err != nil {
+		return err
+	}
+
+	return errors.New("redirect URI is not allowed")
```
louischan-oursky, line 164: I think we should return nil here? If validateRedirectURI returns nil, that means the redirect URI is valid.

pkg/lib/oauth/handler/resolve.go
```diff
@@ -101,3 +102,64 @@ func validateRedirectURI(
...
+	}
+
+	err = validatePostSettingsRedirectURI(client, httpProto, httpOrigin, domainWhitelist, redirectURI)
+	if err != nil {
+		return nil, protocol.NewErrorResponse("invalid_request", err.Error())
+	}
+
+	switch r.SettingsAction() {
+	case "change_password":
+		settingsRedirectURI := &url.URL{Path: "/settings/change_password"}
```
louischan-oursky, line 129: We should follow the same logic used in UIURLBuilder to build the URI.

1. UIURLBuilder.Build currently build authentication URL. Let's rename it to BuildAuthenticationURL
2. Add UIURLBuilder.BuildSettingsURL for building settings URL.

IniZio:
@tung2744 @louischan-oursky Updated to reflect latest settings action spec, still adding test cases for new response type and grant type 🙏

pkg/auth/webapp/service2.go
```diff
@@ -538,6 +551,16 @@ func (s *Service2) afterPost(
...
 		if e, ok := graph.GetAuthenticationInfoEntry(); ok {
+			code, err := s.generateSettingsActionGrant(redirectURI, e)
+			if err != nil {
+				panic(err)
+			}
+
+			if code != "" {
+				e = &authenticationinfo.Entry{
+					ID: code,
+				}
```
louischan-oursky, line 562: Why do we need this block?

pkg/auth/webapp/service2.go
```diff
@@ -564,6 +587,68 @@ func (s *Service2) deriveFinishRedirectURI(session *Session, graph *interaction.
...
 }
 
+func (s *Service2) generateSettingsActionGrant(redirectURI string, e *authenticationinfo.Entry) (string, error) {
+	entry, err := s.OAuthSessions.Get(e.OAuthSessionID)
+	if err != nil {
+		return "", err
+	}
+
+	req := entry.T.AuthorizationRequest
+	if req.ResponseType() != "settings_action" {
```
louischan-oursky, line 597: I think the literal here is incorrect. Use the defined constant.

pkg/auth/webapp/service2.go
```diff
@@ -564,6 +587,68 @@ func (s *Service2) deriveFinishRedirectURI(session *Session, graph *interaction.
...
+	}
+
+	ss := session.GetSession(s.Request.Context())
+	code, _, err := s.SettingsActionGrantService.CreateSettingsActionGrant(&handler.CreateSettingsActionGrantOptions{
+		Authorization:        authz,
+		IDPSessionID:         ss.SessionID(),
+		AuthenticationInfo:   authenticationInfo,
+		IDTokenHintSID:       idTokenHintSID,
+		RedirectURI:          req.RedirectURI(),
+		AuthorizationRequest: req,
```
louischan-oursky, line 643: This is overly complicated. Why do we need to create code here?

pkg/lib/interaction/nodes/do_ensure_session.go
```diff
@@ -221,6 +222,21 @@ func (n *NodeDoEnsureSession) GetEffects() ([]interaction.Effect, error) {
 				}
 			}
 
+			return nil
+		}),
+		interaction.EffectOnCommit(func(ctx *interaction.Context, graph *interaction.Graph, nodeIndex int) error {
```
louischan-oursky, line 227: Why do we need to add this?

pkg/lib/interaction/nodes/settings_action_end.go
```diff
@@ -0,0 +1,46 @@
...
+type NodeSettingsActionEnd struct{}
+
+func (n *NodeSettingsActionEnd) Prepare(ctx *interaction.Context, graph *interaction.Graph) error {
+	return nil
+}
+
+func (n *NodeSettingsActionEnd) GetEffects() ([]interaction.Effect, error) {
+	return []interaction.Effect{
+		interaction.EffectOnCommit(func(ctx *interaction.Context, graph *interaction.Graph, nodeIndex int) error {
+			entry, err := ctx.OAuthSessions.Get(ctx.OAuthSessionID)
```
louischan-oursky, line 27: Should do nothing if OAuthSession is not found (it could happen. We can change password directly in the settings page)

pkg/lib/interaction/nodes/settings_action_end.go
```diff
@@ -0,0 +1,46 @@
...
+			if err != nil {
+				return err
+			}
+
+			if entry.T.AuthorizationRequest.ResponseType() != "settings_action" {
```
louischan-oursky, line 32: No need to check. We can always save the result event if it was not a settings action in the first place.

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -25,15 +25,31 @@ import (
 
 //go:generate mockgen -source=handler_authz.go -destination=handler_authz_mock_test.go -package handler_test
 
+const (
+	CodeResponseType          = "code"
+	NoneResponseType          = "none"
+	SettingsActonResponseType = "settings_action"
```
louischan-oursky, line 31: Name this `urn:authgear:params:oauth:response-type:settings-action`

IniZio: Note to self: Update SDKs after server pr stabilized

pkg/auth/webapp/service2.go
IniZio: Should follow ctrl redirect to consent screen during settings action, removed.

pkg/lib/interaction/nodes/do_ensure_session.go
IniZio: Redundant, removed.

pkg/auth/webapp/service2.go
IniZio: Should not try to override entry ID, instead generate settings action response in consent. Removed.

IniZio:
Tried to create a stimulus controller that can override history items https://github.com/IniZio/authgear-server/tree/3813-settings-action-redirect-web

The script itself works in other flows, but the consent page will block the `history.back()`. Tried reauth as well, seems have same issue where pressing back to consent page will show bad request error.

In this pr, I inserted a `/settings/close` page instead that redirects user back to original redirect URI.

IniZio:
After discussion, will make web behave similarly to other platforms and only guard back-navigation at consent screen instead.

Changes:
- Update consent screen to hold `client_id` and `redirect_uri`. If OAuth session is not found, use `redirect_uri` instead
- Remove `x_authgear_sdk_on_finish` and close screen

pkg/lib/interaction/intents/change_password.go
```diff
@@ -45,9 +45,23 @@ func (i *IntentChangePassword) DeriveEdgesForNode(graph *interaction.Graph, node
 			},
 		}, nil
 	case *nodes.NodeChangePasswordEnd:
-		// Password was not changed, ends the interaction
-		return nil, nil
+		return []interaction.Edge{
+			&nodes.EdgeDoEnsureSession{
+				Mode: nodes.EnsureSessionModeNoop,
+			},
+		}, nil
```
louischan-oursky, line 52: Why do we need to add EdgeDoEnsureSession here?

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -108,7 +125,18 @@ func (h *AuthorizationHandler) Handle(r protocol.AuthorizationRequest) httputil.
...
+		AuthorizationRequest: r,
+	})
+	err := h.OAuthSessionService.Save(oauthSessionEntry)
+	if err != nil {
+		return authorizationResultError{
+			ResponseMode: r.ResponseMode(),
+			Response:     protocol.NewErrorResponse("server_error", "internal server error"),
+		}
+	}
+	redirectURI, errResp := parseAuthzRedirectURI(client, h.UIURLBuilder, h.HTTPProto, h.HTTPOrigin, h.AppDomains, oauthSessionEntry, r)
```
louischan-oursky, line 139: Please do not alter the flow. h.OAuthSessionService.Save was originally used in doHandle. Now it was used more than once.

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -213,7 +242,8 @@ func (h *AuthorizationHandler) doHandleConsent(req *http.Request, withUserConsen
 			RedirectURI: nil,
 		}
 		if errors.As(err, &oauthError) {
-			resultErr.Response = oauthError.Response
+			resultErr := h.prepareErrInvalidOAuthResponse(req, *oauthError)
+			return resultErr, nil
```
louischan-oursky, line 246: Convert this pattern

```
resultErr := authorizationResultError{
  RedirectURI: nil
}
resultErr := h.prepareErrInvalidOAuthResponse(req, *oauthError)
return resultErr, nil
```

To

```
resultErr := h.makeOAuthErrorResponse(req)
```

Rationale
1. Shadowing a local variable and return it in a if block is hard to read.
2. The original code does not have multiple returns. And after this change, sometimes it early returns, sometimes it does not. This is even more hard to read.

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -398,7 +428,7 @@ func (h *AuthorizationHandler) doHandle(
 	// Handle prompt!=none
 	// We must return here.
 	if !slice.ContainsString(uiInfo.Prompt, "none") {
-		endpoint, err := h.UIURLBuilder.Build(client, r, oauthSessionEntry)
+		endpoint, err := h.UIURLBuilder.BuildAuthenticationURL(client, r, oauthSessionEntry)
```
louischan-oursky, line 431: We will remove parseAuthzRedirectURI, and in here, we will choose between BuildAuthenticationURL and BuildSettingsActionURL. In doing this, we preserve the original code flow, and let the reader to reason the code easier.

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -534,6 +570,7 @@ func (h *AuthorizationHandler) validateRequest(
 	if len(allowedResponseTypes) == 0 {
 		allowedResponseTypes = []string{"code"}
 	}
+	allowedResponseTypes = append(allowedResponseTypes, whiteslistedResponseTypes...)
```
louischan-oursky, line 573: line 570 to 572 is redundant now, because we always allow whitelistedResponseTypes. Remove those lines.

pkg/lib/oauth/oidc/ui.go
```diff
@@ -101,6 +101,12 @@ type UIInfoResolver struct {
 }
 
 func (r *UIInfoResolver) SetAuthenticationInfoInQuery(redirectURI string, e *authenticationinfo.Entry) string {
+	consentURI := r.EndpointsProvider.ConsentEndpointURL().String()
+	// Not redirecting to the consent endpoint.
+	// Do not set anything.
+	if redirectURI != consentURI {
+		return redirectURI
+	}
```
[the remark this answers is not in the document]
IniZio: It's in the original code, was removed in a temp commit for the old approach for settings action. New commit added it back.

pkg/lib/interaction/intents/change_password.go
IniZio: Currently it's possible to change password to same as previous value. Or should it be treated as fail for settings action?

tung2744: I think it should be a success, but the developer could enable password history check, which should cause the flow to return api error. That should be handled by password policy already.

louischan-oursky: Does this work? `EdgeDoEnsureSession` should be irrelevant here.

```
case *nodes.NodeChangePasswordEnd:
  // Password was not changed
  return []interaction.Edge{&nodes.EdgeSettingsActionEnd{}}, nil
case *nodes.NodeDoUpdateAuthenticator:
  return []interaction.Edge{&nodes.EdgeSettingsActionEnd{}}, nil
case *nodes.NodeSettingsActionEnd:
  // Intent is finished.
  return nil, nil
```

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -606,3 +640,60 @@ func (h *AuthorizationHandler) generateCodeResponse(
...
+	if client == nil {
+		return resultErr
+	}
+
+	err = validateRedirectURI(client, h.HTTPProto, h.HTTPOrigin, h.AppDomains, redirectURI)
+	if err != nil {
+		return resultErr
+	}
+
+	resultErr.RedirectURI = redirectURI
```
louischan-oursky, line 696: In the places where prepareErrInvalidOAuthResponse is called, redirectURI is already parsed. So prepareErrInvalidOAuthResponse should take the parsed redirect URI instead of *http.Request.

pkg/lib/interaction/intents/change_password.go
IniZio: Should still need, similar to `IntentReauthenticate`. Tried removing it results in invalid `authentication required`

Issue #3826, opened 2024-02-21
AuthUI v2 - Combined Signup or login flow


fungc-io:
## Goal

- Provide an config for the project owner to set the Auth flow and the auth ui to Combined signup/login flow. While it's not configurable in the portal yet. Allow the project to set in the backend server.
- In the AuthUI, there will not be the Create Account/Login switch.
  - If an enduser entered an unregister login ID, they will go to the signup flow
  - Vice versa, if the end-user entered an existing Login ID, they will go to login.

## Design

Figma: https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=742-32101&mode=dev

chpapa:
> While it's not configurable in the portal yet. Allow the project to set in the backend server.

@fungc-io can we have the option in portal as well? I think this is too common that we shouldn't skip it for later

(In fact, once we have it we should make it the default for new project)

I think we can have a simple checkbox above "passkey support" at the login method page?

tung2744:
![Image](https://github.com/authgear/authgear-server/assets/34942210/e9c91a4a-6658-4bdf-9152-ba8e1c2be226)

Maybe we can add a toggle here under the passkey option?

chpapa:
@tung2744 yes exactly what I meant, yet I think this is more important than Passkey so want it to be above.

(But it is a quick fix, maybe we can consult designer later in future)

fungc-io:
thanks, let's do that

tung2744:
- [ ] Update authgear config (3)
    - Add a flag to enable signup login flow
    ```yaml
    ui:
        combine_signup_login_flow: true # true / false
    ```

- [ ] Update portal ui for combined signup login flow config (3)

- [ ] Support combined signup login flow in server (5)
    - Extract logics in signup handler to a new struct `InternalSigninHandler`.
        - `InternalSigninHandler` should expose a method `ServeHTTP(w http.ResponseWriter, r *http.Request, flowType authflow.FlowType)`
    - In signup handler, if config `ui.combine_signup_login_flow` is `true`, Use `signup_login` as flow type to call `InternalSigninHandler.ServeHTTP`
    - In login handler, if config `ui.combine_signup_login_flow` is `true`, call `InternalSigninHandler.ServeHTTP`, else do the original logic.
    - In ui of sign up, hide the login button if `ui.combine_signup_login_flow` is `true`.


[...]

PR #3834, opened 2024-02-23
Implement combined signup login flow

tung2744:
ref #3826

Screenshot 2024-02-23 at 12 09 41 PM


louischan-oursky:
It does not work with `public_signup_disabled: true`

tung2744:
Updated to render the original logic ui if `public_signup_disabled: true`.

louischan-oursky:
Removed the "Remove duplicate translation key" commit. It is because the key was removed already in https://github.com/authgear/authgear-server/pull/3836

Issue #3603, opened 2024-01-10
AuthUI V2 - Dark Mode support


tung2744:
Add dark mode styles for all components and screen.

fungc-io:
- https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=45-26651&mode=dev
- https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=696-10979&mode=design

PR #3835, opened 2024-02-23
Add dark mode support

Ngkaokis:
ref #3603

### DIFF
--- a/authui/src/authflowv2/components/input.css
+++ b/authui/src/authflowv2/components/input.css
@@ -1,5 +1,6 @@
 @layer components {
   :root {
+    --input__bg-color: transparent;
     --input__ring-color: var(--color-neutral-100);
     --input__ring-color--focus: var(--color-primary-theme-primary);
     --input__ring-color--error: var(--color-error);
@@ -41,6 +42,7 @@
 
 .input {
   @apply placeholder:[color:var(--input\_\_placeholder-color)];
+  background-color: var(--input__bg-color);
   padding: var(--input__py) var(--input__px);
   border-radius: var(--input__border-radius);
   color: var(--input__text-color);
--- a/authui/src/authflowv2/components/input.css
+++ b/authui/src/authflowv2/components/input.css
@@ -37,7 +37,9 @@
 }
 
 :root.dark {
-  /* TODO(davis): Dark mode styles */
+  --input__ring-color: var(--color-neutral-500);
+  --input__placeholder-color: var(--color-neutral-400);
+  --input__text-color: var(--color-neutral-100);
 }
 
 .input {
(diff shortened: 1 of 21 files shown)


resources/authgear/templates/en/web/authflowv2/__password_input.html
```diff
@@ -38,7 +38,7 @@
     {{ if $.InputAttrs }}{{ htmlattr $.InputAttrs }}{{ end }}
   >
   <button
-    class="w-5 absolute inset-y-0 ltr:right-4 rtl:left-4"
+    class="input__password-icon"
```
tung2744, line 41: Should we apply this class to in inner `<span>` instead of outer button?

authui/src/authflowv2/components/input.css
```diff
@@ -32,15 +33,20 @@
...
-  /* TODO(davis): Dark mode styles */
+  --input__ring-color: var(--color-neutral-500);
+  --input__placeholder-color: var(--color-neutral-400);
+  --input__text-color: var(--color-neutral-100);
+  --input__password-icon-color: var(--color-neutral-300);
 }
 
 .input {
   @apply placeholder:[color:var(--input\_\_placeholder-color)];
+  background-color: var(--input__bg-color);
```
tung2744, line 49: input should be transparent? (Therefore it is the color of background)

authui/src/authflowv2/components/input.css
```diff
@@ -32,15 +33,20 @@
     );
 
     --input__password-padding-end: 2.375em; /* 38 / 16 = 2.375 */
+    --input__password-icon-color: var(--color-neutral-500);
```
tung2744, line 36: Lets name it ``--input__password-visibility-icon-color` which seems more clear what is this icon.

authui/src/authflowv2/components/otp-input.css
```diff
@@ -53,6 +63,7 @@
       @apply caret-white/0;
       /* iOS treats element with `opacity: 0` as not interactable, set text to transparent instead */
       @apply text-transparent;
+      background-color: var(--otp-input\_\_bg-color);
```
tung2744, line 66: As this is not in tailwind, I think we don't need to escape `__`?

authui/src/authflowv2/components/otp-input.css
```diff
@@ -74,6 +85,7 @@
     width: var(--otp-input\_\_digit-width);
     height: var(--otp-input\_\_digit-height);
     border-radius: var(--otp-input\_\_digit-border-radius);
+    color: var(--otp-input\_\_digit-color);
```
tung2744, line 88: unnecessary escape

Issue #3656, opened 2024-01-19
Review All error message and copywriting in the AuthUI v2


fungc-io:
For all EN, zh-Hant, and zh-Hans
- Fill in all TBC copywriting
- Review all copywriting and make sure they are clear

fungc-io:
@tung2744 ,
please help compile the list of copywriting by the end of tmr for review

tung2744:
@fungc-io Please find the list of copywritings of v2 ui below.
The zh translations are basically translated from en version.

[zh.json](https://github.com/authgear/authgear-server/files/14178831/zh.json)
[en.json](https://github.com/authgear/authgear-server/files/14178832/en.json)

Ngkaokis:
Need to review the copy writing of 
1. verify login link as it's shared by both signup and login
2. Sign up page's `sign up with email` button  

ref #3787

louischan-oursky:
We also want to tidy up the translation.json files a bit.

- Ensure all v2-* keys are placed at the end of the file.
- Some v2 keys are using `.` We do not use `.`. Correct them to use `-`.
- (Optional) Sort the v2 keys alphabetically. Add linter to ensure this ordering.

fungc-io:
@tung2744 @louischan-oursky 
I've reviewed the copywriting
https://docs.google.com/spreadsheets/d/1qZLiazkEJbLZ44NwbZZO7luL7CKLzEi_hsqF__x8j60/edit#gid=137567725

I've left comments on row 6,12,14

Maybe we can document the following translation for future reference too:

| En | Hong Kong | Taiwan | Notes |
| --- | --- | --- | --- |
| Account | 帳戶 | 帳號 | |
| Username | 用戶名稱 | 用戶名稱 | instead of "用戶名"/"使用者名稱"... |
| Email / Email address | 電郵 / 電郵地址 | 電郵 / 電郵地址 | instead of 電子郵件/電子郵件地址
| Passkeys | 通行密匙 | 通行密鑰 | |
| Login with ... | 以...登入 | 使用...登入 | e.g. 以 Apple 帳戶登入, 使用 Apple 帳號登入 |
| a verification code | 一組驗證碼 | 一組驗證碼| instead of 一"段"驗證碼 | 
| a message | 一則訊息 | 一則訊息 | instead of 一條 | 
| link | 連結 | 連結 | instead of "鏈接" |
| QR Code | QR Code | QR Code | instead of 二維碼 |
[...]

PR #3836, opened 2024-02-23
Add script for extracting v2 translation to csv

tung2744:
ref #3656

Expected usage:
export: `npm run -s export-v2-translations > translation.csv`
import: `cat translation.csv | npm run import-v2-translations`


scripts/npm/export-v2-translations.mjs
```diff
@@ -0,0 +1,80 @@
...
+  values: { "output-file": outputFile },
+} = parseArgs({
+  options: {
+    "output-file": {
+      short: "f",
+      type: "string",
+      default: "output/v2-translations.csv",
+    },
+  },
+});
```
louischan-oursky, line 21: Remove the -f option and output to stdout.

tung2744: @louischan-oursky 

If running it by npm, there will be some logs in stdout, which pollute the output file.
```
> node@0.1.0 export-v2-translations
> node export-v2-translations.mjs
```

louischan-oursky: Try `npm run -s <script>`

tung2744: Then the script must run with specific argument to work, which seems a bit weird?

louischan-oursky: I am very surprised that npm run its output to stdout, not stderr.

PR #3837, opened 2024-02-26
Specify settings action

IniZio:
ref #3813

This spec is a reincarnation of #2886. 

Changes from previous spec:
- Use new response type similar to authorization_code wit new grant type instead of internal settings action code
- Treat settings action as part of the authorization flow instead of plain webview

### DIFF
--- a/docs/specs/oidc.md
+++ b/docs/specs/oidc.md
@@ -17,6 +17,7 @@
     + [acr_values](#acr_values)
     + [code_challenge_method](#code_challenge_method)
     + [nonce](#nonce)
+    + [x_settings_action](#x_settings_action)
   * [Token Request](#token-request)
     + [grant_type](#grant_type)
     + [id_token_hint](#id_token_hint-1)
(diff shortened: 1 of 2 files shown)


docs/specs/oidc.md
```diff
@@ -229,11 +237,14 @@ To mitigate replay attacks, provide a `nonce` in the authentication request. Aut
 - `urn:authgear:params:oauth:grant-type:biometric-request`
 - `urn:authgear:params:oauth:grant-type:id-token`
 - `urn:authgear:params:oauth:grant-type:authorization_code`
+- `urn:authgear:params:oauth:grant-type:settings-action`
 
 `urn:authgear:params:oauth:grant-type:anonymous-request` is for authenticating and issuing tokens directly for anonymous user.
 
 `urn:authgear:params:oauth:grant-type:biometric-request` is for authenticating and issuing tokens directly for users with Biometric identity.
 
+`urn:authgear:params:oauth:grant-type:settings-action` is issued upon completion of the settings action, such as change password. The token request must include `x_settings_action` with the value of the settings action.
```
louischan-oursky, line 246: > The token request must include `x_settings_action` with the value of the settings action.

Any reason for this requirement?

docs/specs/oidc.md
```diff
@@ -631,6 +642,67 @@ When the app session token is consumed:
...
+After redirecting back to the app, An code of grant type `urn:ietf:params:oauth:grant-type:settings-action` will be returned to the app. The app can exchange the code to prove the completion of the settings action.
+
+The following flow charts show how the settings actions work.
+
+```mermaid
+flowchart LR
+    subgraph /oauth2/authorize
+        authz[authz endpoint]
+    end
+    subgraph settings[e.g. /settings/change_email]
```
louischan-oursky, line 662: Use /change_password as an example (because it is the first settings action implemented)

docs/specs/oidc.md
IniZio: Should not be needed, removed.

Issue #3823, opened 2024-02-21
Consider adding identifier for each data type in authflow


tung2744:
Currently, same action could have different data returned in different steps.
For example, when `action.type` is `create_authenticator`, and `authentication` is `secondary_oob_otp_sms`, there are two possible data types can be returned: `NodeVerifyClaimData` and `IntentSignupFlowStepCreateAuthenticatorData`.
If `NodeVerifyClaimData` is returned, user should input a otp. If `IntentSignupFlowStepCreateAuthenticatorData` is returned, user should provide a `target`.

Currently, in our webapp, we check the internal type of the data and do corresponding logic. However, it will be impossible to do so for external integrations which call the http api.

Should we consider to add a `type` to each data type we returned?

IniZio:
Had same thought when exploring authflow testing as well. Thinking if we could provide identifier for next steps and corresponding schemas?

PR #3841, opened 2024-02-26
Give a type name to all data

tung2744:
ref #3823 

- Merged those data which are having same purpose into one.
- Give names to each data type.
- Update authflow api spec and doc.

### DIFF
--- a/docs/specs/authentication-flow.md
+++ b/docs/specs/authentication-flow.md
@@ -882,7 +882,7 @@ The data contains recovery codes of the user.
 
 `recovery_codes`: The recovery codes of the user.
 
-#### oob_channels_data
+#### select_oob_otp_channels_data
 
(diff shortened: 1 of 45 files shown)


docs/specs/authentication-flow.md
```diff
@@ -841,3 +842,104 @@ Supertokens requires the developer to host a backend server to interactive with
...
+
+`options`: The list of creatable authenticators.
+
+#### view_recovery_code_data
+
+The data contains recovery codes of the user.
+
+`recovery_codes`: The recovery codes of the user.
+
+#### oob_channels_data
```
louischan-oursky, line 885: Rename this to `select_oob_otp_channels_data`

tung2744: Updated, thanks!

Issue #3517, opened 2023-11-28
API to Import Users and create User


fungc-io:
## Problem

To add the users to the Authgear project using the API rather than the users to signup by themselves.
There are a few problems with the existing "createUser" mutation:

- Users are added one-by-one
- It does not support create by the password hash, it can only be used when the password is known to the system. This forbid the migration of users from other systems using the API.
- The password must conform to the password policy even when using the API

### Use cases to consider
- In Company A:
  - Migrate the users from a legacy system to Authgear. They want to keep the user attributes and User ID. They can export the data into a CSV. 
- In Company B: 
  - User is created with a weaker password by the admin, when the user log in the 1st time, they must set up their own password which conform to the stricter policy

## Appetite

2 weeks

### importUsers

- Introduce a new "importUsers" mutation, which can take a list of users data and create users in bulk.
- When import users, developers can provide
  - User ID (Optional)
  - Login ID & its type (email/phone/username)
  - email_verified and phone_verified
  - Password in plain text; Or to provide the Hashed password with hashing_method (Optional)
  - The standard attributes & custom attributes (Optional)
 - The pre-create and created hook are skipped when using the "importUsers" mutation
 - User with the same "User ID" will be updated. The developers can use this API to bulk update the attributes of the users.
 - Return in the response if the users failed to import with reasons, e.g. duplicated email/phone/username in the system.

### createUser

- User is created one-by-one, not in bulk
- When creating users, developers can provide
  - Login ID & its type (email/phone/username)
  - email_verified and phone_verified
  - Password in plain text, it will be checked against the password policy
    - Or to provide the Hashed password with hashing_method, the password policy will not be checked.
  - The standard attributes & custom attributes
  - The pre-create and created hook will be executed when using the "createUser" mutation

## Rabbit Hole
- We should support batch import considering there maybe hundreds of thousands of users
- The dev should get feedback on which rows cannot be imported and what validation it violated

fungc-io:
@louischan-oursky 
Some directions from todays meeting for reference
- We expect the user to upload a pre-process the data and create a JSON file conform to a data schema before importing
- A easy to understand sample JSON file should be provided
- The design should not be a replica of the Authgear internal structure, keywords like “Authenticators”, ”Identity”,… are difficult to understand. `mfa_sms:[]` is an acceptable field name.
- We need another API for batch delete/anonymization
- Please consider pre and post processing TS hook for import

PR #3854, opened 2024-02-27
Specify user import

louischan-oursky:
ref #3517

### DIFF
--- a/docs/specs/user-import.md
+++ b/docs/specs/user-import.md
@@ -32,7 +32,7 @@ The input is a JSON document. Here is an example.
 {
   "upsert": true,
   "identifier": "email",
-  "users": [
+  "records": [
     {
       "preferred_username": "louischan",
       "email": "louischan@oursky.com",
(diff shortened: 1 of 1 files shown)


docs/specs/user-import.md
```diff
@@ -0,0 +1,233 @@
...
+  "id": "some_opaque_string",
+  "created_at": "2024-01-01T00:00:00.000Z"
+  "status": "completed",
+  "summary": {
+    "total": 100,
+    "inserted": 50,
+    "updated": 49,
+    "failed": 1
+  },
+  "errors": [
```
tung2744, line 161: Can we include the item index here, so user can easily find the item?

louischan-oursky:
Note to myself: Add warnings for IGNORED attributes.

docs/specs/user-import.md
louischan-oursky: JSON path seems overkill. I added `index` instead.

louischan-oursky:
Removed support for plain password. The rationale is, if they know the plain password, they can easily compute the bcrypt hash of it. So taking plain password is unnecessary (and it is insecure)

tung2744:
@louischan-oursky I guess we can merge this?

Issue #3850, opened 2024-02-26
Should automatically remove leading 0 from local phone number in phone input


fungc-io:
In many countries such as UK and Taiwan,
The local phone number is preceded by a leading `0` when used locally. It's normal for user to type the `0` into the phone number field when using a local service.

e.g. a number `+44 7712 123456` is usually entered as `07712 123456`

It's treated as invalid in the new AuthUI.

However in the old implementation `+44 07712 123456` would be recognized as `+44 7712 123456` correctly

![Image](https://github.com/authgear/authgear-server/assets/15207470/dddcc695-f7ac-4fd8-84e0-a0f2d08f5997)

tung2744:
It seems instead of handling in authui, we should handle it in server.

The phone number should be normalized in server side.

tung2744:
I wanted to fix it in server side, however, I found that it is related to the PhoneChecker https://github.com/authgear/authgear-server/blob/main/pkg/lib/authn/identity/loginid/type_checkers.go#L281 which calls the format checker https://github.com/authgear/authgear-server/blob/main/pkg/util/validation/formats.go#L64.
And this checker is used in many places to validate phone inputs.
So to prevent changing behavior of many places, I think I will fix it in ui.

Issue #3853, opened 2024-02-26
The country code appears in the input field when the format is wrong


fungc-io:
When the entered number is in wrong format, e.g. a 8 digit number is entered to +44 phone, the `+44` appears on submit.

I'm not sure the cause of it, because i don't see the same problem when "Hong Kong" is chosen.
For example
- In the country list, pick "Hong Kong"
- Enter a 7 digit or 9 digit number,
- The bug do not happen.

https://github.com/authgear/authgear-server/assets/15207470/06c982e7-1583-4f77-8009-9ef455b6ccce

Issue #3852, opened 2024-02-26
The type-ahead country code should be hidden upon input


fungc-io:
Old implementation, the country code is hidden after certain digits are entered

https://github.com/authgear/authgear-server/assets/15207470/0e764b11-b223-4788-8f33-8701ed555b3e

New implementation, the country code will not go away

https://github.com/authgear/authgear-server/assets/15207470/d8ac2ab4-8c94-4a7f-91ce-66b68537b696

louischan-oursky:
I doubt whether changing the text when the user is still editing is a good UX. Maybe we can remove the country calling code when the input is blur?

tung2744:
Agree. And actually the old auth ui also won't remove the country code immediately.

fungc-io:
@tung2744 oh right, i've just tested that again.
Why the behavior is different in Portal and old AuthUI?

tung2744:
It seems we did some special handling on portal.

fungc-io:
🤔 , ok
Let's trim the country code only on blur

PR #3860, opened 2024-02-27
Some phone input fixes

tung2744:
ref #3850 #3852 #3853

### DIFF
--- a/authui/src/authflowv2/customSelect.ts
+++ b/authui/src/authflowv2/customSelect.ts
@@ -340,7 +340,9 @@ export class CustomSelectController extends Controller {
 
     const fragment = document.createDocumentFragment();
 
-    const filteredOptionsValues = this.filteredOptions.reduce<Set<string>>(
+    const filteredOptions = this.filteredOptions;
+
+    const filteredOptionsValues = filteredOptions.reduce<Set<string>>(
       (valuesSet, option, _idx, _arr): Set<string> => {
         valuesSet.add(option.value);
         return valuesSet;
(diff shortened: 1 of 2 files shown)


louischan-oursky:
I found another bug.

1. Click on the country select, type in 853, press enter to select it.
2. Focus the text input, type in +852. See a warning in the console saying "Trying to select an option which does not exist"

tung2744:
Fixed #3850 in this pr too.

tung2744:
> I found another bug.
> 
> 1. Click on the country select, type in 853, press enter to select it.
> 2. Focus the text input, type in +852. See a warning in the console saying "Trying to select an option which does not exist"

Fixed in [b6854e0](https://github.com/authgear/authgear-server/pull/3860/commits/b6854e02bbef235c1092f56846bf248fed3e73df)

But the warning still appears on initial load, maybe let me open another issue to see how to fix that.

louischan-oursky:
But now I cannot press enter to select the first item in the filtered list :(

louischan-oursky:
Can we refactor so that

1. When the search box is focus, the list shows filtered items.
2. When the search box is focus, the first item in the list is selected.
3. When the search box is focus, pressing enter selects the first item (if there is at least one item). Otherwise, it is no-op.
4. Calling the API to select an item WILL NOT be affected by the search box. I think we need to maintain two lists. One is full, another one is filtered.

tung2744:
> But now I cannot press enter to select the first item in the filtered list :(

Fixed in bfea421

We actually already have two lists, the bug is caused by I forgot to compute the correct selected item in search mode.

louischan-oursky:
The handling of highlighted item in the list still has problem.

Proposed solution

- Introduce `private highlightIndex: number` in customSelect
- When searchTarget value changes, set highlightIndex to 0
- When Up or Down is hit, decrement or increment highlightIndex with bound checking

tung2744:
@louischan-oursky Updated again, thanks!

PR #3870, opened 2024-02-28
Update v2 translations

tung2744:
ref #3656


louischan-oursky:
Added a few commits to make the key names more consistent

louischan-oursky:
Since I have updated the keys, the spreadsheet is now outdated. @tung2744 @fungc-io

Issue #3845, opened 2024-02-26
Cannot search "Hong Kong" when "Hon" is entered


fungc-io:
Seems only happens in `Hong Kong`, it's not shown when only 3 characters are entered
I tested "Mac" would get "Macao", "Cha" would get "Chad", "Uni" would get "United Kingdom"

"Hong Kong" is only shown when "Hong" (4 characters) are entered

PR #3875, opened 2024-02-29
Fix country list keyword search includes dom element

IniZio:
ref #3845


louischan-oursky:
This change makes anything starts with `+` results in empty result set.

Issue #3627, opened 2024-01-16
[Portal, Localization] Blank screen if switch to a newly added language before save


joyzoursky:
## Environment
- Environment: Staging
- App: Any
- Browser: Chrome on Macbook Pro

## Description
If user add a localization on portal, and before save, switch to that language on the dropdown, it throws error and shows a blank screen.

## Steps to reproduce
1. Login to admin portal
2. Go to "Localization"
3. Click on the gear icon to add a language, e.g. Chinese (Traditional)
4. Set it as default
5. On the dropdown besides, switch to that newly added language
6. See blank screen (with error in console log)

## Expected behavior
Should either show the display of changed language, or do not let user choose the language before save.

## Screenshots
<img width="1472" alt="Screenshot 2024-01-16 at 5 08 55 PM" src="https://github.com/authgear/authgear-server/assets/12539486/233fce86-3804-4426-b5e7-f173b763ed71">

https://github.com/authgear/authgear-server/assets/12539486/540b9970-2043-436c-a34b-c9962e130de4

fungc-io:
Hi @newman
I think the UX is a bit weird now we don't show the newly added language before save.

This is a design flaw that need to be fixed by changing how a language is added. Without changing the design much now, how about showing the newly added lang as a disabled option.

e.g. After I added "Arabic" 

<img src="https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/915695c1-cd0b-4757-9f95-fedf73dcf5e5/0e54f65b-4283-436b-95dc-a36d954833e1?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi85MTU2OTVjMS1jZDBiLTQ3NTctOWY5NS1mZWRmNzNkY2Y1ZTUvMGU1NGY2NWItNDI4My00MzZiLTk1ZGMtYTM2ZDk1NDgzM2UxIiwiaWF0IjoxNzA5OTE1ODQyLCJleHAiOjMzMjgwNDc1ODQyfQ.ft_F4Armqeu_rZsShsYTCZFTt41QUD325val-qrgpyg " alt="image.png" width="392" height="172" />

And provide help text under the selection box to remind the user to save the changes

[...]

PR #3881, opened 2024-02-29
Fix portal handling for empty default localization

IniZio:
ref #3627

### DIFF
--- a/pkg/util/template/translation.go
+++ b/pkg/util/template/translation.go
@@ -81,18 +81,44 @@ func (t *translationJSON) ViewResources(resources []resource.ResourceFile, rawVi
 func (t *translationJSON) UpdateResource(_ context.Context, all []resource.ResourceFile, fileToUpdate *resource.ResourceFile, data []byte) (*resource.ResourceFile, error) {
 	path := fileToUpdate.Location.Path
 
-	// Compute requestedLangTag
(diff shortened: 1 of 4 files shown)


louischan-oursky:
The fix should be 

> do not let user choose the language before save.

louischan-oursky (commented):
- Remove the prop `supportedLanguages` of ResourcesConfigurationContent in UISettingsScreen.tsx. It is because `supportedLanguages` duplicates with `configForm.state.supportedLanguages`
- Remove the prop `supportedLanguages` of ResourcesConfigurationContent in LocalizationConfigurationScreen.tsx. Same reason.
- Remove `supportedLanguages` from ManageLanguageWidget. Add `supportedLanguagesForDropdown` and `supportedLanguagesForEditing`

IniZio:
The root cause was not actually because the language was selected before save, it was because currently a language with empty resource is initialized on frontend with default language.

Since user selected a new language as default, it will fallback the resource json with itself which is still null, ending in this error.

IniZio:
After discussion, will use en as the last fallback instead of empty string.

louischan-oursky:
1. Implement the changes described in https://github.com/authgear/authgear-server/pull/3881#pullrequestreview-1915907751
2. Cannot remove language.
Screenshot 2024-03-06 at 12 38 16

Issue #3824, opened 2024-02-21
Review restrictions Portal API on modifying html templates or css


fungc-io:
Currently, projects can have their own css and html templates overridden by app specific resources. However, it seems we should have some restrictions on it, or else any project can modify the template / css to remove watermark or even host a unrelated static site.

Maybe we should only allow apps with paid plan to do so?

fungc-io:
- Use allowlist to allow which resources can be override by the users

tung2744:
Tested and it will be blocked by csp.

PR #3887, opened 2024-02-29
Prevent overriding html in app fs except email htmls

tung2744:
ref #3824


pkg/util/template/resource.go
```diff
@@ -25,6 +25,9 @@ type HTML struct {
 	Name string
 	// ComponentDependencies is the HTML component templates this template depends on.
 	ComponentDependencies []*HTML
+	// IsFindAllowedInFs returns an boolean indicating if the Fs can be used to find this resources
+	// For example, if you don't want a html inside the app fs to be used, it should return false.
+	IsFindAllowedInFs FsFilter
```
louischan-oursky, line 30: FindResourcesFsFilter is a better name. What do you think?

tung2744: Renamed to FindResourcesFsFilter

Issue #3916, opened 2024-03-04
Role and group - api integration


Ngkaokis:
- [ ]  graphql query
- [ ]  graphql mutation

PR #3899, opened 2024-03-01
Roles and Groups - Part 2

Ngkaokis:
ref #3830 

require #3839

PR #3920, opened 2024-03-05
Integrate roles and groups api

Ngkaokis:
ref #3916

require #3899


tung2744:
@Ngkaokis Please rebase :adore:

PR #3930, opened 2024-03-06
Setup Authflow test

IniZio:
ref #3773

> Currently still working on doc, will probably create another pr for it instead 🙏


e2e/.env
```diff
@@ -0,0 +1,85 @@
...
+ADMIN_LISTEN_ADDR=0.0.0.0:4002
+
+MAIN_INTERNAL_LISTEN_ADDR=0.0.0.0:14000
+RESOLVER_INTERNAL_LISTEN_ADDR=0.0.0.0:14001
+ADMIN_INTERNAL_LISTEN_ADDR=0.0.0.0:14002
+
+DATABASE_URL=postgres://postgres:postgres@db:5432/postgres?sslmode=disable
+AUDIT_DATABASE_URL=postgres://postgres:postgres@db:5432/postgres?sslmode=disable
+
+REDIS_URL=redis://redis:6379/0
```
louischan-oursky: It seems that this section of the file overrides the variables declared above. To avoid confusion, remove or comment out the overridden variables above.

.github/workflows/ci.yaml
```diff
@@ -13,144 +13,183 @@ on:
...
-    - run: npm run stylelint
-      working-directory: ./authui
-      if: ${{ !cancelled() }}
-    - run: npm run prettier
-      working-directory: ./authui
-      if: ${{ !cancelled() }}
-    - run: npm run build
-      working-directory: ./authui
-      if: ${{ !cancelled() }}
+  # authui-test:
```
louischan-oursky: Do not comment out existing jobs.

cmd/authgear/Dockerfile
```diff
@@ -22,14 +22,14 @@ RUN make build BIN_NAME=authgear TARGET=authgear GIT_HASH=$GIT_HASH
...
-COPY . .
-RUN make authui GIT_HASH=$GIT_HASH
+# ARG GIT_HASH
+# WORKDIR /usr/src/app
+# COPY ./scripts/npm/package.json ./scripts/npm/package-lock.json ./scripts/npm/
+# RUN cd ./scripts/npm && npm ci
+# COPY ./authui/package.json ./authui/package-lock.json ./authui/
+# RUN cd ./authui && npm ci
+# COPY . .
+# RUN make authui GIT_HASH=$GIT_HASH
```
louischan-oursky: Do not comment existing code.

cmd/authgear/cmd/cmdinternal/e2e.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+var cmdInternalE2EImportUser = &cobra.Command{
+	Use:   "e2e import-users",
+	Short: "Import users for e2e tests",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		binder := authgearcmd.GetBinder()
+
+		configSourceDir := binder.GetString(cmd, authgearcmd.ArgConfigSourceDir)
+		jsonPath := configSourceDir + "/users.json"
```
louischan-oursky: Just read from the first argument. This makes the command easier to use. You can use `Args: cobra.ExactArgs(1)` to validate there is only one argument.

pkg/lib/deps/deps_provider.go
```diff
@@ -187,3 +208,12 @@ var BackgroundDependencySet = wire.NewSet(
 	clock.DependencySet,
 	globaldb.DependencySet,
 )
+
+var End2EndDependencySet = wire.NewSet(
+	appRootDeps,
```
louischan-oursky, line 212: Instead of adding e2e stuff here, Make appRootDeps public and define End2EndDependencySet in the e2e package.

pkg/lib/config/identity.go
```diff
@@ -281,7 +281,8 @@ var _ = Schema.Add("OAuthSSOProviderType", `
 		"azureadb2c",
 		"adfs",
 		"apple",
-		"wechat"
+		"wechat",
+		"mock"
```
louischan-oursky, line 285: Do not add `mock` here. The baseline is do not modify existing code to facilitate the e2e test.

pkg/lib/config/config.go
```diff
@@ -360,3 +360,22 @@ func Parse(inputYAML []byte) (*AppConfig, error) {
 func PopulateDefaultValues(config *AppConfig) {
 	SetFieldDefaults(config)
 }
+
+func Export(config *AppConfig) ([]byte, error) {
```
louischan-oursky, line 364: Move this to e2e package.

cmd/portal/cmd/cmdinternal/e2e.go
```diff
@@ -0,0 +1,39 @@
...
+	return log.NewFactory(log.LevelInfo)
+}
+
+var cmdInternalE2E = &cobra.Command{
+	Use:   "e2e",
+	Short: "End2End commands",
+}
+
+var cmdInternalE2ECreateConfigSource = &cobra.Command{
+	Use:   "create-configsource",
```
louischan-oursky, line 20: How is this command different from the existing command we have?

cmd/portal/e2e/configsource.go
```diff
@@ -1,18 +1,22 @@
...
 
 	"io/fs"
 
+	"dario.cat/mergo"
 	"github.com/authgear/authgear-server/cmd/portal/internal"
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/lib/config/configsource"
 	"github.com/authgear/authgear-server/pkg/lib/infra/task"
 	cp "github.com/otiai10/copy"
+	"sigs.k8s.io/yaml"
```
louischan-oursky, line 19: The order of imports is stdlib -> 3rd party -> own module.

e2e/tests/authflow/models.go
```diff
@@ -1,10 +1,21 @@
 package tests
 
 type TestCase struct {
-	Name    string       `yaml:"name"`
-	Project string       `yaml:"project"`
-	Steps   []Step       `yaml:"steps"`
-	Before  []BeforeHook `yaml:"before"`
+	Name               string             `yaml:"name"`
+	Path               string             `yaml:"path"`
+	Focus              bool               `yaml:"focus"`
```
louischan-oursky, line 6: What is Focus?

pkg/lib/authn/otp/form.go
```diff
@@ -52,6 +52,15 @@ func (f Form) GenerateCode(cfg *config.TestModeConfig, featureCfg *config.TestMo
 		}
 		return c.Generate()
 	case secretcode.LinkOTPSecretCodeType:
+		if cfg.Email.Enabled {
+			if r, ok := cfg.Email.MatchTarget(target); ok {
+				fixedCode := r.FixedCode
+				if fixedCode == "" {
+					fixedCode = c.Generate()
+				}
```
louischan-oursky, line 60: We should follow this order

1. If the rule has fixed code, use it
2. Otherwise, if feature config has enabled deterministic code, use it (c.GenerateDeterministic(userID))
3. Otherwise, use random code.

e2e/.env
IniZio: Should remove from run.sh and Makfile

e2e/fixtures/signup/authgear.yaml
```diff
@@ -0,0 +1,428 @@
...
+  forgot_password:
+    email:
+    - channel: email
+      otp_form: code
+    phone:
+    - channel: sms
+      otp_form: code
+  implementation: authflowv2
+  phone_input:
+    allowlist:
```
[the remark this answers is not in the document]
IniZio: Latest code uses e2e's base `authgear.yaml`, shortened the base config instead.

e2e/tests/authflow/models.go
IniZio: It's created like https://pkg.go.dev/github.com/smartystreets/goconvey/convey#FocusConvey focus on one test case

pkg/lib/deps/deps_provider.go
tung2744: I thought we are calling the api through http so we don't need additional dependencies? e.g. IP should be resolved based on the request.

IniZio: Yes the test itself part is simple http client 

This dependency set is for setting up project and user fixtures per test case

e2e/tests/authflow/testcase.go
```diff
@@ -0,0 +1,168 @@
...
+				t.Errorf("failed to execute custom SQL: %v", err)
+				return
+			}
+		default:
+			t.Errorf("unknown before hook type: %s", beforeHook.Type)
+		}
+	}
+
+	hostName := httputil.HTTPHost(fmt.Sprintf("%s.portal.localhost:4000", appID))
+	client := authflowclient.NewClient(ctx, "localhost:4000", hostName)
```
tung2744, line 93: In my opinion, e2e tests should not use authflowclient. It should create a new client which call the authgear server like calling it from another project.

As authflowclient is not used except in e2e, I suggest rename it to e2e client and move to e2e package.

e2e/tests/authflow/models.go
louischan-oursky: Please leave a comment above Focus for future readers.

pkg/lib/authflowclient/data.go
```diff
@@ -0,0 +1,215 @@
...
+	return m, nil
+}
+
+func CastVerifyClaimOrCreateAuthenticator(rawData json.RawMessage) (*DataVerifyClaim, *DataCreateAuthenticator, error) {
+	dataMap, err := asMap(rawData)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	_, okOTPForm := dataMap["otp_form"]
```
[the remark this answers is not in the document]
IniZio: As per suggestion https://github.com/authgear/authgear-server/pull/3930#discussion_r1547705369, updated the client to be specific to e2e

e2e/tests/authflow/client/client.go
```diff
@@ -1,4 +1,4 @@
-package authflowclient
+package e2eclient
```
louischan-oursky, line 1: Rename the directory to match the package name.

e2e/migrations
```diff
@@ -0,0 +1 @@
+../migrations/
```
louischan-oursky, line 1: This symlink is unnecessary. We can change the cwd directory before we execute `authgear` or `authgear-portal`.

e2e/resources
```diff
@@ -0,0 +1 @@
+../resources/
```
louischan-oursky, line 1: Is this necessary?

pkg/lib/deps/deps_provider.go
louischan-oursky: So can we remove these changes?

pkg/lib/authn/otp/form.go
```diff
@@ -35,30 +35,50 @@ func (f Form) GenerateCode(cfg *config.TestModeConfig, featureCfg *config.TestMo
...
-					fixedOTP = c.Generate()
-				}
-				return c.GenerateFixed(fixedOTP)
+		return f.generateOOBOTPCode(c, cfg, featureCfg, target)
+	case secretcode.LinkOTPSecretCodeType:
+		return f.generateLinkOTPCode(c, cfg, featureCfg, target, userID)
+	default:
+		panic("unknown otp form")
+	}
+}
```
louischan-oursky, line 44: Please add tests for GenerateCode to ensure we do not have regression.

By looking at the code, I see an interesting behavior.

1. Test mode is enabled
2. A rule is matched.
3. The rule does not specify fixed code
4. Feature config test mode is disabled.
5. But Feature config test mode has fixed code defined.
6. The fixed code is taken from the feature config (but it is disabled).

e2e/tests/authflow/authflow_test.go
```diff
@@ -0,0 +1,39 @@
+package tests
```
louischan-oursky, line 1: I propose the following directory structure

```
/e2e
  /tests
    /issue_3583
    /issue_3585
    ...
  /cmd
    main.go
    ...
  /pkg
    /e2eclient
      client.go
      ...
    /testrunner
      matcher.go
      testcase.go
      ...
```

e2e/migrations
IniZio: Currently `authgear` and `authgear-portal` rely uses env file from cwd though. Ok to update the their cmd logic for this case?

louischan-oursky (commented):
- Keep the feature config disabled not respect fix.
- Revert all other changes in pkg
- Place a default authtgear.features.yaml in e2e test, like we have a default authgear.yaml
- Make the input a Go template. The data should be `{{ linkOTPCode "email" "user@example.com" }}`

IniZio:
Updated to use deterministic link OTP

louischan-oursky (approved):

Issue #3902, opened 2024-03-04
Role - Create Role Form


Ngkaokis:
![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/d8bd91b3-9b1c-45c6-9105-7e96b0d5ff61/80582493-1dbc-4883-8406-2d235ec2ad9e?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi9kOGJkOTFiMy05YjFjLTQ1YzYtOTEwNS03ZTk2YjBkNWZmNjEvODA1ODI0OTMtMWRiYy00ODgzLTg0MDYtMmQyMzVlYzJhZDllIiwiaWF0IjoxNzA5NTM1NjE4LCJleHAiOjMzMjgwMDk1NjE4fQ.KdnowEcP39D5TEmaFC8GYYp0ZaqoKBTvlYUOiLx7Eng "=1253x")

tung2744:
@fung Sorry, I forgot what is our decision about the generated value on the role name field? Do we still generate it from the key?

fungc-io:
No, only in Group. No need to setup key-name relation in Roles

PR #3936, opened 2024-03-06
Portal - Implement create roles

tung2744:
ref #3902

Screenshot 2024-03-06 at 8 14 13 PM

### DIFF
--- a/portal/src/AppRoot.tsx
+++ b/portal/src/AppRoot.tsx
@@ -225,7 +225,7 @@ const AppRoot: React.VFC = function AppRoot() {
                   }
                 />
                 <Route
-                  path="add-roles"
+                  path="add-role"
                   element={
                     <Suspense fallback={<ShowLoading />}>
                       <AddRolesScreen />
--- a/portal/src/AppRoot.tsx
+++ b/portal/src/AppRoot.tsx
@@ -10,6 +10,9 @@ import CookieLifetimeConfigurationScreen from "./graphql/portal/CookieLifetimeCo
 import { AppContext } from "./context/AppContext";
 
 const RolesScreen = lazy(async () => import("./graphql/adminapi/RolesScreen"));
+const AddRolesScreen = lazy(
+  async () => import("./graphql/adminapi/AddRolesScreen")
+);
 const GroupsScreen = lazy(
   async () => import("./graphql/adminapi/GroupsScreen")
 );
@@ -218,6 +221,14 @@ const AppRoot: React.VFC = function AppRoot() {
                   element={
                     <Suspense fallback={<ShowLoading />}>
                       <RolesScreen />
+                    </Suspense>
+                  }
+                />
+                <Route
+                  path="add-roles"
+                  element={
+                    <Suspense fallback={<ShowLoading />}>
+                      <AddRolesScreen />
                     </Suspense>
                   }
                 />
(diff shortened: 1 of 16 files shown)


portal/src/AppRoot.tsx
```diff
@@ -221,6 +224,14 @@ const AppRoot: React.VFC = function AppRoot() {
                     </Suspense>
                   }
                 />
+                <Route
+                  path="add-roles"
```
Ngkaokis, line 228: should be `add-role`. 
same as `add-user`

portal/src/graphql/adminapi/AddRolesScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+            parentJSONPointer=""
+            type="text"
+            label={renderToString("AddRolesScreen.roleName.title")}
+            value={formState.roleName}
+            onChange={onFormStateChangeCallbacks.roleName}
+          />
+          <WidgetDescription className="mt-2">
+            <FormattedMessage id="AddRolesScreen.roleName.description" />
+          </WidgetDescription>
+        </div>
```
Ngkaokis, line 147: should be `name` go first

https://linear.app/authgear/issue/DEV-404/roles-and-groups-configuration-in-portal#comment-e5e223c7

portal/src/FormContainer.tsx
```diff
@@ -2,126 +2,74 @@ import React, { useCallback, useContext, useMemo, useState } from "react";
...
     // If the form contains a CodeEditor, dialog dismiss animation does not play.
     // Defer the dismissal to ensure dismiss animation.
     setTimeout(() => setIsResetDialogVisible(false), 0);
-  }, [reset]);
+  }, [onReset]);
 
-  const allowSave = formCanSave !== undefined ? formCanSave : isDirty;
-  const disabled = isUpdating || !allowSave;
+  // const allowSave = formCanSave !== undefined ? formCanSave : isDirty;
+  // const disabled = isUpdating || !allowSave;
```
Ngkaokis, line 62: Remove the comment

Ngkaokis (commented):
Seems have bug

And I think although we don't want the command bar now, we still want to add back the top margin to keep the spacing, as the layout would shift when navigate to other page

https://github.com/authgear/authgear-server/assets/63200569/d03ffb4a-7d84-40e6-bccb-dd6ba5f70dc0

tung2744:
> And I think although we don't want the command bar now, we still want to add back the top margin to keep the spacing

I think that will be too much spacing. @fungc-io What do you think?

tung2744:
The bug is fixed in e68d276. Thanks!

portal/src/locale-data/en.json
```diff
@@ -122,6 +122,7 @@
   "errors.bad-nft-collection": "This NFT collection is either not supported or is from another network.",
   "errors.alchemy-protocol": "We have some problem with this NFT collection. Please contact us at {ExternalLink, react, children{hello@authgear.com} href{mailto:hello@authgear.com}}.",
   "errors.deno-hook.typecheck": "{pre, react, children{{message}}}",
+  "errors.roles.key.validation.format": "Invalid role key. Only english alphbets, digits, colons (:) and underscores (_) are accepted.",
```
Ngkaokis, line 125: The error message seems not correct when I enter key `123`

tung2744: @Ngkaokis This is actually because the regex in server side doesn't allow digits as the initial character. 
It seems differ from [the spec](https://github.com/authgear/authgear-server/pull/3556/files#diff-dff982576a46d7877aac0dfe34a729c8bee1d062c2e992a59093600d7502e922R22).
@louischan-oursky Should we allow digit as the first character?

louischan-oursky: Use the implementation as the reference. Please update the spec and the error message to match the implementation.

PR #3937, opened 2024-03-07
[TBC] Combine uppercase and lowercase password policy

IniZio:


louischan-oursky:
Why is this needed?

IniZio:
It's still being confirmed in issue, let me convert back to draft first to avoid confusion.

IniZio:
Will keep separated.

Issue #3903, opened 2024-03-04
Role - List View


Ngkaokis:
<img src="https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/cbd185f4-7111-4933-abb3-9c4ac6c6bbc8/75e2ccd4-e7b9-4c21-ab94-e015fb5506c4?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi9jYmQxODVmNC03MTExLTQ5MzMtYWJiMy05YzRhYzZjNmJiYzgvNzVlMmNjZDQtZTdiOS00YzIxLWFiOTQtZTAxNWZiNTUwNmM0IiwiaWF0IjoxNzA5NTM0NzEyLCJleHAiOjMzMjgwMDk0NzEyfQ.e85On2pSYyZey2sp1qRy91K5NmD76XVJFneWW_hW1tQ " alt="image.png" width="892" height="526" />

PR #3938, opened 2024-03-07
Add RolesScreen list view

Ngkaokis:
ref #3903 

Issue: Tried to adjust `flexGrow` of other columns but it's still always only the first column's width is adjusted

image


Ngkaokis:
> What about setting maxWidth?

That's work if we don't want to adjust columns base on proportion. Let me confirm with designer

pkg/admin/facade/rolesgroups.go
```diff
@@ -80,9 +82,13 @@ func (f *RolesGroupsFacade) ListRoles(options *rolesgroups.ListRolesOptions, pag
 		return nil, nil, err
 	}
 
+	count, err := f.RolesGroupsQueries.CountRoles()
+	if err != nil {
+		return nil, nil, err
+	}
+
```
tung2744, line 89: ListRoles supports `SearchKeyword`, so CountRoles should also support it?

pkg/admin/facade/rolesgroups.go
```diff
@@ -119,9 +125,13 @@ func (f *RolesGroupsFacade) ListGroups(options *rolesgroups.ListGroupsOptions, p
 		return nil, nil, err
 	}
 
+	count, err := f.RolesGroupsQueries.CountGroups()
```
tung2744, line 128: Same as roles, it should support SearchKeyword?

portal/src/graphql/adminapi/RolesScreen.module.css
```diff
@@ -1,12 +1,19 @@
 .content {
-  @apply mt-11;
+  height: 100%;
```
tung2744, line 2: This element's parent is a flex container, therefore use flex's method to layout. Use `flex: 1 1 0px` here in desktop, and `flex: 1 0 auto` in mobile.

portal/src/graphql/adminapi/RolesScreen.tsx
```diff
@@ -64,36 +105,92 @@ const RolesScreen: React.VFC = function RolesScreen() {
...
-        <NavBreadcrumb className="block" items={items} />
-      </div>
-      {isEmpty ? (
-        <RolesScreenEmptyState className={styles.widget} />
-      ) : (
-        // TODO: implement list view
-        <RolesScreenEmptyState className={styles.widget} />
-      )}
-    </ScreenContent>
+    <LocalSearchBoxContext.Provider value={localSearchBoxProps}>
```
tung2744, line 163: What is the reason we need this provider, instead of passing the props directly?

tung2744: And also the word `Local` make me think it is searching in local, but it seems not. The search is performed in server side.

tung2744:
Screenshot 2024-03-08 at 12 02 54 PM

It seems the table is too long in narrow screens.

portal/src/graphql/adminapi/RolesScreen.tsx
```diff
@@ -64,36 +105,92 @@ const RolesScreen: React.VFC = function RolesScreen() {
...
+      <ScreenContent className={styles.content} layout="list">
+        <div className={styles.widget}>
+          <div className={styles.titleContainer}>
+            <NavBreadcrumb className="block" items={items} />
+            {!isEmpty ? <CreateRoleButton /> : null}
+          </div>
+          <LocalSearchBox />
```
tung2744, line 170: According to figma, search box shouldn't exist when empty state is showing.

portal/src/graphql/adminapi/RolesScreen.tsx
```diff
@@ -64,36 +105,92 @@ const RolesScreen: React.VFC = function RolesScreen() {
...
+        </div>
+        {isEmpty ? (
+          isSearch ? (
+            <MessageBar className={cn(styles.widget, styles.message)}>
+              <FormattedMessage id="RolesScreen.empty.search" />
+            </MessageBar>
```
tung2744, line 176: Where did you find this design? IMO it should be displayed in the detail list.

pkg/admin/facade/rolesgroups.go
Ngkaokis: It's mainly for counting the total number. Do you think it's better to add it?

portal/src/graphql/adminapi/RolesScreen.tsx
Ngkaokis: I copied it from `UsersScreen` :sosad:

Let me fix it together

portal/src/graphql/adminapi/RolesScreen.tsx
Ngkaokis: Thx. Missed the below design so followed the old design of users list

Ngkaokis:
> It seems the table is too long in narrow screens.

Let me just randomly adjust the min/max width of the columns to make it look fine now may need designer to make a version on mobile

cc @fungc-io 

> There shouldn't be horizontal scrollbar on the outer window. I suggest you to set "Show scroll bars" to "Always" in your system preference, so you will be easier to spot this kind of problem.

Not sure if I set it correct, but I can't see the scroll bar you see

Screenshot 2024-03-11 at 12 50 00 PM

portal/src/graphql/adminapi/RolesScreen.tsx
Ngkaokis: Not so sure how to handle the empty state in mobile view? :sosad:

Expect it to be static or scroll with columns?

https://github.com/authgear/authgear-server/assets/63200569/f2370bc9-ab27-4ca8-967e-01eed44a51c1

tung2744:
> Not sure if I set it correct, but I can't see the scroll bar you see

@Ngkaokis 

I tried in your latest commit, it seems fixed.

portal/src/graphql/adminapi/RolesScreen.tsx
tung2744: If scroll with columns, there is a chance that user cannot see it.
But static looks a bit weird too. Maybe @fungc-io What do you think?

pkg/admin/facade/rolesgroups.go
tung2744: @Ngkaokis I think pagination will fail if the total count doesn't respect the search keyword?

Ngkaokis: If we stick we the behaviour as users list, searching should not show pagination and would query all results (although we limit the search result to 10 in users list for performance issue. But I think we don't have issue with roles and groups query)

tung2744: Oh I see. Lets keep it return null if search keyword exist first to be consistent with user query. And lets confirm it with @fungc-io later.

tung2744:
@Ngkaokis Sorry the scrollbar problem still exist, please try a narrow screen size (but not mobile), and add more rows.
Screenshot 2024-03-11 at 2 01 10 PM

portal/src/graphql/adminapi/RolesScreen.module.css
```diff
@@ -1,12 +1,19 @@
 .content {
-  @apply mt-11;
+  flex: 1 1 0;
```
tung2744, line 2: To fix the scrollbar problem, it seems you need overflow: hidden; here

portal/src/graphql/adminapi/RolesScreen.tsx
```diff
@@ -64,34 +86,83 @@ const RolesScreen: React.VFC = function RolesScreen() {
...
+        </div>
+        <SearchBox {...searchBoxProps} />
       </div>
       {isEmpty ? (
         <RolesScreenEmptyState className={styles.widget} />
       ) : (
-        // TODO: implement list view
-        <RolesScreenEmptyState className={styles.widget} />
+        <RolesList
+          className={styles.widget}
```
tung2744, line 157: To fix the scrollbar problem, it seems you need min-height: 0 here

Ngkaokis: overflow: auto seems more intuitive to me

portal/src/graphql/adminapi/RolesScreen.module.css
Ngkaokis: I think should just add `overflow: auto;` here

tung2744: No, because the scrollbar should appear in content of CommandBarContainer.

fungc-io:
@Ngkaokis @tung2744 
When no result,
Hiding the columns and show a message bar like this in the "User" list seems looks better imo
image

tung2744:
@fungc-io Then other lists should follow this design too?
Such as https://www.figma.com/file/bhUac10bqlsRVXF5HiwCjW/Q1-%2F-2024?type=design&node-id=42-5664&mode=design&t=iFDAUZQOpvTa85MR-4

fungc-io:
@tung2744 
yes, like the audit log,
image

it looks better on all screen sizes
Thanks for pointing out the design problem 🙇

PR #3943, opened 2024-03-07
Use origin instead of base url

louischan-oursky:


pkg/lib/endpoints/endpoints.go
```diff
@@ -35,13 +34,8 @@ func (e *Endpoints) BaseURL() *url.URL {
...
 	// Because String() will add leading / to make the URL legal.
-	base.Path = "/"
-	return base
-}
-
-func (e *Endpoints) urlOf(relPath string) *url.URL {
-	u := e.BaseURL()
-	u.Path = path.Join(u.Path, relPath)
+	u := e.Origin()
+	u.Path = path.Join("/", relPath)
```
tung2744, line 38: I think `relPath` means relative path, but now it becomes absolute path.

louischan-oursky: Yes. All existing usages actually result in absolute path.

louischan-oursky: And this is the method for composing rel path. https://pkg.go.dev/net/url#URL.JoinPath (Pretty new, it was introduced in 1.19)

Issue #3939, opened 2024-03-07
OAuth callback does not handle AuthflowV2


IniZio:
For example with following authgear config

```
authentication_flow:
  login_flows:
  - name: default
    steps:
    - one_of:
      - identification: oauth
        steps:
        - one_of:
          - authentication: primary_oob_otp_sms
          - authentication: primary_oob_otp_email
          type: authenticate
    type: identify

ui:
  implementation: authflowv2
```

After OAuth, user gets redirected to authflowv1 instead of authflowv2

It should be due to `pkg/auth/handler/webapp/sso_callback.go` not handling authflowv2 switch

tung2744:
The problem is AuthflowController of SSOCallbackHandler is always using navigator of v1.

PR #3948, opened 2024-03-08
Handle ui implementation in sso callback

IniZio:
ref #3939

Wanted to make a switcher like implementation_switcher middleware, but seems weird to make navigator aware of current `ui.implementation`


louischan-oursky:
superseded by https://github.com/authgear/authgear-server/pull/3950

PR #3950, opened 2024-03-11
Fix OAuth callback does not handle authflowv2

louischan-oursky:
ref #3939

### DIFF
--- a/pkg/auth/handler/webapp/authflow_controller.go
+++ b/pkg/auth/handler/webapp/authflow_controller.go
@@ -194,7 +194,7 @@ func (c *AuthflowController) HandleOAuthCallback(w http.ResponseWriter, r *http.
 		return
 	}
 
-	s, err := c.getWebSession(r)
+	s, err := c.Sessions.Get(state.WebSessionID)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


louischan-oursky:
Comparing to https://github.com/authgear/authgear-server/pull/3948, the difference are

- It seems that the browsers no longer include web_session cookies in the request to /sso/callback. Therefore, we cannot retrieve the web session for further processing. But we already include web_session_id in state. So we just use it to retrieve the web session.
- Fixed an issue that implementation switcher middleware is not mounted at the root of the webapp chain. This causes /sso/callback to always see authflowv1, even authflowv2 is in effect.
- Actually applied implementation switcher to /sso/callback.

pkg/auth/handler/webapp/authflow_controller.go
```diff
@@ -194,7 +194,7 @@ func (c *AuthflowController) HandleOAuthCallback(w http.ResponseWriter, r *http.
 		return
 	}
 
-	s, err := c.getWebSession(r)
+	s, err := c.Sessions.Get(state.WebSessionID)
```
tung2744, line 197: The session should be in cookie, is there any reason we need to get it from the state?
I remembered that WebSessionID is for wechat login.

Issue #3931, opened 2024-03-06
Log and show unknown errors


IniZio:
- [X] `~PublicSignup~`\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\~ error is rendered as \\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\~`~error-unknown~`\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\~, should have it's dedicated error message instead\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\~

* It's due to a bug in oauth flow, should be fixed in [https://github.com/authgear/authgear-server/pull/3948](https://github.com/authgear/authgear-server/pull/3948)

- [X] Currently `error-unknown` will only show a static "contact us" text, can log the `.Error` in a hidden span for easier debugging
- [X] Add a template function to log the `error-unknown`, e.g. `{{ call $.LogUnknownError .Error }}`

IniZio:
@fung I will check this as well after [DEV-1099](https://linear.app/authgear/issue/DEV-1099/disable-public-signup-blocks-oauth) since the error message for signup blocked is confusing.

PR #3951, opened 2024-03-11
Log unknown viewmodel error

IniZio:
ref #3931

### DIFF
--- a/pkg/auth/handler/webapp/viewmodels/base.go
+++ b/pkg/auth/handler/webapp/viewmodels/base.go
@@ -19,9 +19,16 @@ import (
 	"github.com/authgear/authgear-server/pkg/util/geoip"
 	"github.com/authgear/authgear-server/pkg/util/httputil"
 	"github.com/authgear/authgear-server/pkg/util/intl"
+	"github.com/authgear/authgear-server/pkg/util/log"
 	"github.com/authgear/authgear-server/pkg/util/template"
 	"github.com/authgear/authgear-server/pkg/util/wechat"
 )
+
+type BaseLogger struct{ *log.Logger }
+
+func NewBaseLogger(lf *log.Factory) BaseLogger {
+	return BaseLogger{lf.New("webapp")}
+}
 
 type TranslationService interface {
 	HasKey(key string) (bool, error)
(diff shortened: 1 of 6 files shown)


resources/authgear/templates/en/web/__error.html
```diff
@@ -241,7 +241,12 @@
     {{ else if eq .Error.reason "InvalidWhatsappUser" }}
     <li>{{ template "error-invalid-whatsapp-user" }}</li>
     {{ else }}
-    <li>{{ template "error-unknown" }}</li>
+    <li>
+      {{ template "error-unknown" }}
+      <span id="error-unknown" class="hidden">
+        {{ call $.LogUnknownError .Error }}
+      </span>
```
louischan-oursky, line 248: No need to use a span to wrap the template call. `LogUnknownError` always return empty string. So write this instead `{{ template "error-unknown" }}{{ call $.LogUnknownError .Error }}`

pkg/auth/handler/webapp/viewmodels/base.go
```diff
@@ -257,6 +267,23 @@ func (m *BaseViewModeler) ViewModel(r *http.Request, rw http.ResponseWriter) Bas
...
+			if err != nil {
+				var sessionID string
+				if s := webapp.GetSession(r.Context()); s != nil {
+					sessionID = s.ID
+				}
+
+				m.Logger.WithFields(map[string]interface{}{
+					"session_id":  sessionID,
+					"request_uri": r.URL.RequestURI(),
+				}).Errorf("unknown error: %v", err)
```
louischan-oursky, line 280: `err` here is the JSON serialization of APIError. So we should log all fields of it. Remove `session_id` and `request_uri`.

```
m.Logger.WithFields(err).Errorf("unknown error: %v", err)
```

pkg/auth/handler/webapp/viewmodels/base.go
```diff
@@ -257,6 +267,23 @@ func (m *BaseViewModeler) ViewModel(r *http.Request, rw http.ResponseWriter) Bas
...
+
+				return fmt.Sprintf("%v", err)
```
louischan-oursky, line 282: Always return empty string.

louischan-oursky:
Build failed.

PR #3953, opened 2024-03-11
Add id token and access token tests

IniZio:
ref #3944


pkg/lib/oauth/oidc/id_token_test.go
```diff
@@ -48,4 +59,104 @@ func TestSID(t *testing.T) {
...
+
+	Convey("IssueIDToken and VerifyIDTokenHint", t, func() {
+		ctrl := gomock.NewController(t)
+
+		now := time.Date(2020, 1, 1, 0, 0, 0, 0, time.UTC)
+
+		secrets := &config.OAuthKeyMaterials{
+			Set: jwk.NewSet(),
+		}
+		jwkKey := utilsecrets.GenerateRSAKey(now, rand.SecureRand)
```
louischan-oursky, line 71: Instead of generating a 2048-bit RSA key pair per test, generate it once and paste it here in PEM format. You can find a example in jwkutil/pem_test.go You can just copy the `PrivateKeyPEM` in jwkutil/pem_test.go and use it here.

pkg/lib/oauth/token_encoding_test.go
```diff
@@ -0,0 +1,86 @@
...
+func TestAccessToken(t *testing.T) {
+	Convey("EncodeAccessToken and DecodeAccessToken", t, func() {
+		ctrl := gomock.NewController(t)
+
+		now := time.Date(2020, 1, 1, 0, 0, 0, 0, time.UTC)
+
+		secrets := &config.OAuthKeyMaterials{
+			Set: jwk.NewSet(),
+		}
+		jwkKey := utilsecrets.GenerateRSAKey(now, rand.SecureRand)
```
louischan-oursky, line 28: Same here.

Issue #3905, opened 2024-03-04
Role Detail - Groups Tab


Ngkaokis:
- [ ] empty state
- [ ] groups list view
- [ ] search bar
- [ ] add to group dialog
  - [ ] combobox
- [ ] remove group dialog

PR #3954, opened 2024-03-12
Implement groups in role

tung2744:
ref #3905

requires https://github.com/authgear/authgear-server/pull/3946

- Add group to role will be implemented in another pr


portal/src/components/roles-and-groups/RolesAndGroupsEmptyView.tsx
```diff
@@ -0,0 +1,48 @@
...
+  text: React.ReactNode;
+}> = function CreateButton(props) {
+  const { className, href, text } = props;
+  return (
+    <PrimaryButton
+      href={href}
+      className={className}
+      text={text}
+      iconProps={{ iconName: "Add" }}
+    />
```
Ngkaokis: I thought we don't want to use `href` to avoid page from full reload?

portal/src/components/roles-and-groups/GroupsEmptyView.tsx
```diff
@@ -0,0 +1,29 @@
...
+        className={className}
+        icon={<img src={iconGroups} />}
+        title={<FormattedMessage id="GroupsEmptyView.title" />}
+        description={<FormattedMessage id="GroupsEmptyView.description" />}
+        button={
+          <ReactRouterLink.CustomComponent
+            component={RolesAndGroupsEmptyView.CreateButton}
+            to={`/project/${appID}/user-management/groups/add-group`}
+            text={<FormattedMessage id="GroupsEmptyView.button.text" />}
+          />
```
Ngkaokis, line 25: Not sure if we want to do it this way, seems introducing a redundant layer to me. 
At least I think a common usage would be wrapping `ReactRouterLink.CustomComponent`  inside `ReactRouterLink`?

portal/src/components/roles-and-groups/RolesAndGroupsEmptyView.tsx
```diff
@@ -1,24 +1,26 @@
...
+  onClick?: MouseEventHandler<HTMLButtonElement | HTMLAnchorElement>;
   text: React.ReactNode;
-}> = function CreateButton(props) {
-  const { className, href, text } = props;
+}) {
+  const { className, href, onClick, text } = props;
   return (
     <PrimaryButton
       href={href}
+      onClick={onClick}
```
Ngkaokis: How fluent ui handle it when both href and onClick present?

portal/src/model/group.ts
```diff
@@ -0,0 +1,39 @@
...
+      if (groupKey.includes(keyword)) {
+        return true;
+      }
+      if (groupName?.includes(keyword)) {
+        return true;
+      }
+    }
+    return false;
+  });
+}
```
Ngkaokis, line 39: We do client side filtering here?

portal/src/components/roles-and-groups/RolesAndGroupsEmptyView.tsx
tung2744: We can have href, but we need to handle the onClick event and preventDefault to prevent the page reload. You can see implementation of `ReactRouterLink.CustomComponent` to know about how I did it.

portal/src/components/roles-and-groups/GroupsEmptyView.tsx
tung2744: > At least I think a common usage would be wrapping ReactRouterLink.CustomComponent inside ReactRouterLink?

Why we will use it in this way? When using `ReactRouterLink.CustomComponent`, we don't need `ReactRouterLink`.

In fact, I introduced `ReactRouterLink.CustomComponent` because the `component` prop of `ReactRouterLink` doesn't have a correct typing. It assumed you must pass in a component which have same props of a `<a>`, which make it difficult to use.

So I separated `ReactRouterLink.CustomComponent` from `ReactRouterLink`. `ReactRouterLink` now only renders to `<a>`. And `ReactRouterLink.CustomComponent` only renders to your provided `component`, and it accepts all props your component accepts.

portal/src/components/roles-and-groups/RolesAndGroupsEmptyView.tsx
tung2744: Actually, a `<a>` element have click event too. So it is same as a `<a>` having href and onClick handler. If you preventDefault() in the onClick handler, the navigation will be stopped.

portal/src/model/group.ts
tung2744: Because I doesn't find a api which search groups inside roles. We can confirm in standup.

portal/src/components/roles-and-groups/GroupsEmptyView.tsx
Ngkaokis: If that so, sounds like `ReactRouterLink` is not needed anymore as well.

> When using ReactRouterLink.CustomComponent, we don't need ReactRouterLink

What I mean is I think there is a common convention that we wrap component `X.Y` inside `X`? 
For example
```
<Button>
  <Button.Label>
  </Button.Label>
</Button>
```

tung2744: The problem is it will be difficult to make a type safe interface for `as="a"` or `as="button"`. So I didn't support it.

tung2744:
@Ngkaokis Updated, thanks!

PR #3963, opened 2024-03-12
Implement add role to group

tung2744:
ref #3905 

https://github.com/authgear/authgear-server/assets/34942210/197e666b-e1df-44ce-915d-bf2ba157279a

### DIFF
--- a/portal/src/components/roles-and-groups/AddRoleGroupsDialog.tsx
+++ b/portal/src/components/roles-and-groups/AddRoleGroupsDialog.tsx
@@ -10,7 +10,13 @@ import {
   FormattedMessage,
   Context as MessageContext,
 } from "@oursky/react-messageformat";
-import React, { useCallback, useContext, useMemo, useState } from "react";
+import React, {
+  useCallback,
+  useContext,
+  useMemo,
+  useState,
+  useEffect,
+} from "react";
 import PrimaryButton from "../../PrimaryButton";
 import DefaultButton from "../../DefaultButton";
 import ErrorDialog from "../../error/ErrorDialog";
@@ -65,6 +71,9 @@ export const AddRoleGroupsDialog: React.VFC<AddRoleGroupsDialogProps> =
     roleGroups,
   }) {
     const { renderToString } = useContext(MessageContext);
+    const existingGroupIDs = useMemo(() => {
+      return new Set(roleGroups.map((group) => group.id));
+    }, [roleGroups]);
 
     const [groupTags, setGroupTags] = useState<GroupTag[]>([]);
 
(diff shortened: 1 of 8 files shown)


portal/src/components/roles-and-groups/AddRoleGroupsDialog.tsx
```diff
@@ -0,0 +1,232 @@
...
+      refetchRole,
+      roleID,
+      onDismiss,
+    ]);
+
+    useEffect(() => {
+      // Reset states on isHidden changes
+      setGroupTags([]);
+    }, [isHidden]);
+
```
Ngkaokis, line 175: I think it's more intuitive to just put this in onDismiss?

Ngkaokis (commented):
Wrong style

image

tung2744:
> Wrong style, should be same width as input

I think we should follow fluent ui's behavior. Or let me see see if there is an option to match the width with input.

portal/src/components/roles-and-groups/AddRoleGroupsDialog.tsx
tung2744: onDimiss doesn't really mean it is dismissed, it is actually a "request of dismiss".
The flag that controls it is `isHidden`.

Ngkaokis:
> I think we should follow fluent ui's behavior. Or let me see see if there is an option to match the width with input.

Just set width: 100% on the container?

portal/src/components/roles-and-groups/AddRoleGroupsDialog.tsx
Ngkaokis: How about `onDimissed`? Instead of resetting tags when the flag is toggle (and also adding a extra useEffect), resetting the tags when the dialog is dimissed seems eaiser to understand.

tung2744:
If you inspect the callout, you will find that it is actually not a children of the input.
"w-full" result in this:
Screenshot 2024-03-13 at 1 41 12 PM

Ngkaokis:
> If you inspect the callout, you will find that it is actually not a children of the tagPickerContainer.
"w-full" result in this:

Then something like the following?

```
        pickerCalloutProps={{
          className: "w-full",
          target: tagPickerRef,
          calloutMaxWidth: tagPickerRef.current?.offsetWidth,
        }}

```

tung2744:
> tagPickerRef.current?.offsetWidth

This only captures the current width at the moment the component renders. It could fail when you resize the window.

I've tried to make another fix in [94d7aac](https://github.com/authgear/authgear-server/pull/3963/commits/94d7aac6ed97bb300d2e70717b0c89adf3534c90)

Also updated to use onDismissed to reset the states.

Ngkaokis:
image

Got covered by the overlay :sosad:

tung2744:
I reverted the commit first. I think we should simply follow the default behavior of fluent ui if it takes too much effort to customize.

Ngkaokis:
Expose maxWidth props and set it in dialog component? We hardcode the max width of dialog anyway

Ngkaokis:
> I reverted the commit first. I think we should simply follow the default behavior of fluent ui if it takes too much effort to customize.

Ok, I merge first after ci and may be open another issue for it

tung2744:
Actually, using `max-width` to set the callout width will fail if the dialog is rendered as smaller than the max width.
e.g. max width is 510px, but the dialog rendered as 500px due to screen size. But the callout will remain 510px (Because it's width is 100% screen size, but dialog has some margins so it will not occupy 100% of the screen width).

What we actually want is [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver), which I don't want to introduce it due to a minor styling issue.

Issue #3957, opened 2024-03-12
Add and parse password expiry config


IniZio:
```yaml
password:
  # The 3 cases can be turned on individually.
  # However, the precedence is deny_login > force_change > prompt_change.
  expiry:
    # In this case, the authflow will result in a dead end, with an error telling that the password is expired and the login is denied.
    deny_login:
      enabled: true
      since_last_update_in_days: 90
    # In this case, the authflow will enter the change_password step (The existing one). To proceed, the end-user must change the password.
    force_change:
      enabled: true
      since_last_update_in_days: 60
    # In this case, the authflow will enter the change_password step. But this step now supports taking an input to skip the password update.
    prompt_change:
      enabled: true
      since_last_update_in_days: 30
```

> Will only have `renew` for now, `warning` and `reject` will be excluded

PR #3964, opened 2024-03-12
Add config schema for password expiry

IniZio:
ref #3957

### DIFF
--- a/pkg/lib/authn/authenticator/password/deps.go
+++ b/pkg/lib/authn/authenticator/password/deps.go
@@ -28,7 +28,7 @@ func ProvideChecker(
 	}
 	if !*featureCfg.Password.Expiry.ForceChange.Disabled {
 		checker.PwExpiryForceChangeEnabled = cfg.Expiry.ForceChange.IsEnabled()
-		checker.PwExpiryForceChangeSinceLastUpdateInDays = cfg.Expiry.ForceChange.SinceLastUpdateInDays
+		checker.PwExpiryForceChangeSinceLastUpdate = cfg.Expiry.ForceChange.SinceLastUpdate
 	}
 	if !*featureCfg.Password.Policy.MinimumGuessableLevel.Disabled {
 		checker.PwMinGuessableLevel = cfg.Policy.MinimumGuessableLevel
--- a/pkg/lib/authn/authenticator/password/deps.go
+++ b/pkg/lib/authn/authenticator/password/deps.go
@@ -27,7 +27,7 @@ func ProvideChecker(
 		checker.PasswordHistoryEnabled = cfg.Policy.IsEnabled()
 	}
 	checker.PwExpiryForceChangeEnabled = cfg.Expiry.ForceChange.IsEnabled()
-	checker.PwExpiryForceChangeSinceLastUpdate = cfg.Expiry.ForceChange.SinceLastUpdate
+	checker.PwExpiryForceChangeSinceLastUpdate = cfg.Expiry.ForceChange.DurationSinceLastUpdate
 	if !*featureCfg.Password.Policy.MinimumGuessableLevel.Disabled {
 		checker.PwMinGuessableLevel = cfg.Policy.MinimumGuessableLevel
 	}
--- a/pkg/lib/authn/authenticator/password/deps.go
+++ b/pkg/lib/authn/authenticator/password/deps.go
@@ -26,10 +26,8 @@ func ProvideChecker(
 		checker.PwHistoryDays = cfg.Policy.HistoryDays
 		checker.PasswordHistoryEnabled = cfg.Policy.IsEnabled()
 	}
-	if !*featureCfg.Password.Expiry.ForceChange.Disabled {
-		checker.PwExpiryForceChangeEnabled = cfg.Expiry.ForceChange.IsEnabled()
-		checker.PwExpiryForceChangeSinceLastUpdate = cfg.Expiry.ForceChange.SinceLastUpdate
-	}
+	checker.PwExpiryForceChangeEnabled = cfg.Expiry.ForceChange.IsEnabled()
+	checker.PwExpiryForceChangeSinceLastUpdate = cfg.Expiry.ForceChange.SinceLastUpdate
 	if !*featureCfg.Password.Policy.MinimumGuessableLevel.Disabled {
 		checker.PwMinGuessableLevel = cfg.Policy.MinimumGuessableLevel
 	}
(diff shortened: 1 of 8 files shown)


pkg/lib/authenticationflow/declarative/data_password_policy.go
```diff
@@ -10,6 +10,15 @@ type PasswordPolicyHistory struct {
 	Days    int  `json:"days,omitempty"`
 }
 
+type PasswordPolicyExpiry struct {
```
louischan-oursky, line 13: It seems to me that there is no need to tell password expiry to the authflow caller. The information we expose in authflow is for the caller to implement the UI. Password expiration details normally will not be shown in the UI. Most of the time, the end-user just enter their password and let the server to tell them whether the password is still valid.

So do not report password expiry in authflow data.

pkg/lib/config/authenticator.go
```diff
@@ -58,34 +79,70 @@ var _ = Schema.Add("PasswordPolicyConfig", `
 		"minimum_guessable_level": { "type": "integer" },
 		"excluded_keywords": { "type": "array", "items": { "type": "string" } },
 		"history_size": { "type": "integer" },
-		"history_days": { "$ref": "#/$defs/DurationDays" }
+		"history_days": { "$ref": "#/$defs/DurationDays" },
+		"expiry": { "$ref": "#/$defs/PasswordExpiryConfig" }
```
louischan-oursky, line 83:
```yaml
password:
  # The original policy configuration.
  # I think password expiry should not be put in it.
  # The things in policy is more about password requirement.
  # If we can rename it, it should be called "requirement" instead.
  policy:
    min_length: 8
  # This force_change is for password.policy.
  force_change: true

  # The 3 cases can be turned on individually.
  # However, the precedence is deny_login > force_change > prompt_change.
  expiry: # "expiration" is American English, while "expiry" is British English. But we have been using "expiry" in the config. So let's stick with it.
    # In this case, the authflow will result in a dead end, with an error telling that the password is expired and the login is denied.
    deny_login:
      enabled: true
      since_last_update_in_days: 90
    # In this case, the authflow will enter the change_password step (The existing one). To proceed, the end-user must change the password.
    force_change:
      enabled: true
      since_last_update_in_days: 60
    # In this case, the authflow will enter the change_password step. But this step now supports taking an input to skip the password update.
    prompt_change:
      enabled: true
      since_last_update_in_days: 30
```

pkg/lib/config/authenticator.go
```diff
@@ -86,6 +88,57 @@ func (c *PasswordPolicyConfig) SetDefaults() {
...
+}
+`)
+
+var _ = Schema.Add("PasswordExpiryForceChangeConfig", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"enabled": { "type": "boolean" },
+		"since_last_update_in_days": { "$ref": "#/$defs/DurationDays" }
```
louischan-oursky: Use DurationString. Refer to RateLimitConfig to see how the schema of RateLimitConfig is written in a way that `since_last_update_duration` is only required when `enabled` is true.

pkg/lib/config/authenticator.go
```diff
@@ -86,6 +88,57 @@ func (c *PasswordPolicyConfig) SetDefaults() {
...
+}
+
+func (c *PasswordExpiryConfig) SetDefaults() {
+	if c.ForceChange == nil {
+		c.ForceChange = &PasswordExpiryForceChangeConfig{}
+	}
+}
+
+type PasswordExpiryForceChangeConfig struct {
+	Enabled               bool         `json:"enabled"`
```
louischan-oursky: omitempty

pkg/lib/config/feature_authenticator.go
```diff
@@ -65,3 +67,44 @@ func (c *PasswordPolicyItemFeatureConfig) SetDefaults() {
 		c.Disabled = newBool(false)
 	}
 }
+
+var _ = FeatureConfigSchema.Add("PasswordExpiryFeatureConfig", `
```
louischan-oursky, line 71: No need to add feature config for password expiry.

pkg/lib/config/authenticator.go
```diff
@@ -120,25 +120,15 @@ func (c *PasswordExpiryConfig) SetDefaults() {
...
-	Enabled               bool         `json:"enabled"`
-	SinceLastUpdateInDays DurationDays `json:"since_last_update_in_days,omitempty"`
+	Enabled         bool           `json:"enabled"`
+	SinceLastUpdate DurationString `json:"since_last_update,omitempty"`
 }
 
 func (c *PasswordExpiryForceChangeConfig) IsEnabled() bool {
-	return c.Enabled && c.SinceLastUpdateInDays > 0
+	sinceLastUpdate, _ := c.SinceLastUpdate.MaybeDuration()
+	return c.Enabled && sinceLastUpdate > 0
```
louischan-oursky, line 129: It should be `c.Enabled && ok && sinceLastUpdate > 0`.

louischan-oursky:
Like pkg/lib/config/testdata/rate_limit_tests.yaml, add tests for the new config.

IniZio:
It's currently added in `pkg/lib/config/testdata/config_tests.yaml`, or prefer splitting into separate test files?

Will add more test on the conditional require nevertheless

pkg/lib/config/authenticator.go
louischan-oursky: - Rename `since_last_update` to `duration_since_last_update`.
- Copy the above config and paste it as comment around `PasswordExpiryConfig` for future reference.

pkg/lib/authn/authenticator/password/deps.go
```diff
@@ -26,6 +26,10 @@ func ProvideChecker(
 		checker.PwHistoryDays = cfg.Policy.HistoryDays
 		checker.PasswordHistoryEnabled = cfg.Policy.IsEnabled()
 	}
+	if !*featureCfg.Password.Expiry.ForceChange.Disabled {
+		checker.PwExpiryForceChangeEnabled = cfg.Expiry.ForceChange.IsEnabled()
+		checker.PwExpiryForceChangeSinceLastUpdateInDays = cfg.Expiry.ForceChange.SinceLastUpdateInDays
+	}
```
louischan-oursky: Remove changes to authn/authenticator/password/checker.go and authn/authenticator/password/deps.go

louischan-oursky:
superseded by https://github.com/authgear/authgear-server/pull/3968

PR #3955, opened 2024-03-12
Add group screen empty state

Ngkaokis:
ref #3903

Issue #3912, opened 2024-03-04
Groups - Delete Group


Ngkaokis:
![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/3c0f7dc2-1d4f-4785-a957-18fb0afdfd3e/550f8f36-ae5c-40c2-a1cc-b11c31c4ee9a?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi8zYzBmN2RjMi0xZDRmLTQ3ODUtYTk1Ny0xOGZiMGFmZGZkM2UvNTUwZjhmMzYtYWU1Yy00MGMyLWExY2MtYjExYzMxYzRlZTlhIiwiaWF0IjoxNzA5NTM1NjcyLCJleHAiOjMzMjgwMDk1NjcyfQ.Xnof-QgYrOVbMUer9hi88qiyUU3DL9O8ZR6oD1Ek6nk "=1047x")

Issue #3911, opened 2024-03-04
Groups - List View


Ngkaokis:
![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/effaa010-1941-4702-9bdf-633ddfe78f19/dc3c0d1c-d0bf-44e3-afc3-b5da33c13ff1?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi9lZmZhYTAxMC0xOTQxLTQ3MDItOWJkZi02MzNkZGZlNzhmMTkvZGMzYzBkMWMtZDBiZi00NGUzLWFmYzMtYjVkYTMzYzEzZmYxIiwiaWF0IjoxNzA5NTM1Njc1LCJleHAiOjMzMjgwMDk1Njc1fQ.xGAA5CBpZerY9qcWfmOY6KOW2bad7SHWRZRrbn11eP8 "=1065x")

PR #3965, opened 2024-03-12
Implement Groups List and Group Delete

Ngkaokis:
ref #3911 #3912 

require #3955

### DIFF
--- a/portal/src/graphql/adminapi/DeleteGroupDialog.tsx
+++ b/portal/src/graphql/adminapi/DeleteGroupDialog.tsx
@@ -1,10 +1,4 @@
-import React, {
-  useCallback,
-  useContext,
-  useEffect,
-  useMemo,
-  useState,
-} from "react";
+import React, { useCallback, useContext, useMemo } from "react";
 import {
   Dialog,
   DialogFooter,
@@ -17,6 +11,7 @@ import PrimaryButton from "../../PrimaryButton";
 import DefaultButton from "../../DefaultButton";
 import ErrorDialog from "../../error/ErrorDialog";
 import { useDeleteGroupMutation } from "./mutations/deleteGroupMutation";
+import { useSnapshotData } from "../../hook/useSnapshotData";
 
 export interface DeleteGroupDialogData {
   groupID: string;
@@ -28,18 +23,6 @@ interface DeleteGroupDialogProps {
   data: DeleteGroupDialogData | null;
   onDismiss: (isDeleted: boolean) => void;
   onDismissed?: () => void;
-}
-
-function useSnapshotData(data: DeleteGroupDialogData | null) {
-  // Keep the latest non-null data, because the dialog has transition animation before dismiss.
-  // During the transition, we still need the data. However, the parent may already changed the props.
-  const [snapshot, setSnapshot] = useState<DeleteGroupDialogData | null>(data);
-  useEffect(() => {
-    if (data !== null) {
-      setSnapshot(data);
-    }
-  }, [data]);
-  return snapshot;
 }
 
 const DeleteGroupDialog: React.VFC<DeleteGroupDialogProps> =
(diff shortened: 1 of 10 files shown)


portal/src/RoleAndGroupsLayout.tsx
```diff
@@ -9,8 +10,13 @@ import {
...
 } from "./ErrorMessageBar";
 
-interface RoleAndGroupsLayoutProps {
+export interface RoleAndGroupsHeaderProps {
   breadcrumbs: BreadcrumbItem[];
+  subitem?: JSX.Element | null;
+}
+
+interface RoleAndGroupsLayoutProps {
+  header: RoleAndGroupsHeaderProps;
```
tung2744, line 19: Usually we won't nest props, because this make it difficult to use (Because inline objects doesn't not have a stable reference, usually you need to use `useMemo` to create the object).

Prefer flatten the props and name it `headerBreadcrumbs` and `headerSubItems`

portal/src/RoleAndGroupsLayout.module.css
```diff
@@ -24,7 +24,11 @@
 }
 
 .header {
-  @apply self-stretch;
+  @apply flex flex-row justify-between;
+}
+
+.breadcrumb {
+  @apply flex-1-0-auto;
```
tung2744, line 31: I think what you want is:
- Take all remaining spaces (grow = 1)
- But not more than remaining spaces (shrink = 1)

Therefore what you want should be flex-1.

portal/src/RoleAndGroupsLayout.tsx
```diff
@@ -9,8 +10,13 @@ import {
   ErrorMessageBarContextProvider,
 } from "./ErrorMessageBar";
 
-interface RoleAndGroupsLayoutProps {
+export interface RoleAndGroupsHeaderProps {
   breadcrumbs: BreadcrumbItem[];
+  subitem?: JSX.Element | null;
```
tung2744, line 15: Use `React.ReactNode` instead of JSX.Element. Remove `/* global JSX */` at the top of this file as well.

portal/src/graphql/adminapi/DeleteGroupDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+  groupKey: string;
+}
+
+interface DeleteGroupDialogProps {
+  data: DeleteGroupDialogData | null;
+  onDismiss: (isDeleted: boolean) => void;
+  onDismissed?: () => void;
+}
+
+function useSnapshotData(data: DeleteGroupDialogData | null) {
```
tung2744, line 33: :sadfrog: instead of copying this hook again, can we create a new file to implement this hook in a generic way and reuse it?

portal/src/graphql/adminapi/GroupsList.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const onRenderGroupRow = React.useCallback(
+    (props?: IDetailsRowProps) => {
+      if (props == null) {
+        return null;
+      }
+      const targetPath = isGroupListItem(props.item)
+        ? `/project/${appID}/user-management/groups/${props.item.id}/details`
+        : ".";
+      return (
+        <Link to={targetPath}>
```
tung2744, line 124: Give this Link a "contents" class so that it doesn't participate in layout.

portal/src/graphql/adminapi/GroupsList.module.css
```diff
@@ -0,0 +1,48 @@
...
+  line-height: 20px;
+}
+
+.description {
+  white-space: normal;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  font-size: 12px;
+  line-height: 16px;
+}
```
tung2744, line 48: All lists in roles and groups should share same styles, maybe we should create common components to contain these styles.

We can do it in separated issue.

Ngkaokis: opened https://github.com/authgear/authgear-server/issues/3967

Issue #3956, opened 2024-03-12
Add password expiration checker in authflow and interaction level


IniZio:
- [X] Add `checkPasswordExpiry` in `ValidateCurrentPassword` (`pkg/lib/authn/authenticator/password/checker.go`)

* Compare `updated_at` in `_auth_authenticator` (`pkg/lib/authn/authenticator/password/store.go`)

- [ ] Interaction
- [X] Authflow

Questions:

* How to return warning?

```
{
  "warnings": [
    {
      "message": "Password will expire in [n] days."
    }
  ]
}
```

Update:

Will exclude warnings in current sprint. Also warning will add extra step instead.

louischan-oursky:
If we think in terms of authflow, it is not just a warning. The action the end-user can take is 1) Change the password now, 2) Skip. So it is a step in the authflow. We have this step (change_password) already. But change_password currently does not allow skipping. We also have another step (prompt_create_passkey) which allows skipping. Therefore, password expiration warning in authflow is change_password with skip.

PR #3968, opened 2024-03-13
Add password expiration checker

IniZio:
ref #3956

dependencies:
- [x] #3964


pkg/lib/authenticationflow/declarative/intent_login_flow_step_authenticate.go
```diff
@@ -40,7 +40,7 @@ func (i *IntentLoginFlowStepAuthenticate) GetJSONPointer() jsonpointer.T {
 
 var _ IntentLoginFlowStepChangePasswordTarget = &IntentLoginFlowStepAuthenticate{}
 
-func (i *IntentLoginFlowStepAuthenticate) GetPasswordAuthenticator(_ context.Context, _ *authflow.Dependencies, flows authflow.Flows) (info *authenticator.Info, ok bool) {
+func (i *IntentLoginFlowStepAuthenticate) GetPasswordAuthenticator(_ context.Context, _ *authflow.Dependencies, flows authflow.Flows) (info *authenticator.Info, changeRequired bool) {
```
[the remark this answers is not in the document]
IniZio: I don't think `ok` had any meaning other than password authenticator need update?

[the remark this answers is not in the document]
tung2744: 🤔 Then it seems `GetPasswordAuthenticator` should be named as `GetChangeRequiredPasswordAuthenticator`, because it will not return the password authenticator if it doesn't require change, even it exist in the flow.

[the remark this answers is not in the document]
IniZio: Or maybe I move the info assign outside of the condition then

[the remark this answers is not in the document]
louischan-oursky: Let's rename GetPasswordAuthenticator to GetChangeRequiredPasswordAuthenticator to match its behavior.

pkg/lib/authn/authenticator/password/checker.go
```diff
@@ -250,6 +260,23 @@ func (pc *Checker) checkPasswordHistory(password, authID string) (*Policy, error
 	return nil, nil
 }
 
+func (pc *Checker) checkPasswordExpiry(password, authID string) error {
```
louischan-oursky, line 263: This checking should be done in password.Provider#Authenticate, rather than here.

Rationale
- password.Provider#Authenticate has the *authenticator.Password already.
- password.Checker is intended to enforce password requirement, not password expiry.

pkg/lib/authn/authenticator/password/checker.go
```diff
@@ -290,7 +290,7 @@ func (pc *Checker) ValidateNewPassword(userID string, plainPassword string) erro
 }
 
 // ValidateCurrentPassword should be used when the user authenticates.
-func (pc *Checker) ValidateCurrentPassword(plainPassword string) error {
+func (pc *Checker) ValidateCurrentPassword(userID string, plainPassword string) error {
```
louischan-oursky, line 293: Remove this change.

pkg/lib/authn/authenticator/password/deps.go
```diff
@@ -4,15 +4,18 @@ import (
...
+	"github.com/authgear/authgear-server/pkg/util/clock"
 )
 
 func ProvideChecker(
 	cfg *config.AuthenticatorPasswordConfig,
 	featureCfg *config.AuthenticatorFeatureConfig,
-	s CheckerHistoryStore,
+	hs CheckerHistoryStore,
+	pas *Store,
+	c clock.Clock,
```
louischan-oursky, line 15: Remove this change.

pkg/lib/authn/authenticator/password/deps.go
```diff
@@ -4,15 +4,18 @@ import (
...
 ) *Checker {
 	checker := &Checker{
-		PasswordHistoryStore: s,
+		PasswordHistoryStore: hs,
```
louischan-oursky, line 18: Remove this change.

pkg/lib/authn/authenticator/password/provider.go
```diff
@@ -124,12 +125,17 @@ func (p *Provider) Authenticate(a *authenticator.Password, password string) (req
...
 	}
 
-	if notAllowedErr := p.PasswordChecker.ValidateCurrentPassword(password); notAllowedErr != nil {
+	if validateErr := p.PasswordChecker.ValidateCurrentPassword(a.UserID, password); validateErr != nil {
 		if p.Config.ForceChange != nil && *p.Config.ForceChange {
 			requireUpdate = true
 		}
 	}
 
+	if expiryErr := p.Expiry.Validate(password, a.ID); expiryErr != nil {
```
louischan-oursky, line 134: It should take `a` instead of `password`. We just need to check the password being used is expired, not all password the user has.

pkg/lib/authn/authenticator/service/service.go
```diff
@@ -513,7 +514,11 @@ func (s *Service) verifyWithSpec(info *authenticator.Info, spec *authenticator.S
 		plainPassword := spec.Password.PlainPassword
 		a := info.Password
 		requireUpdate, err = s.Password.Authenticate(a, plainPassword)
-		if err != nil {
+		var apiError = apierrors.AsAPIError(err)
+		if apiError != nil && apiError.Kind == password.PasswordExpiryForceChange {
+			// Proceed next auth step but require update
+			err = api.ErrPasswordExpiryForceChange
```
louischan-oursky, line 520: This pattern of reporting information is very easy to miss. It requires to the caller to pay attention that the error sometimes does not mean an error. 

Let's change the interface so that the caller must be aware of the possible outcome.

Originally: `(info *authenticator.Info, requireUpdate bool, err error)`
Now: `(info *authenticator.Info, result *VerifyResult, err error)`

```golang
type VerifyResult struct {
  PasswordPolicyForceChange bool
  PasswordExpiryForceChange bool
}
```

With this change, this PR should be greatly simplified. We just propagate VerifyResult and we can differentiate the case.

I also notice that we need to add a piece of data in the `change_password` step to tell the caller why the change is needed. The step is using `new_password_data`. Let's add a new data `force_change_password_data`. This new data contains a new field `reason`. The valid values for `reason` is `policy` or `expiry`.

IniZio: If that's the case seems better to propagate to workflow / interaction to decide what next action is?

pkg/lib/authenticationflow/declarative/data_change_password.go
```diff
@@ -0,0 +1,9 @@
+package declarative
+
+type PasswordChangeReason string
+
+const (
+	PasswordChangeReasonPolicy PasswordChangeReason = "policy"
+	// nolint: gosec // This is not a credential
+	PasswordChangeReasonExpiryForceChange PasswordChangeReason = "expiry_force_change"
```
louischan-oursky, line 8: Just name it `expiry`.

pkg/lib/authenticationflow/declarative/data_force_change_password.go
```diff
@@ -0,0 +1,18 @@
+package declarative
+
+import authflow "github.com/authgear/authgear-server/pkg/lib/authenticationflow"
+
+type ForceChangePasswordData struct {
+	TypedData
+	PasswordPolicy *PasswordPolicy       `json:"password_policy,omitempty"`
+	Reason         *PasswordChangeReason `json:"reason,omitempty"`
```
louischan-oursky, line 8: Rename this from `reason` to `force_change_reason`

So it now reads `force_change_reason: policy` and `force_change_reason: expiry`.

pkg/lib/authn/authenticator/password/expiry.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type Expiry struct {
+	ForceChangeEnabled         bool
+	ForceChangeSinceLastUpdate config.DurationString
+	Clock                      clock.Clock
+}
+
+func (pe *Expiry) Validate(authenticator *authenticator.Password, password string) error {
+	if IsSamePassword(authenticator.PasswordHash, password) {
```
louischan-oursky, line 17: No need to verify the password once more. The password should have been verified before this method is called.

Comparing password is an expensive operation, we do not want to spend time doing it twice.

IniZio:
Updated the naming 🙏

louischan-oursky:
Build failed.

IniZio:
Yes, ci in local didn't have this error so checking why.

pkg/lib/authn/authenticator/password/expiry.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/authgear/authgear-server/pkg/util/clock"
+)
+
+type Expiry struct {
+	ForceChangeEnabled         bool
+	ForceChangeSinceLastUpdate config.DurationString
+	Clock                      clock.Clock
+}
+
+func (pe *Expiry) Validate(authenticator *authenticator.Password, password string) error {
```
louischan-oursky, line 16: Do not need to take `password`.

IniZio:
Seems `go generate` is bugged sometimes, it successfully generated `pkg/admin/wire_gen.go` after I rebased again.

PR #3977, opened 2024-03-14
Add password expiry config to portal 

IniZio:
ref #3958

dependencies:
- [x] #3964

image

### DIFF
--- a/portal/src/graphql/portal/PasswordSettings.tsx
+++ b/portal/src/graphql/portal/PasswordSettings.tsx
@@ -760,7 +760,7 @@ export default function PasswordSettings<T extends State>(
         label={renderToString(
           "LoginMethodConfigurationScreen.password.expiry.force-change-since-last-update.label"
         )}
-        value={passwordExpiryForceChangeDays?.toFixed(0)}
+        value={passwordExpiryForceChangeDays?.toFixed(0) ?? ""}
         onChange={onChangeExpiryForceChangeDays}
         onBlur={onBlurExpiryForceChangeDays}
       />
--- a/portal/src/graphql/portal/PasswordSettings.tsx
+++ b/portal/src/graphql/portal/PasswordSettings.tsx
@@ -36,6 +36,7 @@ import {
   parseNumber,
   tryProduce,
 } from "../../util/input";
+import { formatDuration, parseDuration } from "../../util/duration";
 
 export enum ResetPasswordWithEmailMethod {
   Link = "link",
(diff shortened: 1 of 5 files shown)


portal/src/graphql/portal/PasswordSettings.tsx
```diff
@@ -408,6 +429,30 @@ export default function PasswordSettings<T extends State>(
...
+      setState((s) =>
+        produce(s, (s) => {
+          s.authenticatorPasswordConfig.expiry ??= {};
+          s.authenticatorPasswordConfig.expiry.force_change ??= {};
+          s.authenticatorPasswordConfig.expiry.force_change.duration_since_last_update =
+            tryProduce(
+              s.authenticatorPasswordConfig.expiry.force_change
+                .duration_since_last_update,
+              () => {
+                const num = parseNumber(value);
```
louischan-oursky, line 446: It does not allow me to delete all text, and re-enter a new number. When I delete the last digit, the toggle becomes disabled, and the text field is blurred.

I think the behavior should be

1. Only allow digits to be entered in the text field.
2. Allow the text field to be empty.
3. If the text field is blurred and it is empty, then change the toggle to false.
4. If the text field is blurred and it is 0, then change the toggle to false.
5. When the toggle is false, the text field should be empty, not contain a 0.

portal/src/locale-data/en.json
```diff
@@ -551,6 +551,11 @@
   "PasswordPolicyConfigurationScreen.history-size.label": "No. of password history to keep",
   "PasswordPolicyConfigurationScreen.excluded-keywords.label": "Keywords to be excluded",
 
+  "LoginMethodConfigurationScreen.password.expiry": "Password Expiry",
+  "LoginMethodConfigurationScreen.password.expiry.description": "Configure actionss upon password expiry.",
```
louischan-oursky, line 555: actions. Remove the extra s.

portal/src/graphql/portal/PasswordSettings.tsx
```diff
@@ -408,6 +425,44 @@ export default function PasswordSettings<T extends State>(
...
+      );
+    },
+    [setState]
+  );
+
+  const onBlurExpiryForceChangeDays = useCallback(() => {
+    setState((s) =>
+      produce(s, (s) => {
+        if (!passwordExpiryForceChangeDays) {
+          s.authenticatorPasswordConfig.expiry ??= {};
```
louischan-oursky, line 456: I think we can just set `expiry` to undefined.

Issue #3882, opened 2024-02-29
Implement /api/users/import as a synchronous API first


fungc-io:
- [X] Implement insert
- [X] `preferred_username`, `email`, `phone_number`
- [X] `email_verified`, `phone_number_verified`
- [X] standard attributes
- [X] custom attributes
- [X] `disabled`
- [X] `roles`, `groups`
- [X] `password`
- [X] `mfa.email`
- [X] `mfa.phone_number`
- [X] `mfa.password`
- [X] `mfa.totp`
- [X] Implement update
- [X] Re-index the user

PR #3978, opened 2024-03-14
User import - Part 1

louischan-oursky:
ref #3882

### DIFF
--- a/pkg/admin/transport/handler_user_import.go
+++ b/pkg/admin/transport/handler_user_import.go
@@ -1,9 +1,13 @@
 package transport
 
 import (
+	"context"
 	"net/http"
 
+	"github.com/authgear/authgear-server/pkg/api"
+	"github.com/authgear/authgear-server/pkg/lib/userimport"
 	"github.com/authgear/authgear-server/pkg/util/httproute"
+	"github.com/authgear/authgear-server/pkg/util/httputil"
 )
 
 func ConfigureUserImportRoute(route httproute.Route) httproute.Route {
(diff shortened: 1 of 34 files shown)


pkg/admin/transport/handler_user_import.go
```diff
@@ -0,0 +1,50 @@
...
+
+type UserImportHandler struct {
+	JSON              JSONResponseWriter
+	UserImportService UserImportService
+}
+
+func (h *UserImportHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var err error
+	var request userimport.Request
+	err = httputil.BindJSONBody(r, w, userimport.RequestSchema.Validator(), &request)
```
tung2744, line 30: Do we have a limit on the json size?

louischan-oursky: Yes. pkg/util/httputil/json.go:16

louischan-oursky: But it seems too large for user import. Maybe I refactor a bit to make it configurable.

tung2744:
@louischan-oursky I've tried the api, and found there seems a bug in upsert.
If you set `disabled: false`, and trying to upsert non-disabled user, you will see an error
```
        "errors": [
          {
            "name": "Invalid",
            "reason": "InvalidAccountStatusTransition",
            "message": "invalid account status transition: normal -> normal",
            "code": 400,
            "info": {
              "from": "normal",
              "to": "normal"
            }
          }
        ]
```

pkg/admin/transport/handler_user_import.go
louischan-oursky: Fixed.

louischan-oursky:
Fixed the account status bug.

Issue #3914, opened 2024-03-04
Group Detail - Roles Tab


Ngkaokis:
- [ ] roles tab
  - [ ] empty state
  - [ ] roles list view
  - [ ] search bar
  - [ ] add to role dialog
    - [ ] combobox
  - [ ] remove role dialog

PR #3979, opened 2024-03-14
Implement group details roles tab

Ngkaokis:
ref #3914 

require #3971


portal/src/components/roles-and-groups/DeleteGroupRoleDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+  groupKey: string;
+}
+
+interface DeleteGroupRoleDialogProps {
+  data: DeleteGroupRoleDialogData | null;
+  onDismiss: (isDeleted: boolean) => void;
+  onDismissed?: () => void;
+}
+
+function useSnapshotData(data: DeleteGroupRoleDialogData | null) {
```
tung2744, line 37: Use the common function instead of duplicating one.

portal/src/util/rolesAndGroups.ts
```diff
@@ -0,0 +1,41 @@
+export interface Searchable {
+  id: string;
+  key: string;
+  name?: string | null;
+}
+
+export function searchRolesAndGroups<T extends Searchable>(
```
tung2744, line 7: How roles / groups can be searched should be a model logic of roles / groups, therefore I put it in `model/XX.ts`. 

I prefer keeping them as two separated functions `searchGroups` and `searchRoles`.
You can extract some common logics to util, such as the preprocessing of searchKeyword.

portal/src/locale-data/en.json
```diff
@@ -184,6 +184,20 @@
...
   "GroupsList.delete-group": "Delete Group",
 
+  "GroupRolesList.column.name": "Name",
+  "GroupRolesList.column.key": "Role Key",
+  "GroupRolesList.column.action": "Action",
+  "GroupRolesList.actions.remove": "Remove",
+  "GroupRolesList.empty": "No roles in group",
+
+  "DeleteGroupRoleDialog.title": "Remove Role",
+  "DeleteGroupRoleDialog.description": "Are you sure you want to remove \"{groupName}\" from the \"{roleName}\" group?",
```
tung2744, line 194: from the \"{roleName}\" role?

tung2744:
Added two more commits:
1. Move the search util function into the correct file.
2. Make the search result more accurate.

Issue #3848, opened 2024-02-26
Country list may be broken if the search box is scrolled away


fungc-io:
## Steps to reproduce 

- Use WebKit Webview on iOS
- open the country list in phone input
- Drag the area around the search box and scroll up.
- The country list will be broken and cannot see the searchbox again.

The problem is not severe because 1) the user can select any country and re-open the list,  it will reset automatically & 2) it's a very specific action that is not easy to trigger,
It's more a aesthetic problem.

![Image](https://github.com/authgear/authgear-server/assets/15207470/758cfc8e-f2c7-4050-8a31-ff01e0ec51ac)

https://github.com/authgear/authgear-server/assets/15207470/167bc9a3-90f5-4b85-b024-eff8a5b26e2e

fungc-io:
Possible fix: Dismiss keyboard (unfocus) when tapping outside the text box

IniZio:
Still got edge cases with this fix, moving back.

PR #3996, opened 2024-03-18
Dismiss keyboard on scrolling in fixed element

IniZio:
ref #3848


louischan-oursky:
Right now, when the popup is opened, it is possible to scroll the body under it, resulting a weird behavior.

Approach to explore
- See if we use `<dialog>`, the bug is fixed already.
- If `<dialog>` fixes the bug, that verify that we can feature-detect the support of `<dialog>` element.
- Feature-detect `<dialog>` and use it. If it does not support `<dialog>`, then leave the bug unfixed.

authui/src/authflowv2/dismissKeyboard.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * On iOS, the keyboard is treated as overlay rather than part of the viewport.
+ * This causes an extra padding to be added to the bottom of the viewport when the keyboard is open.
+ *
+ * This controller listens for scroll events and dismisses the keyboard when user scrolls on
+ * the body or non-`overflow:auto/scroll` element.
+ */
+export class DismissKeyboardOnScrollController extends Controller {
+  connect() {
+    window.addEventListener("scroll", this.dismissKeyboard);
```
tung2744, line 12: Use `capture` to listen to all possible scroll events?

tung2744:
For the dialog scrolling problem, I think it could be solved by adding a backdrop which covers the whole screen. And pressing the backdrop should close the dialog too.

IniZio:
A backdrop will not help preventing body scroll especially on iOS. react-aria needed quite some workaround for this

https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/overlays/src/usePreventScroll.ts#L80-L106

IniZio:
> Right now, when the popup is opened, it is possible to scroll the body under it, resulting a weird behavior.

Actually tested on physical device again didn't reproduce this, even for "No results found ase".

authui/src/authflowv2/dismissKeyboard.ts
IniZio: Currently set `capture` to false to avoid closing keyboard when scrolling in the options container. 

Otherwise the keyboard will be dismissed when user is typing in searchbar since there is autoscrolling

tung2744: I see

authui/src/authflowv2/bodyScrollLock.ts
```diff
@@ -0,0 +1,26 @@
+import { Controller } from "@hotwired/stimulus";
+
+export class BodyScrollLockController extends Controller {
+  static targets = ["dialog"];
```
louischan-oursky, line 4: This seems unused. And Stimulus does have something called targets https://stimulus.hotwired.dev/reference/targets

authui/src/authflowv2/customSelect.ts
```diff
@@ -26,6 +27,14 @@ export class CustomSelectController extends Controller {
...
   };
 
+  get bodyScrollLock() {
+    const ctr = this.application.getControllerForElementAndIdentifier(
+      this.element,
+      "body-scroll-lock"
+    );
+    return ctr as BodyScrollLockController | null;
+  }
+
```
louischan-oursky, line 37: Follow this guide to coordinate these two controllers. https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events

Ultimately, it should look like

```
<div data-controller="custom-select body-scroll-lock"
  data-action="custom-select:open->body-scroll-lock#lock custom-select:close->body-scroll-lock#unlock">
```

authui/src/authflowv2/bodyScrollLock.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  scrollPosition = 0;
+
+  lock() {
+    this.scrollPosition = document.documentElement.scrollTop;
+    document.body.style.overflow = "hidden";
+    document.body.style.position = "fixed";
+    document.body.style.top = `-${this.scrollPosition}px`;
+    document.body.style.width = "100%";
+    document.body.style.height = visualViewport?.height
```
louischan-oursky, line 14: Use `window.visualViewport`. It is more explicit.

authui/src/authflowv2/bodyScrollLock.ts
```diff
@@ -0,0 +1,26 @@
...
+      ? `${visualViewport.height}px`
+      : "100vh";
+  }
+
+  unlock() {
+    document.body.style.removeProperty("overflow");
+    document.body.style.removeProperty("position");
+    document.body.style.removeProperty("top");
+    document.body.style.removeProperty("width");
```
louischan-oursky, line 23: If body originally has some of these property defined, removing them will break things. Consider using insertRule() https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule

In lock(), we create a style element and append it to the body. In unlock(), we remove the style element we created in lock() from the body.

IniZio: Updated 🙏

Issue #3980, opened 2024-03-14
Users - Groups Tab


Ngkaokis:
Should just reuse the component from roles and groups

Or should refactor those to be reusable if thy are not 

- [ ] empty view
- [ ] list view
  - [ ] search box
  - [ ] add group button
- [ ] delete group dialog
- [ ] add group dialog

PR #3997, opened 2024-03-18
Implement User groups tab

Ngkaokis:
ref #3980 

pending question: do we need to respect user page's grid layout

### DIFF
--- a/portal/src/components/roles-and-groups/list/UserDetailsScreenGroupListContainer.tsx
+++ b/portal/src/components/roles-and-groups/list/UserDetailsScreenGroupListContainer.tsx
@@ -1,0 +1,111 @@
+import { useQuery } from "@apollo/client";
+import { SearchBox } from "@fluentui/react";
+import {
+  FormattedMessage,
+  Context as MessageContext,
+} from "@oursky/react-messageformat";
+import React, { useContext, useState, useCallback, useMemo } from "react";
+import ShowError from "../../../ShowError";
+import ShowLoading from "../../../ShowLoading";
+import {
+  GroupsListQueryQuery,
+  GroupsListQueryQueryVariables,
+  GroupsListQueryDocument,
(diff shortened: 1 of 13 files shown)


tung2744:
> do we need to respect user page's grid layout

Per discussion in standup, we will keep the 8 column layout for now.

portal/src/components/roles-and-groups/list/UserDetailsScreenGroupListContainer.tsx
```diff
@@ -0,0 +1,162 @@
...
+            onChange={onChangeSearchKeyword}
+            onClear={onClearSearchKeyword}
+          />
+          <PrimaryButton
+            text={<FormattedMessage id="UserDetailsScreen.groups.add" />}
+            onClick={showAddRoleDialog}
+          />
+        </header>
+        <UserGroupsList
+          className="flex-1 min-h-0"
```
tung2744, line 140: Screenshot 2024-03-18 at 7 40 49 PM

When the screen is too short, the list could become completely hidden.
So lets give it a minimum height, say 200px.
Also, use `flex-1-0-auto` to ensure it doesn't shrink beyond the minimum height.

portal/src/graphql/adminapi/UserDetailsScreen.module.css
```diff
@@ -2,8 +2,25 @@
...
 
 .siweEnabledTabWarningMessageBar {
   @apply mt-5;
 }
+
+.screenContent {
+  @apply flex-1-0-auto;
+
+  grid-template-rows: min-content minmax(min-content, 1fr);
+  grid-auto-columns: 1fr;
```
tung2744, line 17: It seems this `grid-auto-columns: 1fr;` has no effect, because `grid-template-columns: repeat(12, 80px);` already defines it has 12 columns, and 80px each.

Issue #3989, opened 2024-03-14
Implement a Redis queue for user import


louischan-oursky:
https://redis.com/glossary/redis-queue/

---

Producer
  - LPUSH: enqueue a job
    - LPUSH "queue" "Task1"

Consumer
  - BRPOP: dequeue a job
    - BRPOP "queue" 30

The Reliable Queues pattern is not implemented because house keeping the processing list is complicated.

Issue #3883, opened 2024-02-29
Convert user import to async


fungc-io:
- Convert /_api/admin/users/import to async
- Implement /_api/admin/users/import/ID

PR #3998, opened 2024-03-18
Make user import API async

louischan-oursky:
ref #3989 
ref #3883


louischan-oursky:
This PR is missing the redaction of secret in the response. Let me add it back.

pkg/lib/infra/redisqueue/producer.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+const (
+	idAlphabet string = base32.Alphabet
+	idLength   int    = 32
+)
+
+var rng *rand.Rand = corerand.SecureRand
+
+var TTL = 24 * time.Hour
```
tung2744, line 26: It seems there shouldn't be TTL on the tasks. They should be processed if they exist.

pkg/lib/infra/redisqueue/task.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+type TaskStatus string
+
+const (
+	TaskStatusPending   TaskStatus = "pending"
+	TaskStatusCompleted TaskStatus = "completed"
+)
+
+type Task struct {
```
tung2744, line 16: Should there be a `type` key here to indicate the task type?
(but we have one type of task only for now)

pkg/lib/infra/redisqueue/producer.go
louischan-oursky: We need TTL so that it will be deleted automatically. We do not store them indefinitely.

pkg/lib/infra/redisqueue/task.go
louischan-oursky: Right now it the type is determined by the queue name. A queue only stores one type of task. And each queue has a specific TaskProcessor to process it.

pkg/lib/infra/redisqueue/producer.go
tung2744: 🤔 Finished tasks can be removed automatically, However, it seems incompleted tasks shouldn't be deleted.

Or actually the task result should be stored into db?
And we can also keep track the history of import tasks. One days seems too short for a import result?
And as it is read only it shouldn't cause large loadings on the db?

louischan-oursky: > Finished tasks can be removed automatically, However, it seems incompleted tasks shouldn't be deleted.

If the task is not dequeued within 1 day, then yes, it is a problem. Otherwise, the task will be saved with status update with a TTL of 24 hours. For our usage on hand, it should be fine.

> actually the task result should be stored into db?
I tend to avoid housekeeping the task results.

> One days seems too short for a import result?
Auth0 also keeps the result for 1 day https://auth0.com/docs/manage-users/user-migration/bulk-user-imports#job-timeouts

> Or maybe we need audit logs?
We can add it later.

tung2744: My concern is if there is any problem occurred causing the consumer stopped to consume tasks from the queue, and if the tasks will disappear automatically after one day, it could be difficult for us to discover the problem.

louischan-oursky: The consumer now does not recover from panic so a panic will cause it to stop. Let me fix it.

tung2744: I think there could be other unknown reason / bugs that causing the consumer to stop consuming tasks, so I guess we might want some kind of monitoring to ensure the consumer is working. I think we can open another issue for it?

louischan-oursky: https://github.com/authgear/authgear-server/issues/4025

Issue #3873, opened 2024-02-28
Add phone number to onboarding wizard


fungc-io:
## Problem

We would love to contact the signed up user to discuss their use cases.

## Solution

* In "Create your first project wizard", add an additional field for an optional phone number.

![SCR-20240228-jbj.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/29b84116-e1b2-4ff1-88cd-580a63fa257d/6dba3f14-8be2-499e-a8b1-fcfadf974df7?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi8yOWI4NDExNi1lMWIyLTRmZjEtODhjZC01ODBhNjNmYTI1N2QvNmRiYTNmMTQtOGJlMi00OTllLWE4YjEtZmNmYWRmOTc0ZGY3IiwiaWF0IjoxNzA5MjA0MjE0LCJleHAiOjMzMjc5NzY0MjE0fQ.jOWn8AgKHi4UzHlU7_eAJYPEBTGsUyeXPf6BZqWp12w "=1272x")

* Copywriting:

> **Would you mind to share us your phone number?**
>
> \\\[phone input\\\]

* There will be no validation on the phone number
* The country code is auto-selected based on IP.
* Add the phone number as a custom attribute in the `accounts` app, so it will be shown in the `user.updated` webhook for further integration

chpapa:
A few thoughts:

1. Maybe a shorter copywriting: "Would you like to discuss your needs over phone?"
2. As phone no is optional, let's not do validation? (So we don't block users who don't care)

fungc-io:
Ok, the description is edited

fungc-io:
@newman Seems the country code selection is not based on my IP address.

It's showing Afghanistan for in when i test it

fungc-io:
The location fix should be included in the next release, @louischan can you help review and merge in on Tuesday morning?

fungc-io:
@newman 

There is some problem with the phone input. I should be able to put in random number without validation.

Problems:

1. It's blocked by the custom attribute validation in the first attempt?
2. In the second attempt, it will say the project ID is already created. However, when i refresh the page, i will see the user does not own any project.

[...]

PR #4001, opened 2024-03-18
Accept phone number on create project

IniZio:
ref #3873

### DIFF
--- a/pkg/portal/graphql/app_mutation.go
+++ b/pkg/portal/graphql/app_mutation.go
@@ -15,6 +15,7 @@ import (
 	"github.com/authgear/authgear-server/pkg/lib/config/configsource"
 	"github.com/authgear/authgear-server/pkg/lib/tutorial"
 	"github.com/authgear/authgear-server/pkg/portal/appresource"
+	"github.com/authgear/authgear-server/pkg/portal/model"
 	"github.com/authgear/authgear-server/pkg/portal/session"
 	"github.com/authgear/authgear-server/pkg/util/graphqlutil"
 )
@@ -389,7 +390,7 @@ var _ = registerMutationField(
 		Resolve: func(p graphql.ResolveParams) (interface{}, error) {
 			input := p.Args["input"].(map[string]interface{})
 			appID := input["id"].(string)
-			phoneNumber := input["phoneNumber"].(*string)
+			phoneNumber := input["phoneNumber"].(string)
 
 			gqlCtx := GQLContext(p.Context)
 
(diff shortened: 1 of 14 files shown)


pkg/portal/graphql/app_mutation.go
```diff
@@ -385,6 +390,7 @@ var _ = registerMutationField(
 		Resolve: func(p graphql.ResolveParams) (interface{}, error) {
 			input := p.Args["input"].(map[string]interface{})
 			appID := input["id"].(string)
+			phoneNumber := input["phoneNumber"].(string)
```
louischan-oursky, line 393: I set the custom attribute to be of type Phone Number. This type will ensure the phone number is in E.164 format. But the phone number actually submitted to Admin API is just the national number (that is, `98765432` instead of `+85298765432`).

pkg/portal/model/onboard.go
```diff
@@ -0,0 +1,5 @@
+package model
+
+type OnboardEntry struct {
+	PhoneNumber string `json:"onboard_phone_number"`
```
louischan-oursky, line 4: Rename the json key to `onboarding_survey_phone_number`.

pkg/portal/graphql/app_mutation.go
IniZio: Updated to include dial code :pray:

portal/src/PhoneTextField.tsx
```diff
@@ -12,7 +12,7 @@ export interface PhoneTextFieldProps {
   pinnedList?: string[];
   allowlist?: string[];
   inputValue: string;
-  onChange: (validValue: string, inputValue: string) => void;
+  onChange: (validValue: string, inputValue: string, dialCode?: string) => void;
```
louischan-oursky, line 15: No need to add dialCode. `validValue` is already the E.164 value we want.

Revert the changes in this file.

portal/src/graphql/portal/CreateProjectScreen.tsx
```diff
@@ -106,10 +108,8 @@ function CreateProjectScreenContent(props: CreateProjectScreenContentProps) {
   );
 
   const onChangePhoneNumber = useCallback(
-    (_e, newValue) => {
-      if (newValue != null) {
-        setState((prev) => ({ ...prev, phoneNumber: newValue }));
-      }
+    (_e, phoneNumber: string, dialCode?: string) => {
```
louischan-oursky, line 111: The `_e` is actually the E.164 phone number we need. Refer to PhoneScreen to see PhoneScreen stores both values for different purpose.

Issue #4007, opened 2024-03-18
Copy writing updates for Groups


fungc-io:
## No Groups state

More explanation:

> Create groups to represent the types of users that access your applications. Users in a group share the same roles.
>
> For example, all users in "Marketing Manager" group will automatically possess the admin right of a marketing app

![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/87de207a-1668-447d-875e-da37141fbf67/c49861e6-3022-4d66-97dc-a69d98594241?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi84N2RlMjA3YS0xNjY4LTQ0N2QtODc1ZS1kYTM3MTQxZmJmNjcvYzQ5ODYxZTYtMzAyMi00ZDY2LTk3ZGMtYTY5ZDk4NTk0MjQxIiwiaWF0IjoxNzEwNzczNjY3LCJleHAiOjMzMjgxMzMzNjY3fQ.Nizbb2eJXR6bsF920W66UPrDa5BUaKFdqA_615HyNIA "=1322x")

## Group Key

Mention when group keys are used

> The Group Key is a unique identifier for this group when referenced from Admin API and Hooks

![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/32e8c567-ff44-4e96-8463-f156f147d838/3661c541-3c1f-4781-a5d1-dc6c01459e22?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi8zMmU4YzU2Ny1mZjQ0LTRlOTYtODQ2My1mMTU2ZjE0N2Q4MzgvMzY2MWM1NDEtM2MxZi00NzgxLWE1ZDEtZGM2YzAxNDU5ZTIyIiwiaWF0IjoxNzEwNzczNjY3LCJleHAiOjMzMjgxMzMzNjY3fQ.7PLyTeCMGZ5n9gY8RkvUlSQtGk-KAVOSqOyJILBWGtI "=418x")

## Add role to group button

[...]

PR #4012, opened 2024-03-19
Update copy writings of groups empty view

tung2744:
ref #4007


tung2744:
Merge directly as it is copy writing updates only

Issue #3981, opened 2024-03-14
Users - Roles Tab


Ngkaokis:
Should just reuse the component from roles and groups

Or should refactor those to be reusable if thy are not

- [ ] empty view
- [ ] list view
  - [ ] show groups of the role
    - [ ] role directly assigned to user should be one row
    - [ ] role associated with groups should be merged to one row
  - [ ] search box
  - [ ] add role button
- [ ] delete role dialog
- [ ] add role dialog

PR #4013, opened 2024-03-19
Add user detail roles tab

Ngkaokis:
ref #3981 

https://github.com/authgear/authgear-server/assets/63200569/47fa93cb-1e80-4ffb-a5fd-3ea85704cd9f


portal/src/components/roles-and-groups/list/common/ActionButtonCell.tsx
```diff
@@ -4,26 +4,36 @@ import ActionButton from "../../../../ActionButton";
...
         text={
-          <Text className={styles.actionButtonText} theme={themes.destructive}>
+          <Text
+            // TO BE CONFIRMED: How the themes color work for different states
+            className={cn(
+              styles.actionButtonText,
+              disabled && "text-[#BFBFC3]"
+            )}
+            theme={themes.destructive}
+          >
```
Ngkaokis, line 29: @tung2744 Not so sure the correct way to set the disabled color here. When to use this themes and when to not use it?

tung2744: It seems it should be
```
      <ActionButton
        text={text}
        styles={{
          label: { fontWeight: 600 },
          labelDisabled: { color: "#BFBFC3" },
        }}
        className={styles.actionButton}
        theme={themes.destructive}
        onClick={onClick}
        disabled={disabled}
      />
```

portal/src/components/roles-and-groups/list/common/RolesAndGroupsBaseList.tsx
```diff
@@ -50,6 +50,15 @@ function RolesAndGroupsBaseList<T>(
     emptyText,
   } = props;
 
+  // NOTE: Avoid DetailsList automatically take key column of item as key of react list, causing duplicated key error
+  // Ref https://github.com/microsoft/fluentui/blob/19195df9f17f287bf4c66dd30453dd0d0a0ced93/packages/react/src/components/DetailsList/DetailsList.base.tsx#L1499
+  const getKey = useCallback((item: any, index?: number) => {
+    const baseKey = index?.toString() ?? Math.random().toString();
```
tung2744, line 56: We should not generate random string as key, because that defeat the purpose of "key".
What about simply fallback to 0? (Because index should always exist?)

portal/src/components/roles-and-groups/list/UserDetailsScreenRoleListContainer.tsx
```diff
@@ -0,0 +1,186 @@
...
+            onChange={onChangeSearchKeyword}
+            onClear={onClearSearchKeyword}
+          />
+          <PrimaryButton
+            text={<FormattedMessage id="UserDetailsScreen.roles.add" />}
+            onClick={showAddRoleDialog}
+          />
+        </header>
+        <UserRolesList
+          className="flex-1 min-h-0"
```
tung2744, line 164: Same as the groups list in the previous pr,
1. Give it a minimum height of 200px to avoid it disappearing in short screens.
1. It should be `flex-1-0-auto` so that it will not be shorter than 200px.

portal/src/graphql/adminapi/UserDetailsScreen.tsx
```diff
@@ -424,10 +425,11 @@ const UserDetails: React.VFC<UserDetailsProps> = function UserDetails(
           />
         </PivotItem>
         <PivotItem
+          className={"h-full pt-8"}
```
tung2744, line 428: `height: x%` only works if the container has a explicitly specified height.
https://developer.mozilla.org/en-US/docs/Web/CSS/height#formal_definition

The container of this element doesn't have a explicitly specified height, so prefer not to use `h-full` here.

Same in L453 in this file.

portal/src/components/roles-and-groups/list/UserDetailsScreenRoleListContainer.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  const totalCount = rolesQueryData?.roles?.totalCount ?? 0;
+
+  if (totalCount === 0) {
+    return <RolesEmptyView />;
+  }
+
+  return (
+    <>
+      <section className={cn("flex flex-col h-full", className)}>
```
tung2744, line 149: Prefer not to use `h-full` here, as mentioned in https://github.com/authgear/authgear-server/pull/4013/files#r1531547090

portal/src/graphql/adminapi/UserDetailsScreen.tsx
Ngkaokis: a bit doubt on this statement `If the height of the containing block is not specified explicitly (i.e., it depends on content height)`. As the parent container of the element is a flex-child using `flex-1`, seems it's not depending on content height?

tung2744: I think "it depends on content height" is only one case. Height of `flex-1` is affected by the remaining spaces of the container, so I don't think it is "specified explicitly".
In fact, some old browsers such as safari 13 / 14 will not work with this style if I remember correctly (I seems working in the latest versions tho)

tung2744: I've searched a bit on this issue, and found some related discussions:
https://stackoverflow.com/a/61942083
https://stackoverflow.com/questions/33636796/chrome-safari-not-filling-100-height-of-flex-parent

And my understanding is:
- Originally, height: x% cannot be resolved correrctly if the container doesn't specify the height explicitly (height: 100px or something like that)
- But now, modern browsers start accepting `height: 100%` under a certain conditions (Such as using `align-items: stretch`).
- An always working solution is still using nested flex containers.

Ngkaokis:
@tung2744 Updated :adore2x:

Issue #3984, opened 2024-03-14
User Search - Support filters users by role key


Ngkaokis:
ref `pkg/admin/graphql/query.go`

- [ ] evaluate if the indices are used properly

tung2744:
Suspected query:

```
EXPLAIN ANALYZE SELECT * FROM _auth_user au
LEFT JOIN _auth_user_group aug ON (aug.user_id = au.id)
LEFT JOIN _auth_group user_group ON (user_group.id = aug.group_id)
LEFT JOIN _auth_group_role agr ON (agr.group_id = user_group.id)
LEFt JOIN _auth_user_role aur ON (aur.user_id = au.id)
LEFT JOIN _auth_role user_role ON (user_role.id = aug.group_id)
LEFT JOIN _auth_role group_role ON (group_role.id = agr.group_id)
WHERE au.app_id = 'my-app' AND (
  user_group.key ILIKE '%test%' OR user_group.name ILIKE '%test%' OR
  user_role.key ILIKE '%test%' OR user_role.name ILIKE '%test%' OR
  group_role.key ILIKE '%test' OR group_role.name ILIKE '%test%'
)
ORDER BY au.created_at DESC LIMIT 100;
```

Some findings:

* `_auth_user` is missing index on `app_id`, filtering with app_id result in sequential scan.
* `_auth_user` is missing index on `_created_at` and `_last_login_at`, which are the two columns used to sort users in the list user api.

tung2744:
* Tried to add ts_vector index in `_auth_group` and `_auth_role`, and filter with `@@` operator with `ts_query`. However, the index is not used.
* Tried to add gin index in `_auth_group.key` and `_auth_role.key` by converting them to arrays, but the index not used as well.

PR #4014, opened 2024-03-19
Support filter users by role keys

tung2744:
ref #3984

### DIFF
--- a/pkg/lib/authn/user/store.go
+++ b/pkg/lib/authn/user/store.go
@@ -10,6 +10,7 @@ import (
 	sq "github.com/Masterminds/squirrel"
 	"github.com/lib/pq"
 
+	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/lib/infra/db"
 	"github.com/authgear/authgear-server/pkg/lib/infra/db/appdb"
 	"github.com/authgear/authgear-server/pkg/util/clock"
@@ -34,6 +35,7 @@ type Store struct {
 	SQLBuilder  *appdb.SQLBuilderApp
 	SQLExecutor *appdb.SQLExecutor
 	Clock       clock.Clock
+	AppID       config.AppID
 }
 
 func (s *Store) Create(u *User) (err error) {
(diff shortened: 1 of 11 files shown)


pkg/lib/authn/user/store.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/Masterminds/squirrel"
+	sq "github.com/Masterminds/squirrel"
```
louischan-oursky, line 11: This line duplicates with line 10.

pkg/lib/authn/user/store.go
```diff
@@ -256,6 +289,8 @@ func (s *Store) QueryPage(listOption ListOptions, pageArgs graphqlutil.PageArgs)
 	if err != nil {
 		return nil, 0, err
 	}
+	stmt, args, err := query.ToSql()
+	fmt.Println("tung DEBUG SQL", stmt, args, err)
```
louischan-oursky: Remove debug log. Consider adding debug print to SQLExecutor. So that when LOG_LEVEL=debug, the SQL got printed when executed.

tung2744: I've removed the debug log in later commits. There might be too much noise when `LOG_LEVEL=debug` if we print all sqls? Prefer keep the current behavior or add another config for printing sqls.

louischan-oursky: OK.

pkg/lib/authn/user/store.go
tung2744: Removed, thanks!

pkg/lib/authn/user/store.go
```diff
@@ -244,10 +244,38 @@ func (s *Store) Count() (uint64, error) {
...
+			pq.Array(listOption.RoleKeys),
+		))
+		orCondititions = append(orCondititions, sq.Expr(
+			"group_role.key = ANY (?)",
+			pq.Array(listOption.RoleKeys),
+		))
+	}
+
+	if len(orCondititions) > 0 {
+		query = query.Where(orCondititions)
```
Ngkaokis, line 278: Actually running the sql gives multiple duplicated row :derpthink:

tung2744: Fixed in https://github.com/authgear/authgear-server/pull/4014/commits/b0387339dc44ba1a0cc065ef62d99af2db544e66

@louischan-oursky It seems the library we are using (squirrel) doesn't support CTE and using sub query in WHERE. https://github.com/Masterminds/squirrel/issues/299
So I've to write the query using SQL here :sadfrog:

louischan-oursky: Looks good to me.

Issue #3983, opened 2024-03-14
Users List - role and groups column


Ngkaokis:
- [X] Query `groups` and `effectiveRoles` in usersListQuery
- [X] Role: show first assigned role + the count of remaining role (include derived roles)
- [X] Group: show first assigned group + the count of remaining group
- [X] only show the columns when there are defined roles / groups

![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/eae5e53e-912d-4daa-b0be-f60db5af8519/98c14eff-a9e0-4c17-8f1b-d7b524858af6?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi9lYWU1ZTUzZS05MTJkLTRkYWEtYjBiZS1mNjBkYjVhZjg1MTkvOThjMTRlZmYtYTllMC00YzE3LThmMWItZDdiNTI0ODU4YWY2IiwiaWF0IjoxNzEwMzk3NDcyLCJleHAiOjMzMjgwOTU3NDcyfQ.YWhbDQ0rPZ92E0jKs6AmHYGntuaYbU7Gwst9QODBSB4 "=1344x")

PR #4026, opened 2024-03-20
Add columns roles and groups to users list

Ngkaokis:
ref #3983

### DIFF
--- a/pkg/lib/rolesgroups/store_group_user.go
+++ b/pkg/lib/rolesgroups/store_group_user.go
@@ -21,7 +21,8 @@ func (s *Store) ListGroupsByUserIDs(userIDs []string) (map[string][]*Group, erro
 	).
 		From(s.SQLBuilder.TableName("_auth_user_group"), "ug").
 		Join(s.SQLBuilder.TableName("_auth_group"), "g", "ug.group_id = g.id").
-		Where("ug.user_id = ANY (?)", pq.Array(userIDs))
+		Where("ug.user_id = ANY (?)", pq.Array(userIDs)).
+		OrderBy("ug.created_at")
 
 	return s.queryGroupsWithUserID(q)
 }
@@ -37,7 +38,8 @@ func (s *Store) ListGroupsByUserID(userID string) ([]*Group, error) {
 	).
 		From(s.SQLBuilder.TableName("_auth_user_group"), "ug").
 		Join(s.SQLBuilder.TableName("_auth_group"), "g", "ug.group_id = g.id").
-		Where("ug.user_id = ?", userID)
+		Where("ug.user_id = ?", userID).
+		OrderBy("ug.created_at")
 
 	return s.queryGroups(q)
 }
--- a/pkg/lib/rolesgroups/store_group_user.go
+++ b/pkg/lib/rolesgroups/store_group_user.go
@@ -28,12 +28,12 @@ func (s *Store) ListGroupsByUserIDs(userIDs []string) (map[string][]*Group, erro
 }
 
 func (s *Store) ListGroupsByUserID(userID string) ([]*Group, error) {
-  userGroups, err := s.ListGroupsByUserIDs([]string{userID})
-  if err != nil {
-    return nil, err
-  }
+	userGroups, err := s.ListGroupsByUserIDs([]string{userID})
+	if err != nil {
+		return nil, err
+	}
 
-  return userGroups[userID], nil
+	return userGroups[userID], nil
 }
 
 func (s *Store) ListUserIDsByGroupID(groupID string, pageArgs graphqlutil.PageArgs) ([]string, uint64, error) {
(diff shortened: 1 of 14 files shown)


pkg/lib/rolesgroups/store_group_user.go
```diff
@@ -37,7 +38,8 @@ func (s *Store) ListGroupsByUserID(userID string) ([]*Group, error) {
 	).
 		From(s.SQLBuilder.TableName("_auth_user_group"), "ug").
 		Join(s.SQLBuilder.TableName("_auth_group"), "g", "ug.group_id = g.id").
-		Where("ug.user_id = ?", userID)
+		Where("ug.user_id = ?", userID).
```
tung2744, line 41: Not related to this issue, but it seems this `ListGroupsByUserID` can simply call `ListGroupsByUserIDs` with a single user id instead of duplicating the implementation.

portal/src/components/roles-and-groups/list/common/ActionButtonCell.tsx
```diff
@@ -2,22 +2,26 @@ import React from "react";
...
 import ActionButton from "../../../../ActionButton";
 import styles from "./ActionButtonCell.module.css";
-import { Text } from "@fluentui/react";
+import { ITheme, Text } from "@fluentui/react";
 import { useSystemConfig } from "../../../../context/SystemConfigContext";
 
 interface ActionButtonCellProps {
   text: string;
   onClick?: (e: any) => void;
+  theme?: ITheme;
```
tung2744, line 11: Prefer to expose a prop `variant: "destructive" | "default"` instead of accepting a theme which is difficult to understand how to use it correctly.

tung2744:
I've rebased authgear/roles-and-groups-portal, please rebase the related prs

pkg/lib/rolesgroups/store_group_user.go
Ngkaokis: Actually yes, any concern for that? @louischan-oursky

louischan-oursky: No concern.

portal/src/components/roles-and-groups/list/common/ActionButtonCell.tsx
```diff
@@ -8,10 +8,11 @@ interface ActionButtonCellProps {
   text: string;
   onClick?: (e: any) => void;
   disabled?: boolean;
+  variant?: "destructive" | "actionButton";
```
tung2744, line 11: Name `actionButton` as `default`, because your component is already a `ActionButtonCell`.

portal/src/graphql/adminapi/UsersList.tsx
```diff
@@ -271,53 +326,78 @@ const UsersList: React.VFC<UsersListProps> = function UsersList(props) {
...
-                className={styles.actionButton}
-                theme={theme}
-                onClick={(event) => onUserActionClick(event, item)}
-                text={children}
-              />
-            </div>
+            <ActionButtonCell
+              variant={variant}
+              onClick={(event) => onUserActionClick(event, item)}
+              text={text}
```
tung2744, line 358: Extract it to new function instead of disable complexity linting.

portal/src/graphql/adminapi/UsersList.tsx
```diff
@@ -271,53 +326,78 @@ const UsersList: React.VFC<UsersListProps> = function UsersList(props) {
...
+          const text =
+            item.deleteAt != null
+              ? renderToString("UsersList.cancel-removal")
+              : item.isAnonymized
+              ? ""
+              : item.anonymizeAt != null
+              ? renderToString("UsersList.cancel-anonymization")
+              : item.isDisabled
+              ? renderToString("UsersList.reenable-user")
+              : renderToString("UsersList.disable-user");
```
tung2744, line 352: Actually `variant` and `text` are coupled to each other, they should be generated in the same logic.

```typescript
let variant
let text

if  (item.deleteAt != null) {
  variant = "default"
  text = renderToString("UsersList.cancel-removal")
} else if (item.anonymizeAt != null) {
  variant = "default"
  text = renderToString("UsersList.cancel-anonymization")
} else if () { ... }
```

Ngkaokis:
Updated :adore2x: @tung2744

portal/src/graphql/adminapi/UsersList.tsx
```diff
@@ -270,54 +325,103 @@ const UsersList: React.VFC<UsersListProps> = function UsersList(props) {
...
+        text = renderToString("UsersList.cancel-removal");
+      } else if (item.isDisabled) {
+        variant = "default";
+        text = renderToString("UsersList.reenable-user");
+      } else if (item.isAnonymized) {
+        variant = "destructive";
+        text = "";
+      } else if (item.anonymizeAt != null) {
+        variant = "destructive";
+        text = renderToString("UsersList.cancel-anonymization");
```
pkong-ds, line 346: @Ngkaokis Do you remember why you write L341 and L344 as 2 conditions?

Is there a possibility where `!isAnonymized && anonymizeAt != null`?

pkong-ds: Oh I see. When the anonymization is scheduled, but the scheduled time is yet to come

Issue #4016, opened 2024-03-19
Show Roles in Group listing


fungc-io:
## Problem

As a user, I would like to see the roles inside a group from the list easily

## Design

In the "Groups" list inside the user details, show the **role names** inside a group.

**Spec**

* The role names are shown, not the role keys
* Show the roles beyond the first 7 as "+N", e.g. if there are 14 roles under the group, show > "abc, def, ..., + 7".

![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/236ff1f2-3b87-42bb-a5ee-3fc549d8c94c/480dfd2d-ed19-4b8f-9a49-56f4d84f2906?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi8yMzZmZjFmMi0zYjg3LTQyYmItYTVlZS0zZmM1NDlkOGM5NGMvNDgwZGZkMmQtZWQxOS00YjhmLTlhNDktNTZmNGQ4NGYyOTA2IiwiaWF0IjoxNzEwOTExNzYzLCJleHAiOjMzMjgxNDcxNzYzfQ.beDtmGkyAEG4T3rqO7srdHXsN-b-EkDxL-9CAux0lKE "=3024x")

PR #4034, opened 2024-03-20
Display roles of group in user groups list

Ngkaokis:
ref #4016

image


portal/src/components/roles-and-groups/list/UserGroupsList.tsx
```diff
@@ -141,6 +161,22 @@ export const UserGroupsList: React.VFC<UserGroupsListProps> =
...
+                ? "-"
+                : item.roles.items
+                    ?.slice(0, 3)
+                    .map((item) => item.name)
+                    .join(", ");
+            const addtionalInfo =
+              item.roles.totalCount > 3 ? ` +${item.roles.totalCount - 3}` : "";
+            return (
+              <BaseCell>
+                <Text className="whitespace-normal text-inherit">{`${text}${addtionalInfo}`}</Text>
```
tung2744, line 176: `whitespace: normal` seems to be the default, why do we need it here?
And I think this cell should have same text style as other cells (TextCell), why we need `text-inherit`?

Ngkaokis: Seems the detailed list has injected styles into cell to make it not wrap.
For the color part, actually it's other cells not using `Text` but `div` for wrapping the text, let me update it.

tung2744: It seems we should not do wrapping in detailed list. If there is a very long role name, will it break the layout?

Ngkaokis: image

It will just make the row taller which look fine to me

tung2744: But it is unlimited, that means, if the text is long enough it can occupy the whole screen.
I also found the same problem in the description cell of roles and groups.
Please use line clamp to limit the maximum line to 2:
https://tailwindcss.com/docs/line-clamp#truncating-multi-line-text

Ngkaokis: Then seems like `+5` at last would seldom been displayed

But in terms of UX, I think user want to see something like `Human Resource, ..., +5` rather than `Human Resource , Next...`?

tung2744: You are right. What about something like this?
```
<p class="flex">
  <span class="line-clamp-2">Human Resource Manager, IT Admin</span>
  <span class="flex items-end">+5</span>
</p>
```

Ngkaokis: It looks ok when the line is clamped but strange if it's not. We can't determine how to position this `+5` 

image


image

Ngkaokis: Maybe I just use `line-clamp-2` first.
See if @fungc-io is ok with chop off the `+5` at last?

fungc-io: The "+5" is better following the last role name instead of showing on the side. Is that very hard to do?

tung2744: I think a simple approach would be to display the +X at a new line. i.e.

```
Human Resource Manager, HR
Admin, Some Admin
+3
```
```
Human Resource Manager, HR
Admin, Some Admin, Some other...
+3
```

IniZio: Maybe add a tooltip instead of +3?

Issue #3986, opened 2024-03-14
User Search - Roles and Groups Dropdown


Ngkaokis:
- [ ]  filter user by roles and groups
- [ ] clear result when search keyword is entered and vice versa
- [ ] hide it there is no roles / groups

PR #4035, opened 2024-03-20
Implement roles and groups filtering

tung2744:
ref #3986 
requires https://github.com/authgear/authgear-server/pull/4014

Screenshot 2024-03-20 at 6 57 00 PM
Screenshot 2024-03-20 at 6 56 33 PM

### DIFF
--- a/pkg/admin/graphql/query.go
+++ b/pkg/admin/graphql/query.go
@@ -50,6 +50,9 @@ var query = graphql.NewObject(graphql.ObjectConfig{
 				"groupKeys": &graphql.ArgumentConfig{
 					Type: graphql.NewList(graphql.NewNonNull(graphql.String)),
 				},
+				"roleKeys": &graphql.ArgumentConfig{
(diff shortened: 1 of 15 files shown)


Ngkaokis (commented):
I remember @fungc-io mentioned want the role/group filter auto expand to display the whole word?

image

tung2744:
I think there must be a maximum, maybe I set it to minimum 100px (by figma), maximum 300px (same as search bar).
And in mobile, I think it should always occupy half of the column so I think we shouldn't let it expand.

tung2744:
Here is an video demonstrating the behavior in different screen size:

https://github.com/authgear/authgear-server/assets/34942210/ec6b33df-9c30-429e-9c35-863903f57070

Ngkaokis:
Merge first. Can open another issue if there is other styling issue

fungc-io:
The filter box and responsive layout LGTM, 
thanks @tung2744 @Ngkaokis !

Issue #3969, opened 2024-03-13
Role and groups - Tag picker styles follow up


Ngkaokis:
[https://github.com/authgear/authgear-server/pull/3963#pullrequestreview-1933181055](https://github.com/authgear/authgear-server/pull/3963#pullrequestreview-1933181055)

Ngkaokis:
Tried to look into the callout props. Seems nothing to help adjusting the width. 

Any concern I go ahead to implement ResizeObserver? @tung2744

tung2744:
@Ngkaokis What about `onLayerMounted`?
If that doesn't work please use `ResizeObserver`.

Ngkaokis:
> What about onLayerMounted?

Seems work! Originally thought it's something only triggered on the very first place :derpthink:

PR #4038, opened 2024-03-21
Set tag picker callout width equals to tag picker

Ngkaokis:
ref #3969

https://github.com/authgear/authgear-server/assets/63200569/c1a1a264-f752-4404-8417-8c1193205a9f


portal/src/StyledTagPicker.tsx
```diff
@@ -96,6 +97,13 @@ const StyledTagPicker: React.VFC<StyledPickerProps> = function StyledTagPicker(
         }}
         pickerCalloutProps={{
           target: tagPickerRef,
+          calloutWidth: calloutWidth,
+          onLayerMounted: () => {
+            setCalloutWidth(tagPickerRef.current?.offsetWidth ?? 0);
+          },
```
tung2744, line 103: Use useCallback to wrap this callback.

Issue #3982, opened 2024-03-14
Users List - pin the first column on scroll


Ngkaokis:
![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/4261c645-56be-4210-9d5f-f03b31617d13/21ddb0b3-6868-4a9d-ba85-4e7bac8104c9?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi80MjYxYzY0NS01NmJlLTQyMTAtOWQ1Zi1mMDNiMzE2MTdkMTMvMjFkZGIwYjMtNjg2OC00YTlkLWJhODUtNGU3YmFjODEwNGM5IiwiaWF0IjoxNzEwMzk2OTUxLCJleHAiOjMzMjgwOTU2OTUxfQ.JiedrPR08Z_8mySioiaIUXOFZ6dgqG54B1pexfWMpUU "=1285x")

PR #4039, opened 2024-03-22
Pin the first column of Users List

Ngkaokis:
ref #3982

https://github.com/authgear/authgear-server/assets/63200569/b8a7229c-2b7a-467d-8298-612c50e8b3c7

### DIFF
--- a/portal/src/graphql/adminapi/UsersList.module.css
+++ b/portal/src/graphql/adminapi/UsersList.module.css
@@ -73,3 +73,11 @@
 .anonymizedUserLabel {
   font-style: italic;
 }
+
+.list div[role="columnheader"]:first-child,
+.list div[role="gridcell"]:first-child {
+  position: sticky;
+  left: 0;
+  background-color: white;
+  z-index: 1;
+}
--- a/portal/src/graphql/adminapi/UsersList.module.css
+++ b/portal/src/graphql/adminapi/UsersList.module.css
@@ -4,7 +4,7 @@
 }
 
 .listWrapper {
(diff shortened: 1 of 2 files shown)


tung2744:
What about vertical scrolling of the list?

Ngkaokis:
> What about vertical scrolling of the list?

Can see the demo video. I think just let it scroll as usual?

tung2744:
In your video, the scrollbar appears on the outer body. What I mean is the vertical scrolling "of the table".
But I've tested it and seems working too.

BTW, it seems listWrapper of UsersList is using `flex: 1 0 0%` therefore the scrollbar appeared on incorrect position.

Ngkaokis:
Find problem in mobile screen size. Need to think think how to display the first column :sosad:

 @fungc-io 


https://github.com/authgear/authgear-server/assets/63200569/d7d41168-1898-4ac2-a38d-e547da0e7e0d

tung2744:
I guess we should disable the pin in mobile.

fungc-io:
Right, we better disable the pin if the screen witdth is less than 2 columns.
i.e. if there is not enough space for 1st column + 1 more.

Ngkaokis:
> we better disable the pin if the screen witdth is less than 2 columns.
i.e. if there is not enough space for 1st column + 1 more.

I think we can't determine such situation :derpthink:
just disable it on mobile breakpoint should be enough

fungc-io:
> I think we can't determine such situation :derpthink:
> just disable it on mobile breakpoint should be enough

Yes, that works 👍

PR #4049, opened 2024-03-25
Immediate apply on destructive language change

IniZio:
ref #3627

### DIFF
--- a/portal/src/graphql/portal/ManageLanguageWidget.tsx
+++ b/portal/src/graphql/portal/ManageLanguageWidget.tsx
@@ -4,7 +4,6 @@ import React, {
   useMemo,
   useState,
   useEffect,
-  CSSProperties,
 } from "react";
 import cn from "classnames";
 import {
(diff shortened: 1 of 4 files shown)


portal/src/graphql/portal/ManageLanguageWidget.tsx
```diff
@@ -387,8 +435,15 @@ const ManageLanguageWidget: React.VFC<ManageLanguageWidgetProps> =
...
+        const style: CSSProperties | undefined = option?.disabled
+          ? {
+              fontStyle: "italic",
+              color: theme.semanticColors.disabledText,
+            }
+          : undefined;
+
         return (
-          <Text>
+          <Text style={style}>
```
louischan-oursky, line 446: `style` is the ordinary style attribute of a HTML element. Fluent UI supports `styles`. `styles` can be a function. You can receive `theme` from the arguments. Then you can save calling `useTheme()` manually.

Issue #4031, opened 2024-03-20
Support filtering by roles / groups using elasticsearch


tung2744:
TO BE FILLED

tung2744:
@fungc-io We have admin api logs in other mutations, but not in roles and groups mutations. Do we need to add them back?

fungc-io:
Oh, right, i forgot this part too. Yes, they are neccessary

PR #4051, opened 2024-03-25
Use elasticsearch to filter users by roles and groups

tung2744:
ref #4031 

To test the pr, you need to run this command to update the index:
```
go run ./cmd/authgear internal elasticsearch update-index
```

TODO:

- [x] Update tests

### DIFF
--- a/pkg/api/event/nonblocking/adminapi_mutation_anonymizeuser_executed.go
+++ b/pkg/api/event/nonblocking/adminapi_mutation_anonymizeuser_executed.go
@@ -38,11 +38,11 @@ func (e *AdminAPIMutationAnonymizeUserExecutedEventPayload) ForAudit() bool {
 }
 
 func (e *AdminAPIMutationAnonymizeUserExecutedEventPayload) RequireReindexUserIDs() []string {
-	return []string{}
+	return nil
 }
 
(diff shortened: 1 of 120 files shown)


pkg/api/event/nonblocking/adminapi_mutation_anonymizeuser_executed.go
```diff
@@ -37,8 +37,8 @@ func (e *AdminAPIMutationAnonymizeUserExecutedEventPayload) ForAudit() bool {
 	return true
 }
 
-func (e *AdminAPIMutationAnonymizeUserExecutedEventPayload) ReindexUserNeeded() bool {
-	return false
+func (e *AdminAPIMutationAnonymizeUserExecutedEventPayload) RequireReindexUserIDs() []string {
```
louischan-oursky, line 40: Returning nil is better, just to avoid allocation.

pkg/lib/elasticsearch/service.go
```diff
@@ -144,6 +145,7 @@ func (s *Service) QueryUser(
 	if err != nil {
 		return nil, nil, err
 	}
+	fmt.Println("bodyJSONBytes", string(bodyJSONBytes))
```
louischan-oursky, line 148: Remove this.

tung2744:
Updated with some new test cases.

Issue #4047, opened 2024-03-24
Users Login IDs missing in the user list


fungc-io:
In the recent update on the users list for "Roles and groups" in user list. The username, email, and phone gone missing.

![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/331ac3e3-de2d-48e4-8275-a0b2934b7dae/6e7a3e17-b567-41c3-96e3-4a445b6e5746?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi8zMzFhYzNlMy1kZTJkLTQ4ZTQtODI3NS1hMGIyOTM0YjdkYWUvNmU3YTNlMTctYjU2Ny00MWMzLTk2ZTMtNGE0NDViNmU1NzQ2IiwiaWF0IjoxNzExMzA3Mzg2LCJleHAiOjMzMjgxODY3Mzg2fQ.A7F7gdITZzVmoeSNtO81Xt5uAkkgcpbcAN6akIs4GPg "=1136x")

PR #4052, opened 2024-03-25
Pass missing column data to cell renderer

Ngkaokis:
ref #4047 

image

### DIFF
--- a/portal/src/graphql/adminapi/UsersList.tsx
+++ b/portal/src/graphql/adminapi/UsersList.tsx
@@ -368,7 +368,7 @@ const UsersList: React.VFC<UsersListProps> = function UsersList(props) {
     }
     return (
       <BaseCell>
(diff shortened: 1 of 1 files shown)


tung2744:
@Ngkaokis Looks like wrong issue too

portal/src/graphql/adminapi/UsersList.tsx
```diff
@@ -411,8 +411,9 @@ const UsersList: React.VFC<UsersListProps> = function UsersList(props) {
         case "roles": {
           return renderRoleCell(item);
         }
-        default:
-          return renderDefaultCell(item);
+        default: {
+          return renderDefaultCell(item, column);
```
tung2744, line 415: If column cannot be omitted, change the signature of `renderDefaultCell` to `renderDefaultCell(item: UserListItem, column: IColumn | undefined)` to force the caller to pass column to the function.

Issue #3992, opened 2024-03-14
LoginID normalizer should normalize phone number to e164 format


tung2744:
The current login id normalizer doesn't normalize phone numbers.

[https://github.com/authgear/authgear-server/blob/main/pkg/lib/authn/identity/loginid/normalizer.go#L24](https://github.com/authgear/authgear-server/blob/main/pkg/lib/authn/identity/loginid/normalizer.go#L24)

Currently, the http handlers will validate the phone number and ensure it is in e164 format, but we should also normalize it in normalizer and do not assume the input is always correct.

Should normalize it to e164 format.

Issue #4015, opened 2024-03-19
Lower phone number valid check criteria in the server


fungc-io:
## Problem

* There are 2 validation offered by the phone library:
  * isPossible: Length matched for a given country code, **ONLY length** is checked.
    * e.g. `+852 1xxx xxxx` is a **possible** number
  * isValid: The number is allocated/assigned in the numbering plan.
    * e.g. `+852 535x xxxx` is **invalid.** (See: [HK Numbering plan PDF](https://www.ofca.gov.hk/filemanager/ofca/tc/content_311/no_plan.pdf))
    * e.g. `+1(370)xxx-xxx` is **invalid** (See: [North American numbering plan](https://en.wikipedia.org/wiki/List_of_North_American_Numbering_Plan_area_codes))
* When testing the integration, the user may want to manually import a set of test data into the system. It's common to use the range of numbers of possible length but invalid during test. e.g. `+852 32xx xxxx` are possible but invalid numbers.
* This also requires us to update the library rather frequently. For instance in a recent incident, the numbering scheme in HK was updated so the authority is now issuing numbers starting with `4` but our definition was not up-to-date to include these numbers.

## Appetite

2 weeks

## Solution

* Add an option in the portal for the admin to choose if they want to accept invalid numbers
  * `[ ] Validate phone number using length and prefix information`
  * Default on for SaaS, but off for ONCE and other versions
* Handle the error reported by SMS gateway for invalid and blocked phone number.

PR #4062, opened 2024-03-26
Add phone normalizer

Ngkaokis:
ref #3992

### DIFF
--- a/pkg/lib/authn/identity/loginid/normalizer.go
+++ b/pkg/lib/authn/identity/loginid/normalizer.go
@@ -118,8 +118,7 @@ type PhoneNumberNormalizer struct {
 }
 
 func (n *PhoneNumberNormalizer) Normalize(loginID string) (string, error) {
-	phoneNumberParser := &phone.LegalParser{}
-	e164, err := phoneNumberParser.ParseInputPhoneNumber(loginID)
+	e164, err := phone.LegalParser.ParseInputPhoneNumber(loginID)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/lib/authn/identity/loginid/normalizer.go
+++ b/pkg/lib/authn/identity/loginid/normalizer.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/lib/config"
+	"github.com/authgear/authgear-server/pkg/util/phone"
 )
 
 type Normalizer interface {
@@ -31,6 +32,8 @@ func (f *NormalizerFactory) NormalizerWithLoginIDType(loginIDKeyType model.Login
 		return &UsernameNormalizer{
 			Config: f.Config.Types.Username,
 		}
+	case model.LoginIDKeyTypePhone:
+		return &PhoneNumberNormalizer{}
 	}
 
 	return &NullNormalizer{}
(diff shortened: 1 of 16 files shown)


pkg/lib/authn/identity/loginid/normalizer.go
```diff
@@ -111,6 +114,22 @@ func (n *UsernameNormalizer) ComputeUniqueKey(normalizeLoginID string) (string,
 	return normalizeLoginID, nil
 }
 
+type PhoneNumberNormalizer struct {
+}
+
+func (n *PhoneNumberNormalizer) Normalize(loginID string) (string, error) {
+	e164, err := phone.ParseCombinedPhoneNumber(loginID)
```
tung2744, line 121: ParseCombinedPhoneNumber will check if the phone number is valid, but I think as a normalizer, we don't need to check the validity of the input. So we can simply call `phonenumbers.Parse` and `phonenumbers.Format`.

louischan-oursky: I want to clarify the situation here.

In pkg/util/phone/phone.go,

- ParseCombinedPhoneNumber: Call phonenumbers.Parse and phonenumbers.IsValidNumber. So it accepts only valid phone numbers.
- ParseE164ToCallingCodeAndNumber: Same as ParseCombinedPhoneNumber
- Parse: Same as ParseCombinedPhoneNumber
- EnsureE164: Same as ParseCombinedPhoneNumber

So if we use any functions from pkg/util/phone, then only valid phone numbers are accepted.

Also in this PR, the normalizer used in stdattrs was switched to use the login ID normalizer. If we change the behavior of login ID normalizer, then the behavior in stdattrs will change as well.

I suggest we keep the current behavior of only accepting valid phone numbers, leaving the behavior change in https://github.com/authgear/authgear-server/issues/4015

tung2744: Actually, I think stdattrs should not use login id normalizer, because it is not necessary to be a login id?

And I think the goal of this issue is to normalize login id, but not to add additonal validations. So what I expect is not only `IsValid`, but `IsPossible` is also not checked in the normalizer, so it is not related to #4015.

louischan-oursky:
```golang
type Parser interface {
	ParseInputPhoneNumber(input string) (e164 string, err error)
	ParseCountryCallingCodeAndNationalNumber(countryCallingCode string, nationalNumber string) (e164 string, err error)
	SplitE164(input string) (countryCallingCode string, nationalNumber string, err error)
	CheckE164(s string) error
}

// What is Legal? libphonenumbers.Parse does not return err.
// Login ID uses legalParser
var LegalParser Parser
// Other places uses LegalAndValidParser.
// What is Legal and valid? libphonenumbers.Parse does not return err, and isPhoneValid returns true.
var LegalAndValidParser Parser
```

pkg/util/phone/legal_parser.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"regexp"
+	"strconv"
+
+	"github.com/nyaruka/phonenumbers"
+)
+
+// Login ID normalizer uses LegalParser
+// What is Legal? phonenumbers.Parse does not return err.
+type LegalParser struct{}
```
tung2744, line 12: I think we should instantiate instance of LegalParser and LegalAndValidParser, so when using them we don't need to create new instance.

And also, there should be an interface `Parser` which is implemented by LegalParser and LegalAndValidParser.

```golang
var LegalParser Parser = legalParser{}
var LegalAndValidParser Parser = legalAndValidParser{}
```

louischan-oursky: Also, no need to put LegalParser and LegalAndValidParser in the wire dependency graph. They are just utilities we can use directly without dependency injection.

louischan-oursky: Please tidy up the comment.

The doc of struct in Go should start with the struct name. So the doc should read

> LegalParser parses a legal phone number. A legal phone number is a phone number that passes phonenumbers.Parse().

pkg/util/phone/legal_and_valid_parser.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"regexp"
+	"strconv"
+
+	"github.com/nyaruka/phonenumbers"
+)
+
+// Other places except Login ID normalizer uses LegalAndValidParser.
+// What is Legal and Valid? phonenumbers.Parse does not return err, and phonenumbers.IsValidNumber returns true.
+type LegalAndValidParser struct{}
```
louischan-oursky, line 12: > LegalAndValidParser parses a legal and valid phone number. A legal and valid phone number is a phone number that passes phonenumbers.Parse() and phonenumbers.IsValidNumber().

Issue #4043, opened 2024-03-22
Add sensible rate limit for authflow per IP


fungc-io:

tung2744:
Maybe admin APIs should have a sensible rate limit per ip / per project too? @fungc-io

PR #4065, opened 2024-03-26
Implement RateLimitMiddleware

Ngkaokis:
ref #4043


tung2744:
I think we should also change `UsageLimitExceeded` error to return 429 (TooManyRequests) instead of 503 @louischan-oursky what do you think?

louischan-oursky:
UsageLimitExceeded is a bit different. The number of requests may not be too many, but the usage limit is low. 429 is more about the user sends too many requests to the server.

Issue #3718, opened 2024-01-26
Received WhatsApp OTP in Simplified Chinese when locale is zh-HK


fungc-io:
I set my ui_locales to zh-HK, the AuthUI is showing the correct language.

But the language is wrong in the whatsapp OTP message for verification:

![Image](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/60ef6f72-357d-4fcb-80aa-33ffb3fde8a7/97e7f71c-16d8-498f-bfed-c1b2218a7a22?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi82MGVmNmY3Mi0zNTdkLTRmY2ItODBhYS0zM2ZmYjNmZGU4YTcvOTdlN2Y3MWMtMTZkOC00OThmLWJmZWQtYzFiMjIxOGE3YTIyIiwiaWF0IjoxNzA5NzE2NDE2LCJleHAiOjMzMjgwMjc2NDE2fQ.VIdrDE2fLS4z8kx4hxV1IlpKxvz428CjzIQf-BtuzOk)

tung2744:
@fungc-io Tested in local, staging and production, but cannot reproduce the problem. Do you have a specific project which can reproduce it?

tung2744:
After further testing, if `ui_locales=zh-Hant`, and `zh-HK` `zh-TW` `zh-CN` are enabled for the project, the whatsapp message will be in simplified chinese.

However, if `ui_locales=zh-HK`, the problem doesn't happen.

fungc-io:
Thanks @tung2744 for your investigation

I've just tested it again on staging too and seems the problem is gone. Christina reported the same problem before, let me check with her which project was it.

Please put this on hold

tung2744:
I think it is the same problem as https://github.com/authgear/authgear-server/issues/3454

For now, if the frontend always use `zh-HK` instead of `zh-Hant`, the locale should be resolved correctly.

fungc-io:
🤔 right, it maybe related. Coz they reported the user' system was in "Traditional Chinese" it's a possibility that the device is declaring `zh-hant` in Accept-Language instead of zh-hant-HK/zh-hant-TW/zh-HK/zh-TW…

fungc-io:
[...]

Issue #3454, opened 2023-11-03
Script in UI locale resolves incorrectly


fungc-io:
- When ui_locale is zh-Hant, it is treated as zh-Hant-HK in the translation and zh-Hans-CN in the templates
- When ui_locale is zh-Hans, it is also treated as zh-Hant-HK in the translation and zh-Hans-CN in the templates

louischan-oursky:
https://go.dev/play/p/q_haFShsPgG

tung2744:
Interestingly, I got a different result:
https://goplay.tools/snippet/VQH4Qs4MXY4

What could be the difference?

tung2744:
Oh, it is the order of supported tag. If we remove `zh-TW` from supported tags (Which has `Exact` as confidence`, and put `zh-CN` before `zh-HK`, `zh-CN` will be selected instead of `zh-HK`.

louischan-oursky:
The doc of Match says when there is no match, it returns the first item in the supported tags.

tung2744:
If you put `en-US` as the first item, it won't be matched. The problem is the library didn't consider the confidence level if it is  not `Exact`. (`zh-CN` is `Low` when matched with `zh-Hant`, and `zh-HK` is `High`, but still `zh-CN` were selected)

louischan-oursky:
Maybe related? https://github.com/golang/text/blob/master/language/match.go#L106

PR #4067, opened 2024-03-26
Use a better matching algorithm to match languages

tung2744:
ref #3718 #3454 

Benchmark result

```
goos: darwin
goarch: amd64
pkg: github.com/authgear/authgear-server/pkg/util/intl
cpu: VirtualApple @ 2.50GHz
BenchmarkMatch-8       	   56115	     21034 ns/op
BenchmarkBestMatch-8   	   48363	     24730 ns/op
```

### DIFF
--- a/pkg/util/template/translation_test.go
+++ b/pkg/util/template/translation_test.go
@@ -433,10 +433,10 @@ func TestTranslationResource(t *testing.T) {
 				data, err = read(er)
 				So(err, ShouldBeNil)
 				So(data, ShouldEqual, compact(`{
-					"app.name": { "LanguageTag": "en", "Value": "en app.name in fs A" },
+					"app.name": { "LanguageTag": "zh-HK", "Value": "zh-HK app.name in fs A" },
 					"email.default.sender": { "LanguageTag": "jp", "Value": "no-reply+jp@app.com" },
 					"some-key-1": { "LanguageTag": "jp", "Value": "jp some-key-1 in fs C" },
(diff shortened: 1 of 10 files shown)


pkg/util/template/translation_test.go
```diff
@@ -433,10 +433,10 @@ func TestTranslationResource(t *testing.T) {
 				data, err = read(er)
 				So(err, ShouldBeNil)
 				So(data, ShouldEqual, compact(`{
-					"app.name": { "LanguageTag": "en", "Value": "en app.name in fs A" },
+					"app.name": { "LanguageTag": "zh-HK", "Value": "zh-HK app.name in fs A" },
 					"email.default.sender": { "LanguageTag": "jp", "Value": "no-reply+jp@app.com" },
 					"some-key-1": { "LanguageTag": "jp", "Value": "jp some-key-1 in fs C" },
-					"some-key-2": { "LanguageTag": "en", "Value": "en some-key-2 in fs A" }
+					"some-key-2": { "LanguageTag": "zh-HK", "Value": "zh-HK some-key-2 in fs A" }
```
tung2744, line 439: I'm not sure if this change is correct.
It seems default tag was set as `zh-HK` so the fallback should be `zh-HK`?

louischan-oursky: Correct the `jp` in this file to `ja`.

tung2744:
Updated, thanks!

Issue #4054, opened 2024-03-25
Error when creating user using email


Ngkaokis:
![Screenshot 2024-03-25 at 3.23.40 PM.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/2e51f4c6-3cdc-433a-8490-f8da5a7998cc/a4b2fd0e-4e67-417b-82c1-91036196fd68?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi8yZTUxZjRjNi0zY2RjLTQzM2EtODQ5MC1mOGRhNWE3OTk4Y2MvYTRiMmZkMGUtNGU2Ny00MTdiLTgyYzEtOTEwMzYxOTZmZDY4IiwiaWF0IjoxNzExMzU4NDI4LCJleHAiOjMzMjgxOTE4NDI4fQ.Vf5nU9lRK_jmIZCZslhiYKFOyoGKWYkHJYjzoW324Ic "=2555x")

PR #4068, opened 2024-03-27
Fix create user error

Ngkaokis:
ref #4054

### DIFF
--- a/portal/src/graphql/adminapi/AddUserScreen.tsx
+++ b/portal/src/graphql/adminapi/AddUserScreen.tsx
@@ -77,16 +77,33 @@ function isPasswordNeeded(
     return false;
   }
 
+  const filterAuthenticators = (allowedTypes: PrimaryAuthenticatorType[]) => {
+    return primaryAuthenticators.filter((authenticator) =>
+      allowedTypes.includes(authenticator)
+    );
+  };
+
+  let relatedAuthenticators: PrimaryAuthenticatorType[];
   switch (loginIdKeySelected) {
     case "email":
-      return !primaryAuthenticators.includes("oob_otp_email");
+      relatedAuthenticators = filterAuthenticators([
+        "oob_otp_email",
+        "password",
+      ]);
+      break;
     case "phone":
-      return !primaryAuthenticators.includes("oob_otp_sms");
+      relatedAuthenticators = filterAuthenticators(["oob_otp_sms", "password"]);
+      break;
     case "username":
-      return true;
+      relatedAuthenticators = filterAuthenticators(["password"]);
+      break;
     default:
-      return false;
+      relatedAuthenticators = filterAuthenticators([]);
   }
+
+  return (
+    relatedAuthenticators.length > 0 && relatedAuthenticators[0] === "password"
+  );
 }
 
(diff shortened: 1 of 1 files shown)


portal/src/graphql/adminapi/AddUserScreen.tsx
```diff
@@ -77,11 +77,30 @@ function isPasswordNeeded(
     return false;
   }
 
+  const enablePasswordField = (authenticator: PrimaryAuthenticatorType) => {
```
tung2744, line 80: Actually, should we always enable password field is password authenticator is enabled?
@fungc-io

fungc-io: I don't think the password is needed if it's not compulsory. But why was @Ngkaokis experiencing the error if pw is not needed?

Is this solved by changing the design to

> Email
> [Email input]
>
> Set password for this user?
> [Toggle: Yes/No - default "No"]
>
> <-- if set pw is yes -->
> Password
> [ Password input ]
> </ --end if -- >

Ngkaokis: It occurs when we enable both passwordless and password authenticator, and prioritize password authenticator first.

Front end side disable the password field as long as we enable passwordless authenticator.

But backend side raise error when password authenticator is the first one and no password is given.

I fixed it but syncing the password field disable logic as the backend one.

tung2744: So the criteria of enabled password field should be the first primary authenticator is "password", and not related to other authenticators?

Ngkaokis: There maybe a situation that

1. Passwordless by email
2. Password
3. Passwordless by phone

In such situation, create user with email doesn't need password but create with phone does need. 

So need to check the order as well

tung2744: The problem here is actually:
1. When create user by email, the possibly created primary authenticators are `["oob_otp_email", "password"]`. And in this case "password" is not the first primary authenticator.
2. When create user by phone, the possibly created primary authenticators are `["password", "oob_otp_sms"]`, so "password" becomes the first primary authenticator.

So I suggest to modify this part so that it is easier to understand the logic behind:

```typescript
fuction filterAuthenticators(allowedTypes: PrimaryAuthenticatorType[]):  PrimaryAuthenticatorType[] {
  // TODO: Filter primary authenticators by the provided `allowedTypes`
}

// ... In isPasswordNeeded
const relatedAuthenticators: PrimaryAuthenticatorType[]
switch (loginIdKeySelected) {
  case "email":
    relatedAuthenticators = filterAuthenticators(["oob_otp_email", "password", "passkey"])
  case "phone":
    relatedAuthenticators = filterAuthenticators(["oob_otp_sms", "password", "passkey"])
  case "username":
    relatedAuthenticators = filterAuthenticators(["password", "passkey"])
  default:
    relatedAuthenticators = filterAuthenticators([])
}

return relatedAuthenticators.length > 0 && relatedAuthenticators[0] === "password"
```

Issue #4057, opened 2024-03-25
"Create" button should be disabled when required fields are not filled


fungc-io:
When the group "Name" is not filled, the "Create" button should remain disabled.

[Screen Recording 2024-03-25 at 16.07.00.mov](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/557e8f78-3814-4cff-907b-2f219ac37d72/475f3d79-9762-4dc2-b237-4c8bd012b01a?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi81NTdlOGY3OC0zODE0LTRjZmYtOTA3Yi0yZjIxOWFjMzdkNzIvNDc1ZjNkNzktOTc2Mi00ZGMyLWIyMzctNGM4YmQwMTJiMDFhIiwiaWF0IjoxNzExMzgzMDY0LCJleHAiOjMzMjgxOTQzMDY0fQ.irCxlqh90-yddSJwBHTqyIs6XKNMGFzCnpC9HGFzIs4)

![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/d4739f64-1061-45fd-9b5c-ba63f10a5b11/539fa009-177e-4f51-82ba-b8ad7f37aeec?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi9kNDczOWY2NC0xMDYxLTQ1ZmQtOWI1Yy1iYTYzZjEwYTViMTEvNTM5ZmEwMDktMTc3ZS00ZjUxLTgyYmEtYjhhZDdmMzdhZWVjIiwiaWF0IjoxNzExMzgzMTE1LCJleHAiOjMzMjgxOTQzMTE1fQ.swoOyHllDVjTzHUqGHXMWaiD3kn_1YKfiX8jI63lZNU "=493x")

Same problem in "Roles"

[...]

PR #4077, opened 2024-03-28
Disable create button when required fields are not filled

Ngkaokis:
ref #4057

### DIFF
--- a/portal/src/graphql/adminapi/AddGroupScreen.tsx
+++ b/portal/src/graphql/adminapi/AddGroupScreen.tsx
@@ -188,6 +188,11 @@ const AddGroupScreen: React.VFC = function AddGroupScreen() {
     validate,
   });
 
+  const canSave = useMemo(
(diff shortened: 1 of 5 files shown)


portal/src/graphql/adminapi/AddGroupScreen.tsx
```diff
@@ -188,6 +188,11 @@ const AddGroupScreen: React.VFC = function AddGroupScreen() {
     validate,
   });
 
+  const canSave = useMemo(
+    () => form.state.groupName !== "",
```
tung2744, line 192: I think we need to check groupKey too, same as AddRoleScreen?

tung2744:
The edit forms should also be updated?

portal/src/graphql/adminapi/AddGroupScreen.tsx
Ngkaokis: groupKey is auto filled when it's empty

Issue #3918, opened 2024-03-04
Further improvement on copywriting


fungc-io:
Now the AuthUI v2 is available on production for testing, this issue is used for collecting feedback on the copywriting:

1. `Login to [Project] to continue to [Application]` is too convoluted
   1. suggested to change to : `Login to [Application]`

fungc-io:
@fungc-io to check the copywriting in Sign up too

fungc-io:
@newman,

checked and no need to change the message for sign up, keep:

| Sign up for {AppName} | 建立您的 {AppName} 帳戶 | 建立您的 {AppName} 帳號 |

For login:

* en: Login to {ClientName} || Login to {AppName}
* zh-hant: 登入至 {ClientName} || 登入至 {AppName}}

louischan-oursky:
I wonder why AppName is used in Sign up, while ApplicationName is used in Log in.

tung2744:
@fungc-io Does it means we always use project name instead of client name in all cases? (signup, login, signup login)

fungc-io:
No, i meant to use project name in signup and signup/login; and client name in login.

imo the user is login to the app and signup to the project… 🤔 

do you guys think it's simpler to just "signup/login to the project"?

louischan-oursky:
It is easier to explain if we stick with project name by default.

fungc-io:
all right, let's do that 🙌 @louischan @tung2744 @newman 

Signup, login and signup/login to project name

PR #4080, opened 2024-03-28
Shorten login page copywriting

IniZio:
ref #3918 

Login|Signup|Login+Signup
-|-|-
Screen Shot 2024-03-28 at 14 46 06|Screen Shot 2024-03-28 at 15 35 18|Screen Shot 2024-03-28 at 15 38 46


resources/authgear/templates/en/translation.json
```diff
@@ -848,7 +848,7 @@
   "v2-signup-login-title": "Welcome",
   "v2-signup-login-subtitle": "Sign up or Log in to {AppName}",
   "v2-login-title": "Welcome",
-  "v2-login-subtitle": "{ClientName, select, null {Login to {AppName}} other {Login to {AppName} to continue to {ClientName}}}",
+  "v2-login-subtitle": "{ClientName, select, null {Login to {AppName}} other {Login to {ClientName}}}",
```
tung2744, line 851: Can we change "v2-signup-login-subtitle" and "v2-signup-subtitle" so that it will use ClientName if given too?

Issue #4078, opened 2024-03-28
Wrong fill color for delete button in Roles and Groups


Ngkaokis:
Poblem

- The fill color code for the delete button should be #fbfcff.

Expectation

- The fill color code for the delete button should be #ffffff.

PR #4081, opened 2024-03-28
Set DefaultButton default bg color to white

Ngkaokis:
ref #4078

### DIFF
--- a/portal/src/DefaultButton.tsx
+++ b/portal/src/DefaultButton.tsx
@@ -1,8 +1,9 @@
-import React from "react";
+import React, { useMemo } from "react";
 import {
   IButtonProps,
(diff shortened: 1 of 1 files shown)


portal/src/DefaultButton.tsx
```diff
@@ -13,8 +14,16 @@ export interface DefaultButtonProps
...
   props: DefaultButtonProps
 ) {
+  const { styles, ...rest } = props;
+
+  const _styles: IButtonStyles = {
+    root: {
+      backgroundColor: "#ffffff",
+    },
+    ...styles,
+  };
```
tung2744, line 24: Use useMemo to memo the object.

PR #4082, opened 2024-03-28
Add audit logs for roles and groups mutations

tung2744:
ref #4044


pkg/api/event/nonblocking/adminapi_mutation_deletegroup_executed.go
```diff
@@ -2,14 +2,17 @@ package nonblocking
...
 )
 
 const (
 	AdminAPIMutationDeleteGroupExecuted event.Type = "admin_api.mutation.delete_group.executed"
 )
 
 type AdminAPIMutationDeleteGroupExecutedEventPayload struct {
-	AffectedUserIDs []string `json:"-"`
+	Group        model.Group `json:"group"`
+	GroupRoleIDs []string    `json:"role_ids"`
```
louischan-oursky, line 14: The json key should be `group_role_ids`.

pkg/api/event/nonblocking/adminapi_mutation_deleterole_executed.go
```diff
@@ -2,14 +2,18 @@ package nonblocking
...
 
 const (
 	AdminAPIMutationDeleteRoleExecuted event.Type = "admin_api.mutation.delete_role.executed"
 )
 
 type AdminAPIMutationDeleteRoleExecutedEventPayload struct {
-	AffectedUserIDs []string `json:"-"`
+	AffectedUserIDs []string   `json:"-"`
+	Role            model.Role `json:"role"`
+	RoleGroupIDs    []string   `json:"group_ids"`
```
louischan-oursky, line 15: The key should be `role_group_ids`.

pkg/api/event/nonblocking/adminapi_mutation_addgrouptoroles_executed.go
```diff
@@ -10,6 +10,8 @@ const (
 
 type AdminAPIMutationAddGroupToRolesExecutedEventPayload struct {
 	AffectedUserIDs []string `json:"-"`
+	GroupKey        string   `json:"group_key"`
+	RoleKeys        []string `json:"role_keys"`
```
louischan-oursky, line 14: I think we should include ids instead keys in the event. Using IDs make investigation easier. Also, the IDs can be used to build links that link back to the role / group details page.

tung2744:
Updated, thanks!

louischan-oursky (approved):

Issue #4058, opened 2024-03-25
All roles or groups should be shown when there is no keywords


fungc-io:
In "Add roles to group" in the roles or groups details, when the input is focused but nothing in the textbox,

* There should be a dropdown which shows all options in alphabetical order. (Or the first 100)
* The list is scrollable, height = 4 items
* This give hints to the user what to type here.

![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/3e133a7d-917b-43a4-a76e-11a122e446c9/8c79d514-9d3e-483e-ad9f-f8afb65a1ebc?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi8zZTEzM2E3ZC05MTdiLTQzYTQtYTc2ZS0xMWExMjJlNDQ2YzkvOGM3OWQ1MTQtOWQzZS00ODNlLWFkOWYtZjhhZmI2NWExZWJjIiwiaWF0IjoxNzExMzgzOTY5LCJleHAiOjMzMjgxOTQzOTY5fQ.3E_JWGlv12mgBZcf4W9b_x4LMYmeQlu\_-tTEOLCFIx0 "=575x")

PR #4086, opened 2024-03-28
Resolve suggestions on open tag picker dialog

Ngkaokis:
ref #4058

### DIFF
--- a/portal/src/components/roles-and-groups/dialog/AddUserRolesDialog.tsx
+++ b/portal/src/components/roles-and-groups/dialog/AddUserRolesDialog.tsx
@@ -68,7 +68,12 @@ export const AddUserRolesDialog: React.VFC<AddUserRolesDialogProps> =
         const selectedRoleIDs = new Set(
           selectedTags?.map((tag) => (tag as RoleTag).role.id)
         );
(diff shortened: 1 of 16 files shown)


portal/src/components/roles-and-groups/dialog/AddUserRolesDialog.tsx
```diff
@@ -68,7 +68,12 @@ export const AddUserRolesDialog: React.VFC<AddUserRolesDialogProps> =
         const selectedRoleIDs = new Set(
           selectedTags?.map((tag) => (tag as RoleTag).role.id)
         );
-        const result = await refetch({ searchKeyword: filter, pageSize: 100 });
+        const excludedIDs = new Set([...selectedRoleIDs, ...existingRoleIDs]);
+        const result = await refetch({
+          searchKeyword: filter,
+          excludedIDs: [...excludedIDs],
```
IniZio, line 74: Is being able to select a role already assigned to user due to group expected?

Ngkaokis: Yes, it doesn't have such restriction

PR #4087, opened 2024-04-02
Guard against removing other user's authenticator

IniZio:
ref DEV-1153

### DIFF
--- a/pkg/lib/interaction/nodes/remove_authenticator.go
+++ b/pkg/lib/interaction/nodes/remove_authenticator.go
@@ -1,6 +1,7 @@
 package nodes
 
 import (
+	"github.com/authgear/authgear-server/pkg/api"
 	"github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/lib/authn/authenticator"
 	"github.com/authgear/authgear-server/pkg/lib/interaction"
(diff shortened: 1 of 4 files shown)


pkg/lib/interaction/nodes/remove_authenticator.go
```diff
@@ -35,6 +36,14 @@ func (e *EdgeRemoveAuthenticator) Instantiate(ctx *interaction.Context, graph *i
 		return nil, err
 	}
 
+	if info.UserID != graph.MustGetUserID() {
```
tung2744, line 39: Prefer to check in authenticator.Service

IniZio: Authenticator facade is used by other transport e.g. userimport. In those cases deleting from other/no user id should be legitimate?

tung2744: I see.
BTW, is there same problem in delete identity?

IniZio: Found have similar issue in remove and update identity as well. Let me update that thanks!

pkg/lib/interaction/nodes/do_update_identity.go
```diff
@@ -46,6 +46,14 @@ func (n *NodeDoUpdateIdentity) Prepare(ctx *interaction.Context, graph *interact
...
 	return []interaction.Effect{
 		interaction.EffectRun(func(ctx *interaction.Context, graph *interaction.Graph, nodeIndex int) error {
+			if n.IdentityBeforeUpdate.UserID != graph.MustGetUserID() {
+				return api.NewInvariantViolated(
+					"IdentityNotBelongToUser",
+					"identity does not belong to the user",
+					nil,
+				)
+			}
+
```
louischan-oursky, line 56: This checking should be done in update_identity_end.go:22

louischan-oursky (approved):

PR #4094, opened 2024-04-03
Expose partial phone number

louischan-oursky:
ref https://github.com/authgear/authgear-server/issues/3873

### DIFF
--- a/portal/src/PhoneTextField.tsx
+++ b/portal/src/PhoneTextField.tsx
@@ -3,6 +3,11 @@ import React, { createRef } from "react";
 import { Text, Label } from "@fluentui/react";
 import intlTelInput from "intl-tel-input";
 import { SystemConfigContext } from "./context/SystemConfigContext";
+import {
+  cleanRawInputValue,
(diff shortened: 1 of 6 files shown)


louischan-oursky:
This PR remove the validation from the frontend. The partial value is passed to the server. But there are some caveats

1. The input does not forbid invalid characters. So we could save something like `+852abcd`
2. If the input is partial, then spaces are not trimmed, like `+852 9876 43`
3. If the input is valid, then the value is a E.164 without spaces.

fungc-io:
1. Use a pattern to only allow */#/+/digits
2. Trim in frontend before save

tung2744:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern

I found that pattern seems doesn't enforce the value to be matching the pattern, it only affect validation.

louischan-oursky:
Yes, I also realized that. Added a commit to fix that.

tung2744 (approved):

Issue #4066, opened 2024-03-26
Research for auto merge OAuth and Login ID approaches


fungc-io:
For #261 ([DEV-20](https://linear.app/authgear/issue/DEV-20/allow-configuration-of-handling-conflicted-oauth-and-email-address))

PR #4096, opened 2024-04-05
Specify account linking

tung2744:
ref #4066

### DIFF
--- a/docs/specs/account-linking.md
+++ b/docs/specs/account-linking.md
@@ -1,4 +1,4 @@
-# Identity Provider Account Linking
+# Account Linking
 
 - [Introduction](#introduction)
 - [Configuration](#configuration)
--- a/docs/specs/account-linking.md
+++ b/docs/specs/account-linking.md
@@ -92,7 +92,7 @@ Currently, only `oauth` is supported in account linking. However, account linkin
 
 We define linkings between the new oauth identity and any existing identities using the `oauth_claim` and `user_profile_attribute` fields.
 
-- `oauth_claim`: An object containing a json pointer, specified in `oauth_claim.pointer`, pointing to a claim of the incoming oauth user profile. Note that, for oidc compatible providers, this pointer is used to access value from the oidc claims, which is from the user info endpoint. For non-oidc compatible providers, please read the [SSO Providers](#todo) document for the corresponing logics authgear implemented to obtain a user profile from the provider.
+- `oauth_claim`: An object containing a json pointer, specified in `oauth_claim.pointer`, pointing to a claim of the incoming oauth user profile. Note that, for oidc compatible providers, this pointer is used to access value from the oidc claims, which is from the user info endpoint. For non-oidc compatible providers, please read the [SSO Providers](#./sso-providers.md) document for the corresponing logics authgear implemented to obtain a user profile from the provider.
 
 - `user_profile_attribute`: An object containing a json pointer, specified in `user_profile_attribute.pointer`, pointing to a value of the user profile of an existing authgear user, or the attribute of an existing authgear identity. For the meaning of attribute of authgear identity, please read the [Identity Attribute](#identity-attribute) section.
 
@@ -185,7 +185,7 @@ OAuth identities also contributes attributes of the user profile, but it is more
 
 ### The built-in oauth identity standard attributes
 
-For each supporteed oauth provider types, authgear has implemented a built-in standard attribute mappings. You could find the mappings of each provider in the [SSO Providers](#todo) document.
+For each supporteed oauth provider types, authgear has implemented a built-in standard attribute mappings. You could find the mappings of each provider in the [SSO Providers](#./sso-providers.md) document.
 
 ### Customizing the oauth identity attributes
 
--- a/docs/specs/account-linking.md
+++ b/docs/specs/account-linking.md
@@ -10,6 +10,9 @@
 - [Action on conflict](#action-on-conflict)
 - [Login and Link Flow](#login-and-link-flow)
 - [Q&A](#qa)
+  - [Why we need to login the user before linking the account?](#why-we-need-to-login-the-user-before-linking-the-account)
+  - [Why we need to continue the original signup flow instead of simply adding the oauth identity to the user?](#why-we-need-to-continue-the-original-signup-flow-instead-of-simply-adding-the-oauth-identity-to-the-user)
+  - [Why `on_conflict` is only available on `identification: oauth` but not other login ids such as `identification: email`?](#why-on_conflict-is-only-available-on-identification-oauth-but-not-other-login-ids-such-as-identification-email)
 - [References](#references)
 
 ## Introduction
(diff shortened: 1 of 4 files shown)


tung2744:
@louischan-oursky Updated according to our latest discussion, thanks!

docs/specs/idp-account-linking.md
```diff
@@ -12,47 +12,60 @@ Here is an example of the configuration:
...
+  on_conflict:
+    account_linking:
+      oauth: "merge"
+  oauth:
+    providers:
+      - alias: google
+        client_id: exampleclientid
+        type: google
+        account_linking:
+          enabled: false
```
louischan-oursky, line 24: What does `enabled: false` mean? Remember we have to be back-forward compatible. That is, the default config should match the current behavior, that is, error in case of conflict.

If `enabled: false` means `account_linking.oauth=error`, it could be confusing.

docs/specs/idp-account-linking.md
```diff
@@ -12,47 +12,60 @@ Here is an example of the configuration:
...
+
+- The `identity.on_conflict.account_linking.oauth` config was added to specify the behavior if an oauth account linking was triggered. The possible values are:
+  - `"error"`: Reject the signup with an error. This is the default.
+  - `"merge"`: Merge the new oauth identity to the existing user after an authentication process.
 
 ## Trigger conditions and actions
 
 - Whenever an oauth identity is used, and there is no existing user for that identity.
 
   1. If `account_linking.enabled` is not `true`, do nothing. Else,
```
louischan-oursky, line 45: Step 1 is not backward compatible. It is because account_link.enabled is false by default. The backward compatible behavior should be `"error"`.

docs/specs/idp-account-linking.md
```diff
@@ -12,47 +12,60 @@ Here is an example of the configuration:
...
-    - alias: google
-      client_id: exampleclientid
-      type: google
-      account_linking:
-        enabled: true
-        idp_claim_key: "email"
-        match_against_claim_key: "email"
+  on_conflict:
+    account_linking:
+      oauth: "merge"
```
louischan-oursky, line 17: I guess the key `oauth` means From OAuth. That is, this config is about the scenario that the triggering identity is an OAuth Identity.

This config, I believe, can handle the following cases

1. Triggering=OAuth, existing=OAuth
2. Triggering=OAuth, existing=Login ID

Then, I wonder why do we need to separate the config of From OAuth and From Login ID, because if we did not make the distinction, the cases would be

1. Triggering=OAuth, existing=OAuth
2. Triggering=OAuth, existing=Login ID
3. Triggering=Login ID, existing=OAuth (The handling of this case should be the same as case 1)

docs/specs/idp-account-linking.md
tung2744: @louischan-oursky `enabled: false` means the oauth account will never be considered as matching an existing identity.

So I guess the default should be `enabled: true` with `email` for some of the provider, but some of them can still remain `enabled: false` as default such as wechat. Maybe let me list out the appropriate default values.

docs/specs/idp-account-linking.md
tung2744: For backward compatibility, modify the default value as mentioned in https://github.com/authgear/authgear-server/pull/4096#discussion_r1555413513

docs/specs/idp-account-linking.md
tung2744: > Triggering=Login ID, existing=OAuth (The handling of this case should be the same as case 1)

I doubt if it is the case. We probably want to block and ask the user to login with oauth if this is this case, while we want to automatically merge accounts in case 1 and 2.

tung2744: Proposed change:

```
identity:
  on_conflict:
    signup:
      oauth: "merge" # "error" | "merge"
      login_id: "error" # "error" is the only supported value at the moment.
```

docs/specs/idp-account-linking.md
tung2744:
```
link_by_claim: "email" # nullable
```

tung2744:
- Renamed some config keys
- Use pointer to point to a claim
- Changed default `link_by.pointer` to `/email`
- Added more options of `identity.on_conflict.signup` and documented their corresponding behavior

docs/specs/idp-account-linking.md
```diff
@@ -12,47 +12,74 @@ Here is an example of the configuration:
...
+    - If `identity.on_conflict.signup=create_new_account`, create a new user with that oauth identity.
+    - If `identity.on_conflict.signup=hook`, trigger a hook and use the result to determine what to do.
+  - If there is no match, create a new user.
 
-- Whenever a new user create an identity during signup.
-  1. For every `identity.providers` with `account_linking.enabled` equal to `true`, do:
-  - 1.1. Use `account_linking.match_against_claim_key` to obtain a value from the new user. If the value is empty, do nothing.
-  - 1.2. Use `account_linking.idp_claim_key` to query users with an oauth identity of the current provider matching the value in 1.1.
-  2. (TBC) If there is at least one user found in 1, block the login and ask user to signin with with the found identity.
+- Whenever a new user creates an new identity during signup.
```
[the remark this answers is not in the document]
tung2744: > How is this section different from the section starting from line 46?

Basically they are same, this spec separated the two case to show that this steps can handle these two cases.

IniZio:
Looks like we have accounting_linking design in login flow, need to update as well?

Can find `account_linking:` in https://github.com/authgear/authgear-server/blob/main/docs/specs/authentication-flow.md#loginflow

tung2744:
I think I will remove that from the current authflow spec, because it differs from the current spec.

tung2744:
Some updates was made to the spec:

- Refined the `link_by` section:
    - Separated the config for matching existing authgear accounts and the external idp claims.
        - Use `identity` to match existing authgear users.
        - Use `idp_claim` to match a claim inside the external idp user info.
    - Removed the `pointer` key.
    - Document the default (Or auto) behavior.
- Added the Login and Link Flow section, which describes what exactly the "login_and_link" action will do.
- Added an Q&A section, which written down some questions and answers we've discussed in meetings.

Maybe @chpapa can also take a look if you got time, thanks!

tung2744:
Further updated the provider configs.

docs/specs/idp-account-linking.md
```diff
@@ -0,0 +1,528 @@
...
+        - authentication: primary_password
+          steps:
+            - type: create_authenticator
+              one_of:
+                - authentication: secondary_totp
+                  steps:
+                    - type: view_recovery_code
+```
+
+7. As the next step of the original flow is create primary password authenticator, the user will need to created primary authenticator.
```
louischan-oursky, line 455: to create

docs/specs/idp-account-linking.md
```diff
@@ -0,0 +1,528 @@
...
+
+- User A
+  - Email Identity: a@example.com
+  - OAuth Identity: Google (email:a@example.com)
+  - Primary Password Authenticator
+  - Secondary TOTP Authenticator
+
+## Q&A
+
+- Why we need to login the user before linking the account?
```
louischan-oursky, line 470: Use ### for each question. And use a paragraph for each answer.

```
### Why ...

Answer in a paragraph.
```

Remember to re-generate the TOC.

docs/specs/authentication-flow.md
```diff
@@ -275,34 +275,6 @@ login_flows:
   # If the condition does not hold, this step is no-op.
   - type: change_password
     target_step: step1
-
```
louischan-oursky, line 278: We still have some account linking content in this document. Please remove them. Also add a backlink to account-linking.md to tell the reader to find the new document.

tung2744:
Updated again for some config renaming and also documented account linking of login ids, as discussed in the product meeting.

tung2744:
Added a spec to document the current supported sso providers

docs/specs/account-linking.md
```diff
@@ -0,0 +1,421 @@
...
+
+```yaml
+identity:
+  oauth:
+    providers:
+      - alias: adfs
+        client_id: exampleclientid
+        type: adfs
+        user_profile_mapping:
+          - from_claim:
```
louischan-oursky, line 202: It should be `oauth_claim`

docs/specs/account-linking.md
```diff
@@ -0,0 +1,421 @@
...
+              pointer: "/primary_phone"
+            to_attribute:
```
louischan-oursky, line 204: It should be `user_profile_attribute`.

docs/specs/sso-providers.md
```diff
@@ -0,0 +1,216 @@
...
+
+The facebook user profile is obtained by this api:
+https://graph.facebook.com/v11.0/me?fields=id,email,first_name,last_name,middle_name,name,name_format,picture,short_name
+
+Then, we will construct a user profile from the response by the following process:
+
+1. Map the following fields from the response into the user profile.
+
+- From `email` to `email`
+- From `first_name` to `first_name`
```
louischan-oursky, line 91: to `given_name`.

docs/specs/sso-providers.md
```diff
@@ -0,0 +1,216 @@
...
+- From `last_name` to `last_name`
```
louischan-oursky, line 92: to `family_name`.

docs/specs/account-linking.md
chpapa: actually `user_profile` might be good enough? As we defined `user_profile` = standard attributes + custom attributes elsewhere already.

Issue #4019, opened 2024-03-19
Enable `X-Frame-Options Header` in portal


fungc-io:

IniZio:
Should separate between portal and authui

PR #4097, opened 2024-04-05
Add x-frame-options header middleware

IniZio:
ref #4019


pkg/lib/config/ui.go
```diff
@@ -43,6 +44,8 @@ type UIConfig struct {
 	Implementation UIImplementation `json:"implementation,omitempty"`
 	// ForgotPassword is the config for the default auth ui
 	ForgotPassword *UIForgotPasswordConfig `json:"forgot_password,omitempty"`
+	// IframeDisabled disallows embedding in iframe
+	IframeDisabled bool `json:"iframe_disabled,omitempty"`
```
louischan-oursky, line 48: After reading https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors, we see that they are relevant.

In pkg/lib/web/csp.go, we see that we automatically put custom UI URI origin in the list of frame ancestors. We also see that we have been outputing `frame-ancestors 'none'` already.

Therefore, we should output the relevant headers in a SINGLE middleware. So in this PR, we actually do not need to introduce a new config. We just need to output `x-frame-options: DENY` when we output `frame-ancestors 'none'`. The changes of this PR should be very simple. Please add tests.

louischan-oursky (approved):

louischan-oursky:
I added a few commit to fix the portal still have no x-frame-options: DENY being included in the response.

PR #4101, opened 2024-04-08
Setup OAuth proxy for e2e test

IniZio:
ref #3773

dependencies:
- [x] #3930


IniZio:
Will add redirect step to allow testing more oaouth cases in future e.g. oauth error.

Example config:
```yaml
- action: oauth_redirect
  to: data.result.url
  redirect_uri: http://mock
  output:
    result: |
      {
        "code": "[[string]]"
      }

- action: input 
  input: |
    {
      "code": {{ prevstep.output.result.code }}
    }
```

louischan-oursky:
You can rebase once first.

IniZio:
Updated to run oauth redirect

pkg/util/httputil/client.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"net/http"
+	"net/url"
+
+	"os"
+
+	"github.com/authgear/authgear-server/pkg/lib/config"
```
louischan-oursky, line 11: util cannot import lib. Create a new struct in this package instead.

pkg/util/httputil/client.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func NewHTTPClient(env *config.EnvironmentConfig) *http.Client {
+	client := http.DefaultClient
```
louischan-oursky, line 15: Do not modify DefaultClient. According to https://pkg.go.dev/net/http#Client a zero value of Client is good enough.

pkg/lib/authn/sso/access_token.go
```diff
@@ -78,6 +78,7 @@ func (r AccessTokenResp) TokenType() string {
 }
 
 func fetchAccessTokenResp(
+	client *http.Client,
```
louischan-oursky, line 81: Make a separate commit for switching to non DefaultClient.

pkg/lib/config/environment.go
```diff
@@ -50,6 +50,10 @@ type EnvironmentConfig struct {
 	DevMode DevMode `envconfig:"DEV_MODE" default:"false"`
 	// LogLevel sets the global log level
 	LogLevel string `envconfig:"LOG_LEVEL" default:"warn"`
+	// HTTPProxy sets the HTTP proxy
+	HTTPProxy string `envconfig:"HTTP_PROXY"`
+	// SSLCertFile sets additional SSL certificate file
+	SSLCertFile string `envconfig:"SSL_CERT_FILE"`
```
louischan-oursky, line 56: The name of these variables is too general, making people think Authgear will support them. But in fact, they are used in e2e test only. I suggest we make the authgear command to take 2 extra CLI options, `--e2e-http-proxy-url` and `--e2e-tls-ca-cert-path`. When the CLI options are given, then the HTTP client used by OAuth will be configured accordingly.

pkg/util/httputil/deps.go
```diff
@@ -4,6 +4,7 @@ import "github.com/google/wire"
 
 var DependencySet = wire.NewSet(
 	NewJSONResponseWriterLogger,
+	NewHTTPClient,
```
louischan-oursky, line 7: Wire inject by type. Adding NewHTTPClient will make it available in the graph, other components can then use it. But it is not our intention. So add a new type `type OAuthHTTPClient *http.Client` in `pkg/lib/authn/sso`. Add `NewOAuthHTTPClient` in `pkg/lib/authn/sso/deps.go`. Make OAuthProviderFactory to use OAuthHTTPClient instead of *http.Client.

louischan-oursky (changes requested):

pkg/util/httputil/client.go
IniZio: Moved http client inside `authn/sso`

pkg/util/httputil/deps.go
IniZio: Think this will make `OAuthHTTPClient` unable to use methods of `*http.Client` 

If I use anonymous field though, I won't be able to intercept it with `gock.InterceptClient`.

e2e/pkg/testrunner/testcase.go
```diff
@@ -249,13 +249,19 @@ func execTemplate(cmd *End2EndCmd, prev *StepResult, content string) (string, er
 	tmpl := texttemplate.New("")
 	tmpl.Funcs(makeTemplateFuncMap(cmd))
 
-	_, err := tmpl.Parse(content)
+	// Treat ${{ as {{ for oneline templates
+	_, err := tmpl.Parse(strings.ReplaceAll(content, "${{", "{{"))
```
louischan-oursky, line 253: Do not add this.

Instead, we can write

```yaml
to: |-
  {{ .some.value }}
```

pkg/util/httputil/deps.go
IniZio: Updated to use custom struct for sso, thanks!

louischan-oursky (approved):

PR #4117, opened 2024-04-10
Example testcase for forgot password

IniZio:


IniZio:
Used as demo, will organize with other test cases later.

Issue #4121, opened 2024-04-12
Respect flow group allowlist for default UI


IniZio:
* Include computed group struct in uiinfo or oauthsession data
* Update all webapp handlers' \`flowName := "default"\` to use given group struct

Issue #4120, opened 2024-04-12
Add config parsing + validation for authflow selection


IniZio:
* Validate groups in allowlist must exist in `ui.authentication_flows`
* All flow types are required in each \`ui.authentication.groups\`

PR #4129, opened 2024-04-16
Add flow group configuration for default UI

IniZio:
ref #4120 / DEV-1215
ref #4121 / DEV-1216


pkg/lib/config/config.go
```diff
@@ -327,6 +330,33 @@ func (c *AppConfig) validateLockout(ctx *validation.Context) {
...
+		}
+
+		groupNames[group.Name] = struct{}{}
+	}
+
+	// Ensure client's group allowlist is valid
+	for i, client := range c.OAuth.Clients {
+		if client.AuthenticationFlowGroupAllowlist == nil {
+			continue
+		}
```
louischan-oursky, line 381: This if condition is redundant. `range nil` means the loop has no iterations.

pkg/lib/config/ui.go
```diff
@@ -43,6 +44,14 @@ type UIConfig struct {
...
 	// ForgotPassword is the config for the default auth ui
 	ForgotPassword *UIForgotPasswordConfig `json:"forgot_password,omitempty"`
+	// AuthenticationFlow configures ui behaviour of authentication flow under default auth ui
+	AuthenticationFlow *UIAuthenticationFlowConfig `json:"authentication_flow,omitempty"`
+}
+
+func (c *UIConfig) SetDefaults() {
+	if c.AuthenticationFlow == nil {
+		c.AuthenticationFlow = &UIAuthenticationFlowConfig{}
+	}
```
louischan-oursky: SetDefaults is only necessary when you want to assign values to strings, ints, or booleans. Pointer to struct will be populated by `PopulateDefaultValues`. So remove this SetDefaults()

pkg/lib/config/config_test.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func TestAppConfig(t *testing.T) {
-	Convey("AppConfig", t, func() {
+	FocusConvey("AppConfig", t, func() {
```
louischan-oursky, line 17: Remove Focus.

pkg/lib/config/config_test.go
```diff
@@ -88,7 +88,7 @@ identity:
 			So(cfg, ShouldResemble, cfg2)
 		})
 
-		Convey("parse validation", func() {
+		FocusConvey("parse validation", func() {
```
louischan-oursky, line 91: Remove Focus.

pkg/lib/config/config_test.go
```diff
@@ -111,7 +111,7 @@ identity:
 					panic(err)
 				}
 
-				Convey(testCase.Name, func() {
+				FocusConvey(testCase.Name, func() {
```
louischan-oursky, line 114: Remove Focus.

pkg/lib/config/oauth.go
```diff
@@ -115,16 +115,35 @@ var _ = Schema.Add("AuthenticationFlowAllowlist", `
...
+			"type": "string",
+			"enum": [
+				"signup",
+				"promote",
+				"login",
+				"signup_login",
+				"reauth",
+				"account_recovery"
+			],
+			"minLength": 1
```
louischan-oursky, line 139: This `minLength` is redundant. We have `enum` already.

pkg/lib/config/oauth.go
```diff
@@ -212,17 +231,44 @@ type OAuthClientConfig struct {
...
+const (
+	AuthenticationFlowAllowlistFlowTypeSignup          AuthenticationFlowAllowlistFlowType = "signup"
+	AuthenticationFlowAllowlistFlowTypePromote         AuthenticationFlowAllowlistFlowType = "promote"
+	AuthenticationFlowAllowlistFlowTypeLogin           AuthenticationFlowAllowlistFlowType = "login"
+	AuthenticationFlowAllowlistFlowTypeSignupLogin     AuthenticationFlowAllowlistFlowType = "signup_login"
+	AuthenticationFlowAllowlistFlowTypeReauth          AuthenticationFlowAllowlistFlowType = "reauth"
+	AuthenticationFlowAllowlistFlowTypeAccountRecovery AuthenticationFlowAllowlistFlowType = "account_recovery"
+)
+
+func (flow AuthenticationFlowAllowlistFlow) Compare(other AuthenticationFlowAllowlistFlow) bool {
```
louischan-oursky, line 270: We do not need this method. Struct in Golang is equal to each other if all fields are equal.

pkg/lib/config/ui.go
```diff
@@ -45,7 +45,7 @@ type UIConfig struct {
 	// ForgotPassword is the config for the default auth ui
 	ForgotPassword *UIForgotPasswordConfig `json:"forgot_password,omitempty"`
 	// AuthenticationFlow configures ui behaviour of authentication flow under default auth ui
-	AuthenticationFlow *UIAuthenticationFlowConfig `json:"authentication_flow,omitempty"`
+	AuthenticationFlow UIAuthenticationFlowConfig `json:"authentication_flow,omitempty"`
```
louischan-oursky, line 48: It should be a pointer. Any reason for this change?

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -381,6 +382,11 @@ func (h *AuthorizationHandler) doHandle(
 		return nil, err
 	}
 
+	err = h.validateUIImplementation(client, r)
```
louischan-oursky, line 385: This validation should not be performed here. Instead, it should be performed in pkg/auth/handler/api/authenticationflow_v1_create.go and pkg/auth/handler/webapp/authflow_controller.go

In the two places I mention, OAuthSession is read. OAuthSession may now contain x_authentication_flow_group which be validated.

pkg/auth/handler/webapp/authflow_forgot_password.go
```diff
@@ -226,7 +225,6 @@ func (h *AuthflowForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http
 
 	h.Controller.HandleStartOfFlow(w, r, webapp.SessionOptions{}, authflow.FlowReference{
 		Type: authflow.FlowTypeAccountRecovery,
-		Name: flowName,
```
louischan-oursky, line 229: To avoid confusion, let's change all argument type from FlowReference to FlowType. It is because the name is now derived by the controller. It eliminates the possibility that someone later on passes a name and thinks name is not respected.

louischan-oursky (commented):
1. Change the spec and the implementation to use `{ type, name }` in defining a group. This saves us from adding a new field when we introduce a new flow type in the future.
2. The group allow list should be a list of `{ name }` struct, not a list of string.

pkg/lib/oauth/handler/handler_authz.go
IniZio: Actually authflow controller already has the checking as well. I added this extra logic so that sdk will receive this error code before webapp is opened. 

Maybe I remove this extra checking instead then.

pkg/lib/config/ui.go
```diff
@@ -130,3 +133,67 @@ func (c *UIForgotPasswordConfig) SetDefaults() {
...
+var _ = Schema.Add("UIAuthenticationFlowGroup", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"required": [
+		"name",
+		"flows"
+	],
+	"properties": {
+		"name": { "type": "string", "minLength": 1 },
```
louischan-oursky, line 156: Use the JSON schema AuthenticationFlowObjectName for name.

louischan-oursky:
@IniZio I added more failing tests and added a fix.

louischan-oursky (approved):

Issue #4131, opened 2024-04-16
Translation for major East/South-East Asian and European languages


fungc-io:
## Goal

To provide basic support in the AuthUI for major languages in East/South-East Asian and European languages. So the languages can be used out-of-the-box instead of users' need to translate the file manually.

## Approach

Use Claude to Translate the transtion.json file.

The model I used was `claude-3-haiku-20240307`

Prompt I tried:

```plaintext
- translate the values in the JSON from English into {the language}.
- Don't translate the phrase "Passkey", keep it as is. 
- Return only the JSON and nothing else
```

Due to token limits, the file should be translated 60 lines at a time.

## Appetite

2 days

## East/South-East Asian

* Chinese (Hong Kong)
* Chinese (Taiwan)
* Chinese (China)
* Korean
* Japanese
* Vietnamese
* Thai
* Malay
* Filipino (Tagalog)
* Indonesian

## European

* English
* French
* Spanish (Spain)
  * Variant: Spanish (Latin America)
* Italian
* Portuguese (Portugual)
  * Variant: Portuguese (Brazil)
* German
* Polish
* Dutch
* Greek

fungc-io:
@louischan <br>There are 2 empty entry in the language list, probably from "ES" and "PT"

<img src="https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/560daa73-3579-47c6-852b-3123016ba1b7/2892f373-3f80-409c-9ea3-98198e50e55b?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi81NjBkYWE3My0zNTc5LTQ3YzYtODUyYi0zMTIzMDE2YmExYjcvMjg5MmYzNzMtM2Y4MC00MDljLTllYTMtOTgxOThlNTBlNTViIiwiaWF0IjoxNzE0MTE4NzM0LCJleHAiOjMzMjg0Njc4NzM0fQ.dbSH0toN__0v1BIO1afINDSGafg_PWpKtbPu6-R4q_g " alt="image.png" width="1110" height="1450" />

---

There are some problem with the language resolver. It does not respect the "Accept-Language" and is always using the "Default" language.

**Accept-Language: ko (Korean) in Request Header:**

[...]

PR #4154, opened 2024-04-22
Generate translations

IniZio:
ref #4131 / DEV-1222


resources/authgear/templates/ja/translation.json
```diff
@@ -0,0 +1,1041 @@
...
+  "territory-XA": "疑似アクセント",
+  "territory-XB": "撮影のバイド",
+  "territory-XK": "コソボ",
+  "territory-YE": "イエメン",
+  "territory-YT": "マヨット",
+  "territory-ZA": "南アフリカ",
+  "territory-ZM": "ザンビア",
+  "territory-ZW": "ジンバブエ",
+  "territory-ZZ": "不明な地域",
+  "language-af": "アフリカーンス語",
```
louischan-oursky, line 776: language-* should be kept untranslated. The names are intentionally kept untranslated so that people who know that language can identify it.

resources/authgear/templates/ja/translation.json
```diff
@@ -0,0 +1,1041 @@
...
+  "account-deletion-delete-account-button-label": "アカウントを削除",
+  "account-deletion-success-title": "アカウント削除のスケジュール",
+  "account-deletion-success-description": "アカウントが無効化され、データは<span data-date=\"{rfc3339}\" data-date-type=absolute data-date-date-style=long data-date-time-style>{date, date, long}</span>に削除されます",
+  "account-status-contact-support": "カスタマーサポートにお問い合わせください。",
+  "account-status-contact-support-with-link": "<a class=\"link\" target=\"_blank\" href={customerSupportLink}>カスタマーサポート</a>に連絡してください。",
+  "gender-female": "女性",
+  "gender-male": "男性",
+  "tutorial-signup-login": "これは、プロジェクトの<b>サインアップ&ログインページ</b>です。サインアップして、アプリの最初のユーザーになってみましょう!",
+  "tutorial-settings": "<b>アカウント設定ページ</b>の準備ができました。ユーザーはここでセキュリティとプロファイルを管理できます。完全な認証体験を提供するためにこのページをアプリに統合してください。",
+  "territory-AC": "アセンション島",
```
louischan-oursky, line 513: We have a script scripts/npm/generate-territory-translation-json.js to obtain the localization from CLDR. Let's make new locales follow that.

scripts/python/generate_translations.py
```diff
@@ -0,0 +1,187 @@
...
+
+def make_update_locale_fn(chunk_size):
+  def update_locale(locale):
+    try:
+      logging.info(f'{locale} | Starting translation for {locale} ({LOCALE_DICT[locale]})')
+
+      # HTML template translation
+      update_translation(
+        locale=locale,
+        default_translation_file=f'../../resources/authgear/templates/en/translation.json',
```
louischan-oursky, line 162: This script should allow the user to choose the base language. For example, to generate the zh-CN translation, we want to use our zh-HK as the base language.

louischan-oursky (commented):
The language list provided by Fung does not match our list 100%. I think we have to make some adjustments.

1. In our list, we have `es-419` and `es`. Taking this as reference https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-localenames-full/main/en/languages.json#L185 I think we should provide `es-419` and `es-ES` at the moment.
2. In our list, we have `pt` and `pt-PT`. Taking this as reference https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-localenames-full/main/en/languages.json#L498 I think we should provide `pt-PT` and `pt-BR` at the moment.
3. I see `fil` in https://github.com/unicode-org/cldr-json/tree/main/cldr-json/cldr-localenames-full/main/fil but no `tl`. So I guess `fil` is preferred over `tl`.

Let's add `es-ES` and `pt-PT` to AvailableLanguages. Since we added two more langauges, we have to run devtools/languagename/main.go to generate the names. Copy the new names to translation.json

resources/authgear/templates/ja/translation.json
IniZio: Seems cldr does not provide territory nor language names for es-ES and pt-BR, maybe I fallback to es and pt respectively?

louischan-oursky (approved):

PR #4159, opened 2024-04-23
Complete account linking in authflow

tung2744:
ref https://github.com/authgear/authgear-server/issues/4134 https://github.com/authgear/authgear-server/issues/4146
ref DEV-1223 DEV-1231

Issue #4109, opened 2024-04-09
Modify AuthUI to support linking accounts


fungc-io:
![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/7a6c1b0a-2fa3-47f4-b911-ace2726eb2d7/3994d7f7-ad6a-40bf-bcc2-f7aa02282ed2?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi83YTZjMWIwYS0yZmEzLTQ3ZjQtYjkxMS1hY2UyNzI2ZWIyZDcvMzk5NGQ3ZjctYWQ2YS00MGJmLWJjYzItZjdhYTAyMjgyZWQyIiwiaWF0IjoxNzEyNjgyMjcxLCJleHAiOjMzMjgzMjQyMjcxfQ.qdr3QGENljgwJEsOSKp2enzqAminFIKC1ILrPAcdyQ0 "=786x")

Figma: [https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=2365-15938&mode=design&t=LXscrPWftY0rV2Wc-4](https://www.figma.com/file/LRE8EfCaTfYxC0lk5AeMmE/New-Auth-UI?type=design&node-id=2365-15938&mode=design&t=LXscrPWftY0rV2Wc-4)

---

- [ ] Add `prompt=login` to provider

fungc-io:
@tung2744 @louischan @newman see if the design above works?

fungc-io:
@tung2744 @joyz i've updated the design based on our discussion today. Any thoughts?:

[...]

PR #4164, opened 2024-04-24
Implement account linking in authui

tung2744:
ref DEV-1204
ref https://github.com/authgear/authgear-server/issues/4109
requires https://github.com/authgear/authgear-server/pull/4159

- It seems we already implemented prompt=login if that provider support that, so nothing have to be changed.

### DIFF
--- a/pkg/lib/config/authentication_flow.go
+++ b/pkg/lib/config/authentication_flow.go
@@ -1441,12 +1441,16 @@ const (
 )
 
 type AuthenticationFlowAccountLinkingJSONPointer struct {
-	Pointer jsonpointer.T `json:"pointer,omitempty"`
+	Pointer string `json:"pointer,omitempty"`
+}
+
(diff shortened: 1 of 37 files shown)


pkg/lib/config/authentication_flow.go
```diff
@@ -1441,12 +1441,16 @@ const (
 )
 
 type AuthenticationFlowAccountLinkingJSONPointer struct {
-	Pointer jsonpointer.T `json:"pointer,omitempty"`
+	Pointer string `json:"pointer,omitempty"`
```
louischan-oursky, line 1444: I don't quite get what this bug is about from the commit title. Can you elaborate?

tung2744: @louischan-oursky I found the problem on adding e2e test. Basically the problem is if I parse `/email` into a `jsonpointer.T`, then I serialize the config. It will become `#/email`.

louischan-oursky: I see.

PR #4170, opened 2024-04-25
Make account linking config global to all flows

tung2744:
ref DEV-1246

The problem:

Starting at this example config:
```yaml
authentication_flow:
  signup_flows:
    - name: signup_flow_1
      account_linking:
        oauth:
        - alias: adfs
          oauth_claim:
            pointer: "/preferred_username"
          user_profile:
            pointer: "/preferred_username"
          action: login_and_link
          login_flow: default
      steps:
        - name: identify
          type: identify
          one_of:
            - identification: email
            - identification: oauth
  promote_flows:
    - name: promote_flow_1
      steps:
        - name: identify
          type: identify
          one_of:
            - identification: email
            - identification: oauth
```

The linking logics, (like `oauth_claim.pointer`, `user_profile.pointer`, which are originally the link_by part), is defined in the signup flow. And in the above example, promote flow didn't define any "link by what". But this seems doesn’t make sense because this kind of “link by what” logic should be shared across the flows. If they are not same in all flows, expected accounts could be created by the promote flow, which in signup flow it is a conflict.
So what this pr is trying to do is to change the “link by what” logics to a global config to all auth flows.

This pr is trying to fix this problem by:
1. Change `authentication_flow.default_account_linking` to `authentication_flow.account_linking`. Now it is not only a default, you must define this section to use account linking.
2. Limit the overridable fields inside a specific flow. Only `login_flow` and `action` is overridable.

And also, moved per flow config to per step, and added a new usecase to describe the reason.

### DIFF
--- a/docs/specs/account-linking.md
+++ b/docs/specs/account-linking.md
@@ -167,7 +167,7 @@ authentication_flow:
 
 In the above example, an `account_linking` object was added inside the signup flow `signup_flow_1`, and it overrides some configs in the `authentication_flow.account_linking`.
 
-The first step to define an override is to give a `name` to the object you want to override. In the above example, `name: adfs_link_by_email` was added to an object inside `authentication_flow.account_linking.oauth`. `name` can be any string, it is used to reference this object in the later configs. 
+The first step to define an override is to give a `name` to the object you want to override. In the above example, `name: adfs_link_by_email` was added to an object inside `authentication_flow.account_linking.oauth`. `name` can be any string, it is used to reference this object in the later configs.
 
 Then, we add one item in `account_linking.oauth` inside the signup flow `signup_flow_1`, which has `name: adfs_link_by_email`. This means the config inside that item will override any config in `authentication_flow.account_linking.oauth` with `name` equal to `adfs_link_by_email`.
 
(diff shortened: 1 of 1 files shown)


docs/specs/account-linking.md
```diff
@@ -120,35 +120,62 @@ We define the action to link the new oauth identity with the existing identity's
...
+        - name: identify
+          type: identify
+          one_of:
+            - identification: email
+              steps:
+                - name: authenticate_primary_email
+                  type: create_authenticator
+                  one_of:
+                    - authentication: primary_password
+            - identification: oauth
```
tung2744, line 156: Actually this design is a bit uncomfortable to me.
Considering this example:

```yaml
authentication_flow:
  account_linking:
    oauth:
      - alias: adfs
        oauth_claim:
          pointer: "/preferred_username"
        user_profile:
          pointer: "/preferred_username"
        action: login_and_link
        login_flow: default
      - alias: adfs
        oauth_claim:
          pointer: "/email"
        user_profile:
          pointer: "/email"
        action: login_and_link
        login_flow: default
  signup_flows:
    - name: signup_flow_1
      account_linking:
        oauth:
          - alias: adfs # This will override the account_linking configs for adfs above 
            action: link_without_login_when_verified
            login_flow: login_flow_1
      steps:
        - name: identify
          type: identify
          one_of:
            - identification: email
              steps:
                - name: authenticate_primary_email
                  type: create_authenticator
                  one_of:
                    - authentication: primary_password
            - identification: oauth
```

The problem here is:
1. One object inside `signup_flows.account_linking.oauth` overridden two object inside `authentication_flow.account_linking`. Because the two object are both `alias: adfs`.
2. Cannot define different `action` for different pointers.

For 1, I am thinking about if we should limit the number of config per alias to 1 only, but this seems making an assumption that one provider should only have one logic (One pair of pointers) to link accounts.

For 2, it seems acceptable as there should be no normal usecase to do different action when using different pointer?

tung2744: Another option would be, just keep `oauth_claim` and `user_profile` in the flow specific config. And each if you want to override something, you should copy the whole object from `authentication_flow.account_linking`.

louischan-oursky: Given

1. We expect the matching logic is consistent for the whole project. Thus the following part should be shared by the whole project

```yaml
# No login_flow in the item. So the default is to use the login_flow of the same name.
account_linking:
  oauth:
  - alias: adfs
    oauth_claim:
      pointer: "/preferred_username"
    user_profile:
      pointer: "/preferred_username"
    action: login_and_link
  - alias: adfs
    oauth_claim:
      pointer: "/email"
    user_profile:
      pointer: "/email"
    action: login_and_link
```

2. For each match, we want to allow the specific signup flow / promote flow to specify the action, and the login flow to use. Thus, I think the items are missing a name attribute for other part of the config to reference them. Once we have name, I think we should flatten the items into a single list.

```yaml
# This is now moved to top-level, not under authentication_flow
account_linking:
- name: asdf_preferred_username
  type: oauth
  alias: adfs
  oauth_claim:
    pointer: "/preferred_username"
  user_profile:
    pointer: "/preferred_username"
  action: login_and_link
- name: asdf_email
  type: oauth
  alias: adfs
  oauth_claim:
    pointer: "/email"
  user_profile:
    pointer: "/email"
  action: login_and_link
- name: email
  type: login_id
  key: email
  action: login_and_link
```

3. In case the signup flow wants to override, use the name to reference the item

```yaml
signup_flows:
- name: default
  account_linking:
    override:
    - name: asdf_preferred_username
       login_flow: another_login_flow
       action: link_without_login_when_verified
```

tung2744:
Updated according to louis's suggestions :adore:

docs/specs/account-linking.md
```diff
@@ -388,14 +426,20 @@ The above configs defined the account linking behavior of two login id types:
...
+| ------------- | ------------------------------------------------------------------------------------------------------ |
+| `email`       | <pre><code>user_profile:<br/>&nbsp;&nbsp;pointer: "/email"<br/>action: error</code></pre>              |
 | `phone`       | <pre><code>user_profile:<br/>&nbsp;&nbsp;pointer: "/phone_number"<br/>action: error</code></pre>       |
 | `username`    | <pre><code>user_profile:<br/>&nbsp;&nbsp;pointer: "/preferred_username"<br/>action: error</code></pre> |
 
 Therefore when not specified, all attempts to link a new login id to existing users or idenitities will result in error.
 
+## Account Linking in Promote Flow
+
+Account linking can occur in promote flow, however, only `error` is allowed to be the action of account linking.
```
chpapa, line 439: why?

tung2744: In promote flow, there is an existing user already. And if we allow actions such as "login_and_link", it means we support login to an existing user during promote flow. But login to an existing user in promote flow is not allowed at the moment.

chpapa: I think we should allow it as promote flow UX-wise is equals to login or signup from end users' perspective.

If their behaviour is different, it will be very difficult for developers to use this function at all. 

If it is technically impossible to do so (e.g. not sure which login ID to use), there maybe some other design issues on our system.

louischan-oursky (commented):
1. Move `account_linking` from under `authentication_flow` to the top-level.
2. Remove `login_flow` from the global config.

louischan-oursky (approved):

PR #4191, opened 2024-05-02
Specify authflow prioritized identity

tung2744:
ref DEV-1272

### DIFF
--- a/docs/specs/authentication-flow-prioritized-identity.md
+++ b/docs/specs/authentication-flow-prioritized-identity.md
@@ -2,23 +2,25 @@
 
 - [Introduction](#introduction)
 - [Usecases](#usecases)
-  - [Force user to use a preferred authentication method if available](#force-user-to-use-a-preferred-authentication-method-if-available)
+  - [Force users to use a safer OAuth Provider (such as Google Login)](#force-users-to-use-a-safer-oauth-provider-such-as-google-login)
+  - [Internal ADFS always have priority over username login for security purpose](#internal-adfs-always-have-priority-over-username-login-for-security-purpose)
 - [Behaviors](#behaviors)
   - [When there are multiple levels of priority](#when-there-are-multiple-levels-of-priority)
   - [When the user has only some of the prioritized identities](#when-the-user-has-only-some-of-the-prioritized-identities)
+
 ## Introduction
 
 It is common that some identities are considered more secure, or more preferable comparing with other identities of the same user. Apps might want to have a way to specify the priority of different types of identities in a login flow, so when a user has another identity which is more preferable, it will be used for identification inside a login flow.
 
 ## Usecases
 
-### Force user to use a preferred authentication method if available
+### Force users to use a safer OAuth Provider (such as Google Login)
 
 Assume you have a portal app which supports two signup methods: Email with Password, and Google oauth login.
 
 User of the portal can signup to the portal by one of the two methods, and login with that method afterwards.
 
-If the user has signed up with email with password, the user can connect to a Google account at any time after he logged in to the portal (For example, through [Account Linking](./account-linking.md)). In your persepctive, Google oauth login is the preferred login method because it is more secure. Therefore once user connected their google account, the portal should only accept google oauth login for the same account, and do not accept email with password logins.
+If the user has signed up with email with password, the user can connect to a Google account at any time after he logged in to the portal (For example, through [Account Linking](./account-linking.md)). In your perspective, Google oauth login is the preferred login method because it is more secure. Therefore once user connected their google account, the portal should only accept google oauth login for the same account, and do not accept email with password logins.
 
 Use the following config to acheive the goal:
 
@@ -52,7 +54,7 @@ Whenever an identification method was chosen in an `identify` step, the `priorit
    ```json
    {
      "name": "Invalid",
-     "reason": "UnpreferredIdentity",
+     "reason": "PrioritizedIdentityRequired",
      "message": "please use another identification method",
      "code": 400,
      "info": {
@@ -119,7 +121,7 @@ Looking at the existing identities of Alice, she has a Google oauth identity, wh
 ```jsonc
 {
   "name": "Invalid",
-  "reason": "UnpreferredIdentity",
+  "reason": "PrioritizedIdentityRequired",
   "message": "please use another identification method",
   "code": 400,
   "info": {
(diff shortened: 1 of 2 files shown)


docs/specs/authentication-flow-prioritized-identity.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Usecases
+
+### Force user to use a preferred authentication method if available
+
+Assume you have a portal app which supports two signup methods: Email with Password, and Google oauth login.
+
+User of the portal can signup to the portal by one of the two methods, and login with that method afterwards.
+
+If the user has signed up with email with password, the user can connect to a Google account at any time after he logged in to the portal (For example, through [Account Linking](./account-linking.md)). In your persepctive, Google oauth login is the preferred login method because it is more secure. Therefore once user connected their google account, the portal should only accept google oauth login for the same account, and do not accept email with password logins.
```
louischan-oursky, line 21: In your s/persepctive/perspective

docs/specs/authentication-flow-prioritized-identity.md
```diff
@@ -0,0 +1,268 @@
...
+```yaml
+authentication_flow:
+  login_flows:
+    - name: default
+      steps:
+        - name: identify
+          type: identify
+          one_of:
+            - identification: oauth
+              priority: 1 # <- Configure priority here
```
louischan-oursky, line 34: Adding priority here does not allow me to specify "Login with Google" is preferred over "Login with X". I think we need to allow specifing priority for each alias. A few problems then arise.

1. The developer wants to specify priority for a alias. So she must write out the alias. For aliases that were not written out, what is the meaning?
  - Aliases that are not written out are omitted from this identification. This enables another feature that is not implemented yet.
  - Aliases that are not written out are still included in this identification. Their priority is 0.

tung2744: It seems the problem here is we didn't consider different oauth method as different identification options, but they should be in fact separated options. My suggestion is:

```yaml
steps:
  - type: identify
    one_of:
        - identification: oauth
          priority: 2
          alias: adfs
        - identification: oauth
          priority: 1
          alias: google
        - identification: oauth
          priority: 0
```

1. Add `alias` to `identification: oauth`. If `alias` is set, only provider of that alias will be included as option.
2. If `alias` is not set, it means "all" provider.
3. `priority` applies to the specified oauth provider if `alias` is set, or apply to all providers if `alias` is not set.

louischan-oursky: From your example config, due to the presence of the 3rd item, all OAuth providers are shown as options. The actual number of options could be more than 3. It is not intuitive.

What about

```yaml
steps:
- type: identify
  one_of:
  - identification: oauth
    aliases: # For exposing only a subset of configured OAuth providers.
    - alias: google
    - alias: adfs
    priorities: # Specify the priority
    - alias: google
      priority: 1
  - identification: email
  - identification: phone

```

tung2744: Some questions in you example:
1. How to specify priority for email and phone? I guess it will be something like this:

```yaml
steps:
- type: identify
  one_of:
  - identification: oauth
    aliases: # For exposing only a subset of configured OAuth providers.
    - alias: google
    - alias: adfs
    priorities: # Specify the priority
    - alias: google
      priority: 1
  - identification: email
     priority: 1 # <--
  - identification: phone
```

but this is a bit confusing. The identification object now has two fields, `priorities` which is for oauth, and `priority` for other identifications.

2. How to specify priority for "any provider"?

> From your example config, due to the presence of the 3rd item, all OAuth providers are shown as options. The actual number of options could be more than 3. It is not intuitive.

This is actually a problem exist in the current config. Anyway, for backward compatibility, we should support `identification: oauth` without a specified `aliases`, so this problem is actually unavoidable. In your example, this is also possible:

```yaml
steps:
- type: identify
  one_of:
  - identification: oauth
     priorities: # Specify the priority
    - # Not sure how to write
      priority: 1
  - identification: email
  - identification: phone
```

And the resulting number of options is still not same as the number of rows of `identification`.

docs/specs/authentication-flow-prioritized-identity.md
```diff
@@ -0,0 +1,268 @@
...
+- [Behaviors](#behaviors)
+  - [When there are multiple levels of priority](#when-there-are-multiple-levels-of-priority)
+  - [When the user has only some of the prioritized identities](#when-the-user-has-only-some-of-the-prioritized-identities)
+## Introduction
+
+It is common that some identities are considered more secure, or more preferable comparing with other identities of the same user. Apps might want to have a way to specify the priority of different types of identities in a login flow, so when a user has another identity which is more preferable, it will be used for identification inside a login flow.
+
+## Usecases
+
+### Force user to use a preferred authentication method if available
```
chpapa, line 15: BTW @tung2744 this is not a "use case", you're literally just writing out what the feature want to do instead of mentioning a use case.

Use cases could be:
- Force users to use a safer OAuth Provider (such as Google Login)
- Migrating users from email identities to phone identities
- Internal ADFS always have priority over email login for security purpose

Than in each use case it would be easier to describe what's needed exactly, and easier for us to consider what might be missing in the config.

chpapa: cc: @louischan-oursky 

(p.s. I paid special attention on it this time since Tung is new to it)

docs/specs/authentication-flow-prioritized-identity.md
chpapa: p.s. right now portal have no way to config these authflow, so if something seriously wrong in our existing design, we can actually change it and no need to consider backward compatibilities (we can always modify all existing authflow config after we change it) -- this is intended. @louischan-oursky @tung2744

docs/specs/authentication-flow-prioritized-identity.md
```diff
@@ -0,0 +1,268 @@
...
+  "code": 400,
+  "info": {
+    "PreferredIdentitifications": [
+      {
+        "identification": "oauth",
+        "provider_type": "google",
+        "alias": "google"
+      },
+      {
+        "identification": "phone"
```
chpapa, line 204: Why show both oauth/google and phone here? Isn't that we should show the "ONE" identification method that this users could use?

docs/specs/authentication-flow-prioritized-identity.md
tung2744: Then I think if we really think `number of options != number of identification` is a problem, I suggest we simply deprecate `identification: oauth` without `alias`. i.e., we always ask developer to provide a `alias` if `identification: oauth` is used.

And for migration of old data, for simplicity, we can just leave the current behavior of `identification: oauth` without `alias`, but we will stop generating config in this way, and also remove it from documentation.

docs/specs/authentication-flow-prioritized-identity.md
tung2744: Then I think the use case here should be "Force users to use a safer OAuth Provider (such as Google Login)". Let me update this section.

> Migrating users from email identities to phone identities

However I am not sure if this usecase should be handled by this feature. Maybe I will not include it in the spec at the moment.

docs/specs/authentication-flow-prioritized-identity.md
tung2744: My consideration is, if we only return a list of options which the user has, it seems we exposed more imformation about what identities the user has.
However, after rethink it seems we already exposed the imformation when we have different behavior on user has / doesn't have the corresponding identity, so looks like it is not very critical.
Let me change it to "Only return the options that user can choose", which is more suitable for the usecase.

docs/specs/authentication-flow-prioritized-identity.md
```diff
@@ -0,0 +1,268 @@
...
+  "login_id": "alice@example.com"
+}
+```
+
+`identification: email` will be selected as the identification method, which has `priority` of `1`. `phone` and `oauth` which has priority higher than the current selected identification method `email` will be included in the `PreferredIdentitifications` array, while `username` will be excluded.
+
+```jsonc
+{
+  "name": "Invalid",
+  "reason": "UnpreferredIdentity",
```
louischan-oursky, line 193: Maybe `PrioritizedIdentityRequired`. It is more consistent with the term (`priority`) we use in the config, and also convey the meaning that the prioritized identification is required.

docs/specs/authentication-flow-prioritized-identity.md
louischan-oursky: > And for migration of old data, for simplicity, we can just leave the current behavior of identification: oauth without alias, but we will stop generating config in this way, and also remove it from documentation.

Agree. We should keep supporting it.

On the other hand, this deprecated config must not co-exist with the new one. It is a config error if such case happens.

tung2744:
Updated, thanks!
Changes:

1. Fixed some typos.
2. Renamed the title of the usecase section, and added one usecase.
3. Use `PrioritizedIdentityRequired` as error reason.
4. Added `alias` to `identification: oauth`.

louischan-oursky:
Closed this to avoid piling up pending PRs.

PR #4196, opened 2024-05-03
Update base image to bookworm

louischan-oursky:
ref DEV-1251


.dockerignore
```diff
@@ -5,5 +5,11 @@
 **/node_modules
 **/.parcel-cache
 **/parcel-bundle-reports
+**/venv
 **/var
+**/resources/**/generated
```
tung2744, line 10: Should we use whitelist instead of blacklist in dockerignore to avoid including unwanted files in the image?
It could be easy to forgot updating the dockerignore file when adding new things.

louischan-oursky: I tried to switch to an allowlist but it seems that when a pattern is negated, the pattern has a high precedence. That would make `**/` wildcards has no effect.

I am going to set up a command to compare the build context with `git ls-files` instead.

louischan-oursky: Updated.

tung2744 (approved):

PR #4205, opened 2024-05-08
Further reduce noise

louischan-oursky:
ref DEV-1251

### DIFF
--- a/pkg/api/apierrors/skip_logging.go
+++ b/pkg/api/apierrors/skip_logging.go
@@ -3,6 +3,7 @@ package apierrors
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"net/http"
 	"syscall"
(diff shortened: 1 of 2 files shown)


pkg/api/apierrors/skip_logging.go
```diff
@@ -63,7 +71,12 @@ func IgnoreError(err error) (ignore bool) {
...
 	if errors.Is(err, syscall.EPIPE) {
-		// syscall.EPIPE is "write: broken pipe"
+		// syscall.EPIPE is "broken pipe"
+		ignore = true
+	}
+
+	if errors.Is(err, syscall.ECONNREFUSED) {
+		// syscall.ECONNREFUSED is "connection refused"
 		ignore = true
 	}
```
tung2744, line 81: It seems `connection refused` is not something we can safely ignore.
If it occurs once or twice, and fixed by retrying, then it is ok to ignore. But if it is caused by some other problems we should fix, such as firewall configs or networking issue, the error shouldn't be ignored.

And also it looks like it depends on context. If it is from the hook sink, this can be simply ignored. However, if it is from redis / db, looks like we should have some retries and finally throw an error which shouldn't be ignored.

louischan-oursky: No longer ignore syscall errors.

tung2744 (approved):

Issue #4212, opened 2024-05-10
Should not translate "DELETE" in the account delete message


fungc-io:
Check translation and AI prompt to ensure this "DELETE" keyword is not translated.

![image.png](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/c5318e80-264f-466c-896c-0921c338f50f/ef497334-fdf4-40d2-b6f8-24b451be4f88?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi9jNTMxOGU4MC0yNjRmLTQ2NmMtODk2Yy0wOTIxYzMzOGY1MGYvZWY0OTczMzQtZmRmNC00MGQyLWI2ZjgtMjRiNDUxYmU0Zjg4IiwiaWF0IjoxNzE1NjgxMTU3LCJleHAiOjMzMjg2MjQxMTU3fQ.4Tb_bDv\_-0mpqWI2T3_B09Qk1P8aIdDtrJWCs68KbHA "=869x")

PR #4220, opened 2024-05-16
Fix delete confirmation text being translated

randy1yu:
ref #4212 
ref DEV-1293


louischan-oursky (commented):
I picked Italian to test. And found that it will crash in the confirmation delete page with Italian. I further discover that in the it/translation.json, there are some malformed Simplified Chinese translation causing the crash. Can you re-generate the Italian translation?

randy1yu:
> I picked Italian to test. And found that it will crash in the confirmation delete page with Italian. I further discover that in the it/translation.json, there are some malformed Simplified Chinese translation causing the crash. Can you re-generate the Italian translation?

Regenerating for few times but some result seems still wrong, I think it's because ai misunderstood locale code `it`.
Maybe it's helpful to add locale name to the prompt.

louischan-oursky:
We have `name` in the mapping https://github.com/authgear/authgear-server/blob/main/scripts/python/generate_translations.py#L17

But the name is not canonical. The canonical names can be obtained from `scripts/npm/node_modules/cldr-localenames-modern/main/en/languages.json`. Let's copy the names from `scripts/npm/node_modules/cldr-localenames-modern/main/en/languages.json` to our mapping. And then update the prompt to use the canonical name instead of the 2-letter language identifier. See if this approves anything.

randy1yu:
Updated

> And then update the prompt to use the canonical name instead of the 2-letter language identifier. See if this approves anything.

It looks better for Italian.

louischan-oursky (approved):

PR #4221, opened 2024-05-17
Use redirect_uri to return error if client id cannot be resolved

tung2744:
ref dev-1282


louischan-oursky:
I remember why we have the current behavior. The current behavior is intentional. If redirect URI is blindly followed, then Authgear is vulnerable to an attack called "open redirect". Therefore, what we really is to disable a client instead of removing it.

tung2744:
I see, let me close this first.

Issue #4181, opened 2024-04-29
Move cookie consent widget to else where


fungc-io:
* Add a new option here for "Cookie Preference" which will trigger the OSANO drawer

  <img src="https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/6f83604d-20d6-475f-8083-212f3ceb56bb/45b0fb23-6f1d-40e6-9e21-a9a991b0c95c?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi82ZjgzNjA0ZC0yMGQ2LTQ3NWYtODA4My0yMTJmM2NlYjU2YmIvNDViMGZiMjMtNmYxZC00MGU2LTllMjEtYTlhOTkxYjBjOTVjIiwiaWF0IjoxNzE0Mzg3MTQ1LCJleHAiOjMzMjg0OTQ3MTQ1fQ.xTJbBAJdKeAF1S-2NSKBGzKEUneKL3No5mvwJKLC864 " alt="image.png" width="177" height="113" />
  * Icon: "cookies" icon by fluent<br>ref: [https://github.com/microsoft/fluentui-system-icons/blob/main/icons_regular.md](https://github.com/microsoft/fluentui-system-icons/blob/main/icons_regular.md)

[...]

PR #4222, opened 2024-05-17
Move Osano widget to header nav

randy1yu:
ref #4181
ref DEV-1256

截圖 2024-05-17 下午4 16 03

### DIFF
--- a/portal/src/ScreenHeader.tsx
+++ b/portal/src/ScreenHeader.tsx
@@ -199,6 +199,14 @@ const ScreenHeader: React.VFC<ScreenNavProps> = function ScreenHeader(props) {
       });
   }, [redirectURI, reset]);
 
+  const onClickCookiePreference = useCallback(() => {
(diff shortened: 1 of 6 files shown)


portal/src/ScreenHeader.tsx
```diff
@@ -173,6 +175,12 @@ interface ScreenNavProps {
   showHamburger?: boolean;
 }
 
+registerIcons({
+  icons: {
+    Cookies: <Cookies16Regular />,
+  },
+});
```
louischan-oursky, line 182: Please move this call of registerIcons to src/index.tsx. By convention, we do all initialization in that file.

louischan-oursky (changes requested):

portal/src/ScreenHeader.tsx
randy1yu: Updated

louischan-oursky (approved):

PR #4224, opened 2024-05-17
Support marking password as expired on user import

tung2744:
ref DEV-1304

### DIFF
--- a/pkg/util/validation/formats.go
+++ b/pkg/util/validation/formats.go
@@ -267,7 +267,7 @@ type FormatDateTime struct{}
 func (FormatDateTime) CheckFormat(value interface{}) error {
 	str, ok := value.(string)
 	if !ok {
-		return fmt.Errorf("date-time must be string")
+		return nil
 	}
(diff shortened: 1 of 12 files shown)


pkg/util/validation/formats.go
```diff
@@ -261,6 +262,22 @@ func (FormatTimezone) CheckFormat(value interface{}) error {
 	return nil
 }
 
+type FormatDateTime struct{}
+
+func (FormatDateTime) CheckFormat(value interface{}) error {
```
louischan-oursky, line 267: Include some tests for the new format.
- With integer second
- With some milliseconds
- With UTC timezone
- With some explicit timezone offset

tung2744: Added some tests, thanks!

louischan-oursky:
@tung I added a commit to improve the JSON schema. Previously a unknown key is silently ignored.

louischan-oursky (approved):

Issue #3863, opened 2024-02-27
Unclear error message when no country is selected in the portal


fungc-io:
In the portal, in phone number authentication method settings,
if no country is selected,
The current error message shows "Expect at least 1 item", which is unclear for the user

It should be changed to: `Please select at least one country for the phone number settings`

![Image](https://uploads.linear.app/860e83ae-05f0-4c4b-bd93-8a4b8d251836/29dbda3c-5101-4306-b1ef-7582f37f09fd/617276c0-6f64-4ee3-955d-521c811f4e58?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg2MGU4M2FlLTA1ZjAtNGM0Yi1iZDkzLThhNGI4ZDI1MTgzNi8yOWRiZGEzYy01MTAxLTQzMDYtYjFlZi03NTgyZjM3ZjA5ZmQvNjE3Mjc2YzAtNmY2NC00ZWUzLTk1NWQtNTIxYzgxMWY0ZTU4IiwiaWF0IjoxNzE1MDgwOTg0LCJleHAiOjMzMjg1NjQwOTg0fQ.hgQ3YJ9D0KQBf4ECAC2pHAjt1up32PSKAkId2C7U_oU)

PR #4228, opened 2024-05-21
Add specific error message for ui/phone_input/allowlist

randy1yu:
ref #3863 
ref DEV-401


louischan-oursky:
@randy1yu Turned the original error handling mechanism does not support customizing the message. I refactored ErrorParseRule to support this.

louischan-oursky (approved):

PR #4234, opened 2024-05-22
Separate create / update / delete config of login ids

tung2744:
ref DEV-1313

Will open another pr for oauth identity config.


pkg/lib/config/identity.go
```diff
@@ -232,8 +238,17 @@ func (c *LoginIDKeyConfig) SetDefaults() {
 	if c.Key == "" {
 		c.Key = string(c.Type)
 	}
-	if c.ModifyDisabled == nil {
-		c.ModifyDisabled = newBool(false)
+	if c.Deprecated_ModifyDisabled == nil {
+		c.Deprecated_ModifyDisabled = newBool(false)
+	}
+	if c.UpdateDisabled == nil {
+		c.UpdateDisabled = c.Deprecated_ModifyDisabled
```
louischan-oursky, line 245: Although we will probably not change these boolean fields, it is more correct to write

```golang
if c.UpdateDisabled == nil {
  b := *c.Deprecated_ModifyDisabled
  c.UpdateDisabled = &b
}
```

So that each field has its own pointer to a distinct boolean value. Changing one will not accidentally change others.

tung2744: Updated, thanks!

louischan-oursky (approved):

PR #4246, opened 2024-05-28
Block duplicate login id creation

tung2744:
ref DEV-1314


tung2744:
Not needed now, closing

PR #4251, opened 2024-05-29
Create onboarding survey UI

toshinari123:
Sorry for lateness, was gonna do it last night but fell asleep
- on mobile, having each button in its own row is a bit ugly due to there being a very long button "project manager / product manager" so changed to a cluster of button with about 2 buttons per row
- logo background colour need to change
- tried to be typesafe but ended up overcomplicating a bit
- implemented button group so the logic can be flexible with processChoice callback
- very weird issue where if button is toggled and you hover over the border the text colour changes


louischan-oursky (commented):
I took a brief look. The focus was on how you customize a Fluent UI component. I added a commit to show you my expectation on customizing a component. Hope you can learn it by this example.

I did not spend too much time on understanding the type functions you wrote. I think they are too complicated. If your goal is to write safe code, you should write simple code instead. AND you should write unit tests. The logic in the page is actually very simple, it does not deserve the complicated type functions. Let's put more effort on the UI code instead.

toshinari123:
For now, the `(Optional)`s are hardcoded into the labels. Should I wrap FormTextField and FormPhoneTextField into CanOptionalFormTextField and CanOptionalFormPhoneTextField and add my own custom boolean attribute `optional`?

I can also customize FluentUI to add `(Optional)` to the labels when `required` is false, but I don't think I can get rid of the red asterisks that comes after fields where `required` is true.

portal/src/OnboardingSurveyLayout.tsx
```diff
@@ -61,13 +93,20 @@ export default function SurveyLayout(
...
+  const theme = useTheme();
+  const bodyStyles = useMemo(() => {
+    return {
+      backgroundColor: theme.semanticColors.bodyStandoutBackground,
+      height: "100vh",
+    };
+  }, [theme]);
   return (
-    <div className={styles.root}>
+    <div style={bodyStyles}>
```
louischan-oursky, line 104: We use className to customize CSS for non-fluent UI element. Move height to the CSS file.

portal/src/OnboardingSurveyLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const theme = useTheme();
+  const styles = useMemo(() => {
+    return {
+      root: {
+        "font-size": "x-large",
+        "font-weight": 600,
+        "white-space": "pre-line",
+        "text-align": "center",
+        color: theme.semanticColors.bodyText,
+        "margin-bottom": "20px",
```
toshinari123, line 41: font-size, font-weight -> variant
color -> styles (need runtime information)
other -> className

portal/src/OnboardingSurveyLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+export function SurveySubtitle(props: SurveySubtitleProps): React.ReactElement {
+  const theme = useTheme();
+  const styles = useMemo(() => {
+    return {
+      root: {
+        "font-size": "medium",
+        "font-weight": 400,
+        "white-space": "pre-line",
+        "text-align": "center",
+        color: theme.semanticColors.bodySubtext,
```
toshinari123, line 65: same apply here.

portal/src/OnboardingSurveyLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+  }, [theme]);
+  return (
+    <Text styles={styles} variant="small" block={true}>
+      {props.children}
+    </Text>
+  );
+}
+
+export interface SurveyLayoutProps {
+  title: string;
```
toshinari123, line 77: It should be ReactNode

portal/src/OnboardingSurveyLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+  subtitle: string;
```
toshinari123, line 78: It should be ReactNode.

portal/src/OnboardingSurveyLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+  backButtonDisabled: boolean;
+  primaryButton: React.ReactNode;
+  secondaryButton: React.ReactNode;
```
toshinari123, line 81: Remove this prop. It is actually back button,

portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -0,0 +1,491 @@
...
+    />
+  )
+}
+*/
+
+interface ChoiceButtonGroupProps {
+  prefix: string;
+  choices: readonly string[];
+  state: Record<string, boolean>;
+  setChoice: (choice: Record<string, boolean>) => void;
```
toshinari123, line 68: onChangeChoices: (availableChoices: string[], selectedChoices: string[])  void

portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -0,0 +1,491 @@
...
+      className={styles.compoundChoiceButton}
+      onClick={onClick}
+    />
+  )
+}
+*/
+
+interface ChoiceButtonGroupProps {
+  prefix: string;
+  choices: readonly string[];
```
toshinari123, line 66: availableChoices: string[]

portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -0,0 +1,491 @@
...
+  state: Record<string, boolean>;
```
toshinari123, line 67: selectedChoices: string[]

portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -0,0 +1,491 @@
...
+  setChoice: (choice: Record<string, boolean>) => void;
+  processChoice: (
+    oldChoice: Record<string, boolean>,
+    selectedChoice: Readonly<string>
+  ) => Record<string, boolean>;
```
toshinari123, line 72: Given that we actually only need single-choice and multiple-choice, replace this prop with `mode?: "single_choice" | "multiple_choice"` The default is `"single_choice"`

portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -0,0 +1,491 @@
...
+  Button: ComponentType<DefaultButtonProps>;
+  isIndividualComponent: boolean;
```
toshinari123, line 74: Instead of this prop, we separate ChoiceGroupVariantLabel, ChoiceGroupVariantQuestion, ChoiceGroupVariantMutipleChoice

portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -0,0 +1,491 @@
...
+    allFalse(roleChoices);
+  const [roleChoicesState, setRoleChoicesState] = useState(defaultRoleChoices);
+  const empty = useMemo(() => {
+    return (
+      Object.entries(roleChoicesState).reduce(
+        (acc, [_, v]) => acc + (v ? 1 : 0),
+        0
+      ) === 0
+    );
+  }, [roleChoicesState]);
```
toshinari123, line 169: Thanks to selectedChoices, we can just check the .length

portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -0,0 +1,491 @@
...
+      checked={pressed}
+      className={styles.compoundChoiceButton}
+      onClick={onClick}
+    />
+  )
+}
+*/
+
+interface ChoiceButtonGroupProps {
+  prefix: string;
```
toshinari123, line 65: Remove prefix because it promotes a pattern that causes scanning of translation keys to fail.

portal/src/OnboardingSurveyLayout.tsx
```diff
@@ -1,18 +1,25 @@
...
   const { renderToString } = useContext(Context);
-
+  const theme = useTheme();
+  const logoStyles = useMemo(() => {
+    return {
+      fill: theme.semanticColors.bodyText,
+    };
+  }, [theme]);
   return (
     <img
```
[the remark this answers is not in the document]
toshinari123: FluentUI Image flickers on pressing next or back, reverted back to img

portal/src/OnboardingSurveyLayout.tsx
toshinari123: After a bit of thinking, I think this is better:
- if its reactnode, i would have to make SurveyTitle and SurveySubtitle in OnboardingSurveyScreen.tsx instead of OnboardingSurveyLayout.tsx; it would make sense if different pages have different titles and subtitles elements, but this is not the case.
- the nextbutton and backbuttons are reactnode because the `onClick`, `disabled` and `text` needs to be supplied; also the nextbutton is different in the last page (finish suvey) so it makes sense for them to be ReactNodes

portal/src/OnboardingSurveyScreen.tsx
toshinari123: sorry forgot to ask, what about the buttons on let's say the first page where you select your role? If i need to not use prefix, I would have to hard code the entire button group and the 7 buttons it has. (Dev, IT, PM, PD, Market, Owner, Other; see en.json below)

toshinari123:
https://linear.app/authgear/issue/DEV-1331/ui-implementation-for-onboarding-survey

louischan-oursky (commented):
SCR-20240607-mkxs

Maybe you need to use grid to make each row share the same height, in mobile layout?

---

When I click "Finish", the local storage does not clear. It should.

toshinari123:
| mobile:grid-cols-1 | mobile:grid-cols-2 |
| - | - |
| image | image |

Which is preferred? (Both kinda need to scroll a bit)

louischan-oursky:
The 1-column one is better.

toshinari123:
localstorage is also cleared now

toshinari123:
tasks left:
- phone number input cant customize border color
- I hard coded the (optional) into the label (`en.json`) refer to https://github.com/authgear/authgear-server/pull/4251#issuecomment-2144895644

portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -660,8 +722,8 @@ function Step4(_props: StepProps) {
       if (otherReason !== "") setLocalStorage("otherReason", otherReason);
       //TODO: save to somewhere
       //eslint-disable-next-line no-console
-      console.log(localStorage.getItem("authgear-onboarding-survey"));
-      //clear local storage?
+      console.log(localStorage.getItem(localStorageKey), "Sdfsf");
```
louischan-oursky, line 725: Remove this log.

PR #4298, opened 2024-06-12
Specify Pre-Authenticated URL

tung2744:


docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+This document is drafted based on [OpenID Connect Native SSO for Mobile Apps 1.0](https://openid.net/specs/openid-connect-native-sso-1_0.html).
+
+## The Flow
+
+```mermaid
+sequenceDiagram
+    participant app1 as Mobile App #35;1
+    participant Web as Web App #35;2 <br/>(Cookie Based)<br/>https://www.example.com
+    participant Authgear as Authgear <br/> https://auth.example.com
+    app1->>+Authgear: [1]GET https://auth.example.com/oauth2/authorize<br/>client_id=1<br/>scope=x_device_browser_sso<br/>redirect_uri=app://redirect
```
louischan-oursky, line 29: Maybe we should rename the scope to `x_device_browser_session_token`. If `x_device_browser_session_token` is granted in the scope, then device browser session token can be exchanged in the token endpoint with Token Exchange mechanism.

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+    Authgear->>-app1: [6]Response with<br/>device_browser_session_token=DEVICE_BROWER_SESSION_TOKEN
+    app1->>Authgear: [7]Open URL In Browser: https://auth.example.com/oauth2/authorize?<br/>client_id=2<br/>id_token_hint=ID_TOKEN<br/>device_browser_session_token=DEVICE_BROWER_SESSION_TOKEN<br/>redirect_uri=https://www.example.com/redirect<br/>state=STATE<br/>prompt=none<br/>response_type=token<br/>response_mode=cookie
+    Authgear->>Web: [8]Redirect to https://www.example.com/redirect?state=STATE<br/>Set Cookie at=ACCESS_TOKEN
+    Note right of Web: Web now authenticated as Alice by cookie
+```
+
+Step [1] to [4] are the standard OpenID Connect authorization_code flow with the following extensions. In step [1], the `x_device_browser_sso` is specified signifying that the client is requesting a `device_secret` to be returned when the code is exchanged for tokens.
```
louischan-oursky, line 41: Remember to rename all ocurrences of `x_device_browser_sso`

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+
+After step [4], Mobile App #1 stores the device_secret and id_token in the protected device storage accessible only to Mobile App #1.
+
+Step [5] invokes the /token endpoint with the token exchange profile passing the id_token obtained from the device storage, the client_id and the device secret. A device_browser_session_token is then returned in step [6].
+
+Step [7] uses the obtained device_browser_session_token to open /authorize in a browser, with the id_token used in step [5], and a redirect_uri pointing to Web App #2.
+
+Step [8] redirects the browser to Web App #2, with a session cookies set. The session cookie can then be used for authorizing Web App #2 as it was logged in as the same user of Mobile App #1.
```
louischan-oursky, line 49: Avoid using the term "session cookie", which normally means HTTP session cookie, a cookie without expiration. Let's use "with a cookie representing the user session set", "The cookie can then be ..."

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Native App Authorization Extensions
+
+Same as the [Native SSO](https://openid.net/specs/openid-connect-native-sso-1_0.html#name-native-app-authorization-ex), except that the new scope value supported is `x_device_browser_sso`.
+
+## Token Exchange Profile for App Initiated SSO to Web
+
+This section profiles the OAuth2 Token Exchange [RFC8693] spec and describes the processing rules used to exchange a previous authentication for a device browser session token, which can then be used to exchange for a browser session of another Web App.
```
louischan-oursky, line 57: Add back the link to RFC8693 `[RFC8693](https://datatracker.ietf.org/doc/html/rfc8693)`

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
+# App Initiated SSO to Web
```
louischan-oursky, line 1: App-initiated SSO to Web

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+
+```
+urn:authgear:params:oauth:token-type:device-browser-session-token
+```
+
+This token type identifier refers to the device browser session token mentioned in this spec. It can be used as the `requested_token_type` parameter in token exchange request.
+
+### Token Exchange Request
+
+When a mobile app wants to initiate SSO to web, it makes a /token endpoint request to obtain a device browser session token with the following parameters:
```
louischan-oursky, line 77: "When a mobile app wants to perform App-initiated SSO to Web". Stay consistent of our terms.

louischan-oursky, line 77: ", it makes a request to the token endpoint to ..."

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+
+- `grant_type`: Must be `urn:ietf:params:oauth:grant-type:token-exchange`
+- `subject_token`: The id_token obtained by the mobile app
+- `subject_token_type`: Must be `urn:ietf:params:oauth:token-type:id_token`
+- `actor_token`: The device_secret obtained by the mobile app
+- `actor_token_type`: Must be `urn:x-oath:params:oauth:token-type:device-secret`
+- `requested_token_type`: Optional. If given, the value must be `urn:authgear:params:oauth:token-type:device-browser-session-token`.
```
louischan-oursky, line 84: I see it is optional in Native SSO and RFC8693, but I think it is required if the client wants us to issue a device browser session token. Otherwise, the behavior is the same Native SSO, we issue access token.

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+&requested_token_type=urn%3Aauthgear%3Aparams%3Aoauth%3Atoken-type%3Adevice-browser-session-token
+```
+
+### Token Exchange Response
+
+The Token Exchange response for this profile has the following characteristics:
+
+- `device_secret`: Required. A new device secret must be returned. The mobile app should update its device secret with the new secret.
+- `x_device_browser_session_token`: Required. The device browser session token which can then be used to exchange for a browser cookie.
+- `expires_in`: Required. When the device browser session token expires.
```
louischan-oursky, line 106: The token response is not compliant to the spec. https://datatracker.ietf.org/doc/html/rfc8693#name-successful-response We must include `access_token`, `issued_token_type `, and `token_type`.

> access_token
REQUIRED. The security token issued by the authorization server in response to the token exchange request. The access_token parameter from [Section 5.1](https://www.rfc-editor.org/rfc/rfc6749#section-5.1) of [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)] is used here to carry the requested token, which allows this token exchange protocol to use the existing OAuth 2.0 request and response constructs defined for the token endpoint. The identifier access_token is used for historical reasons and the issued token need not be an OAuth access token.

Although the key is `access_token`, it does not have to be an OAuth access token. In our case the token response should be

```
{
  "access_token": "DEVICE_BROWSER_SESSION_TOKEN",
  "issued_token_type": "urn:authgear:params:oauth:token-type:device-browser-session-token",
  "token_type": "DPoP"
}
```

Note that the `token_type` is `DPoP`, if DPoP is used. Otherwise, it is `Bearer`. This also applies to any `token_type` we return in other kind of Token responses.

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+{
+  "device_secret":"casdfgarfgasdfg",
+  "x_device_browser_session_token": "zxczxczxc",
+  "expires_in":300
+}
+```
+
+## Exchanging for browser cookie of a session with the authorization endpoint
+
+A web browser can open the /authorize endpoint to exchange for a valid browser session cookie with the following parameters:
```
louischan-oursky, line 124: "open the authorization endpoint"

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+
+- `client_id`: Required. The client_id of the obtained session.
+- `id_token_hint`: Required. The id_token obtained by the mobile app.
+- `x_device_browser_session_token`: Required. The device browser session token obtained from token exchange response.
+- `scope`: Optional. Scope of the session of the browser session cookie. If not provided, it will be same as the scope of the session which generates this browser session cookie (i.e. The offline grant of the mobile app).
+- `prompt`: Optional. If provided, must be `none`.
+- `response_type`: Required. Must be one of the following values:
+  - `session`: A long lived token will be set in the cookie, which shares the same lifetime of the refresh_token. However, it will not live longer than an IDP session.
+  - `token`: A short lived token will be set in the cookie, which shares the same lifetime of an access_token.
```
louischan-oursky, line 133: Implementation-wise is to better to require the client to specify it is performing App-initiated SSO to Web. So `response_type` should either be `app_initiated_sso_to_web session` or `app_initiated_sso_to_web token`.

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Exchanging for browser cookie of a session with the authorization endpoint
+
+A web browser can open the /authorize endpoint to exchange for a valid browser session cookie with the following parameters:
+
+- `client_id`: Required. The client_id of the obtained session.
+- `id_token_hint`: Required. The id_token obtained by the mobile app.
+- `x_device_browser_session_token`: Required. The device browser session token obtained from token exchange response.
+- `scope`: Optional. Scope of the session of the browser session cookie. If not provided, it will be same as the scope of the session which generates this browser session cookie (i.e. The offline grant of the mobile app).
+- `prompt`: Optional. If provided, must be `none`.
```
louischan-oursky, line 130: Make it required for `response_type` including `app_initiated_sso_to_web`

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+
+device_secret and x_device_browser_session_token should always be bound to a device.
+
+[OAuth 2.0 Demonstrating Proof of Possession](https://datatracker.ietf.org/doc/html/rfc9449) is implemented for such binding.
+
+The DPoP Proof MAY be provided when making requests to the /token endpoint. If DPoP Proof is provided, device_secret and x_device_browser_session_token returned in the response will be bound to the public key provided in the DPoP Proof.
+
+## Changes In SDKs
+
+- `authenticate()` will accept a new option `isSSOToWebEnabled: boolean`. When `true`, the `x_device_browser_sso` will be added to the authorize request.
```
louischan-oursky, line 152: `isAppInitiatedSSOToWebEnabled`. Stay consistent with the term.

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+- `constructAuthenticatedURL(): URL` will be added.
+  - Parameters:
+    - `redirectURI: URI`: The uri you want to open in authenticated state.
```
louischan-oursky, line 155: Client ID is required. The client ID of the web app.

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,179 @@
...
+device_secret and x_device_browser_session_token should always be bound to a device.
+
+[OAuth 2.0 Demonstrating Proof of Possession](https://datatracker.ietf.org/doc/html/rfc9449) is implemented for such binding.
+
+The DPoP Proof MAY be provided when making requests to the /token endpoint. If DPoP Proof is provided, device_secret and x_device_browser_session_token returned in the response will be bound to the public key provided in the DPoP Proof.
+
+## Changes In SDKs
+
+- `authenticate()` will accept a new option `isSSOToWebEnabled: boolean`. When `true`, the `x_device_browser_sso` will be added to the authorize request.
+- `constructAuthenticatedURL(): URL` will be added.
```
louischan-oursky, line 153: `makeAppInitiatedSSOToWebURL`.

docs/specs/app-initiated-sso-to-web.md
tung2744: Are we going to rename it to `x_app_initiated_sso_to_web`?

docs/specs/app-initiated-sso-to-web.md
tung2744: As the sdk is in the app, mentioning "AppInitiated" looks a bit redundant, but I understand that it is easier to for the user to find the config if they have the same name.

docs/specs/app-initiated-sso-to-web.md
tung2744: Let me name it `x_device_browser_session_token` for now

louischan-oursky: When we correct our Token Response, we include the token in the key `access_token`.

The only remaining place `x_device_browser_session_token` will appear is in the authorization endpoint. When I think it again, it is not a good practice to have side effects in GET request. It is more appropriate to have side effects in POST request.

Ideally, we should perform two token exchanges.

1. Token Exchange 1: id token + device secret -> a short live token
2. Token Exchange 2: the short live token -> cookie

Token Exchange 2 has to happen in the webview. But this design is not compatible with ASWebAuthenticationSession. But if we assume App-initiated-sso-to-web is only available for webview, then it is not a problem. We can offer a method in the SDK to return a script to be run in the webview to perform Token Exchange 2.

tung2744:
Updated for all comments above.

docs/specs/app-initiated-sso-to-web.md
tung2744: >  When I think it again, it is not a good practice to have side effects in GET request.

However, the /authorize endpoint always has side effects such as setting oauth session cookies. I think setting the device browser session in this endpoint is more correct.
And actually `/token` never handles cookies.

> But if we assume App-initiated-sso-to-web is only available for webview

I think this assumption may not be correct, as usecase 1 probably want to open the marketing site in a normal browser.

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,182 @@
...
+sequenceDiagram
+    participant app1 as Mobile App #35;1
+    participant Web as Web App #35;2 <br/>(Cookie Based)<br/>https://www.example.com
+    participant Authgear as Authgear <br/> https://auth.example.com
+    app1->>+Authgear: [1]GET https://auth.example.com/oauth2/authorize<br/>client_id=1<br/>scope=x_device_browser_session_token<br/>redirect_uri=app://redirect
+    Note right of Authgear: Alice enter user credentials
+    Authgear->>-app1: [2]Redirect to app://redirect?code=CODE
+    app1->>+Authgear: [3]POST https://auth.example.com/oauth2/token<br/>code=CODE
+    Authgear->>-app1: [4]Response with<br/>access_token=ACCESS_TOKEN<br/>refresh_token=REFRESH_TOKEN<br/>id_token=ID_TOKEN<br/>device_secret=DEVICE_SECRET
+    app1->>+Authgear: [5]POST https://auth.example.com/oauth2/token<br/>grant_type=token-exchange<br/>client_id=2<br/>subject_token_type=urn:ietf:params:oauth:token-type:id_token<br/>subject_token=ID_TOKEN<br/>actor_token_type=urn:x-oath:params:oauth:token-type:device-secret<br/>actor_token=DEVICE_SECRET<br/>requested_token_type=urn:x-oath:params:oauth:token-type:device-browser-session-token
```
louischan-oursky, line 34: - scope: `https://authgear.com/scopes/app-initiated-sso-to-web`
- token identifier: `...:app-initiated-sso-to-web-token`
- parameter name in authorization endpoint: `x_app_initiated_sso_to_web_token`

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,182 @@
...
+urn:authgear:params:oauth:token-type:device-browser-session-token
+```
+
+This token type identifier refers to the device browser session token mentioned in this spec. It can be used as the `requested_token_type` parameter in token exchange request.
+
+### Token Exchange Request
+
+When a mobile app wants to perform App-initiated SSO to Web, it makes a request to the token endpoint to obtain a device browser session token with the following parameters:
+
+- `grant_type`: Must be `urn:ietf:params:oauth:grant-type:token-exchange`
```
tung2744, line 79: Add `scope` here, similar to native sso token exchange request.
https://openid.net/specs/openid-connect-native-sso-1_0.html#name-token-exchange-request

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+## Exchanging for browser cookie of a session with the authorization endpoint
+
+A web browser can open the authorization endpoint endpoint to exchange for a valid browser session cookie with the following parameters:
+
+- `client_id`: Required. The client_id of the obtained session.
+- `id_token_hint`: Required. The id_token obtained by the mobile app.
+- `x_device_browser_session_token`: Required. The device browser session token obtained from token exchange response.
+- `scope`: Optional. Scope of the session of the browser session cookie. If not provided, it will be same as the scope of the session which generates this browser session cookie (i.e. The offline grant of the mobile app).
```
tung2744, line 131: No scope here, because it was specified in previous token exchange step already.

docs/specs/app-initiated-sso-to-web.md
```diff
@@ -0,0 +1,182 @@
...
+
+device_secret and x_device_browser_session_token should always be bound to a device.
+
+[OAuth 2.0 Demonstrating Proof of Possession](https://datatracker.ietf.org/doc/html/rfc9449) is implemented for such binding.
+
+The DPoP Proof MAY be provided when making requests to the /token endpoint. If DPoP Proof is provided, device_secret and x_device_browser_session_token returned in the response will be bound to the public key provided in the DPoP Proof.
+
+## Changes In SDKs
+
+- `authenticate()` will accept a new option `isAppInitiatedSSOToWebEnabled: boolean`. When `true`, the `x_device_browser_session_token` scope will be added to the authorize request.
```
tung2744, line 154: Add `isAppInitiatedSSOToWebEnabled` in `configure()` instead

tung2744:
Rename the whole feature to Pre-Authenticated URL

tung2744:
@louischan-oursky Updated the spec:
1. Renamed the feature.
2. Included native SSO spec.
3. Included DPoP spec.

docs/specs/pre-authenticated-url.md
```diff
@@ -0,0 +1,193 @@
...
+
+A web browser can open the authorization endpoint endpoint to exchange for a valid browser session cookie with the following parameters:
+
+- `client_id`: Required. The client_id of the obtained session.
+- `id_token_hint`: Required. The id_token obtained by the mobile app.
+- `x_pre_authenticated_url_token`: Required. The Pre-Authenticated URL Token obtained from token exchange response.
+- `scope`: Optional. Scope of the session of the browser session cookie. If not provided, it will be same as the scope of the session which generates this browser session cookie (i.e. The offline grant of the mobile app).
+- `prompt`: Required. The value must be `none`.
+- `response_type`: Required. Must be one of the following values:
+  - `app_initiated_sso_to_web token`: An access token will be set in the cookie, which shares the same session of the refresh_token of the native mobile app.
```
louischan-oursky, line 140: This still references the old name.

docs/specs/pre-authenticated-url.md
```diff
@@ -0,0 +1,193 @@
...
+
+Step [7] uses the obtained Pre-Authenticated URL Token to open /authorize in a browser, with the id_token used in step [5], and a redirect_uri pointing to Web App #2.
+
+Step [8] redirects the browser to Web App #2, with a with a cookie representing the user session set. The cookie can then be used for authorizing Web App #2 as it was logged in as the same user of Mobile App #1.
+
+## Native App Authorization Extensions
+
+Same as the [Native SSO](https://openid.net/specs/openid-connect-native-sso-1_0.html#name-native-app-authorization-ex). With a new supported scope `https://authgear.com/scopes/pre-authenticated-url`.
+
+## Token Exchange Profile for App Initiated SSO to Web
```
louischan-oursky, line 56: This still references the old name.

louischan-oursky: @tung2744 In case you missed my comments.

tung2744: @louischan-oursky Sorry forgot to update. Just added a new commit to fix it. Thanks!

louischan-oursky (approved):

PR #4323, opened 2024-06-13
Fix resources/portal/static/img/* were excluded from docker build

louischan-oursky:
The bug

SCR-20240613-ocvb

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,8 +9,11 @@
 **/__pycache__
 **/venv
 **/var
-**/resources/**/generated
-**/resources/portal/static/*
+
+resources/**/generated
+resources/portal/static/*
+!resources/portal/static/img/*
+
 **/__generated_asset.html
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -9,8 +9,11 @@
 **/__pycache__
 **/venv
 **/var
-**/resources/**/generated
-**/resources/portal/static/*
+
+resources/**/generated
+resources/portal/static/*
+!resources/portal/static/img/*
```
tung2744, line 15: Oh why I thought the build pipeline should generate the images?

louischan-oursky: The images inside img is static.

louischan-oursky: The real problem is we use the same directory to contain generated assets and static assets. But that is another issue.

tung2744 (approved):

PR #4332, opened 2024-06-13
Give some sensible timeout to e2e clients

tung2744:
Lets see what exactly is timed out


tung2744:
It seems more difficult to know what is happening, will close first.

Issue #4325, opened 2024-06-13
[Captcha] Implement the new captcha configuration


louischan-oursky:
* \[ \] Keep the old configuration intact
* \[ \] Introduce the new configuration
  * The Go structs
  * The JSON schema
  * Write tests to test for valid and invalid configurations

pkong-ds:
> Keep the old configuration intact

old config use key `captcha`, which conflicts with new config key
Maybe mark old one as `captcha_legacy`?

louischan-oursky:
No. You add new fields to it directly. Just leave the legacy `provider` unchanged.

PR #4337, opened 2024-06-14
Implement new captcha configuration

pkong-ds:
ref #4325 

@louischan-oursky thanks, copied items from issue below

>  Keep the old configuration intact
>  Introduce the new configuration
>>        The Go structs
>>        The JSON schema
>>        Write tests to test for valid and invalid configurations

### DIFF
--- a/pkg/lib/config/captcha.go
+++ b/pkg/lib/config/captcha.go
@@ -5,14 +5,42 @@ var _ = Schema.Add("CaptchaConfig", `
 	"type": "object",
 	"additionalProperties": false,
 	"properties": {
+		"enabled": { "type": "boolean" },
+		"providers": {
+			"type": "array",
+			"items": { "$ref": "#/$defs/CaptchaProvider" }
+		},
 		"provider": { "$ref": "#/$defs/LegacyCaptchaProvider" }
+	}
+}
+`)
+
+var _ = Schema.Add("CaptchaProvider", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"required": ["type", "alias"],
+	"properties": {
+		"type": { "type": "string" },
+		"alias": { "type": "string" },
+		"site_key": { "type": "string" }
 	}
 }
 `)
 
 type CaptchaConfig struct {
 	LegacyProvider *LegacyCaptchaProvider `json:"provider,omitempty"`
+	Enabled        bool                   `json:"enabled,omitempty"`
+	Providers      []*CaptchaProvider     `json:"providers,omitempty"`
 }
+
+type CaptchaProvider struct {
+	Type    string `json:"type,omitempty"`
+	Alias   string `json:"alias,omitempty"`
+	SiteKey string `json:"site_key,omitempty"` // only for some providers
+}
+
+// legacy code below
 
 var _ = Schema.Add("LegacyCaptchaProvider", `
 {
--- a/pkg/lib/config/captcha.go
+++ b/pkg/lib/config/captcha.go
@@ -21,7 +21,7 @@ var _ = Schema.Add("CaptchaProvider", `
 	"additionalProperties": false,
 	"required": ["type", "alias"],
 	"properties": {
-		"type": { "type": "string" },
+		"type": { "type": "string", "enum": ["cloudflare", "recaptchav2"] },
 		"alias": { "type": "string" },
 		"site_key": { "type": "string" }
 	}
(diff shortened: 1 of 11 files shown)


pkg/lib/config/captcha.go
```diff
@@ -5,24 +5,59 @@ var _ = Schema.Add("CaptchaConfig", `
...
+
+var _ = Schema.Add("CaptchaProvider", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"required": ["type", "alias"],
+	"properties": {
+		"type": { "type": "string", "enum": ["cloudflare", "recaptchav2"] },
+		"alias": { "type": "string" },
+		"site_key": { "type": "string" }
```
pkong-ds, line 26: Actually am not very sure about this one: Is there a way to type force additionaled fields based on `type`?

louischan-oursky: Yes, you use allOf, if, then. An example is pkg/lib/config/oauth.go:145

pkong-ds: I see! https://json-schema.org/understanding-json-schema/reference/combining#allOf

pkong-ds: But on golang level, how should `type force struct property based on other struct property` be done?

```
type Pet struct {
  Type PetType `json"type,omitempty"`
}

type PetType string

const (
  PetTypeDog PetType = "dog"
  PetTypeCat PetType = "cat"
)

type PetDog interface {
  Bark() nil
}

type PetCat interface {
  Meow() nil
}

// how should Bark() / Meow() method be required based on Pet.PetType?
// Actually, should dog and cat be their own structs, while pet be an interface?
// But if we implement dog and cat as their own structs, we would need *union* in another struct below? this makes things more complicated?

type Config struct {
  Pet (Dog | Cat) `json"pet"`
}
```

louischan-oursky: You don't, and we don't. See pkg/lib/config/authentication_flow.go:893 The struct contains all fields, only some of them are used in a particular case.

pkong-ds:
Missed secrets, lemme add them 🙏 

```
 [authgear.secrets.yaml](#authgearsecretsyaml)
      - [`type: cloudflare`](#type-cloudflare-1)
      - [`type: recaptchav2`](#type-recaptchav2-1)
```

pkg/lib/config/captcha.go
```diff
@@ -5,24 +5,24 @@ var _ = Schema.Add("CaptchaConfig", `
...
 	"properties": {
-		"provider": { "$ref": "#/$defs/CaptchaProvider" }
+		"provider": { "$ref": "#/$defs/LegacyCaptchaProvider" }
 	}
 }
 `)
 
 type CaptchaConfig struct {
-	Provider *CaptchaProvider `json:"provider,omitempty"`
+	LegacyProvider *LegacyCaptchaProvider `json:"provider,omitempty"`
```
louischan-oursky, line 47: - Rename the field LegacyProvider to Deprecated_Provider
- Rename the type LegacyCaptchaProvider to Deprecated_CaptchaProvider

pkg/lib/config/secret.go
```diff
@@ -265,8 +265,8 @@ const (
 	// nolint: gosec
 	OAuthClientCredentialsKey SecretKey = "oauth.client_secrets"
 	// nolint: gosec
-	CaptchaCloudflareCredentialsKey  SecretKey = "captcha.cloudflare"
-	WhatsappOnPremisesCredentialsKey SecretKey = "whatsapp.on-premises"
+	LegacyCaptchaCloudflareCredentialsKey SecretKey = "captcha.cloudflare"
```
louischan-oursky, line 271: Rename LegacyCaptchaCloudflareCredentialsKey to Deprecated_CaptchaCloudflareCredentialsKey

pkg/lib/config/secret_captcha.go
```diff
@@ -11,11 +11,11 @@ var _ = SecretConfigSchema.Add("CaptchaCloudflareCredentials", `
 }
 `)
 
-type CaptchaCloudflareCredentials struct {
+type LegacyCaptchaCloudflareCredentials struct {
```
louischan-oursky, line 80: Rename CaptchaCloudflareCredentials to Deprecated_CaptchaCloudflareCredentials

pkong-ds:

louischan-oursky (approved):

pkg/lib/config/secret_captcha.go
```diff
@@ -57,12 +61,11 @@ const (
...
 )
 
-func (c *CaptchaProvidersCredentials) SensitiveStrings() []string {
-	sensitiveStrings := make([]string, len(c.Items))
-	for i, cred := range c.Items {
-		sensitiveStrings[i] = cred.SecretKey
-	}
-	return sensitiveStrings
+func (c *CaptchaProvidersCredentials) SensitiveStrings() (sensitiveStrings []string) {
+	sensitiveStrings = slice.Map(c.Items, func(item CaptchaProvidersCredentialsItem) string {
```
louischan-oursky, line 65: The original pattern is OK. No need to rewrite.

Issue #4334, opened 2024-06-13
Move "Custom Email Provider" into "Advanced"


fungc-io:
Move the page in the Nav bar into "Advanced"

Resulting nav bar:

* …
* Branding
  * Design
  * Email/SMS Template
  * Custom Domains
  * Custom Text
* Languages
* User Profile
  * …
* Integrations
* Billing
* Advanced
  * Hook
  * Admin API
  * Account Deletion
  * Account Anonymization
  * Cookie Lifetime
  * Custom Email Provider
* Audit Log
* Portal Admins

fungc-io:
checked ok

PR #4341, opened 2024-06-14
Restructure localisation resources setup pages

cychiuae:
ref #4290 
ref #4291 
ref #4334 

ref DEV-1381
ref DEV-1382
ref DEV-1418

@tung2744

### DIFF
--- a/portal/src/AppRoot.tsx
+++ b/portal/src/AppRoot.tsx
@@ -545,14 +545,6 @@ const AppRoot: React.VFC = function AppRoot() {
               </Route>
             </Route>
             <Route
-              path="smtp"
-              element={
-                <Suspense fallback={<ShowLoading />}>
-                  <SMTPConfigurationScreen />
-                </Suspense>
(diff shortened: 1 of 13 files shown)


tung2744:
https://github.com/authgear/authgear-server/assets/34942210/e3f05124-d775-464a-ba65-55573f1a7a04

Looks like the color picker preview isn't working?

tung2744:
I checked in the main branch and looks like existing issue, maybe we create another issue to handle it.

cychiuae:
> I checked in the main branch and looks like existing issue, maybe we create another issue to handle it.

Seems like only background color picker doesn't work

cychiuae:
> I checked in the main branch and looks like existing issue, maybe we create another issue to handle it.

Production also has this problem. Let me create a issue for it.

cychiuae:
Created #4346 to follow up

tung2744 (approved):

PR #4344, opened 2024-06-17
Provide company name in project name

toshinari123:
https://linear.app/authgear/issue/DEV-1378/use-company-name-for-new-project-name

| in onboarding survey | in create project |
| - | - |
| image | image |

### DIFF
--- a/portal/src/OnboardingSurveyScreen.tsx
+++ b/portal/src/OnboardingSurveyScreen.tsx
@@ -723,8 +723,13 @@ function Step4(_props: StepProps) {
       e.stopPropagation();
       setLocalStorage("reasonChoices", reasonChoicesState);
       if (otherReason !== "") setLocalStorage("otherReason", otherReason);
+      const companyName = getFromLocalStorage("companyName");
       localStorage.removeItem(localStorageKey);
-      navigate("./../../projects/create");
+      if (companyName !== undefined)
+        navigate("./../../projects/create", {
+          state: { company_name: companyName },
+        });
+      else navigate("./../../projects/create");
     },
     [navigate, reasonChoicesState, otherReason]
   );
--- a/portal/src/graphql/portal/CreateProjectScreen.tsx
+++ b/portal/src/graphql/portal/CreateProjectScreen.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useContext, useEffect } from "react";
-import { useNavigate } from "react-router-dom";
+import { useNavigate, useLocation } from "react-router-dom";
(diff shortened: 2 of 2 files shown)


portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -723,8 +723,13 @@ function Step4(_props: StepProps) {
       e.stopPropagation();
       setLocalStorage("reasonChoices", reasonChoicesState);
       if (otherReason !== "") setLocalStorage("otherReason", otherReason);
+      const companyName = getFromLocalStorage("companyName");
       localStorage.removeItem(localStorageKey);
-      navigate("./../../projects/create");
+      if (companyName !== undefined)
+        navigate("./../../projects/create", {
+          state: { company_name: companyName },
```
tung2744, line 730: Use `replace: true` to replace the current history entry?

toshinari123: Is it to prevent users from clicking back and resubmitting the survey? I tried `replace: true`, but the user is still able to go back to the survey; I think what is happening is that since submitting replaced step 4, pressing back returned to step 3, and since it detected localstorage is empty it redirects to step 1

tung2744: Yes you are right.
Ok, so we allow the user to back to the survey by using back, and it will start from step 1.

portal/src/graphql/portal/CreateProjectScreen.tsx
```diff
@@ -61,13 +61,31 @@ interface CreateProjectScreenContentProps {
...
+function processCompanyName(company_name: string): string {
+  return company_name
+    .trim()
+    .split("")
+    .filter((char) => /[a-zA-Z\s]/.exec(char))
+    .join("")
+    .split(" ")
+    .filter((word) => word !== "")
+    .join("-")
+    .toLowerCase();
```
tung2744, line 77: Use camelCase as parameter name?

portal/src/graphql/portal/CreateProjectScreen.tsx
```diff
@@ -78,10 +96,14 @@ function CreateProjectScreenContent(props: CreateProjectScreenContentProps) {
     [createApp]
   );
 
+  const defaultState = makeDefaultState();
+  if (typedState)
+    defaultState.appID = processCompanyName(typedState.company_name);
```
tung2744, line 101: use `useMemo` to memoize the object.

```
const defaultState = useMemo(() => {
  ...

}, [..])
```

tung2744 (approved):

PR #4353, opened 2024-06-18
Add sentry tracking with debug info for csrf errors

elise-ng:
Ref DEV-1337

Sentry log from test project:

CleanShot 2024-06-18 at 21 37 36@2x

(sanity check by removing debug cookie manually in browser)

CleanShot 2024-06-18 at 21 39 38@2x

Note the referer field is auto masked by sentry, and seems override is not working (see linear discussion)

CleanShot 2024-06-18 at 21 38 02@2x

### DIFF
--- a/pkg/auth/webapp/csrf.go
+++ b/pkg/auth/webapp/csrf.go
@@ -1,6 +1,12 @@
 package webapp
 
-import "github.com/authgear/authgear-server/pkg/lib/config"
+import (
+	"net/http"
+
+	"github.com/authgear/authgear-server/pkg/lib/config"
+	"github.com/authgear/authgear-server/pkg/util/duration"
+	"github.com/authgear/authgear-server/pkg/util/httputil"
+)
 
 // CSRFFieldName is the same as the default, but public.
 const CSRFFieldName = "gorilla.csrf.Token"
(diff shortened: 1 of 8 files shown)


elise-ng:
@louischan-oursky plz check first, i'll see if there is any way to keep the referer info tmr

pkg/auth/webapp/csrf.go
```diff
@@ -21,3 +27,44 @@ func NewCSRFCookieDef(cfg *config.HTTPConfig) CSRFCookieDef {
 
 	return def
 }
+
+type CSRFDebugMiddleware struct {
+	Cookies CookieManager
+}
+
+var CSRFDebugCookieMaxAge = int(duration.Short.Seconds())
```
louischan-oursky, line 35: Use duration.UserInteraction, same as the actual CSRF cookie.

pkg/util/sentry/request.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 var HeaderWhiteList = []string{
+	"Origin",
 	"Referer",
```
louischan-oursky, line 11: Try to add it to context in the sentry hook.

pkg/auth/webapp/csrf_middleware.go
```diff
@@ -82,11 +82,19 @@ func (m *CSRFMiddleware) unauthorizedHandler(w http.ResponseWriter, r *http.Requ
 	strictCookie, err := m.Cookies.GetCookie(r, CSRFDebugCookieSameSiteStrictDef)
 	hasStrictCookie := (err == nil && strictCookie.Value == "exists")
 
+	csrfCookie, _ := r.Cookie(m.CookieDef.Name)
+	csrfCookieSizeInBytes := 0
+	if csrfCookie != nil {
+		// do not return value but length only for debug.
+		csrfCookieSizeInBytes = len([]byte(csrfCookie.Value))
+	}
```
louischan-oursky, line 90: The format is documented here https://github.com/gorilla/securecookie/blob/main/securecookie.go#L281 Try to mask the value and the sig, and include it as well.

pkg/auth/webapp/csrf.go
elise-ng: ok

pkg/util/sentry/request.go
elise-ng: seems it would work with sentry allow list actually after some time, re: https://linear.app/authgear/issue/DEV-1337/track-csrf-errors-in-sentry#comment-be44f87b

elise-ng: let's keep the info here instead?

pkg/auth/webapp/csrf_middleware.go
elise-ng: https://github.com/gorilla/securecookie/blob/eae3c1840ec4adda88a4af683ad0f60bb690e7c2/securecookie.go#L320C30-L320C44

seems it strips the name before sending and add it back only during decode, so the payload format is `date|value|mac`

let's mask the value and mac here?

elise-ng: CleanShot 2024-06-19 at 21 46 52@2x

pkg/util/sentry/request.go
louischan-oursky: Sure.

pkg/auth/webapp/csrf_middleware.go
louischan-oursky: OK.

louischan-oursky:
I added a commit to fix a lint error.

louischan-oursky (approved):

PR #4343, opened 2024-06-17
Use posthog in onboarding survey and update custom attributes

toshinari123:
https://linear.app/authgear/issue/DEV-1367/submit-and-store-survey-answers

 - follow the naming in https://linear.app/authgear/document/pitch-onboarding-survey-c6236373e6e5
 - added Object type to gtm value type
 - if phone number is invalid user can still submit, will result in posthog having no phone number field

posthog screenshots:
image
image

PR #4354, opened 2024-06-18
Change navigation uri in OnboardingRedirect

toshinari123:
#4345 

https://linear.app/authgear/issue/DEV-1422/redirect-new-users-to-onboarding-survey-after-signing-up

demo: (note my local branch was not up to date, so company name was not filled in project name)

https://github.com/authgear/authgear-server/assets/68701054/06807bdc-f248-44e6-a22a-59d3d9130ce6

### DIFF
--- a/pkg/portal/graphql/context.go
+++ b/pkg/portal/graphql/context.go
@@ -107,7 +107,7 @@ type TutorialService interface {
 
 type OnboardService interface {
 	SubmitOnboardEntry(entry model.OnboardEntry, actorID string) error
-	CheckCompletion(actorID string) (bool, error)
+	CheckOnboardingSurveyCompletion(actorID string) (bool, error)
 }
 
 type AnalyticChartService interface {
--- a/pkg/portal/graphql/context.go
+++ b/pkg/portal/graphql/context.go
@@ -107,6 +107,7 @@ type TutorialService interface {
 
 type OnboardService interface {
 	SubmitOnboardEntry(entry model.OnboardEntry, actorID string) error
+	CheckCompletion(actorID string) (bool, error)
 }
 
 type AnalyticChartService interface {
--- a/pkg/portal/graphql/viewer.go
+++ b/pkg/portal/graphql/viewer.go
@@ -28,12 +28,6 @@ var viewerSubresolver = func(gqlCtx *Context, id string) (interface{}, error) {
 		user.GeoIPCountryCode = geoipInfo.CountryCode
 	}
 
-	isCompleted, err := gqlCtx.OnboardService.CheckCompletion(id)
-	if err != nil {
-		return nil, err
-	}
-	user.IsOnboardingSurveyCompleted = isCompleted
-
 	return user, nil
 }
 
@@ -63,6 +57,15 @@ var nodeViewer = node(
 			},
 			"isOnboardingSurveyCompleted": &graphql.Field{
 				Type: graphql.Boolean,
+				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
+					user := p.Source.(*model.User)
+					gqlCtx := GQLContext(p.Context)
+					isCompleted, err := gqlCtx.OnboardService.CheckOnboardingSurveyCompletion(user.ID)
+					if err != nil {
+						return nil, err
+					}
+					return isCompleted, nil
+				},
 			},
 		},
 	}),
--- a/pkg/portal/graphql/viewer.go
+++ b/pkg/portal/graphql/viewer.go
@@ -14,6 +14,29 @@ import (
 
 const typeViewer = "Viewer"
 
+var viewerSubresolver = func(gqlCtx *Context, id string) (interface{}, error) {
+	userIface, err := gqlCtx.Users.Load(id).Value()
+	if err != nil {
+		return nil, err
+	}
+
+	user := userIface.(*model.User)
+
+	requestIP := httputil.GetIP(gqlCtx.Request, bool(gqlCtx.TrustProxy))
+	geoipInfo, ok := geoip.DefaultDatabase.IPString(requestIP)
+	if ok {
+		user.GeoIPCountryCode = geoipInfo.CountryCode
+	}
+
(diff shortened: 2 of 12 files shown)


portal/src/OnboardingRedirect.tsx
```diff
@@ -20,7 +20,7 @@ const OnboardingRedirect: React.VFC = function OnboardingRedirect() {
     if (apps && apps.length > 0) {
       navigate("/");
     } else {
-      navigate("/projects/create");
+      navigate("/onboarding-survey");
```
tung2744, line 23: You need to check the custom attribute to determine if the user needs to fill the survey or not.

toshinari123: Should I create a new query in the backend as well? I think it is currently not possible to query the viewer's custom attributes currently in portal. On the other hand, if we keep the current implementation, the only possible issue is that the first-time user skips the onboarding survey by changing the URL to `.../projects/create`, which is highly unlikely.

toshinari123: Actually I guess they can get out in the future if the 404 page (that doesnt exist currently) contains a link back to `/projects` page, and the user tried to skip the onboarding survey by typing random things into the URL, going to the 404 page.

tung2744: > Should I create a new query in the backend as well? I think it is currently not possible to query the viewer's custom attributes currently in portal

I have the following suggestion:

1. Modify the `Viewer` node of the portal api. Add one field `isOnboardingSurveyCompleted: boolean`.
2. In the resolver of `Viewer`, query the user's custom attribute. Check if the survey is completed by existence of that attribute, and compute the corresponding value of `isOnboardingSurveyCompleted`.
3. In frontend, query and check `isOnboardingSurveyCompleted` in the viewer api.

toshinari123: Ooh ok thanks for your helpful suggestion, will follow

pkg/portal/graphql/context.go
```diff
@@ -107,6 +107,7 @@ type TutorialService interface {
 
 type OnboardService interface {
 	SubmitOnboardEntry(entry model.OnboardEntry, actorID string) error
+	CheckCompletion(actorID string) (bool, error)
```
tung2744, line 110: Lets be clear what we completed on the method name: `CheckOnboardingSurveyCompletion`.

pkg/portal/model/user.go
```diff
@@ -1,10 +1,11 @@
 package model
 
 type User struct {
```
tung2744, line 3: We don't need to modify the user model.

pkg/portal/graphql/viewer.go
```diff
@@ -38,6 +61,9 @@ var nodeViewer = node(
 			"geoIPCountryCode": &graphql.Field{
 				Type: graphql.String,
 			},
+			"isOnboardingSurveyCompleted": &graphql.Field{
```
tung2744, line 64: Add a field specific resolver here instead of modifying the user model:

```
	source := p.Source.(*model.User)
	gqlCtx := GQLContext(p.Context)
	// ...
	return isOnboardingSurveyCompleted, nil
},
```

portal/src/OnboardingRedirect.tsx
```diff
@@ -3,11 +3,13 @@ import { useNavigate } from "react-router-dom";
 import ShowLoading from "./ShowLoading";
 import ShowError from "./ShowError";
 import { useAppListQuery } from "./graphql/portal/query/appListQuery";
+import { useViewerQuery } from "./graphql/portal/query/viewerQuery";
 
 const OnboardingRedirect: React.VFC = function OnboardingRedirect() {
   const { loading, error, apps, refetch } = useAppListQuery();
 
   const navigate = useNavigate();
+  const { viewer } = useViewerQuery();
```
tung2744, line 12: We need to handle `loading` of viewer query. In the `useEffect`, do not do the redirect before viewer is loaded.

portal/src/OnboardingRedirect.tsx
```diff
@@ -16,13 +18,15 @@ const OnboardingRedirect: React.VFC = function OnboardingRedirect() {
     if (error != null) {
       return;
     }
-    // redirect to create apps if user doesn't have any apps
-    if (apps && apps.length > 0) {
-      navigate("/");
+    if (
+      (apps === null || apps.length === 0) &&
+      !viewer?.isOnboardingSurveyCompleted
```
tung2744, line 23: If viewer is null (`viewer == null`) and is viewer query is still loading, do nothing.

portal/src/OnboardingRedirect.tsx
```diff
@@ -3,33 +3,60 @@ import { useNavigate } from "react-router-dom";
...
     }
-    // redirect to create apps if user doesn't have any apps
-    if (apps && apps.length > 0) {
-      navigate("/");
+    if (viewer === undefined || viewer === null) {
+      return;
+    }
+    if (
+      (apps === null || apps.length === 0) &&
+      viewer.isOnboardingSurveyCompleted
```
tung2744, line 35: Should be `!viewer.isOnboardingSurveyCompleted`.

toshinari123:
During my local testing, if a new user tries to access accounts.portal.localhost and sign up, they will be redirected to account settings, and cannot go to the portal. Is this a concern or does this have a possibility of happening in production?

tung2744:
Do you mean going to `accounts.portal.localhost` directly? If yes, it is expected that it won't back to portal, because the login is not triggered by portal.

tung2744:
@toshinari123 Sorry one problem remaining.
When testing together with https://github.com/authgear/authgear-server/pull/4343#pullrequestreview-2137443329
I found that `viewer.isOnboardingSurveyCompleted` is not updated after I completed the survey, because the query isn't refetching.
Therefore please refetch the viewer query after `updateCustAttrHook` called.

tung2744 (approved):

PR #4355, opened 2024-06-18
Fix incorrect deprecated captcha credential key

tung2744:
In the current main, the server will crash if we have the legacy captcha credential specified.

@pkong-ds


tung2744:
If seems there is a vulnerability in the `ws` package, which is requried by `ethers`. But we cannot update `ethers` because it was required by `siwe` which we cannot update.

tung2744:
Replace by https://github.com/authgear/authgear-server/pull/4356

I will open another pr for the package update

PR #4356, opened 2024-06-18
Add test cases to check captcha credential 

pkong-ds:
## ~Blocked by~ Supercedes
- #4355


louischan-oursky (commented):
- Check uniqueness of alias. (Refer to OAuth provider)
- Check secret type-specific fields.
- Add a test case: Both legacy and new configuration are present

louischan-oursky (approved):

PR #4282, opened 2024-06-11
Specify Captcha support

louischan-oursky:
Depend on https://github.com/authgear/authgear-server/pull/4399

PR #4358, opened 2024-06-18
Implement bot protection in auth flow handlers

pkong-ds:
## Blocked by
- #4382 

## Notes
This PR only address 3 out of 4 items of DEV-1413
- [x] ~Output captcha information when captcha is enabled in the flow response.~ Outdated, ref https://github.com/authgear/authgear-server/pull/4282#discussion_r1646846518
- [x] Output captcha requirement in branches.
- [x] Store captcha verification status in authenticationflow.Session
- [ ] Enforce captcha in branches. That is, when accepting input in a branch that requires captcha, it should check is the authflow session has captcha verification passed. If not, return the error as specified in the spec.` 

Why? - to prevent conflict of 4th point with another PR #4372

### DIFF
--- a/pkg/lib/authenticationflow/declarative/data_identification.go
+++ b/pkg/lib/authenticationflow/declarative/data_identification.go
@@ -25,6 +25,7 @@ func (IdentificationData) Data() {}
 type IdentificationOption struct {
 	Identification config.AuthenticationFlowIdentification `json:"identification"`
 
+	BotProtection *BotProtectionData `json:"bot_protection,omitempty" nullable:"true"`
 	// ProviderType is specific to OAuth.
 	ProviderType string `json:"provider_type,omitempty"`
 	// Alias is specific to OAuth.
@@ -36,24 +37,27 @@ type IdentificationOption struct {
 	RequestOptions *model.WebAuthnRequestOptions `json:"request_options,omitempty"`
 }
 
-func NewIdentificationOptionIDToken(i config.AuthenticationFlowIdentification) IdentificationOption {
+func NewIdentificationOptionIDToken(i config.AuthenticationFlowIdentification, authflowCfg *config.AuthenticationFlowBotProtection, appCfg *config.BotProtectionConfig) IdentificationOption {
 	return IdentificationOption{
 		Identification: i,
+		BotProtection:  GetBotProtectionData(authflowCfg, appCfg),
 	}
 }
 
-func NewIdentificationOptionLoginID(i config.AuthenticationFlowIdentification) IdentificationOption {
+func NewIdentificationOptionLoginID(i config.AuthenticationFlowIdentification, authflowCfg *config.AuthenticationFlowBotProtection, appCfg *config.BotProtectionConfig) IdentificationOption {
 	return IdentificationOption{
 		Identification: i,
+		BotProtection:  GetBotProtectionData(authflowCfg, appCfg),
 	}
 }
 
-func NewIdentificationOptionsOAuth(oauthConfig *config.OAuthSSOConfig, oauthFeatureConfig *config.OAuthSSOProvidersFeatureConfig) []IdentificationOption {
+func NewIdentificationOptionsOAuth(oauthConfig *config.OAuthSSOConfig, oauthFeatureConfig *config.OAuthSSOProvidersFeatureConfig, authflowCfg *config.AuthenticationFlowBotProtection, appCfg *config.BotProtectionConfig) []IdentificationOption {
 	output := []IdentificationOption{}
 	for _, p := range oauthConfig.Providers {
 		if !identity.IsOAuthSSOProviderTypeDisabled(p.AsProviderConfig(), oauthFeatureConfig) {
 			output = append(output, IdentificationOption{
 				Identification: config.AuthenticationFlowIdentificationOAuth,
+				BotProtection:  GetBotProtectionData(authflowCfg, appCfg),
 				ProviderType:   p.AsProviderConfig().Type(),
 				Alias:          p.Alias(),
 				WechatAppType:  wechat.ProviderConfig(p).AppType(),
@@ -63,9 +67,10 @@ func NewIdentificationOptionsOAuth(oauthConfig *config.OAuthSSOConfig, oauthFeat
 	return output
 }
 
-func NewIdentificationOptionPasskey(requestOptions *model.WebAuthnRequestOptions) IdentificationOption {
+func NewIdentificationOptionPasskey(requestOptions *model.WebAuthnRequestOptions, authflowCfg *config.AuthenticationFlowBotProtection, appCfg *config.BotProtectionConfig) IdentificationOption {
 	return IdentificationOption{
 		Identification: config.AuthenticationFlowIdentificationPasskey,
+		BotProtection:  GetBotProtectionData(authflowCfg, appCfg),
 		RequestOptions: requestOptions,
 	}
 }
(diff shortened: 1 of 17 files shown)


pkong-ds:
@louischan-oursky Can I have you attention to my current way of `injecting config.AuthenticationFlow inside Service`
But I feel like there is a better way, any thoughts? :Adore:

- [1984432 - Inject AuthFlowConfig into auth flow service](https://github.com/authgear/authgear-server/pull/4358/commits/1984432eef9b54acb00ff8e5660dd02c9425d6fc)
- [6ecc5be - Check per-flow captcha enabled flag in flow response](https://github.com/authgear/authgear-server/pull/4358/commits/6ecc5be332699cd230866e999e123e90f4f908df)

pkg/lib/authenticationflow/declarative/intent_lookup_identity_oauth.go
```diff
@@ -34,7 +34,8 @@ func (i *IntentLookupIdentityOAuth) MilestoneIdentificationMethod() config.Authe
 
 func (i *IntentLookupIdentityOAuth) CanReactTo(ctx context.Context, deps *authflow.Dependencies, flows authflow.Flows) (authflow.InputSchema, error) {
 	if len(flows.Nearest.Nodes) == 0 {
-		oauthOptions := NewIdentificationOptionsOAuth(deps.Config.Identity.OAuth, deps.FeatureConfig.Identity.OAuth.Providers)
+		// FIXME: figure out how to get flow.branch.Captcha here
+		oauthOptions := NewIdentificationOptionsOAuth(deps.Config.Identity.OAuth, deps.FeatureConfig.Identity.OAuth.Providers, nil, deps.Config.Captcha)
```
pkong-ds, line 38: question: Which node here should be the target flow.step.branch here?

pkg/lib/authenticationflow/declarative/intent_oauth.go
```diff
@@ -34,7 +34,8 @@ func (i *IntentOAuth) MilestoneIdentificationMethod() config.AuthenticationFlowI
 
 func (i *IntentOAuth) CanReactTo(ctx context.Context, deps *authflow.Dependencies, flows authflow.Flows) (authflow.InputSchema, error) {
 	if len(flows.Nearest.Nodes) == 0 {
-		oauthOptions := NewIdentificationOptionsOAuth(deps.Config.Identity.OAuth, deps.FeatureConfig.Identity.OAuth.Providers)
+		// FIXME: figure out how to get flow.branch.Captcha here
+		oauthOptions := NewIdentificationOptionsOAuth(deps.Config.Identity.OAuth, deps.FeatureConfig.Identity.OAuth.Providers, nil, deps.Config.Captcha)
```
pkong-ds, line 38: question: Which node here should be the target flow.step.branch here?

pkg/lib/authenticationflow/declarative/intent_promote_identity_oauth.go
```diff
@@ -35,7 +35,8 @@ func (i *IntentPromoteIdentityOAuth) MilestoneIdentificationMethod() config.Auth
 
 func (i *IntentPromoteIdentityOAuth) CanReactTo(ctx context.Context, deps *authflow.Dependencies, flows authflow.Flows) (authflow.InputSchema, error) {
 	if len(flows.Nearest.Nodes) == 0 {
-		oauthOptions := NewIdentificationOptionsOAuth(deps.Config.Identity.OAuth, deps.FeatureConfig.Identity.OAuth.Providers)
+		// FIXME: figure out how to get flow.branch.Captcha here
+		oauthOptions := NewIdentificationOptionsOAuth(deps.Config.Identity.OAuth, deps.FeatureConfig.Identity.OAuth.Providers, nil, deps.Config.Captcha)
```
pkong-ds, line 39: question: Which node here should be the target flow.step.branch here?

pkg/lib/authenticationflow/declarative/intent_reauth_flow_step_identify.go
```diff
@@ -35,7 +35,8 @@ func NewIntentReauthFlowStepIdentify(ctx context.Context, deps *authflow.Depende
 	for _, b := range step.OneOf {
 		switch b.Identification {
 		case config.AuthenticationFlowIdentificationIDToken:
-			c := NewIdentificationOptionIDToken(b.Identification)
+			// FIXME: Confirm if reauth flow identify step need captcha (assume no)
+			c := NewIdentificationOptionIDToken(b.Identification, nil, nil)
```
pkong-ds, line 39: confirm: Reauth identify does not support captcha?

pkong-ds:
@louischan-oursky free to go through the 2 FIXME commits with me later? 🙏 Not sure which branch should these intent point to for captcha requirement checking

pkong-ds:
@louischan-oursky this should be ready to reviewed :adore:

Question: Quick search seem no test on the 3 api handlers are available, so did not add tests for extra `bot_protection` fields. Please advise if otherwise 👍

pkg/lib/authenticationflow/session_bot_protection.go
```diff
@@ -0,0 +1,13 @@
+package authenticationflow
+
+type BotProtectionVerificationStatus string
```
louischan-oursky, line 3: We should not store a string only. For example, when we verify successfully with the provider, we want to store the verification response as well. So it should be a struct. Since it is a struct, and sometimes we do not need it, the context value should be a pointer to the struct. We need not to define BotProtectionVerificationStatusDefault anymore. Since a nil pointer to the struct means initial status.

```golang
type BotProtectionVerificationResult struct {
  Outcome BotProtectionVerificationOutcome `json:"outcome,omitempty"`
  Response interface{} `json:"response,omitempty"`
}

type BotProtectionVerificationOutcome string

const (
  BotProtectionVerificationOutcomeFailed = "failed"
  BotProtectionVerificationOutcomeVerified = "verified"
  BotProtectionVerificationOutcomeServiceUnavailable = "service_unavailable"
)
```

pkg/lib/authenticationflow/declarative/data_identification.go
```diff
@@ -25,6 +25,7 @@ func (IdentificationData) Data() {}
 type IdentificationOption struct {
 	Identification config.AuthenticationFlowIdentification `json:"identification"`
 
+	BotProtection *BotProtectionData `json:"bot_protection,omitempty" nullable:"true"`
```
louischan-oursky, line 28: Do not specify `nullable`. It has no use here.

pkg/lib/authenticationflow/session_bot_protection.go
louischan-oursky: Or you omit this change, and add it until you actual implement the verification.

pkong-ds: > Or you omit this change, and add it until you actual implement the verification.

Prefer include in this PR 🙏 rationale - It's not "non-implemented feature`, it's referenced in some intent/node reactions in next PR

> `interface{}`

How about `any`? ref https://github.com/golang/go/issues/49884

pkong-ds:
@louischan-oursky updated, thanks! Added 2 commits, starting from [f53fa58](https://github.com/authgear/authgear-server/pull/4358/commits/f53fa5888fc0544854b9fdd9c7f078f44a61860f)

pkg/lib/authenticationflow/session_bot_protection.go
louischan-oursky: Just keep using the good old `interface{}`

louischan-oursky (approved):

PR #4370, opened 2024-06-18
Support multiple tokens in a single offline grant

tung2744:
ref DEV-1403

### DIFF
--- a/pkg/admin/facade/session.go
+++ b/pkg/admin/facade/session.go
@@ -7,21 +7,21 @@ import (
 )
 
 type SessionManager interface {
-	List(userID string) ([]session.Session, error)
-	Get(id string) (session.Session, error)
-	RevokeWithEvent(session session.Session, isTermination bool, isAdminAPI bool) error
-	TerminateAllExcept(userID string, currentSession session.Session, isAdminAPI bool) error
+	List(userID string) ([]session.ListableSession, error)
+	Get(id string) (session.ListableSession, error)
+	RevokeWithEvent(session session.ListableSession, isTermination bool, isAdminAPI bool) error
+	TerminateAllExcept(userID string, currentSession session.ListableSession, isAdminAPI bool) error
 }
 
 type SessionFacade struct {
 	Sessions SessionManager
 }
 
-func (f *SessionFacade) List(userID string) ([]session.Session, error) {
+func (f *SessionFacade) List(userID string) ([]session.ListableSession, error) {
 	return f.Sessions.List(userID)
 }
 
-func (f *SessionFacade) Get(id string) (session.Session, error) {
+func (f *SessionFacade) Get(id string) (session.ListableSession, error) {
 	return f.Sessions.Get(id)
 }
 
--- a/pkg/admin/facade/session.go
+++ b/pkg/admin/facade/session.go
@@ -9,7 +9,7 @@ import (
 type SessionManager interface {
 	List(userID string) ([]session.ListableSession, error)
 	Get(id string) (session.ListableSession, error)
-	RevokeWithEvent(session session.Session, isTermination bool, isAdminAPI bool) error
+	RevokeWithEvent(session session.SessionBase, isTermination bool, isAdminAPI bool) error
 	TerminateAllExcept(userID string, currentSession session.Session, isAdminAPI bool) error
 }
 
--- a/pkg/admin/facade/session.go
+++ b/pkg/admin/facade/session.go
@@ -9,8 +9,8 @@ import (
 type SessionManager interface {
 	List(userID string) ([]session.ListableSession, error)
 	Get(id string) (session.ListableSession, error)
-	RevokeWithEvent(session session.ListableSession, isTermination bool, isAdminAPI bool) error
-	TerminateAllExcept(userID string, currentSession session.ListableSession, isAdminAPI bool) error
+	RevokeWithEvent(session session.Session, isTermination bool, isAdminAPI bool) error
+	TerminateAllExcept(userID string, currentSession session.Session, isAdminAPI bool) error
 }
 
 type SessionFacade struct {
--- a/pkg/admin/facade/session.go
+++ b/pkg/admin/facade/session.go
@@ -10,7 +10,7 @@ type SessionManager interface {
 	List(userID string) ([]session.ListableSession, error)
 	Get(id string) (session.ListableSession, error)
 	RevokeWithEvent(session session.SessionBase, isTermination bool, isAdminAPI bool) error
-	TerminateAllExcept(userID string, currentSession session.Session, isAdminAPI bool) error
+	TerminateAllExcept(userID string, currentSession session.ResolvedSession, isAdminAPI bool) error
 }
 
 type SessionFacade struct {
--- a/pkg/lib/oauth/authz_service.go
+++ b/pkg/lib/oauth/authz_service.go
@@ -59,7 +59,7 @@ func (s *AuthorizationService) Delete(a *Authorization) error {
 	// delete the offline grants that belong to the authorization
 	for _, sess := range sessions {
 		if offlineGrant, ok := sess.(*OfflineGrant); ok {
-			if offlineGrant.AuthorizationID == a.ID {
+			if offlineGrant.Deprecated_AuthorizationID == a.ID {
 				err := s.OAuthSessionManager.Delete(sess)
 				if err != nil {
 					return err
--- a/pkg/lib/oauth/authz_service.go
+++ b/pkg/lib/oauth/authz_service.go
@@ -10,8 +10,8 @@ import (
 )
 
 type OfflineGrantSessionManager interface {
-	List(userID string) ([]session.Session, error)
-	Delete(session session.Session) error
+	List(userID string) ([]session.ListableSession, error)
+	Delete(session session.ListableSession) error
 }
 
 type AuthorizationService struct {
--- a/pkg/lib/oauth/authz_service.go
+++ b/pkg/lib/oauth/authz_service.go
@@ -57,6 +57,7 @@ func (s *AuthorizationService) Delete(a *Authorization) error {
 	// delete the offline grants that belong to the authorization
 	for _, sess := range sessions {
 		if offlineGrant, ok := sess.(*OfflineGrant); ok {
+			// TODO(DEV-1403): Check all authorization ids?
 			if offlineGrant.AuthorizationID == a.ID {
 				err := s.OAuthSessionManager.Delete(sess)
 				if err != nil {
--- a/pkg/lib/oauth/authz_service.go
+++ b/pkg/lib/oauth/authz_service.go
@@ -19,6 +19,8 @@ type AuthorizationService struct {
 	Store               AuthorizationStore
 	Clock               clock.Clock
 	OAuthSessionManager OfflineGrantSessionManager
+	OfflineGrantService *OfflineGrantService
+	OfflineGrantStore   OfflineGrantStore
 }
 
 func (s *AuthorizationService) GetByID(id string) (*Authorization, error) {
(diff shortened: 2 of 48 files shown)


pkg/lib/oauth/handler/service_token.go
```diff
@@ -39,23 +38,36 @@ func (s *TokenService) IssueOfflineGrant(
...
+
+		// These fields exist in OfflineGrantRefreshToken too
+		// They need to be provided to handled cases that we refer to an offline grant without token hash
+		// Such as using an id_token
 		ClientID:        client.ClientID,
-		IDPSessionID:    opts.IDPSessionID,
-		IdentityID:      opts.IdentityID,
+		Scopes:          opts.Scopes,
+		AuthorizationID: opts.AuthorizationID,
+		TokenHash:       tokenHash,
```
tung2744, line 67: I decided to keep these fields in the top level too, because there are cases that referencing an offline grant without an actual token. (The sid in id_token)

See if any thoughts or concern.

pkg/lib/session/session.go
```diff
@@ -15,6 +15,17 @@ const (
 	TypeOfflineGrant     Type = "offline_grant"
 )
 
+type Session interface {
```
louischan-oursky: Do not use the original name. if we use a new name, other PR that also refer to Session (the old one) will fail to compile. That helps us to spot problems easier.

Let's use `ResovledSession`

pkg/lib/config/secret.go
```diff
@@ -310,7 +310,7 @@ var secretItemKeys = map[SecretKey]secretKeyDef{
 	WATICredentialsKey:                         {"WATICredentials", func() SecretItemData { return &WATICredentials{} }},
 	OAuthClientCredentialsKey:                  {"OAuthClientCredentials", func() SecretItemData { return &OAuthClientCredentials{} }},
 	CustomSMSProviderConfigKey:                 {"CustomSMSProviderConfig", func() SecretItemData { return &CustomSMSProviderConfig{} }},
-	Deprecated_CaptchaCloudflareCredentialsKey: {"LegeacyCaptchaCloudflareCredentials", func() SecretItemData { return &Deprecated_CaptchaCloudflareCredentials{} }},
+	Deprecated_CaptchaCloudflareCredentialsKey: {"Deprecated_CaptchaCloudflareCredentials", func() SecretItemData { return &Deprecated_CaptchaCloudflareCredentials{} }},
```
louischan-oursky, line 313: Rebase and drop this commit.

pkg/lib/oauth/grant_offline.go
```diff
@@ -164,6 +167,33 @@ func (g *OfflineGrant) IsSameSSOGroup(ss session.Session) bool {
...
 }
 
-func (g *OfflineGrant) EqualSession(ss session.Session) bool {
+func (g *OfflineGrant) EqualSession(ss session.SessionBase) bool {
 	return g.SessionID() == ss.SessionID() && g.SessionType() == ss.SessionType()
 }
+
+// TODO(DEV-1403): Use token hash to find the correct session
+func (g *OfflineGrant) ToSession(refreshTokenHash string) session.Session {
+	if refreshTokenHash == "" || refreshTokenHash == g.TokenHash {
```
louischan-oursky, line 180: Use constant time compare to compare hash.

pkg/lib/oauth/grant_offline.go
```diff
@@ -164,6 +167,33 @@ func (g *OfflineGrant) IsSameSSOGroup(ss session.Session) bool {
...
+			OfflineGrant: g,
+			CreatedAt:    g.CreatedAt,
+			TokenHash:    g.TokenHash,
+			ClientID:     g.ClientID,
+			Scopes:       g.Scopes,
+		}
+	}
+
+	for _, token := range g.RefreshTokens {
+		if token.TokenHash == refreshTokenHash {
```
louischan-oursky, line 192: Use constant time compare to compare hash.

pkg/lib/oauth/grant_offline_service.go
```diff
@@ -53,7 +53,6 @@ func (s *OfflineGrantService) IsValid(session *OfflineGrant) (bool, time.Time, e
 	return true, expiry, nil
 }
 
-// TODO(DEV-1403): Ensure the client id is correct
```
louischan-oursky: Using OfflineGrant.Client ID is incorrect?

pkg/lib/oauth/handler/service_token.go
```diff
@@ -58,6 +58,14 @@ func (s *TokenService) IssueOfflineGrant(
...
 		IdentityID:   opts.IdentityID,
 
+		// These fields exist in OfflineGrantRefreshToken too
+		// They need to be provided to handled cases that we refer to an offline grant without token hash
+		// Such as using an id_token
+		ClientID:        client.ClientID,
+		Scopes:          opts.Scopes,
+		AuthorizationID: opts.AuthorizationID,
+		TokenHash:       tokenHash,
+
```
louischan-oursky, line 68: I suggest we do not write these fields anymore. There fields are read-only after this PR is merged.

I also have another suggestion. We rename these fields to private.

1. We can spot places still use these fields and correct them
2. Outside the package, we cannot initialize these fields. So they are read-only.

pkg/lib/oauth/oidc/ui.go
```diff
@@ -226,7 +226,7 @@ func (r *UIInfoResolver) ResolveForAuthorizationEndpoint(
 					canUseIntentReauthenticate = true
 				case session.TypeOfflineGrant:
 					if offlineGrant, ok := sidSession.(*oauth.OfflineGrant); ok {
-						// TODO(DEV-1403): Find the correct scopes of the used token
+						// Use the "root" scopes, because we don't know the tokenHash here
```
louischan-oursky, line 229: Do not expose Scopes.

pkg/lib/oauth/redis/store.go
```diff
@@ -384,6 +384,32 @@ func (s *Store) UpdateOfflineGrantApp2AppDeviceKey(grantID string, newKey string
 	return grant, nil
 }
 
+func (s *Store) UpdateOfflineGrantRefreshTokens(grantID string, refreshTokens []oauth.OfflineGrantRefreshToken, expireAt time.Time) (*oauth.OfflineGrant, error) {
```
louischan-oursky, line 387: The mutex here does not protect concurrent write. Reader A and Reader B could read the offline grant around the same time. Reader A obtains the lock and write .RefreshTokens. Reader B then obtains the lock, and overwrite what Reader A wrote.

pkg/lib/oauth/redis/store.go
```diff
@@ -483,11 +509,16 @@ func (s *Store) ListClientOfflineGrants(clientID string, userID string) ([]*oaut
...
-			continue
+		if offlineGrant.ClientID == clientID {
+			result = append(result, offlineGrant)
+		} else {
+			for _, token := range offlineGrant.RefreshTokens {
+				if token.ClientID == clientID {
+					result = append(result, offlineGrant)
+					break
+				}
+			}
```
louischan-oursky, line 520: Do not expose ClientID and RefreshTokens

pkg/lib/oauth/authz_service.go
```diff
@@ -57,13 +59,33 @@ func (s *AuthorizationService) Delete(a *Authorization) error {
...
 			if offlineGrant.AuthorizationID == a.ID {
 				err := s.OAuthSessionManager.Delete(sess)
 				if err != nil {
 					return err
 				}
 			}
+			newTokens := []OfflineGrantRefreshToken{}
+			isTokenChanged := false
+			// Revoke any sub-tokens using this authorization
+			for _, token := range offlineGrant.RefreshTokens {
```
louischan-oursky, line 71: Do not expose RefreshTokens

pkg/lib/oauth/session_manager.go
```diff
@@ -67,8 +67,30 @@ func (m *SessionManager) TerminateAllExcept(userID string, currentSession sessio
 	for _, ss := range sessions {
 		// skip third party client app refresh token
 		// third party refresh token should be deleted through deleting authorization
-		// TODO(DEV-1403): Check all client id?
 		if _, ok := thirdPartyClientIDSet[ss.ClientID]; ok {
+			// If this is a thrid party offline grant,
+			// revoke any tokens of this offline grant which are used by first party app
+			newTokens := ss.RefreshTokens
+			isTokenChanged := false
+			for _, token := range ss.RefreshTokens {
```
louischan-oursky, line 75: Do not expose ClientID and RefreshTokens

pkg/lib/sessionlisting/listing.go
```diff
@@ -68,9 +68,19 @@ func (s *SessionListingService) FilterForDisplay(sessions []session.ListableSess
...
-		// remove third-party app refresh token
-		// TODO(DEV-1403): Check all client id?
-		if _, ok := thirdPartyClientIDSet[offlineGrant.ClientID]; ok {
+		// remove the offline grant if it is only used in third-party app
+		var isUsedInFirstPartyApp = false
+		if _, ok := thirdPartyClientIDSet[offlineGrant.ClientID]; !ok {
+			isUsedInFirstPartyApp = true
+		}
+
+		for _, token := range offlineGrant.RefreshTokens {
```
louischan-oursky, line 77: Do not expose ClientID and RefreshTokens.

pkg/lib/oauth/resolver.go
```diff
@@ -126,7 +126,10 @@ func (re *Resolver) resolveHeader(r *http.Request) (session.Session, error) {
 			return nil, err
 		}
 
-		authSession = g.ToSession(grant.RefreshTokenHash)
+		as, ok := g.ToSession(grant.RefreshTokenHash)
+		if ok {
+			authSession = as
+		}
```
louischan-oursky, line 132: Should it be an error if we do not assign authSession?

tung2744: I think no, it should be simply "no session".

pkg/lib/oauth/handler/service_token.go
tung2744: However, there are cases that we need these fields. id token reference the offline grant without a token hash.

pkg/lib/oauth/oidc/ui.go
tung2744: In this case we can only use the scopes in the root.

pkg/lib/oauth/resolver.go
louischan-oursky: In this function, we have the following return pattern

1. `return nil, nil`: It means no token is found in HTTP header Authorization
2. `return nil, err`: Something went wrong
3. `return s, nil`: Resolved a session successfully
4. And this new case, `return nil, nil`: It means the token cannot look up a session. It think this is an error? It also conflicts with case 1.

tung2744: I see. So maybe it should be `return nil, ErrInvalidSession`, meaning "header exist, but no session can be found".

pkg/lib/oauth/grant_offline_service.go
tung2744: We should keep using `OfflineGrant.ClientID` here to compute the expiry, but will rename `ClientID` to `InitialClientID`.

pkg/lib/oauth/handler/service_token.go
tung2744: We cannot make these fields private, because it can be deserialized from json. However, I will rename the field with prefix `ReadOnly__`

pkg/lib/oauth/resolver.go
louischan-oursky: Yes. We should do `return nil, ErrInvalidSession` in this case.

pkg/lib/oauth/authz_service.go
tung2744: `RefreshTokens` must be exposed, because the OfflineGrant is not contructed in the same package. However, I will remove usage of `RefreshTokens` outside the package.

tung2744:
I also created a feature branch `feat-app-initiated-sso-to-web` for this feature.

pkg/admin/facade/session.go
```diff
@@ -10,7 +10,7 @@ type SessionManager interface {
 	List(userID string) ([]session.ListableSession, error)
 	Get(id string) (session.ListableSession, error)
 	RevokeWithEvent(session session.SessionBase, isTermination bool, isAdminAPI bool) error
-	TerminateAllExcept(userID string, currentSession session.Session, isAdminAPI bool) error
+	TerminateAllExcept(userID string, currentSession session.ResovledSession, isAdminAPI bool) error
```
louischan-oursky, line 13: typo, It should be ResolvedSession.

tung2744: Fixed with a fixup to [6b34b37](https://github.com/authgear/authgear-server/pull/4370/commits/6b34b375ad1a4c81db4607f154513ce7b8b829cc)

louischan-oursky (approved):

PR #4372, opened 2024-06-18
Implement account linking for incoming login ID

louischan-oursky:
ref DEV-1233


pkg/lib/authenticationflow/milestone.go
```diff
@@ -17,9 +17,13 @@ func FindMilestone[T Milestone](w *Flow) (T, bool) {
...
 }
 
-// This function only find milestones in the provided flow,
-// and will not nest into subflows
-func FindMilestoneInCurrentFlow[T Milestone](w *Flow) (T, bool) {
+// FindMilestoneInCurrentFlow find the last milestone in the flow.
+// It does not recur into sub flows.
+// If the found milestone is a node, then the returned flows is the same as flows.
+// If the found milestone is a intent, then the returned flows is Nearest=intent.
+func FindMilestoneInCurrentFlow[T Milestone](flows Flows) (T, Flows, bool) {
```
tung2744, line 16: In this commit, it seems the new returned `Flows` is always ignored. Did you find any places that we need this new `Flows`?

louischan-oursky: Not always, it is need sometimes, like in https://github.com/authgear/authgear-server/pull/4372/commits/6d78a3dccf400c23c8e1b52f614d82561d732df0#diff-ea1ed68baf6a6ca64bbdb127d2f22167bb59c0a26052e499c514f7e0cf06645dR59

The returned flows will have Nearest updated to match the root of the returned milestone. We can then use the returned milestone and the returned flows to call FindMilestoneInCurrentFlow. It is a more structured find. The previous FindMilestone finds too deep milestone that is irrelevant to the finder.

pkg/lib/authenticationflow/declarative/intent_check_conflict_and_create_identity.go
```diff
@@ -22,6 +22,26 @@ type IntentCheckConflictAndCreateIdenity struct {
...
+
+	// If there are conflicts, then we delegate to MilestoneFlowAccountLinking, which also implements MilestoneFlowCreateIdentity.
+	// We cannot find MilestoneFlowCreateIdentity because that would find this intent, resulting in an infinite recursion.
+	m1, m1Flows, ok := authflow.FindMilestoneInCurrentFlow[MilestoneFlowAccountLinking](flows)
+	if ok {
+		return m1.MilestoneFlowCreateIdentity(m1Flows)
+	}
+
+	return nil, flows, false
+}
```
tung2744, line 44: Should `MilestoneDoCreateIdentity` and `MilestoneFlowCreateIdentity` be one thing?

I am thinking that it should conceptually be:
```
- IntentSignup
  - IntentCreateIdentity (MilestoneFlowCreateIdentity)
    - NodeDoCreateIdentity (MilestoneFlowCreateIdentity)
```

1. From IntentSignup, we find `MilestoneFlowCreateIdentity`, which get `IntentCreateIdentity`.

2. In `IntentCreateIdentity`, it cannot determine if the identity is created or not, so it continue to find `MilestoneFlowCreateIdentity` in the sub-nodes. Which get `NodeDoCreateIdentity`.

3. In `NodeDoCreateIdentity`, it is the one who create the identity, so he knows the identity is created or not, so it implements `MilestoneFlowCreateIdentity` with a correct return value.

tung2744 (approved):
Just have one question above

pkg/lib/authenticationflow/declarative/intent_check_conflict_and_create_identity.go
louischan-oursky: I did not do that because to IntentSignup, what matters to it is whether the create identity flow is spawned, whether the identity is created or not does not matter to IntentSignup.

I also want to keep MilestoneDoCreateIdentity simple. The presence of it means created. Your proposal needs to call the milestone to check if it is created or not. This is quite error-prone. If we fail to call the checking, we incorrectly assume it is created.

tung2744: I see, ok.

PR #4384, opened 2024-06-20
Customise authflow v2 ui

cychiuae:


pkg/lib/web/static_assets.go
```diff
@@ -22,6 +22,8 @@ var AuthgearDarkThemeCSS = resource.RegisterResource(CSSDescriptor{
 var AppLogo = resource.RegisterResource(ImageDescriptor{Name: "app_logo"})
 var AppLogoDark = resource.RegisterResource(ImageDescriptor{Name: "app_logo_dark"})
 var Favicon = resource.RegisterResource(ImageDescriptor{Name: "favicon"})
+var AppBackgroundImage = resource.RegisterResource(ImageDescriptor{Name: "app_background_image"})
+var AppBackgroundImageDark = resource.RegisterResource(ImageDescriptor{Name: "app_background_image_dark"})
```
cychiuae, line 26: Not sure how to increase file size limit for background image

portal/src/graphql/portal/DesignScreen/form.ts
```diff
@@ -0,0 +1,576 @@
...
+        fullyHandled: false,
+      };
+    },
+    [resourceForm.state.resources]
+  );
+
+  const errorRules: ErrorParseRule[] = useMemo(
+    () => [imageSizeTooLargeErrorRule],
+    [imageSizeTooLargeErrorRule]
+  );
```
cychiuae, line 457: Here I copy from old code

portal/src/model/themeAuthFlowV2.ts
```diff
@@ -0,0 +1,35 @@
+type Color = string;
```
louischan-oursky, line 14: export type CSSColor = string;

portal/src/model/themeAuthFlowV2.ts
```diff
@@ -264,3 +320,50 @@ export class StyleCSSVisitor<T> extends CssNodeVisitor {
...
+        prop: styleProperty.propertyName,
+        value: String(styleProperty.getCSSValue()),
+      })
+    );
+  }
+
+  getCSS(): Root {
+    return this.root;
+  }
+}
```
louischan-oursky, line 375: Can we add some tests to demonstrate how this thing works, and to prove it is working as expected?

portal/src/graphql/portal/DesignScreen/Components.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { PropsWithChildren } from "react";
```
louischan-oursky: We now have its own directory, it is better to have one component per file.

portal/src/graphql/portal/DesignScreen/Components.tsx
tung2744, line 1: You can create a directory `portal/src/components/design` and put these components into there.

portal/src/graphql/portal/DesignScreen/Components.tsx
```diff
@@ -0,0 +1,559 @@
...
+}
+
+interface ImagePickerProps {
+  base64EncodedData: string | null;
+  onChange: (
+    image: {
+      base64EncodedData: string;
+      extension: ImageFileExtension;
+    } | null
+  ) => void;
```
tung2744, line 441: Looks like these are duplicated from ImageFilePicker, but these should not be duplicated because they should support same file types.

portal/src/graphql/portal/DesignScreen/form.ts
tung2744: Instead of duplication, create a custom hook in `src/hook` to do it?

pkg/lib/web/static_assets.go
tung2744: Seems no easy way at the moment :sosad:
But it make sense that background image could be larger than 100KB? @fungc-io

tung2744: I suggest to add an interface:
```golang
type SizeLimitDescriptor interface {
  GetSizeLimit() int
}
```

And move the logic that checks the size limit (Search for `ConfigFileMaxSize`) into `Manager.applyUpdates(...)`.
After `manager.Resolve(u.Path)` called, you got the descriptor. Check if it implements `SizeLimitDescriptor`. If yes, use `GetSizeLimit` to get the limit and check the size. Else, check with `ConfigFileMaxSize` as the default limit.

cychiuae: I think I will create another ticket for it. This PR is already too big

tung2744:
I guess the updates are complete?

cychiuae:
> I guess the updates are complete?

Yup. YesYEs

louischan-oursky (approved):

tung2744:
When i test the pr, I found some customization isn't working.
In this pr,
- `—-primary-btn__border-radius` was written to the css file, but it should be `--primary-btn__border-radius`. (The first `-` is different :clown:)

- `-—widget__bg-color` should be `--widget__bg-color` (The second `-`)

- `-—primary-btn__bg-color` should be `--primary-btn__bg-color`

- `--layout-flex-align-items` is not used.

cychiuae:
I also found that problem when I am integrating with iframe. :0)
The dash charactors are changed when copying from linear :sosad:

tung2744:
Will you fix it in this pr, or you prefer to fix it in the inline preview pr?

cychiuae:
> Will you fix it in this pr, or you prefer to fix it in the inline preview pr?

I will fix it later because this pr already too large

tung2744:
ok, let me merge this first.

PR #4385, opened 2024-06-21
Feature flag for template customization

elise-ng:
Ref DEV-1302

tested server change by adding test case.

tested ui by setting the flag in local fs `var/authgear.features.yaml`:

```yaml
messaging:
  template_customization_disabled: true
```

portal screenshot: 

https://github.com/authgear/authgear-server/assets/11769136/bbe2a4e5-637e-4f6f-a8d1-d6f1c3fe23fc


elise-ng:
@louischan-oursky i assume "overriding" the MatchResource methods would be fine & we can treat CustomizableXXX as XXX type during render because it will only use FindResources and ViewResourceswhich does not require separate behavior

lmk if there is better golang tricks for this

pkg/util/template/resource.go
```diff
@@ -30,11 +31,27 @@ type HTML struct {
 	IsFindAllowedInFs FindResourcesFsFilter
 }
 
+type CustomizableHTML struct {
```
louischan-oursky, line 34: I took a closer to the code, I realize that we have one `HTML` descriptor for each template. I was mistaken that one HTML `descriptor` manages all HTML templates. Sorry for confusing you.

here is the picture in my mind.

```golang
// HTML defines a HTML template that CANNOT be customized by the app.
type HTML {
	// Name is the name of template
	Name string
	// ComponentDependencies is the HTML component templates this template depends on.
	ComponentDependencies []*HTML
        // IsFindAllowedInFs is removed. The existing behavior of ExcludeAppFs should be embodied into this descriptor. Please also remove FindResourcesFsFilter
}

// MessageHTML represents a message in HTML. It can be customized by the app.
type MessageHTML struct {
	// Name is the name of template
	Name string
        // Omit ComponentDependencies. We do not need it in this descriptor
}
```

- MatchResources is the same as before. You can just duplicate it for HTML and MessageHTML.
- FindResources of HTML is ExcludeAppFs. Find FindResources of MessageHTML is AnyFs.
- ViewResources is  the same as before. You can just duplicate it for HTML and MessageHTML.
- UpdateResource of HTML always results in an error. UpdateResource of MessageHTML is almost the same as before, with an addition of respecting the new feature config.
- Repeating the same thing for PlainText and MessagePlainText.

elise-ng:
@louischan-oursky i have tested can't update html or txt even if user override readOnly from ui; however user will still be able to update `translations.json` (email subject field), should we change that descriptor as well or would there be side-effect?

pkg/util/template/resource.go
elise-ng: updated, plz check

elise-ng:
from discussion:

- change `pkg/util/template/translation.go`, `processAppTranslationData` if feature flag on, do not save keys with pattern `email.***.subject`

pkg/util/template/resource.go
```diff
@@ -25,34 +26,60 @@ type HTML struct {
...
+var _ resource.Descriptor = &MessageHTML{}
 
-func (t *HTML) templateResource() {}
+func (t *HTML) templateResource()        {}
+func (t *MessageHTML) templateResource() {}
 
 func (t *HTML) MatchResource(path string) (*resource.Match, bool) {
+	if strings.HasPrefix(t.Name, "messages/") {
+		return nil, false
+	}
```
louischan-oursky, line 58: No need to add this condition. We assume the calls to RegisterHTML are correct, that is, no one will ever use RegisterHTML to register message HTML.

pkg/util/template/resource.go
```diff
@@ -25,34 +26,60 @@ type HTML struct {
...
+	return matchTemplatePath(path, t.Name)
+}
+
+func (t *MessageHTML) MatchResource(path string) (*resource.Match, bool) {
+	if !strings.HasPrefix(t.Name, "messages/") {
+		return nil, false
+	}
```
louischan-oursky, line 65: No need to add this condition. We assume the calls to RegisterMessageHTML are correct, that is, no one will ever use RegisterMessageHTML to register regular HTML.

pkg/util/template/resource.go
```diff
@@ -67,41 +94,73 @@ type PlainText struct {
...
+var _ resource.Descriptor = &MessagePlainText{}
 
-func (t *PlainText) templateResource() {}
+func (t *PlainText) templateResource()        {}
+func (t *MessagePlainText) templateResource() {}
 
 func (t *PlainText) MatchResource(path string) (*resource.Match, bool) {
+	if strings.HasPrefix(t.Name, "messages/") {
+		return nil, false
+	}
```
louischan-oursky, line 126: Remove this condition.

pkg/util/template/resource.go
```diff
@@ -67,41 +94,73 @@ type PlainText struct {
...
+	return matchTemplatePath(path, t.Name)
+}
+
+func (t *MessagePlainText) MatchResource(path string) (*resource.Match, bool) {
+	if !strings.HasPrefix(t.Name, "messages/") {
+		return nil, false
+	}
```
louischan-oursky, line 133: Remove this condition.

pkg/util/template/resource.go
```diff
@@ -67,41 +94,73 @@ type PlainText struct {
...
 	return matchTemplatePath(path, t.Name)
 }
 
 func (t *PlainText) FindResources(fs resource.Fs) ([]resource.Location, error) {
+	// Exclude App Fs
+	if fs.GetFsLevel() == resource.FsLevelApp {
+		return []resource.Location{}, nil
+	}
```
louischan-oursky, line 141: PlainText originally does not have FindResourcesFsFilter, so we have to preserve the original behavior. Allow it to read from any Fs. Please remove this condition.

pkg/util/template/resource.go
```diff
@@ -67,41 +94,73 @@ type PlainText struct {
...
 func (t *PlainText) ViewResources(resources []resource.ResourceFile, view resource.View) (interface{}, error) {
 	return viewTextTemplates(t.Name, resources, view)
 }
 
+func (t *MessagePlainText) ViewResources(resources []resource.ResourceFile, view resource.View) (interface{}, error) {
+	return viewTextTemplates(t.Name, resources, view)
+}
+
 func (t *PlainText) UpdateResource(_ context.Context, _ []resource.ResourceFile, resrc *resource.ResourceFile, data []byte) (*resource.ResourceFile, error) {
+	return nil, fmt.Errorf("PlainText resource cannot be updated. Use MessagePlainText resource instead.")
```
louischan-oursky, line 159: Allow PlainText to be updated, same as before.

pkg/util/template/resource.go
elise-ng: removed

pkg/util/template/resource.go
elise-ng: removed

pkg/util/template/resource.go
elise-ng: removed

pkg/util/template/resource.go
elise-ng: removed

pkg/util/template/resource.go
elise-ng: removed

pkg/util/template/resource.go
elise-ng: reverted

elise-ng:
@louischan-oursky updated, plz check again 🙏

louischan-oursky (approved):

PR #4389, opened 2024-06-21
Issue device secret in token endpoint

tung2744:
ref DEV-1404

### DIFF
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -566,8 +566,7 @@ func (h *TokenHandler) handleAnonymousRequest(
 	}
 
 	// TODO(oauth): allow specifying scopes
-	// Note(tung): Do we allow anonymous user to use device_sso and x_app_initiated_sso_to_web?
-	scopes := []string{"openid", oauth.FullAccessScope}
+	scopes := []string{"openid", oauth.OfflineAccess, oauth.FullAccessScope}
 
 	authz, err := h.Authorizations.CheckAndGrant(
 		client.ClientID,
@@ -778,7 +777,7 @@ func (h *TokenHandler) handleBiometricAuthenticate(
 		return nil, err
 	}
 
-	scopes := []string{"openid", oauth.FullAccessScope}
+	scopes := []string{"openid", oauth.OfflineAccess, oauth.FullAccessScope}
 	requestedScopes := r.Scope()
 	if len(requestedScopes) > 0 {
 		err := h.ValidateScopes(client, requestedScopes)
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -294,7 +294,7 @@ func (h *TokenHandler) rotateDeviceSecretIfNeeded(
 	authorizedScopes []string,
 	offlineGrant *oauth.OfflineGrant,
 	resp protocol.TokenResponse) (*oauth.OfflineGrant, error) {
-	if offlineGrant.DeviceSecretHash == "" || !oidc.IsScopeAuthorized(authorizedScopes, oauth.DeviceSSOScope) {
+	if offlineGrant.DeviceSecretHash == "" || !oauth.ContainsAllScopes(authorizedScopes, []string{oauth.DeviceSSOScope}) {
 		// No device secret, no rotation needed.
 		return offlineGrant, nil
 	}
@@ -788,6 +788,10 @@ func (h *TokenHandler) handleBiometricAuthenticate(
 		scopes = requestedScopes
 	}
 
+	if !oauth.ContainsAllScopes(scopes, []string{oauth.OfflineAccess, oauth.FullAccessScope}) {
+		return nil, protocol.NewError("invalid_scope", "offline_access and full-access must be requested")
+	}
+
 	authz, err := h.Authorizations.CheckAndGrant(
 		client.ClientID,
 		info.UserID,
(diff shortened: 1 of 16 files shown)


pkg/lib/oauth/handler/handler_token.go
```diff
@@ -755,8 +757,15 @@ func (h *TokenHandler) handleBiometricAuthenticate(
...
 
-	// TODO(DEV-1404): allow specifying scopes
 	scopes := []string{"openid", oauth.FullAccessScope}
+	requestedScopes := r.Scope()
+	if len(requestedScopes) > 0 {
+		err := h.ValidateScopes(client, requestedScopes)
+		if err != nil {
+			return nil, err
+		}
+		scopes = requestedScopes
```
louischan-oursky, line 785: If the client provides only the app-initiated-sso-to-web scope, then scopes will be missing `openid` and `full-access` scope.

And I realize we are missing the scope `offline_grant` here. `offline_grant` should be present in order to issue an offline grant. It seems that we have this problem in places where scopes are hard-coded.

pkg/lib/oauth/handler/service_token.go
```diff
@@ -211,3 +205,10 @@ func (s *TokenService) ParseRefreshToken(token string) (
 
 	return authz, offlineGrant, tokenHash, nil
 }
+
+func (s *TokenService) IssueDeviceSecret(resp protocol.TokenResponse) (deviceSecret string, hash string) {
```
louischan-oursky, line 209: This method only needs to return deviceSecretHash.

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -290,6 +290,24 @@ func (h *TokenHandler) app2appGetDeviceKeyJWKVerified(jwt string) (jwk.Key, erro
 	return key, nil
 }
 
+func (h *TokenHandler) rotateDeviceSecretIfNeeded(offlineGrant *oauth.OfflineGrant, resp protocol.TokenResponse) (*oauth.OfflineGrant, error) {
+	if offlineGrant.DeviceSecretHash == "" {
+		// No device secret, no rotation needed.
+		return offlineGrant, nil
+	}
```
louischan-oursky, line 297: According to the spec, we should issue device secret if it is invalid, or absent, given that the authorization code has the scope.

It seems that we haven't handle `grant_type=refresh_token` yet.

I think we should rotate the device secret every time the refresh token is used, this includes:
- grant_type=refresh_token, that is, when the refresh token is used to refresh an access token
- reauthenticate
- refresh ID token with our custom grant type

tung2744:
I've read the spec again. 

> If a mobile app requests a device secret via the device_sso scope and a device_secret exists, then the client MUST provide the device_secret on the request to the /token endpoint to exchange code for tokens. 

I think that means, if a device has two app installed, A and B, and both apps will request the `device_sso` scope. Assume A already authenticated (Therefore obtained device_secret), and B uses authorize endpoint to authenticate. When B call token endpoint in authorization code flow, B should also pass the device_secret obtained by A. And the server should rotate the device_secret in this case too.

If I understand it correctly, that means in this case, even B is not authenticated using the new token exchange profile, it should share the same session with A (And therefore same device_secret).

So marking this PR as WIP for the above change.

pkg/lib/oauth/handler/handler_token.go
tung2744: @louischan-oursky 
Lets do a validation instead of implicitly add the scopes?

louischan-oursky: Validate the implicitly added scopes, and the explicitly provided scope.

tung2744:
After thinking deeper, if app A provided the `app-initiated-sso-to-web` scope, and therefore an device_secret was issued. After that, app B used authorize endpoint to authenticate, and call token endpoint to obtain tokens. According to the spec, as device_secret already exist (Obtained by app A), B should provide the device_secret to the token endpoint. Which then we can expect the session of app B will be the same as app A (Because of shared device_secret). Therefore looks like `app-initiated-sso-to-web` implies `device_sso` is enabled?

tung2744:
Per offline discussion, this pr need to:
- Support the device_sso scope.
- When using app-initiated-sso-to-web, device_sso must be provided. Else it is an error.

tung2744:
I kept the device secret in offline grant because now we don't have a case that a single secret can be used in multiple offline grant.

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -750,8 +778,19 @@ func (h *TokenHandler) handleBiometricAuthenticate(
...
+	requestedScopes := r.Scope()
+	if len(requestedScopes) > 0 {
+		err := h.ValidateScopes(client, requestedScopes)
+		if err != nil {
+			return nil, err
+		}
+		scopes = requestedScopes
+	}
+
+	if !oauth.ContainsAllScopes(scopes, []string{oauth.OfflineAccess, oauth.FullAccessScope}) {
```
louischan-oursky, line 791: Line 781 should be `["openid", "offline_access", "full-access"]`

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -543,6 +566,7 @@ func (h *TokenHandler) handleAnonymousRequest(
 	}
 
 	// TODO(oauth): allow specifying scopes
+	// Note(tung): Do we allow anonymous user to use device_sso and x_app_initiated_sso_to_web?
```
louischan-oursky, line 569: No, we do not allow device_sso in anonymous user.

And line 570 should be `["openid", "offline_access", "full-access"]`

tung2744: Let me remove the note.
I've opened https://linear.app/authgear/issue/DEV-1460 for discussion on support of device_sso (And app initiated sso)

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #4390, opened 2024-06-24
[WIP] Implement intent_bot_protection

pkong-ds:
## What's in this PR?
- Add intent_bot_protection and MISC dependencies
- Try implementing `spawn IntentBotProtection subflow if cu1rrent branch require bot_protection`

@louischan-oursky can help review 99bae6a9e5b1e2fdb482b65b66bb251b6ff370c9 🙏 This commit is my attempt to add bot protection logic to an existing intent.

If this commit looks good to you, I will replicate this implementation across all other `flows->steps->branches` that supports bot_protection


pkg/lib/feature/botprotection/errors.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"github.com/authgear/authgear-server/pkg/api/apierrors"
+)
+
+var Failed = apierrors.Invalid.WithReason("Failed")
+
+var ErrVerificationFailed = Failed.NewWithCause("verification failed", apierrors.StringCause("VerificationFailed"))
+
+var ErrVerificationServiceUnavailable = Failed.NewWithCause("verification service unavailable", apierrors.StringCause("VerificationServiceUnavailable"))
```
louischan-oursky, line 11: Do not use cause here. Instead of, define them as follows

```golang
var ErrVerificationFailed = apierrors.Forbidden.WithReason("BotProtectionVerificationFailed").New("bot protection verification failed")

var ErrServiceUnavailable = apierrors.ServiceUnavailable.WithReason("BotProtectionServiceUnavailable").("bot protection service unavailable")
```

pkg/lib/authenticationflow/declarative/intent_bot_protection.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (i *IntentBotProtection) ReactTo(ctx context.Context, deps *authflow.Dependencies, flows authflow.Flows, input authflow.Input) (*authflow.Node, error) {
+	var inputBotProtectionVerification inputBotProtectionVerification
+	if !authflow.AsInput(input, &inputBotProtectionVerification) {
+		return nil, authflow.ErrIncompatibleInput
+	}
+
+	err := deps.BotProtection.Verify(inputBotProtectionVerification.GetBotProtectionType(), inputBotProtectionVerification.GetBotProtectionResponse())
+
+	switch err {
```
louischan-oursky, line 47: First of all, you cannot switch err in this way. Error is Go is a chain of errors.

The standard way to handle error in Go is to use errors.Is and errors.As. Both of them return bool so you can do

```golang
err := MyFunction()
var errB *ErrorB
switch { // This is a switch without subject. This is a valid usage of switch in Go. See https://go.dev/tour/flowcontrol/11
case errors.Is(err, ErrorConstantA):
  break
case errors.Is(err, &errB):
  break
}
```

pkg/lib/authenticationflow/errors.go
```diff
@@ -29,6 +29,12 @@ var ErrNoChange = errors.New("no change")
 // This error originates from CanReactTo and will be propagated to public API.
 var ErrEOF = errors.New("eof")
 
+// ErrBotProtectionVerificationSomeStatus means bot-protection verification is of some status
+// This error can only be returned by IntentBotProtection ReactTo
+var ErrBotProtectionVerificationFailed = errors.New("bot-protection verification failed")
+var ErrBotProtectionVerificationSuccess = errors.New("bot-protection verification success")
+var ErrBotProtectionVerificationServiceUnavailable = errors.New("bot-protection verification service unavailable")
+
```
louischan-oursky, line 37: Never use `errors.New` to construct errors that we will report in the API. `errors.New` appears as unexpected errors in the API. We must use apierror to report expected errors.

pkg/lib/authenticationflow/declarative/bot_protection.go
```diff
@@ -35,3 +35,16 @@ func GetBotProtectionData(authflowCfg *config.AuthenticationFlowBotProtection, a
 	}
 	return nil
 }
+
+func IsBotProtectionRequired(authflowCfg *config.AuthenticationFlowBotProtection, appCfg *config.BotProtectionConfig) bool {
```
louischan-oursky, line 39: This function should call `GetBotProtectionData`, they are nearly identical.

pkg/lib/authenticationflow/errors.go
```diff
@@ -43,6 +43,8 @@ var ErrStepNotFound = apierrors.NotFound.WithReason("AuthenticationFlowStepNotFo
 
 var ErrUnknownFlow = apierrors.BadRequest.WithReason("AuthenticationFlowUnknownFlow").New("unknown flow")
 
+var ErrBotProtectionRequired = apierrors.BadRequest.WithReason("AuthenticationBotProtectionRequired").New("bot-protection required")
```
louischan-oursky, line 46: Use apierrors.Forbidden instead of BadRequest. It is not a bad request.

pkg/lib/authenticationflow/declarative/intent_use_identity_login_id.go
```diff
@@ -52,6 +52,26 @@ func (n *IntentUseIdentityLoginID) CanReactTo(ctx context.Context, deps *authflo
...
+			return nil, err
+		}
+
+		current, err := authflow.FlowObject(flowRootObject, n.JSONPointer)
+		if err != nil {
+			return nil, err
+		}
+		if currentBranch, ok := current.(config.AuthenticationFlowObjectBotProtectionConfigProvider); ok {
+			if IsBotProtectionRequired(currentBranch.GetBotProtectionConfig(), deps.Config.BotProtection) {
+				_, _, botProtectionVerified := authflow.FindMilestoneInCurrentFlow[MilestoneDidVerifyBotProtection](flows)
```
louischan-oursky, line 67: It should read from the context to determine if bot protection is has done already.

pkg/lib/authenticationflow/errors.go
pkong-ds: > that we will report in the API

These errors are not actual errors. They are not expected to be visible to users.

They acts similarly to `ErrEOF` to allow special handling in in caller.

Caller can then mutate the context/session on verification status. 

Note - the actual errors are in [pkg/lib/feature/botprotection/errors.go](https://github.com/authgear/authgear-server/pull/4390/files/8010123425bd85aad19dd9937e09514536e92698..911969044381271e059d16afdb749b5d302fb90b#diff-b17e6ea1c38095c33b5d8f42eba53871f909b88f90f74de31994c624a6ac479a)

Any thoughts?

pkg/lib/authenticationflow/declarative/input_verify_bot_protection.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (i *InputSchemaBotProtectionVerification) SchemaBuilder() validation.SchemaBuilder {
+
+	b := validation.SchemaBuilder{}.Type(validation.TypeObject)
+	required := []string{"type"}
+	b.Required(required...)
+	b.AdditionalPropertiesFalse()
+	b.Properties().Property("type", validation.SchemaBuilder{}.Type(validation.TypeString).Enum(config.BotProtectionProviderTypeCloudflare, config.BotProtectionProviderTypeRecaptchaV2))
+	b.Properties().Property("response", validation.SchemaBuilder{}.Type(validation.TypeString))
```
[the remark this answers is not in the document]
pkong-ds: I think for `cloudflare` & `recaptchav2`, it's safe to say it's a string

Maybe we can do `allOf` -> `if` -> `then` for conditional type of response?

pkong-ds:
Added recaptchav2 & cloudflare clients; Blocked by intent state transition problem, will continue tmr

pkong-ds:
As per offline discussion with @louischan-oursky , realized current `accept.go` logic would not support input validation of bot protection.

Hence, will close this PR and create another one with below approach:

In target branches Intents (or Nodes), add below pseudocode to `ReactTo`

```go
func (i *IntentXYZ) ReactTo(...) (...) {
  var bpResult *BotProtectionResult
  bpRequired = IsBotProtectionRequired(i)
  if bpRequired {
     bpResult, err := botprotectionutils.Verify(i)
  }
  
  // core logic, non-error return
  return nextNode, result.Error()
}
```

PR #4394, opened 2024-06-25
Inline preview customisation styles

cychiuae:
based on #4396


resources/authgear/templates/en/web/authflowv2/__html_head.html
```diff
@@ -22,6 +23,23 @@
...
+{{ else }}
+  --widget__bg-image: ;
+{{ end }}
+}
+{{ if $appBackgroundImageDark }}
+:root.dark {
+  --widget__bg-image: url("{{ $appBackgroundImageDark }}");
+}
+{{ end }}
+</style>
```
cychiuae, line 42: Not sure if this is the best way to implement background image for the widget.

pkg/util/template/func.go
```diff
@@ -120,6 +125,32 @@ func makeInclude(t tpl) func(tplName string, data any) (template.HTML, error) {
...
+		d := make(map[string]interface{})
+		d["Key"] = tranlsationKey
+		d["Value"] = included
+		err = t.ExecuteTemplate(buf, templateTranslationMessageTemplateName, d)
+		// Ignore gosec error because the app developer can actually write any template
+		// But we should be careful that do not pass any user input to this function
+		html := template.HTML(buf.String()) // nolint:gosec
+		return html, err
+	}
+}
```
cychiuae, line 152: Feel free to comment this approach

resources/authgear/templates/en/web/authflowv2/__html_head.html
```diff
@@ -10,6 +10,7 @@
 <meta name="x-cldr-locale" content="{{ $.ResolvedCLDRLocale }}">
 <meta name="x-color-scheme" content="{{ $.ColorScheme }}">
 <meta name="x-sentry-dsn" content="{{ $.AuthUISentryDSN }}">
+<meta name="x-window-message-allowed-origins" content="{{ $.AuthUIWindowMessageAllowedOrigins }}">
```
cychiuae, line 13: Is it a common practice to put auth ui config in meta tag?

resources/authgear/templates/en/web/__translation_message.html
```diff
@@ -0,0 +1,9 @@
+{{ define "__translation_message.html" }}
+<span
+  class="translated-message"
```
tung2744, line 3: This could be a class with style
```
display: contents;
```
To ensure it won't affect any layout.

cychiuae: 👍🏻

authui/src/authflowv2/components/layout.css
```diff
@@ -1,15 +1,33 @@
 @layer components {
+  :root {
+    --layout-flex-align-items: center;
```
tung2744, line 3: The alignment button is not only updating the card alignment, but it should affect all alignments (i.e. text alignment, content alignment).

We have added the `alignment-logo` `alignment-content` attributes for this purpose. You can search these strings in the code base.

And I prefer to define the layout alignments following the existing pattern.

```
.layout--default {
    :root[alignment-content="start"] & {
      @apply items-start;
    }

    :root[alignment-content="center"] & {
      @apply items-center;
    }

    :root[alignment-content="end"] & {
      @apply items-end;
    }
}
```

@fungc-io Please correct me if it is not the case.

portal/src/model/themeAuthFlowV2.ts
```diff
@@ -7,8 +7,42 @@ import {
...
+export const enum CSSVariable {
+  LayoutFixAlignItems = "--layout-flex-align-items",
+  LayoutPaddingLeft = "--layout-padding-left",
+  LayoutPaddingRight = "--layout-padding-right",
+  LayoutBackgroundColor = "--layout__bg-color",
+  LayoutBackgroundImage = "--layout__bg-image",
+  PrimaryButtonBackgroundColor = "--primary-btn__bg-color",
+  PrimaryButtonTextColor = "--primary-btn__text-color",
+  PrimaryButtonBorderRadius = "--primary-btn__border-radius",
+  InputFiledBorderRadius = "--input__border-radius",
```
tung2744, line 43: The border radius is lost after changed to other option and back to middle. Same problem in input border radius.
https://github.com/authgear/authgear-server/assets/34942210/f3d329c3-cbac-4cfe-a3ee-ecf3d817618c

authui/src/authflowv2/components/layout.css
tung2744, line 1: Screenshot 2024-06-28 at 4 40 03 PM

The layout looks weird when there is background color and the screen is small enough. As it is a authui problem, let me open another issue for it.

tung2744: https://linear.app/authgear/issue/DEV-1493/authui-looks-weird-when-screen-height-is-not-tall-enough-and

resources/authgear/templates/en/web/authflowv2/__html_head.html
tung2744: I think this is the current way to do it.

pkg/util/template/func.go
tung2744: LGTM

portal/src/model/themeAuthFlowV2.ts
```diff
@@ -7,8 +7,42 @@ import {
...
+  return option[theme];
+}
+
+export const enum CSSVariable {
+  LayoutFixAlignItems = "--layout-flex-align-items",
+  LayoutPaddingLeft = "--layout-padding-left",
+  LayoutPaddingRight = "--layout-padding-right",
+  LayoutBackgroundColor = "--layout__bg-color",
+  LayoutBackgroundImage = "--layout__bg-image",
+  PrimaryButtonBackgroundColor = "--primary-btn__bg-color",
```
tung2744, line 40: hover and active color also need to be set.
```
--primary-btn__bg-color--hover
--primary-btn__bg-color--active
```

Compute the color by using fluent ui theme generator. We currently use `themeDark` as the active and hover color.

ref: https://developer.microsoft.com/en-us/fluentui?fabricVer=5#/styles/themegenerator

authui/src/authflowv2/components/layout.css
cychiuae: According to design, I don't think this case should use `alignment-content`. Design is only the card is aligned to either side

cychiuae: Screenshot 2024-06-28 at 9 49 13 PM
Screenshot 2024-06-28 at 9 49 19 PM

cychiuae: But I can stick with the existing pattern and introduce another flag `alignment-card`

tung2744: ok. I remembered there was a design on content alignment customization. Maybe I mixed up something else.

tung2744 (approved):

PR #4396, opened 2024-06-25
Use path prefix for previewing pages

cychiuae:


cychiuae:
@tung2744 @louischan-oursky ready for review

pkg/util/httproute/httproute.go
```diff
@@ -41,6 +42,26 @@ func (r Route) WithPathPattern(pathPattern string) Route {
 	return r
 }
 
+// TODO: It may not works as expected if the path's pathern is using `/` for "regex"(?)
```
louischan-oursky, line 45: We do not support regex in Route. What do you mean?

pkg/auth/routes.go
```diff
@@ -236,6 +236,16 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource) *h
 		// anonymous user
 		p.Middleware(newSettingsSubRoutesMiddleware),
 	)
+	webappPagePreviewChain := httproute.Chain(
+		rootChain,
+		p.Middleware(newPublicOriginMiddleware),
+		p.Middleware(newPanicWebAppMiddleware),
+		httproute.MiddlewareFunc(httputil.NoStore),
+		httproute.MiddlewareFunc(webapp.IntlMiddleware),
+		p.Middleware(newWebAppUIParamMiddleware),
```
louischan-oursky, line 245: No need to use newWebAppUIParamMiddleware here.

pkg/auth/webapp/preview.go
```diff
@@ -1,15 +1,12 @@
...
-const (
-	PreviewModeInline = "inline"
+import (
+	"net/http"
+	"strings"
 )
 
-func IsPreviewModeInline(r *http.Request) bool {
-	return r.Method == http.MethodGet && r.URL.Query().Get(PreviewQueryKey) == PreviewModeInline
+const InlinePreviewPathPrefix = "/preview"
```
louischan-oursky, line 8: It should be `/preview/`. So it will not match `/preview_something_added_in_the_future`.

pkg/util/httproute/httproute.go
cychiuae: oh I thought `httproute.Route` support regex. but seems it only supports path param and `*`

pkg/auth/routes.go
cychiuae: I think we need it since we need to support changing locale

pkg/auth/webapp/preview.go
cychiuae: Updated here https://github.com/authgear/authgear-server/pull/4396/commits/6e163cc7098ae8607939a0af458244ce47ccc84e#diff-bf07c7b0ae1992e2a9f6495b630e067dc01e6900cfdbb787d88305e174f042f0R8

cychiuae:
@louischan-oursky Updated thanks.

pkg/auth/routes.go
louischan-oursky: webapp.IntlMiddleware does that job.

louischan-oursky: I removed for you.

cychiuae:
Thanks

louischan-oursky (approved):

PR #4407, opened 2024-06-27
Increase background image size limit in authui v2 customisation

cychiuae:
based on #4394

### DIFF
--- a/pkg/lib/web/image.go
+++ b/pkg/lib/web/image.go
@@ -40,10 +40,12 @@ var preferredExtensions = map[string]string{
 var imageRegex = regexp.MustCompile(`^static/([a-zA-Z0-9-]+)/(.+)\.(png|jpe|jpeg|jpg|gif)$`)
 
 type ImageDescriptor struct {
-	Name string
+	Name      string
+	SizeLimit int
 }
 
 var _ resource.Descriptor = ImageDescriptor{}
+var _ resource.SizeLimitDescriptor = ImageDescriptor{}
 
 func (a ImageDescriptor) MatchResource(path string) (*resource.Match, bool) {
 	matches := imageRegex.FindStringSubmatch(path)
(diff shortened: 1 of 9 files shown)


pkg/portal/appresource/manager_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	apimodel "github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	configtest "github.com/authgear/authgear-server/pkg/lib/config/test"
+	"github.com/authgear/authgear-server/pkg/lib/web"
```
cychiuae, line 16: During development, I found that actually one of the package imports performed a side effect that register `AuthgearYAMLDescriptor` into `resource.DefaultRegistry` so that the manager was able to find `authgear.yaml` in the test cases.

While other static resources cannot be found in the test cases because we didn't import `github.com/authgear/authgear-server/pkg/lib/web` and those static resources weren't registered into default registry.

portal/src/graphql/portal/DesignScreen/form.ts
```diff
@@ -273,7 +293,12 @@ export function useBrandDesignForm(appID: string): BranchDesignForm {
         ),
       },
     };
-  }, [resourceForm, selectedLanguage, configForm.state.fallbackLanguage]);
+  }, [
+    resourceForm,
+    getResourceFormByResourceDefinition,
+    selectedLanguage,
+    configForm.state.fallbackLanguage,
+  ]);
```
cychiuae, line 301: Here will conflict with #4394. I will handle the conflict later

tung2744:
@cychiuae Please resolve conflicts

pkg/lib/web/image.go
```diff
@@ -303,3 +305,7 @@ func (a ImageDescriptor) viewByPath(resources []resource.ResourceFile, path stri
 		Data: bytes,
 	}, nil
 }
+
+func (a ImageDescriptor) GetSizeLimit() int {
+	return a.SizeLimit
```
tung2744, line 310: Give it a default? (100KB)

cychiuae: Is it I check if SizeLimit is empty value, i.e. 0, then return default value 100kb?

tung2744: Yes

tung2744 (approved):

PR #4408, opened 2024-06-27
Exchange cookies using AppInitiatedSSOToWebToken

tung2744:
ref DEV-1406
requires #4403


louischan-oursky:
Rebase.

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -26,9 +26,10 @@ import (
 //go:generate mockgen -source=handler_authz.go -destination=handler_authz_mock_test.go -package handler_test
 
 const (
-	CodeResponseType          = "code"
-	NoneResponseType          = "none"
-	SettingsActonResponseType = "urn:authgear:params:oauth:response-type:settings-action"
+	CodeResponseType                      = "code"
+	NoneResponseType                      = "none"
+	SettingsActonResponseType             = "urn:authgear:params:oauth:response-type:settings-action"
+	AppInitiatedSSOToWebTokenResponseType = "urn:authgear:params:oauth:response-type:app_initiated_sso_to_web token"
```
louischan-oursky, line 34: I think we should define AppInitiatedSSOToWebTokenResponseType as "urn:authgear:params:oauth:response-type:app_initiated_sso_to_web" and then parse the response type as an unordered list of string. if the unordered list of string is ["urn:authgear:params:oauth:response-type:app_initiated_sso_to_web", "token"] then it is the request we want to handle. Otherwise it is invalid response type.

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -103,18 +104,19 @@ type AuthorizationHandler struct {
...
+	ValidateScopes                   ScopesValidator
+	AppSessionTokenService           AppSessionTokenService
+	AuthenticationInfoService        AuthenticationInfoService
+	Clock                            clock.Clock
+	Cookies                          CookieManager
+	OAuthSessionService              OAuthSessionService
+	CodeGrantService                 CodeGrantService
+	SettingsActionGrantService       SettingsActionGrantService
+	ClientResolver                   OAuthClientResolver
+	AppInitiatedSSOToWebTokenService oauth.AppInitiatedSSOToWebTokenService
```
louischan-oursky: Use local interface so that we can generate a mock.

pkg/lib/oauth/error.go
```diff
@@ -5,3 +5,4 @@ import "errors"
 var ErrAuthorizationNotFound = errors.New("oauth authorization not found")
 var ErrAuthorizationScopesNotGranted = errors.New("oauth authorization scopes not granted")
 var ErrGrantNotFound = errors.New("oauth grant not found")
+var ErrUnmatchedClient = errors.New("unmatched client id")
```
louischan-oursky, line 8: unmatched client ID

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -605,7 +633,22 @@ func (h *AuthorizationHandler) validateRequest(
...
+			return protocol.NewError("invalid_request", "id_token_hint is required when using app-initiated-sso-to-web")
+		}
+		if r.AppInitiatedSSOToWebToken() == "" {
+			return protocol.NewError("invalid_request", "x_app_initiated_sso_to_web_token is required when using app-initiated-sso-to-web")
+		}
+		if r.ResponseMode() != "cookie" {
+			return protocol.NewError("invalid_request", "only 'response_mode=cookie' is supported when using app-initiated-sso-to-web")
+		}
 	default:
 		return protocol.NewError("unsupported_response_type", "only 'code' response type is supported")
```
louischan-oursky, line 719: The message is now outdated. Maybe just output the raw response type we receive in the query (that is, the string, not the parsed response type).

```go
return protocol.NewError("unsupported_response_type", fmt.Sprintf("unsupported response_type: %v", rawResponseTypeString))
```

pkg/lib/oauth/handler/service_token.go
```diff
@@ -32,6 +32,8 @@ type TokenService struct {
 	GenerateToken       TokenGenerator
 	Clock               clock.Clock
 	Users               TokenHandlerUserFacade
+
+	AccessGrantService oauth.AccessGrantService
```
louischan-oursky, line 36: Use local interface to generate mock. TokenServiceAccessGrantService

pkg/lib/oauth/error.go
```diff
@@ -6,3 +6,4 @@ var ErrAuthorizationNotFound = errors.New("oauth authorization not found")
 var ErrAuthorizationScopesNotGranted = errors.New("oauth authorization scopes not granted")
 var ErrGrantNotFound = errors.New("oauth grant not found")
 var ErrUnmatchedClient = errors.New("unmatched client id")
+var ErrUnmatchedSession = errors.New("unmatched session id")
```
louischan-oursky, line 9: unmatched session ID

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
 ) (httputil.Result, error) {
-	accessToken, err := h.AppInitiatedSSOToWebTokenService.ExchangeForAccessToken(client, r.AppInitiatedSSOToWebToken())
+	idTokenHint, ok := r.IDTokenHint()
+	if !ok {
+		panic("cannot get id_token_hint, the request should be validated")
+	}
+	idToken, err := h.IDTokenIssuer.VerifyIDTokenWithoutClient(idTokenHint)
 	if err != nil {
-		return nil, protocol.NewError("invalid_grant", "invalid x_app_initiated_sso_to_web_token")
+		return nil, protocol.NewError("invalid_grant", "invalid id_token_hint")
```
louischan-oursky, line 502: invalid_request

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
+	}
+	var sidInt interface{}
+	if sidInt, ok = idToken.Get(string(model.ClaimSID)); !ok {
+		return nil, protocol.NewError("invalid_grant", "required sid in id_token_hint")
```
louischan-oursky, line 506: invalid_request

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
+	}
+	var sid string
+	if sid, ok = sidInt.(string); !ok {
+		return nil, protocol.NewError("invalid_grant", "sid is not a string in id_token_hint")
```
louischan-oursky, line 510: invalid_request

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
+	}
+	_, sessionID, ok := oidc.DecodeSID(sid)
+	if !ok {
+		return nil, protocol.NewError("invalid_grant", "invalid sid format id_token_hint")
```
louischan-oursky, line 514: invalid_request

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
+	}
+
+	accessToken, err := h.AppInitiatedSSOToWebTokenService.ExchangeForAccessToken(
+		client,
+		sessionID,
+		r.AppInitiatedSSOToWebToken(),
+	)
+	if err != nil {
+		if err == oauth.ErrUnmatchedClient {
```
louischan-oursky, line 523: Use errors.Is

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
+			return nil, protocol.NewError("invalid_grant", "incorrect client_id")
```
louischan-oursky, line 524: invalid_request

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
+		}
+		if err == oauth.ErrUnmatchedSession {
```
louischan-oursky, line 526: Use errors.Is

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
+			return nil, protocol.NewError("invalid_grant", "incorrect sid in id_token_hint")
```
louischan-oursky, line 527: invalid_request

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
+		}
+		if err == oauth.ErrGrantNotFound {
```
louischan-oursky, line 529: Use errors.Is

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -482,9 +484,43 @@ func (h *AuthorizationHandler) doHandleAppInitiatedSSOToWeb(
...
+			return nil, protocol.NewError("invalid_grant", "invalid x_app_initiated_sso_to_web_token")
```
louischan-oursky, line 530: invalid_request

tung2744:
@louischan-oursky I found there is a problem is the current device secret rotation mechanism.
Currently, we rotate the device secret when:
1. Refresh tokens
2. Refresh id token

And we always generate a new device_secret.

For 1, only a valid refresh_token is needed. And therefore when user have a valid refresh_token, he can always obtain a new valid device secret by refreshing the tokens. This makes the device_secret meaningless.

For 2, only a valid access_token is needed, which can be obtained by refresh tokens. Therefore, it have the same problem of 1.

So I think we should,

For 1, only rotate the device_secret if a valid device_secret is provided in the request, else treat it as no device_secret provided.
For 2, same as above, only return a new device_secret if a valid device_secret is in the request. Else, only return a new id_token with a `ds_hash` not changed.

pkg/lib/oauth/handler/service_token.go
tung2744: It seems we don't need mock here.

tung2744:
Per offline discussion, we will solve this problem by DPoP, as it binds the device_secret, refresh_token to the device.
Will not change the current behavior.

tung2744:
@louischan-oursky Addressed comments above, thanks!

tung2744:
Added a commit to prevent refreshing id token returning a new device_secret

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -716,7 +716,7 @@ func (h *AuthorizationHandler) validateRequest(
 			return err
 		}
 	default:
-		return protocol.NewError("unsupported_response_type", "only 'code' response type is supported")
+		return protocol.NewError("unsupported response_type: %v", r.ResponseType())
```
louischan-oursky: It should still be `unsupported_response_type` https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1 But we change the error description to `fmt.Sprintf("response_type %v is not supported")`

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -398,6 +398,25 @@ func (h *TokenHandler) rotateDeviceSecret(
...
+
+	if subtle.ConstantTimeCompare([]byte(oauth.HashToken(deviceSecret)), []byte(offlineGrant.DeviceSecretHash)) != 1 {
+		// If the provided device sercet is invalid, do not rotate
+		return offlineGrant, false, nil
+	}
+
+	return h.rotateDeviceSecretIfPossible(authorizedScopes, offlineGrant, resp)
+}
+
+func (h *TokenHandler) rotateDeviceSecretIfPossible(
```
louischan-oursky, line 419: I suggest we name this rotateDeviceSecretIfSufficientScope.

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -398,6 +398,25 @@ func (h *TokenHandler) rotateDeviceSecret(
 }
 
 func (h *TokenHandler) rotateDeviceSecretIfNeeded(
```
louischan-oursky, line 400: I suggest we rename this to rotateDeviceSecretIfDeviceSecretIsPresentAndValid

pkg/lib/oauth/handler/handler_authz.go
tung2744: Oh sorry it seems I deleted something wrong :sosad:

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #4419, opened 2024-07-02
Implement back to your app configuration

cychiuae:
based on #4394


tung2744:
ci failed but it doesn't seems to be related to the change, will merge

tung2744 (approved):

PR #4427, opened 2024-07-04
Allow set `AuthUIWindowMessageAllowedOrigins` with host

cychiuae:


cychiuae:
@tung2744 @louischan-oursky Ready

louischan-oursky:
We do not use the plain `go fmt` to format our code, we use `goimports`. You can run `make fmt` to format the code. Basically, the only difference is the format of the imports. With goimports, we separate the imports into three groups. 1st group is stdlib, 2nd group is third party packages, 3rd group is the packages from our own module.

Let me run `make fmt` for you this time.

louischan-oursky (approved):

PR #4429, opened 2024-07-05
Fix array.at() to adapt older version browser

yyorkchan:


cychiuae:
LGTM

@tung2744

authui/src/authflowv2/customSelect.ts
```diff
@@ -344,7 +344,7 @@ export class CustomSelectController extends Controller {
   renderTrigger() {
     const option =
       this.optionsValue.find((option) => option.value === this.value) ??
-      this.optionsValue.at(0);
+      this.optionsValue[0];
```
tung2744, line 347: ci failed because typescript thinks `this.optionsValue[0]` is always not `undefined`.
It seems you need to explicitly write the type of `option` as `SearchSelectOption | undefined`.

yyorkchan:
Ready, Thanks
@louischan-oursky @tung2744

tung2744:
The ci is still failing because eslint is too clever that it knows the value was never assigned as undefined.

@louischan-oursky  It seems what we want is this:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#checked-indexed-accesses---nouncheckedindexedaccess

But for this issue, try
```typescript
this.optionsValue[0] as (SearchSelectOption | undefined)
```

louischan-oursky:
Seems like we should turn on those strict compiler flags that are not included by `strict` https://www.typescriptlang.org/tsconfig/#compilerOptions

cychiuae:
For this case can turn on https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess

louischan-oursky:
@yyorkchan Need you to run `cd authui && npm run format`

yyorkchan:
Passed all test thanks for the help @tung2744 @louischan-oursky

louischan-oursky (approved):

PR #4430, opened 2024-07-05
Fix access token issued by browser sso enabled and authenticated with app access token session will always be invalid

tung2744:
ref DEV-1406


louischan-oursky:
Per offline discussion, we will add sessionKind (or sessionType?) to authorization code. In grant_type=authorization_code, we will either issue a new offline grant if sessionKind is IDP, or append a refresh token if sessionKind is offline grant.

tung2744:
Let me open another pr to fix that

PR #4431, opened 2024-07-05
Create new refresh token inside original offline grant if offline grant is used as session in select account

tung2744:
ref DEV-1406


louischan-oursky:
`authgear-test` failed.

tung2744:
Fixed, thanks!

louischan-oursky (approved):

PR #4433, opened 2024-07-08
Adjust top bottom min spacing

cychiuae:


authui/src/authflowv2/components/layout.css
```diff
@@ -49,8 +49,8 @@
     .layout__header--default,
     .layout__footer--default {
       @apply flex-1;
-      min-height: 20px;
-      max-height: 160px;
+      min-height: 1.25rem;
+      max-height: 10rem;
```
louischan-oursky, line 53: I think you can use tailwind classes here.

cychiuae:
@louischan-oursky Updated. Thanks!

louischan-oursky:
I added a commit to fix the wrong calculation.

louischan-oursky (approved):

PR #4439, opened 2024-07-09
Fix ios autofill reveal input in otp

yyorkchan:
Demo video:

https://github.com/authgear/authgear-server/assets/79845066/356736ef-a9c8-44ee-b8e2-9bb4d0ca743a

Ignore the internal error, it is to stop submitting the code instantly for demo.

### DIFF
--- a/authui/src/authflowv2/components/otp-input.css
+++ b/authui/src/authflowv2/components/otp-input.css
@@ -20,6 +20,8 @@
     /* stylelint-disable-next-line plugin/no-unsupported-browser-features */
     --otp-input__digit-mask-size: 0.625rem;
 
+    --otp-input__dev-1511-color-transparent: transparent;
+
     --otp-input-error-message__margin-top: 0.5rem;
     --otp-input-error-message__text-color: var(--color-error);
     --otp-input-error-message__font-family: var(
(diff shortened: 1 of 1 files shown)


cychiuae:
To give more context, the yellow background of the input is from system. seems cannot be overriden

authui/src/authflowv2/components/otp-input.css
```diff
@@ -63,6 +63,8 @@
       @apply caret-white/0;
       /* iOS treats element with `opacity: 0` as not interactable, set text to transparent instead */
       @apply text-transparent;
+      /* To solve iOS autofill reveals text color */
+      -webkit-text-fill-color: var(--otp-input__bg-color);
```
louischan-oursky, line 67:
```
--otp-input__dev-1511-color-transparent: transparent;

// https://linear.app/authgear/issue/DEV-1511
// When iOS autofills a code from SMS, the filled text will use -webkit-text-fill-color instead of color as text color, so we need to set both property to transparent.
// Continue to write down why simple transparent does not work. Attach the stackoverflow link here.
-webkit-text-fill-color: var(--otp-input__dev-1511-color-transparent);
```

louischan-oursky:
@yyorkchan Please resolve the conflict. It conflicted with another PR done by you.

yyorkchan:
Resolved. Thanks @louischan-oursky

louischan-oursky (approved):

PR #4448, opened 2024-07-12
Hard to find the customization option when the portal width

yyorkchan:
Now the window start from the top when the screen becomes narrow.
https://github.com/user-attachments/assets/b2d53425-2fd7-46f9-92ef-98a191788ee0

### DIFF
--- a/authui/src/passkey.ts
+++ b/authui/src/passkey.ts
@@ -423,7 +423,6 @@ export class AuthflowPasskeyRequestController extends Controller {
   declare hasButtonTarget: boolean;
 
   connect() {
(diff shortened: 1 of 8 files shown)


louischan-oursky:
What issue does this PR is supposed to fix? I see some commits for fixing nosript, some commits for fix the layout of the Design page

yyorkchan:
Thanks @louischan-oursky

louischan-oursky:
For this case, you should rebase your PR to remove the irrelevant commits, since I do not even start reviewing it. This makes the history more clean. Also, the title of this PR seems clipped. Please correct it.

PR #4456, opened 2024-07-15
Make authui usable in no script

yyorkchan:


pkg/util/template/func_test.go
```diff
@@ -185,7 +185,7 @@ func TestFuncs(t *testing.T) {
 		})
 
 		Convey("With HTML", func() {
-			So(trimHTML(template.HTML("  A  B ")), ShouldEqual, "A  B")
+			So(trimHTML(template.HTML("  A  B ")), ShouldEqual, template.HTML("A  B"))
```
cychiuae, line 188: I think maybe convert this test case to using html tag
`So(trimHTML(template.HTML("  <span>A  B</span> ")), ShouldEqual, template.HTML("<span>A  B</span>"))`

authui/src/authflowv2/phoneInput.ts
```diff
@@ -312,6 +318,9 @@ export class PhoneInputController extends Controller {
   connect() {
     void this.initPhoneCode();
     this.phoneInputTarget.addEventListener("blur", this.handleInputBlur);
+    this.phoneInputTarget.parentElement?.classList.remove("hidden");
+    this.countrySelectInputTarget.parentElement?.classList.remove("hidden");
```
cychiuae, line 322: Maybe target the exact element but not access with `element.parentElement`. Otherwise, it will break when the dom structure is changed

yyorkchan:
Thanks a lot @louischan-oursky

louischan-oursky (approved):

PR #4458, opened 2024-07-15
Update Date format in the portal

yyorkchan:
Note: Cannot check subscription pages layout since it requires stripe api key

### DIFF
--- a/portal/src/util/formatDateOnly.ts
+++ b/portal/src/util/formatDateOnly.ts
@@ -1,0 +1,22 @@
+import { DateTime, DateTimeFormatOptions } from "luxon";
+
+export function formatDateOnly(
+  locale: string,
+  date: Date | string | null,
+  format: DateTimeFormatOptions = DateTime.DATE_MED
+): string | null {
+  let datetime;
+
+  if (date instanceof Date) {
+    datetime = DateTime.fromJSDate(date);
+  } else if (typeof date === "string") {
+    datetime = DateTime.fromISO(date);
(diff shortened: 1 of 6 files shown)


portal/src/util/formatDatetime.ts
```diff
@@ -3,7 +3,7 @@ import { DateTime, DateTimeFormatOptions } from "luxon";
 export function formatDatetime(
   locale: string,
   date: Date | string | null,
-  format: DateTimeFormatOptions = DateTime.DATETIME_SHORT
+  format: DateTimeFormatOptions = DateTime.DATETIME_MED_WITH_SECONDS
```
louischan-oursky, line 6: Remove this argument. The purpose of this change is to forbid arbitrary format. formatDatetime always use DATETIME_MED_WITH_SECONDS

portal/src/util/formatDateOnly.ts
```diff
@@ -0,0 +1,22 @@
+import { DateTime, DateTimeFormatOptions } from "luxon";
+
+export function formatDateOnly(
+  locale: string,
+  date: Date | string | null,
+  format: DateTimeFormatOptions = DateTime.DATE_MED
```
louischan-oursky, line 6: Remove this argument. formatDateOnly always uses DATE_MED

louischan-oursky:
@yyorkchan It failed to pass eslint.

louischan-oursky (approved):

PR #4464, opened 2024-07-17
Fix LegalParser and add more tests to demo the behavior of the two parsers

louischan-oursky:
ref DEV-1139


louischan-oursky:
This PR currently fix the incorrect behavior of LegalParser only.

louischan-oursky:
Not using isPossibleNumber is probably a mistake when we introduced LegalParser. This PR fixes that. @tung2744

tung2744 (approved):

PR #4466, opened 2024-07-17
Fix access token issued can be resolved to a wrong user id

tung2744:
The bug occurs when:
1. You have an access token session of user A
2. You try to authenticate as user B
3. You got an access token of user B (sub=user-b-id), but internally the session id was the session in 1. As a result the server will think you are user A.

ref DEV-1589

### DIFF
--- a/pkg/latte/intent_ensure_session.go
+++ b/pkg/latte/intent_ensure_session.go
@@ -58,7 +58,7 @@ func (i *IntentEnsureSession) ReactTo(ctx context.Context, deps *workflow.Depend
 		mode = EnsureSessionModeCreate
 	}
 
-	authnInfo := sessionToCreate.GetAuthenticationInfo()
+	authnInfo := sessionToCreate.GetAuthenticationInfoByThisSession()
 	authnInfo.ShouldFireAuthenticatedEventWhenIssueOfflineGrant = mode == EnsureSessionModeNoop && i.CreateReason == session.CreateReasonLogin
 	authnInfoEntry := authenticationinfo.NewEntry(
 		authnInfo,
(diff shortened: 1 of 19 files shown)


tung2744:
@louischan-oursky Updated according to our latest discussion, thanks!

pkg/latte/intent_ensure_session.go
```diff
@@ -58,7 +58,7 @@ func (i *IntentEnsureSession) ReactTo(ctx context.Context, deps *workflow.Depend
 		mode = EnsureSessionModeCreate
 	}
 
-	authnInfo := sessionToCreate.GetAuthenticationInfo()
+	authnInfo := sessionToCreate.GetAuthenticationInfoByThisSession()
```
louischan-oursky, line 61: Move 61-63 to the end of this method.

pkg/lib/authenticationflow/declarative/node_do_create_session.go
```diff
@@ -35,7 +35,7 @@ func NewNodeDoCreateSession(ctx context.Context, deps *authflow.Dependencies, fl
 	s, token := deps.IDPSessions.MakeSession(attrs)
 	sessionCookie := deps.Cookies.ValueCookie(deps.SessionCookie.Def, token)
 
-	authnInfo := s.GetAuthenticationInfo()
+	authnInfo := s.GetAuthenticationInfoByThisSession()
```
louischan-oursky, line 38: Refactor this part so that we do not write `s = nil` or `sessionCookie = nil`

pkg/lib/interaction/nodes/do_ensure_session.go
```diff
@@ -44,7 +44,7 @@ func (e *EdgeDoEnsureSession) Instantiate(ctx *interaction.Context, graph *inter
 	sessionToCreate, token := ctx.Sessions.MakeSession(attrs)
 	sessionCookie := ctx.CookieManager.ValueCookie(ctx.SessionCookie.Def, token)
 
-	authenticationInfo := sessionToCreate.GetAuthenticationInfo()
+	authenticationInfo := sessionToCreate.GetAuthenticationInfoByThisSession()
```
louischan-oursky, line 47: Use a conservative way to refactor this part.

tung2744:

louischan-oursky (approved):

PR #4472, opened 2024-07-18
Remove unused `$has_oauth` in login.html

pkong-ds:


louischan-oursky (commented):
signup.html also have this problem. Please fix it as well.

louischan-oursky (approved):

PR #4485, opened 2024-07-22
Specify LDAP support

louischan-oursky:
ref DEV-1578

Tag @cychiuae @yyorkchan


docs/specs/ldap.md
fungc-io, line 1: middle of the "Configuration of LDAP servers" part cannot be rendered

louischan-oursky: Sorry, just fixed it.

docs/specs/ldap.md
```diff
@@ -0,0 +1,133 @@
...
+
+> Why does `identity.ldap.servers.url` allow scheme, host, and port?
+> The LDAP URL, defined in [Section 2 in RFC 4516](https://datatracker.ietf.org/doc/html/rfc4516#section-2), is syntactically different from the URL defined in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986).
+> In particular, a LDAP URL can contain multiple question mark characters.
+> To ease implementation, we do not support the LDAP URL, and require a RFC3986 URL (which is implemented by the standard library net/url package).
+
+- `identity.ldap.servers.base_distinguished_name`: The base distinguished name to construct a Search Request, as defined in [Section 4.5.1 in RFC4511](https://datatracker.ietf.org/doc/html/rfc4511#section-4.5.1).
+- `identity.ldap.servers.relative_distinguished_name_attribute`: The attribute name Authgear should use to construct the search request. For example, if the value is `uid`, and the end-user gives a username of `user1`, and the base distinguished name is `dc=example,dc=com`, then the relative distinguished name is `uid=user1`, and the distinguished name is `uid=user1,dc=example,dc=com`.
```
cychiuae, line 65: I think we need to do a search request to get user's DN in ldap rather just prepend `uid=user1` to the the base DN.

From what I read, usually there will be multiple entry nodes between the root and the entry representing the user in the directory information tree. i.e. Usually there will be one or more organization units in between so that the DN usually would look like `cn=Some User,ou=Department Name,ou=A Region,dc=example,dc=com`

I am not so sure we need to document this or not. Or is it already the implementation detail? 🤔

louischan-oursky: In your example `cn=Some User,ou=Department Name,ou=A Region,dc=example,dc=com`, the `base_distinguished_name` is `ou=Department Name,ou=A Region,dc=example,dc=com`. This base DN is configured by the developer.

What we collect from the end-user are the username and the password. We need to combine the username with the base DN, so I design `relative_distinguished_name_attribute`.

cychiuae: What if users are actually from two different departments and different region? 
.e.g.
1. `cn=A Developer,ou=Engineering Team,ou=Hong Kong,dc=example,dc=com`
2. `cn=A Design,ou=Design Team,ou=United Kingdom,dc=example,dc=com`

In such case, we cannot just simple combine end user input and the base DN. Also, end user input (i.e. `uid` or `email`) may not appears in the DN in the directory. It is up to ldap admin to configure to use either `cn` or other rdn like `uid` to representing the DN

docs/specs/ldap.md
```diff
@@ -0,0 +1,183 @@
...
+    items:
+    - name: default
+      username: authgear
+      password: secret1
+  key: ldap
+```
+
+- `items.name`: To associate a LDAP server in `authgear.yaml`.
+- `items.username`: Optional. The username Authgear uses to authenticate itself to the LDAP server. If it is not provided, then Authgear does not authenticates itself, and assumes the LDAP server allows anonymous requests.
+- `items.password`: Optional. The password Authgear uses to authenticate itself to the LDAP server. If `username` is provided, then `password` is required.
```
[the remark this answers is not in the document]
cychiuae: Should we specific this user is for search request?

louischan-oursky:
@yyorkchan @cychiuae I have added two commits to update the spec.

1. Specify the testLDAPConnection mutation, and the errors
2. Rename oid to attribute name.

docs/specs/ldap.md
```diff
@@ -0,0 +1,226 @@
...
+This section documents the expected errors.
+
+|Description|Name|Reason|Info|
+|---|---|---|---|
+|In testLDAPConnection, if connection cannot be established with the LDAP server|ServiceUnavailable|LDAPConnectionTestFailed|`"cause": "FailedToConnect"`|
+|In testLDAPConnection, if search user credentials are invalid|ServiceUnavailable|LDAPConnectionTestFailed|`"cause": "FailedToBindSearchUser"`|
+|In testLDAPConnection, if base DN is invalid|ServiceUnavailable|LDAPConnectionTestFailed|`"cause": "InvalidBaseDN"`|
+|In testLDAPConnection, if search filter template is invalid|ServiceUnavailable|LDAPConnectionTestFailed|`"cause": "InvalidSearchFilterTemplate"`|
+|In testLDAPConnection, if `user_id_attribute_name` is invalid|ServiceUnavailable|LDAPConnectionTestFailed|`"cause": "InvalidUserIDAttributeName"`|
+|In testLDAPConnection, if endUserUsername is given, but the user is not found|ServiceUnavailable|LDAPConnectionTestFailed|`"cause": "TestingEndUserNotFound"`|
```
yyorkchan, line 223: Add Error for more than one user return

PR #4492, opened 2024-07-24
Debug e2e run fail

IniZio:


louischan-oursky:
Here is an instance of the timeout https://github.com/authgear/authgear-server/actions/runs/10058340878/job/27801204288 The first thing is to find out the condition to reproduce it. Some directions for you to try, adjust number of database connection with envvar.

PR #4494, opened 2024-07-25
Specify 2FA grace period

IniZio:
ref DEV-1614

### DIFF
--- a/docs/specs/2fa-grace-period.md
+++ b/docs/specs/2fa-grace-period.md
@@ -14,13 +14,13 @@ After the grace period, the user will be required to enroll in 2FA before they c
 
 ## Configuration
 
-By default, no grace period is provided. User must enroll in 2FA immediately after [2FA mode](./user-model.md#secondary-authenticator) is set to `required`.
+By default, no grace period is provided. User without 2FA must contact admin to login after [2FA mode](./user-model.md#secondary-authenticator) is set to `required`.
 
 For new users, the grace period starts from the time the user is created, whereas for existing users, the grace period starts from when the grace period is enabled.
 
 ### Global Grace Period
 
-After 2FA mode is set to `required`, a global grace period can be configured as follow:
+Global grace period can be enabled for forcing users to enroll in 2FA upon login.
 
 ```yaml
 authentication:
@@ -45,37 +45,30 @@ It is possible to extend the grace period of a user multiple times or cancel imm
 
 ### Customized Authentication Flow
 
-By default all signup / login flows respect the global 2FA mode configuration. However, you can also override it in the flow configuration:
+Customized authentication flow can use `allow_enrollment: true` to allow enrolling any of the following authenticators before proceeding.
 
 ```yaml
 authentication_flow:
   login_flows:
     - name: internal_user
-      secondary_authentication_mode: "required"
       steps:
-      # ....
-    - name: public_user
-      secondary_authentication_mode: "if_exists"
-      steps:
-      # ...
-
-  signup_flows:
-    - name: default
-      secondary_authentication_mode: "disabled"
-      steps:
-      # ...
-
-  signup_login_flows:
-    - name: default
-      secondary_authentication_mode: "required"
-      steps:
-      # ...
-
-  reauth_flows:
-    - name: default
-      secondary_authentication_mode: "required"
-      steps:
-      # ...
+        - type: identify
+          one_of:
+            - identification: phone
+            - identification: email
+        - type: authenticate
+          one_of:
+            - authentication: primary_password
+        - type: authenticate
+          # Requires user to satisfy one of the following authentication.
+          optional: false # or null
+          # If the end-user has no applicable authentication method,
+          # then enrollment will be required before proceeding.
+          # allow_enrollment by default is false, meaning user with no applicable method beforehand will be blocked from proceeding.
+          allow_enrollment: true
+          one_of:
+            - authentication: secondary_totp
+            - authentication: recovery_code
 ```
 
 When 2FA mode has been set to `required` and first 2FA step is met, user will be required to either use existing 2FA or create a new one.
--- a/docs/specs/2fa-grace-period.md
+++ b/docs/specs/2fa-grace-period.md
@@ -75,7 +75,7 @@ Both `setMFAGracePeriod` and `removeMFAGracePeriod` mutations are indempotent, u
 
 |Description|Name|Reason|Info|
 |---|---|---|---|
-|Invalid Grace Period e.g. in the past|`Invalid`|`GracePeriodInvalid`|`endAt` must be in the future.|
+|Invalid Grace Period<br />i.e. in the past or too far in the future (> 90 days)|`Invalid`|`MFAGracePeriodInvalid`|-|
 
 ### Changes on database schema
 
--- a/docs/specs/2fa-grace-period.md
+++ b/docs/specs/2fa-grace-period.md
@@ -42,12 +42,12 @@ type User {
   mfaGracePeriodEndAt: DateTime
 }
 
-type GrantMFAGracePeriodInput {
+type SetMFAGracePeriodInput {
   userID: ID!
   endAt: DateTime!
 }
 
-type GrantMFAGracePeriodPayload {
+type SetMFAGracePeriodPayload {
   user: User!
 }
 
@@ -60,9 +60,9 @@ type RemoveMFAGracePeriodPayload {
 }
 
 type Mutation {
-  grantMFAGracePeriod(
-    input: GrantMFAGracePeriodInput!
-  ): GrantMFAGracePeriodPayload!
+  setMFAGracePeriod(
+    input: SetMFAGracePeriodInput!
+  ): SetMFAGracePeriodPayload!
   removeMFAGracePeriod(
     input: RemoveMFAGracePeriodInput!
   ): RemoveMFAGracePeriodPayload!
--- a/docs/specs/2fa-grace-period.md
+++ b/docs/specs/2fa-grace-period.md
@@ -45,7 +45,7 @@ It is possible to extend the grace period of a user multiple times or cancel imm
 
 ### Customized Authentication Flow
 
-Customized authentication flow can use `allow_enrollment: true` to allow enrolling any of the following authenticators before proceeding.
+Customized authentication flow can use `enrollment_allowed: true` to allow enrolling any of the following authenticators before proceeding.
 
 ```yaml
 authentication_flow:
@@ -64,11 +64,14 @@ authentication_flow:
           optional: false # or null
           # If the end-user has no applicable authentication method,
           # then enrollment will be required before proceeding.
-          # allow_enrollment by default is false, meaning user with no applicable method beforehand will be blocked from proceeding.
-          allow_enrollment: true
+          # enrollment_allowed by default is false, meaning user with no applicable method beforehand will be blocked from proceeding.
+          enrollment_allowed: true
           one_of:
             - authentication: secondary_totp
             - authentication: recovery_code
 ```
 
-When 2FA mode has been set to `required` and first 2FA step is met, user will be required to either use existing 2FA or create a new one.
+#### Notes
+
+- Currently only secondary authenticators are supported for enrollment.
+- `enrollment_allowed` is ignored when `optional` is `true` to match default behaviour.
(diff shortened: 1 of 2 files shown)


docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Abstract
+
+In order to enforce 2FA, we will provide a grace period for users to enroll in 2FA after they have signed up.
+
+After the grace period, the user will be required to enroll in 2FA before they can sign in.
+
+## Configuration
+
+By default, no grace period is provided. User must enroll in 2FA immediately after [2FA mode](./user-model.md#secondary-authenticator) is set to `required`.
```
IniZio, line 17: Change default behaviour to force enroll or add another mode? 

Auth0 always allows user to enroll on login https://auth0.com/docs/secure/multi-factor-authentication/customize-mfa/customize-mfa-enrollments-universal-login#how-it-works

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,81 @@
...
+  secondary_authentication_mode: "required"
+  secondary_authentication_rollout:
+    per_user_grace_period_days: 10
+```
+
+It is possible to extend the grace period of a user multiple times or cancel immediately through Admin Portal / GraphQL API.
+
+### Customized Authentication Flow
+
+By default all signup / login flows respect the global 2FA mode configuration. However, you can also override it in the flow configuration:
```
IniZio, line 48: Another user-case in mind was for step-up authentication, for example:

1. User login with public_user flow, which does not have 2FA
2. User want to do risky actions
3. User reauth with internal_user flow, which forces 2FA enrollment

docs/specs/2fa-grace-period.md
IniZio: Default behaviour will remain block instead of force enrollment after grace period.

IniZio:
@louischan-oursky Updated to use `enrollment_allowed` flag in custom authflow 🙏

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,74 @@
...
+        - type: authenticate
+          one_of:
+            - authentication: primary_password
+        - type: authenticate
+          # Requires user to satisfy one of the following authentication.
+          optional: false # or null
+          # If the end-user has no applicable authentication method,
+          # then enrollment will be required before proceeding.
+          # allow_enrollment by default is false, meaning user with no applicable method beforehand will be blocked from proceeding.
+          allow_enrollment: true
```
IniZio, line 68: - Should we disallow using it along with `optional: true` for now? Otherwise will need a screen for user to optionally choose to enroll

- Should we have primary authenticators respect `allow_enrollment` as well?

louischan-oursky: 1. Name it `enrollment_allowed`
2. The possible combinations are `enrollment_allowed: true`, `optional: true`, ` `. Use JSON schema to enforce these combinations.

||Has authenticator|Has no authenticator|
|---|---|---|
|enrollment_allowed: true|simply authenticate|force enrollment|
|optional: true|simply authenticate|skip authenticate|
||simply authenticate|show error|

3. Should we have primary authenticators respect allow_enrollment as well? no for now

IniZio: Updated.

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,77 @@
...
+For new users, the grace period starts from the time the user is created, whereas for existing users, the grace period starts from when the grace period is enabled.
+
+### Global Grace Period
+
+Global grace period can be enabled for forcing users to enroll in 2FA upon login.
+
+```yaml
+authentication:
+  secondary_authentication_mode: "required"
+  secondary_authentication_rollout:
```
louischan-oursky, line 28: Name this `secondary_authentication_grace_period`

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,77 @@
...
+    global_grace_period_enabled_since: "2021-01-01T00:00:00Z"
+    global_grace_period_days: 30
```
louischan-oursky, line 30:
```
secondary_authentication_grace_period:
  enabled: true
```

This means indefinite grace period.

```
secondary_authentication_grace_period:
  enabled: true
  end_at: "2021-01-01T00:00:00Z"
```

This means grace period is enabled, and it ends at 2021-01-01T00:00:00Z

---

Remove `global_grace_period_days`

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,77 @@
...
+### Per-user Grace Period
+
+Regardless of the global grace period, a per-user grace period can be granted for 10 days by default, or a custom duration can be specified:
+
+```yaml
+authentication:
+  secondary_authentication_mode: "required"
+  secondary_authentication_rollout:
+    per_user_grace_period_days: 10
+```
```
louischan-oursky, line 42: Remove this section. No need to add this now. Conceptually, it is an argument when calling the new mutation to extend the grace period of a user. Of course, we have validation on the input so that it is not in the past, and it is not too far in the future. I think 90 days is enough.

Can you instead specify the proposed changes to the Admin API here? They are public API that we need to review together.

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,77 @@
...
+        - type: authenticate
+          # Requires user to satisfy one of the following authentication.
+          optional: false # or null
+          # If the end-user has no applicable authentication method,
+          # then enrollment will be required before proceeding.
+          # enrollment_allowed by default is false, meaning user with no applicable method beforehand will be blocked from proceeding.
+          enrollment_allowed: true
+          one_of:
+            - authentication: secondary_totp
+            - authentication: recovery_code
```
louischan-oursky, line 71: Put the table I gave previously here to explain `optional` and enrollment_allowed`.

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+#### Notes
+
+- Currently only secondary authenticators are supported for enrollment.
+- `enrollment_allowed` is ignored when `optional` is `true` to match default behaviour.
```
louischan-oursky, line 77: Then you can remove this note. It will be replaced by the table.

IniZio: Removed.

docs/specs/2fa-grace-period.md
IniZio: Updated.

docs/specs/2fa-grace-period.md
IniZio: Updated to graphql schema

docs/specs/2fa-grace-period.md
IniZio: Updated.

docs/specs/2fa-grace-period.md
```diff
@@ -20,29 +20,55 @@ For new users, the grace period starts from the time the user is created, wherea
...
+
+```gql
+type User {
+  id: ID!
+  # ...
+
+  mfaGracePeriodEndAt: DateTime
+}
+
+type GrantMFAGracePeriodInput {
```
louischan-oursky, line 45: Replace the word "grant" with "set", in input, payload, mutation name, etc.

louischan-oursky (commented):
One thing is missing. Add a section to mention we will add a new column to the user table. Include a SQL statement to describe the addition. I mean `ALTER TABLE ... ADD COLUMN` so that we are clear what changes are introduced in this feature.

docs/specs/2fa-grace-period.md
IniZio: Updated.

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,113 @@
...
+
+After the grace period, the user will be required to enroll in 2FA before they can sign in.
+
+## Configuration
+
+By default, no grace period is provided. User without 2FA must contact admin to login after [2FA mode](./user-model.md#secondary-authenticator) is set to `required`.
+
+For new users, the grace period starts from the time the user is created, whereas for existing users, the grace period starts from when the grace period is enabled.
+
+### Global Grace Period
```
louischan-oursky, line 21: Remove this heading. Move the contents to `## Configuration and the global grace period`

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,113 @@
...
+
+```yaml
+authentication:
+  secondary_authentication_mode: "required"
+  secondary_authentication_grace_period:
+    enabled: true
+    end_at: "2021-01-01T00:00:00Z"
+```
+
+### Per-user Grace Period
```
louischan-oursky, line 33:
```
## Per-user Grace period

...

### Changes on the Admin API

...

### Changes on the database schema

...


```

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,113 @@
...
+
+type RemoveMFAGracePeriodPayload {
+  user: User!
+}
+
+type Mutation {
+  setMFAGracePeriod(
+    input: SetMFAGracePeriodInput!
+  ): SetMFAGracePeriodPayload!
+  removeMFAGracePeriod(
```
louischan-oursky, line 66: Add a comment to document its idempotence.

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,113 @@
...
+type RemoveMFAGracePeriodInput {
+  userID: ID!
+}
+
+type RemoveMFAGracePeriodPayload {
+  user: User!
+}
+
+type Mutation {
+  setMFAGracePeriod(
```
louischan-oursky, line 63: Add a comment to document its idempotence.

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,113 @@
...
+    input: SetMFAGracePeriodInput!
+  ): SetMFAGracePeriodPayload!
+  removeMFAGracePeriod(
+    input: RemoveMFAGracePeriodInput!
+  ): RemoveMFAGracePeriodPayload!
+}
```
louischan-oursky, line 69: Follow https://github.com/authgear/authgear-server/pull/4409/files#diff-d2fe537b463363c0561f08dc27a16fd0fcaa38d9462307196bfcd0e32bd33357R93 and document the expected errors in a table.

For this mutation, Invalid MFAGracePeriodInvalid

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,113 @@
+# 2FA Grace Period
+
+- [Abstract](#abstract)
+- [Configuration](#configuration)
+  - [Global Grace Period](#global-grace-period)
+  - [Per-user Grace Period](#per-user-grace-period)
+  - [Customized Authentication Flow](#customized-authentication-flow)
```
louischan-oursky, line 7: Remember to re-generate TOC

docs/specs/2fa-grace-period.md
```diff
@@ -0,0 +1,113 @@
...
+}
+```
+
+It's stored in the user model as `mfa_grace_period_end_at`.
+
+```sql
+ALTER TABLE _authgear_user ADD COLUMN mfa_grace_period_end_at TIMESTAMP WITHOUT TIME ZONE;
+```
+
+### Customized Authentication Flow
```
louischan-oursky, line 78: `## Changes on the configuration of authentication flow`

In authentication-flow.md, specify there is `enrollment_allowed: boolean`. Add a link there to here.

IniZio: Addressed comments above 🙏

docs/specs/2fa-grace-period.md
```diff
@@ -69,15 +69,25 @@ type Mutation {
...
 ```
 
-It's stored in the user model as `mfa_grace_period_end_at`.
+Both `setMFAGracePeriod` and `removeMFAGracePeriod` mutations are indempotent, users can have grace period granted/revoked multiple times even with same value.
+
+#### Error Response
+
+|---|---|---|---|
+|Invalid Grace Period e.g. in the past|`Invalid`|`GracePeriodInvalid`|`endAt` must be in the future.|
```
louischan-oursky, line 78: The reason should be `MFAGracePeriodInvalid`.

The info column specifies the actual `info` object we will include. For this error, we will not include any info.

The description should also mention that the error is be returned if the given input is too far in the future. > 90 days is considered as far.

IniZio: Updated.

louischan-oursky (approved):

PR #4496, opened 2024-07-25
Dep(#4490): Hide default icon when user remove logo

elise-ng:
Ref DEV-1539
Based on #4490

Set css var logo alignment to hidden.

Note there is a commit fixing an outdated state in setState bug in useForm hooks.

https://github.com/user-attachments/assets/36f20814-9805-4156-832e-dfbb16f435e7


portal/src/graphql/portal/DesignScreen/form.ts
```diff
@@ -540,6 +545,11 @@ export function useBrandDesignForm(appID: string): BranchDesignForm {
         ),
         image
       );
+      resourceMutator.updateCustomisableTheme((prev) => {
+        return produce(prev, (draft) => {
+          draft.logo.alignment = image == null ? "hidden" : undefined;
+        });
+      }, targetTheme);
```
cychiuae, line 552: The problem here is because resourceForm.setState is passing the captured state instead of the latest state.

Please try the following fix.

1. https://github.com/authgear/authgear-server/blob/ba91f02fa57f6986cd79f2164fc9d0bb2159bff3/portal/src/hook/useResourceForm.ts#L152
```diff
  const state = currentState ?? initialState;
  const setState = useCallback(
    (fn: (state: State) => State) => {
+      setCurrentState(s => fn(s ?? initialState));
-      setCurrentState(fn(state));
    },
+    [initialState]
-    [state]
  );
```

cychiuae: I think `useAppConfigForm` and `useAppSecretConfigForm` suffer the same problem. If the fix I purposed works, please help also update them

elise-ng:
@louischan-oursky

louischan-oursky:
@elise-ng 

What we want to do about this issue is documented in https://linear.app/authgear/issue/DEV-1539/should-not-show-any-logo-when-there-is-no-logo#comment-ac6b1ef9

We needs to fix Auth UI first.

elise-ng:
@louischan-oursky added commit for updating the template

elise-ng:
if we wanna remove the logo alignment item in portal, let's do it at the next pr as it use the same parent for logo height

louischan-oursky:
@elise-ng I added 2 commits to fix some problems I found.

1. I asked you to use HasAppSpecificAsset. But it turns that HasAppSpecificAsset only supports CSS descriptor. It always returns false for other descriptor. But luckily, we do not need to use it because we can just remove the builtin app logo. So I added a commit to do do.
2. __header.html and brand-logo.css does quite work with DarkThemeEnabled and LightThemeEnabled. For example, the old  "use-light-logo-in-dark-logo" was incorrectly kept. That contradicts with the new Design page. I refactored a bit to make the code read simpler and easier to understand.

louischan-oursky (approved):

PR #4498, opened 2024-07-26
Add 2FA grace period config

IniZio:
ref DEV-1620

Example config;

```yaml
authentication:
  secondary_authentication_grace_period:
      enabled: true
      end_at: "2024-07-25T05:54:32.221Z"
```
---

`make fmt` changed other files as well e.g. `pkg/lib/oauth/store_grant_mock.go`, maybe I  update `make lint` to cover this?

### DIFF
--- a/pkg/lib/config/authentication.go
+++ b/pkg/lib/config/authentication.go
@@ -1,6 +1,8 @@
 package config
 
 import (
+	"time"
+
 	"github.com/authgear/authgear-server/pkg/api/model"
 )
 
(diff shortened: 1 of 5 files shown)


louischan-oursky:
The make target `check-tidy` is supposed to catch that. But it seems that it fails to do so.

louischan-oursky:
`make check-tidy` was fixed in https://github.com/authgear/authgear-server/pull/4502

PR #4500, opened 2024-07-29
Disable redirect url form when no custom domain is set

elise-ng:
Ref DEV-1538

https://github.com/user-attachments/assets/04f33567-3186-457b-a53c-696d0cfff4b8

@fungc-io do we need handling for existing redirect url values set when user remove custom domain?


elise-ng:
@louischan-oursky

louischan-oursky (approved):

PR #4508, opened 2024-07-29
Allow captcha configuration in signup create_authenticator oob_otp

pkong-ds:
## Blocked by

- #4512 
## What's in this PR?

DEV-1608 requires captcha in `signup > create_authenticator`, which is not supported yet. It is added in this PR

## Uncertainties

1. Only supports `create_authenticator_oob_otp` ? -  asked in https://github.com/authgear/authgear-server/pull/4282#discussion_r1695145534
    - yes
2. Only supports `secondary_oob_otp`? Both in authflow/authUI, I cannot observe when will `create_authenticator > primary_oob_otp` take place
    - no, also need `primary_oob_otp` & `verify`, will also implement in this PR


louischan-oursky (changes requested):
mark this as Request changes.

pkong-ds:
Seems this conflicts with #4512 if I change e2e tests in both PRs. Lemme make this depends on #4512

pkong-ds:
@louischan-oursky in case you missed, this is ready for review 🙏

louischan-oursky (approved):

PR #4510, opened 2024-07-30
Specify change password GraphQL API

IniZio:
ref DEV-1642

PR #4511, opened 2024-07-30
Update change password form on portal

IniZio:
ref DEV-1631

### Screens

**Change password form**

- [ ] Confirming if remove confirm password field

Screenshot 2024-07-30 at 11-01-43 Authgear Portal

image

**Change password email**

Screenshot 2024-07-30 at 8 50 20 AM


pkg/admin/facade/user.go
```diff
@@ -103,15 +103,15 @@ func (f *UserFacade) Create(identityDef model.IdentityDef, password string) (str
...
+func (f *UserFacade) ResetPassword(id string, password string, sendPassword bool, changeOnLogin bool) (err error) {
 	err = f.Users.CheckUserAnonymized(id)
 	if err != nil {
 		return err
 	}
 
 	_, err = f.Interaction.Perform(
 		interactionintents.NewIntentResetPassword(),
-		&resetPasswordInput{userID: id, password: password},
+		&resetPasswordInput{userID: id, password: password, sendPassword: sendPassword, changeOnLogin: changeOnLogin},
```
IniZio, line 114: See if we want to move away from interaction for reset password as well?

pkg/admin/graphql/user_mutation.go
```diff
@@ -95,6 +95,14 @@ var resetPasswordInput = graphql.NewInputObject(graphql.InputObjectConfig{
 			Type:        graphql.NewNonNull(graphql.String),
 			Description: "New password.",
 		},
+		"sendPassword": &graphql.InputObjectFieldConfig{
+			Type:        graphql.Boolean,
+			Description: "Indicate whether to send the new password to the user.",
+		},
+		"forceChangeOnLogin": &graphql.InputObjectFieldConfig{
```
louischan-oursky, line 102: According to the spec, it should be `setPasswordExpired`

pkg/lib/authn/authenticator/password/provider.go
```diff
@@ -99,6 +100,7 @@ func (p *Provider) WithPassword(a *authenticator.Password, password string) (*au
 	}
 
 	newAuthen := p.populatePasswordHash(a, password)
+	newAuthen.ExpireAfter = spec.ExpireAfter
```
louischan-oursky, line 103: Do not always write this field. If `setPasswordExpired=true` in the Admin API, then we should set it to now. Otherwise, leave the original value unchanged. That is, if the original value is a date that is in the past, it is kept unchanged.

You may encounter a problem that you cannot distinguish between "I want to keep the expire_at unchanged" and "I want to set expire_at to NULL". Think about it.

pkg/lib/authn/authenticator/password/provider.go
```diff
@@ -159,9 +161,17 @@ func (p *Provider) Authenticate(a *authenticator.Password, password string) (ver
...
 
-	err := p.Store.UpdatePasswordHash(a)
+	prev, err := p.Store.Get(a.UserID, a.ID)
+	if err != nil {
+		return err
+	}
+
+	if prev.ExpireAfter == a.ExpireAfter {
+		a.ExpireAfter = nil
+	}
```
louischan-oursky, line 172: The baseline is do not fetch the row again. Maybe you need to distinguish between the end-user update password (thus expire_after must be set to NULL), and Admin API change password (thus expire_after may be kept unchanged, or set to a value).

pkg/lib/feature/forgotpassword/errors.go
```diff
@@ -13,3 +13,5 @@ var SendCodeFailed = apierrors.Invalid.WithReason("ForgotPasswordFailed")
 
 var ErrFeatureDisabled = SendCodeFailed.NewWithCause("forgot password is disabled", apierrors.StringCause("FeatureDisabled"))
 var ErrUserNotFound = SendCodeFailed.NewWithCause("specified user not found", apierrors.StringCause("UserNotFound"))
+
+var ErrEmailIdentityNotFound = apierrors.Invalid.WithReason("EmailIdentityNotFound").NewWithCause("email identity not found", apierrors.StringCause("EmailIdentityNotFound"))
```
louischan-oursky, line 17: According to the spec, it should https://github.com/authgear/authgear-server/pull/4510/files#diff-3be2ade1b2238264e843494f297f5b50d16774dcdee45eef65b8fe9de515ae4dR55.

And do not add cause for it. The reason is enough.

pkg/lib/feature/forgotpassword/service.go
```diff
@@ -464,6 +480,64 @@ func (s *Service) resetPassword(target string, otpState *otp.State, newPassword
...
+			return err
+		}
+	}
+
+	return nil
+}
+
 // SetPassword ensures the user identified by userID has the specified password.
 // It perform necessary mutation to make this happens.
 func (s *Service) SetPassword(userID string, newPassword string, sendPassword bool, expireAfter *time.Time) (err error) {
```
louischan-oursky, line 543: The argument list is getting long. Introduce a struct SetPasswordOptions.

portal/src/error/error.ts
```diff
@@ -84,6 +84,11 @@ export interface APIGroupDuplicateKeyError {
   reason: "GroupDuplicateKey";
 }
 
+export interface APIEmailIdentityNotFoundError {
+  errorName: string;
+  reason: "EmailIdentityNotFound";
```
louischan-oursky, line 89: Change accordingly

portal/src/error/error.ts
```diff
@@ -84,6 +84,11 @@ export interface APIGroupDuplicateKeyError {
   reason: "GroupDuplicateKey";
 }
 
+export interface APIEmailIdentityNotFoundError {
```
louischan-oursky, line 87: Change accordingly.

pkg/lib/authn/authenticator/password/provider.go
IniZio: Or should I add an `ResetExpireAfter` flag for existing flows to explicitly reset the expireAFter, then Set will just store latest auth info as is?

pkg/lib/authn/authenticator/password/provider.go
```diff
@@ -93,14 +94,16 @@ func (p *Provider) WithSpec(a *authenticator.Password, spec *authenticator.Passw
...
-	// Return original authenticator pointer
-	if pwd.Compare([]byte(password), a.PasswordHash) == nil {
-		return a, nil
+	newAuthen := a
+
+	if pwd.Compare([]byte(password), a.PasswordHash) != nil {
+		newAuthen = p.populatePasswordHash(a, password)
+	}
+
+	if spec.ResetExpireAfter || spec.ExpireAfter != a.ExpireAfter {
```
louischan-oursky, line 103: `spec.ExpireAfter != a.ExpireAfter` is comparing pointers, which is incorrect.

pkg/lib/feature/forgotpassword/service.go
```diff
@@ -538,9 +538,14 @@ func (s *Service) sendPassword(userID string, password string) error {
 	return nil
 }
 
+type SetPasswordOptions struct {
```
louischan-oursky, line 541: I mean you also move UserID and NewPassword to this struct.

pkg/lib/feature/forgotpassword/errors.go
```diff
@@ -13,3 +13,5 @@ var SendCodeFailed = apierrors.Invalid.WithReason("ForgotPasswordFailed")
 
 var ErrFeatureDisabled = SendCodeFailed.NewWithCause("forgot password is disabled", apierrors.StringCause("FeatureDisabled"))
 var ErrUserNotFound = SendCodeFailed.NewWithCause("specified user not found", apierrors.StringCause("UserNotFound"))
+
+var ErrSendPasswordNoTarget = apierrors.Invalid.WithReason("SendPasswordNoTarget").New("")
```
louischan-oursky, line 17: You still have to specify a message. `"no target to send the password"`

pkg/lib/authn/authenticator/service/service.go
```diff
@@ -370,8 +370,7 @@ func (s *Service) WithSpec(ai *authenticator.Info, spec *authenticator.Spec) (bo
 	switch ai.Type {
 	case model.AuthenticatorTypePassword:
 		a := ai.Password
-		plainPassword := spec.Password.PlainPassword
-		newAuth, err := s.Password.WithPassword(a, plainPassword)
+		newAuth, err := s.Password.WithSpec(a, spec.Password)
```
louischan-oursky, line 373: I suggest we introduce a new method for changing password.

```
func (s *Service) UpdatePassword(ai *authenticator.info, options *UpdatePasswordOptions) (bool, *authenticator.Info, error)

type UpdatePasswordOptions struct {
  PlainPassword string
  PasswordHash string
  # If true, then expireAt is set to ExpireAt.
  # To set NULL, SetExpireAt=true and ExpireAt=nil
  # To keep unchanged, SetExpireAt=false
  SetExpireAt bool
  ExpireAt *time.Time
}
```

And then visit every place where WithSpec is used, and change them to call UpdatePassword if necessary.

In the original WithSpec, panic if password authenticator is received.

louischan-oursky (changes requested):
mark this as Request changes.

pkg/lib/authn/authenticator/service/service.go
IniZio: Also rename current `ResetPassword` to `ResetPasswordByEndUser`

IniZio: @louischan-oursky Renamed.

pkg/latte/node_do_update_authenticator.go
```diff
@@ -22,7 +23,14 @@ func (n *NodeDoUpdateAuthenticator) Kind() string {
...
 		workflow.RunEffect(func(ctx context.Context, deps *workflow.Dependencies) error {
-			return deps.Authenticators.Update(n.Authenticator)
+			if n.Authenticator.Password != nil {
+				return deps.Authenticators.UpdatePassword(n.Authenticator, &password.UpdatePasswordOptions{
+					PasswordHash:   n.Authenticator.Password.PasswordHash,
+					SetExpireAfter: true,
+				})
+			} else {
+				return deps.Authenticators.Update(n.Authenticator)
+			}
```
louischan-oursky, line 33: This is incorrect. in pkg/latte/node_change_password.go, there is a call to WithSpec. I said replace WithSpec with UpdatePassword, when the authenticator passed to WithSpec is a password authenticator.

I expect to list out all WithSpec calls, and find out which one should switch to UpdatePassword.

Please read my original commend again. https://github.com/authgear/authgear-server/pull/4511#discussion_r1699516797

pkg/lib/authenticationflow/declarative/node_do_update_authenticator.go
```diff
@@ -25,7 +26,14 @@ func (*NodeDoUpdateAuthenticator) Kind() string {
...
 		authflow.RunEffect(func(ctx context.Context, deps *authflow.Dependencies) error {
-			return deps.Authenticators.Update(n.Authenticator)
+			if n.Authenticator.Password != nil {
+				return deps.Authenticators.UpdatePassword(n.Authenticator, &password.UpdatePasswordOptions{
+					PasswordHash:   n.Authenticator.Password.PasswordHash,
+					SetExpireAfter: true,
+				})
+			} else {
+				return deps.Authenticators.Update(n.Authenticator)
+			}
```
louischan-oursky, line 36: Again, incorrect. This is not I told you to do.

pkg/lib/authn/authenticator/password/provider.go
```diff
@@ -156,12 +158,34 @@ func (p *Provider) Authenticate(a *authenticator.Password, password string) (ver
...
+
+type UpdatePasswordOptions struct {
+	PlainPassword string
+	PasswordHash  []byte
+
+	ExpireAfter    *time.Time
+	SetExpireAfter bool
+}
+
+func (p *Provider) UpdatePassword(a *authenticator.Password, options *UpdatePasswordOptions) error {
```
louischan-oursky, line 176: I think we messed up here.

Provider.UpdatePassword should be simple. It should only do this

```
now := p.Clock.NowUTC()
a.Updated = now
err := p.Store.UpdatePasswordHash(a)
```

- Rename it to `Update`. It means Update really only updates the authenticator as is. No extra logic.

Provider.WithPassword (or Provider.WithSpec, as you have renamed it) is the real method we want to refactor. We want to support the use cases of different  password update.

louischan-oursky (changes requested):

pkg/lib/authn/authenticator/password/provider.go
```diff
@@ -93,14 +94,15 @@ func (p *Provider) WithSpec(a *authenticator.Password, spec *authenticator.Passw
...
-	// If password is not changed, skip the logic.
-	// Return original authenticator pointer
-	if pwd.Compare([]byte(password), a.PasswordHash) == nil {
-		return a, nil
+	newAuthen := a
+	if pwd.Compare([]byte(password), a.PasswordHash) != nil {
+		newAuthen = p.populatePasswordHash(a, password)
+	}
+
+	if !spec.ExpireAfter.Equal(*a.ExpireAfter) {
```
louischan-oursky, line 102: spec.ExpireAfter can be nil

pkg/lib/authn/authenticator/password/provider.go
```diff
@@ -84,7 +85,7 @@ func (p *Provider) New(id string, userID string, passwordSpec *authenticator.Pas
 	}
 }
 
-// WithSpec return new authenticator pointer if password is changed
+// WithSpec return new authenticator pointer if password or expireAfter is changed
 // Otherwise original authenticator will be returned
 func (p *Provider) WithSpec(a *authenticator.Password, spec *authenticator.PasswordSpec) (*authenticator.Password, error) {
```
louischan-oursky, line 90: Also rename this to UpdatePassword and take UpdatePasswordOptions. You may need to duplicate UpdatePasswordOptions in two packages, since you cannot import service.UpdatePasswordOptions in this package.

pkg/lib/authn/authenticator/password/provider.go
louischan-oursky: Once we take UpdatePasswordOptions here, we do not need to compare ExpireAfter. We just look at SetExpireAfter, if it is true, then we set a.ExpireAfter to UpdatePasswordOptions.ExpireAfter. If it is false, we do not set.

pkg/lib/authn/authenticator/service/service.go
```diff
@@ -391,6 +386,37 @@ func (s *Service) WithSpec(ai *authenticator.Info, spec *authenticator.Spec) (bo
...
+}
+
+func (s *Service) UpdatePassword(ai *authenticator.Info, options *UpdatePasswordOptions) (bool, *authenticator.Info, error) {
+	if ai.Type != model.AuthenticatorTypePassword {
+		panic("authenticator: update password is not supported for type " + ai.Type)
+	}
+
+	changed := false
+	a := ai.Password
+	expireAfter := a.ExpireAfter
```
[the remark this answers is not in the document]
IniZio: Updated to use `SetExpiredAfter` options.

louischan-oursky (approved):

PR #4513, opened 2024-07-30
Specify SAML

tung2744:
ref DEV-1682


tung2744:
@louischan-oursky @chpapa The SAML spec is now ready for review.

@chpapa If you have no time to review the whole spec, please read https://github.com/tung2744/authgear-server/blob/design-saml/docs/specs/saml.md#10 for the whole config design. Please let me know if you want a call.

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+
+- `name_id_format`: Optional. The `NameIDFormat` used in SAML assertions. Default `urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified`.
+- `name_id_attribute_pointer`: Optional. See the above [Configure NameID when using urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified](#3_1_1) section for details. Default `/sub`. Only effective if `name_id_format` is `urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified`.
+
+### <a id="3_3"></a> Unsupported `NameIDFormat`
+
+The following `NameIDFormat`s are not supported at the moment, but they are commonly supported by different SAML IdPs so we would like to support it in the future:
+
+- urn:oasis:names:tc:SAML:2.0:nameid-format:persistent
+- urn:oasis:names:tc:SAML:2.0:nameid-format:transient
```
louischan-oursky, line 304: Document the support of them is backward compatible. In particular, how are we going to support `persistent` with `name_id_attribute_pointer` in a backward compatible way.

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+
+```yaml
+- data:
+    certs:
+      - cert:
+          pem: "-----BEGIN CERTIFICATE-----\nMIIH/TCCBeWgAwIBAgIQaBYE3/..."
+        key:
+          alg: RS256
+          created_at: 1671439710
+          kid: key01
```
louischan-oursky, line 547: This time, we can use the JWK thumbprint as the kid. (This is an implementation details by the way)

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+          use: sig
+  key: saml.idp.signing
+```
+
+- The above secret is for storing the signing key pairs used by Authgear to sign the SAML requests, responses and assertions. The keys are stored in jwk format.
+  - `key`: Must be `saml.idp.signing`.
+  - `data`: Required. An object containing the following fields:
+    - `certs`: Required. A list of objects which represents the signing keypairs and the X.509 certs used to sign the IdP requests, responses and assertions. The minimum number of items in the list is 1. The maximum number of items in the list is 0. The item inside the list MUST contains the following fields:
+      - `cert`: Requried. An object represents a X.509 cert. It MUST contains the following fields:
+        - `pem`: Required. The X.509 cert in pem format. See [Metadata](#7) for details.
```
louischan-oursky, line 566: What fields are required in the X.509 certificate? I wonder if we can just store the private key JWK, and generate the X.509 certificate on the fly?

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+    ua->>+ag: [3] Send <AuthnRequest> to https://example.authgear.cloud/saml2/login
+    Note right of ag: Alice enters user credentials
+    ag->>-ua: [4] Redirect to https://example.com/acs with <Response>
+    ua->>+sp: [5] Send <Response> to https://example.com/acs
+    Note right of sp: Authenticated as Alice
+    sp->>-ua: [6] Response with protected resources https://example.com/flights
+
+```
+
+- The login endpoint `https://example.authgear.cloud/saml2/login/ENTITY_ID` is from the metadata.
```
louischan-oursky, line 59: It seems that we do not have a dedicated section to specify the metadata endpoint.



Another question: How does other authentication software name these endpoints I mean `https://example.authgear.cloud/saml2/login/ENTITY_ID` vs `https://example.authgear.cloud/saml2/service_providers/ENTITY_ID/login`

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+- `<saml:Subject>`
+
+  - If provided, the subject will be passed to Authentication Flow for pre-selecting a user. See [Support Pre-Selecting User by SAML Subject in Authentication Flow](#1_5) for details.
+  - Ref: [SAMLCore](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf) Section 3.3.4.
+
+- `<NameIDPolicy>`
+
+  - `<Format>`: If provided, `NameID` will be returned in the specified format. See [NameID](#3) section for supported formats.
+  - `<AllowCreate>`: Not supported at the moment. The behavior is always `false`, which does not allow creating a new NameID when Subject exist.
+  - `SPNameQualifier`: Not supported. Will be ignored if provided.
```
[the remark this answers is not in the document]
tung2744: Basically the spec assumes all parameter supported, so no.
But it says
> The NameQualifier and SPNameQualifier
attributes SHOULD be omitted unless the element or format explicitly defines their use and semantics.

So maybe I should change the spec to say it should be omitted.

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+
+- `<saml:Subject>`
+
+  - If provided, the subject will be passed to Authentication Flow for pre-selecting a user. See [Support Pre-Selecting User by SAML Subject in Authentication Flow](#1_5) for details.
+  - Ref: [SAMLCore](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf) Section 3.3.4.
+
+- `<NameIDPolicy>`
+
+  - `<Format>`: If provided, `NameID` will be returned in the specified format. See [NameID](#3) section for supported formats.
+  - `<AllowCreate>`: Not supported at the moment. The behavior is always `false`, which does not allow creating a new NameID when Subject exist.
```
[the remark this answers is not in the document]
tung2744: Not exactly, I think it depends on what exactly is the `NameID` (see NameIDFormat). Some NameID could be created without creating a new user (Such as persistent NameID?)
But for now yes it only controls signup.

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+```
+
+- `saml.service_providers`: Required. A list of objects containing the configurations of a Service Provider. The objects in the list contains the following fields:
+  - `id`: Required. A string which is the entity ID of the service provider.
+  - `audience`: Optional in SP Initiated flow. Required when using the IdP Initiated flow. `Audience` of the `<Response>` generated at the end of Browser SSO flow. In SP Initiated flow, if this config is set, the value configured will be used instead of `Issuer` of the `<AuthnRequest>`.
+  - `destination`: Optional in SP Initiated flow. Required when using the IdP Initiated flow. `Destination` of the `<Response>` generated at the end of Browser SSO flow. In SP Initiated flow, if this config is set, the value configured will be used instead of `AssertionConsumerServiceURL` of the `<AuthnRequest>`.
+  - `acs_url`: Optional in SP Initiated flow. Required when using the IdP Initiated flow. The ACS URL to send the `<Response>` to. In SP Initiated flow, if this config is set, and the value of `AssertionConsumerServiceURL` in `<AuthnRequest>` does not match the configured value, an erro will be returned.
```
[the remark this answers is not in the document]
tung2744: I think no. The SP should only have one acs url.

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+    - id: ENTITY_ID
+      audience: https://example.com
+      destination: https://example.com/acs
+      acs_url: https://example.com/acs
+```
+
+- `saml.service_providers`: Required. A list of objects containing the configurations of a Service Provider. The objects in the list contains the following fields:
+  - `id`: Required. A string which is the entity ID of the service provider.
+  - `audience`: Optional in SP Initiated flow. Required when using the IdP Initiated flow. `Audience` of the `<Response>` generated at the end of Browser SSO flow. In SP Initiated flow, if this config is set, the value configured will be used instead of `Issuer` of the `<AuthnRequest>`.
+  - `destination`: Optional in SP Initiated flow. Required when using the IdP Initiated flow. `Destination` of the `<Response>` generated at the end of Browser SSO flow. In SP Initiated flow, if this config is set, the value configured will be used instead of `AssertionConsumerServiceURL` of the `<AuthnRequest>`.
```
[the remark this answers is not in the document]
tung2744: I made a mistake here. `AssertionConsumerServiceURL` must always be configured, no matter in SP or IdP initiated flow. And, we will always validate the `AssertionConsumerServiceURL` in request matches the config. So there is no problem of overriding (Because values are always same)

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+  service_providers:
+    - id: ENTITY_ID
+      audience: https://example.com
+      destination: https://example.com/acs
+      acs_url: https://example.com/acs
+```
+
+- `saml.service_providers`: Required. A list of objects containing the configurations of a Service Provider. The objects in the list contains the following fields:
+  - `id`: Required. A string which is the entity ID of the service provider.
+  - `audience`: Optional in SP Initiated flow. Required when using the IdP Initiated flow. `Audience` of the `<Response>` generated at the end of Browser SSO flow. In SP Initiated flow, if this config is set, the value configured will be used instead of `Issuer` of the `<AuthnRequest>`.
```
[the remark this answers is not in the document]
tung2744: Although we can set `Issuer` as `Audience` by default, it is not a must. SP should be able to customize the expected `Audience` value in the config.

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+```
+
+- The request MUST includes a valid Subject, with a valid NameID.
+- The NameID will be mapped to an existing user. See [NameID](#3) section for related configs.
+
+### <a id="1_6"></a> The Login Endpoint
+
+- The URL is https://example.authgear.cloud/saml2/login/ENTITY_ID. Where `ENTITY_ID` is the `id` of the service provider as specified in the config `saml.service_providers[index].id`. As a result, each service provider will have a independent login endpoint.
+- The metadata for each service provider will includes the above login endpoint.
+- When the endpoint receives a GET or POST request with a `<AuthnRequest>` provided, the [SP Initiated SSO flow](#1_1) will be triggered.
```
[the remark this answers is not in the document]
tung2744: It is defined in the binding spec.
https://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf

Read the HTTP Redirect Binding section which uses the GET method.

docs/specs/saml.md
```diff
@@ -0,0 +1,697 @@
...
+- When the endpoint receives a GET request without a `<AuthnRequest>` provided, the [IdP Initiated SSO flow](#1_2) will be triggered.
+- When the endpoint receives a POST request without a `<AuthnRequest>` provided, it is an error.
+
+#### Selecting Service Provider
+
+- This is not supported at the moment.
```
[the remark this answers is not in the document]
tung2744: Actually, I haven't think about how to support "Selecting Service Provider" yet. For now, to trigger a login, you must make a call to a SSO endpoint, which includes the entity_id in the url. Therefore, we always know which service provider you want to login so we don't have the problem on how to handle this.

docs/specs/saml.md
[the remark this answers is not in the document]
louischan-oursky: So it now means allow signup, and the default is false, which means we do not allow signup by default? Seems contradict with OIDC.

docs/specs/saml.md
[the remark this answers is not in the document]
louischan-oursky: > Basically the spec assumes all parameter supported, so no.

Then I think we should change all `Will be ignored if provided` to `Result in an error`. And document the error in this spec.

docs/specs/saml.md
louischan-oursky: Another question: Maybe we should rename the term `Entity ID` to `Service Provider ID` in this spec. Or the SAML spec uses the term `Entity ID` so we want to be consistent with the SAML spec?

docs/specs/saml.md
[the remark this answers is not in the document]
tung2744: If `<Subject>` is provided in the request (It is optional), then the SP is telling the IdP which user he want to signin, in this case signup is not allowed.
If `<Subject>` is not provided, then signup should still be allowed.
And yes, it is a bit different from the current oidc flow. We might need to support disable signup mode in authui, when the request set `AllowCreate` to false.

docs/specs/saml.md
tung2744: I think `name_id_attribute_pointer` is only for `unspecified` type of NameID. Other NameIDFormat does not read the config.

docs/specs/saml.md
tung2744: I think it need to at least have fields required by a X.509 cert, such as "Not Before", "Not After"...
> I wonder if we can just store the private key JWK, and generate the X.509 certificate on the fly?

If we always generate a new cert, the timestamps mentioned above will always be changing.

docs/specs/saml.md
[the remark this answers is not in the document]
louischan-oursky: leave a quote in the spec to explain this difference with OIDC. Future readers may be interested in this.

docs/specs/saml.md
tung2744: It is here https://github.com/tung2744/authgear-server/blob/design-saml/docs/specs/saml.md#7

Maybe I can document the endpoint too.

 > How does other authentication software name these endpoints

For auth0, it is like https://dev-y17ryy2xy2nvgpk1.us.auth0.com/samlp/lhImeA5MPX4nthJGxb5MCDW5AeK6QEYL

> Maybe we should rename the term Entity ID to Service Provider ID in this spec. Or the SAML spec uses the term Entity ID so we want to be consistent with the SAML spec?


Yes. Entity ID is a term used in the spec. And it will also appear in the metadata. Every "entity", such as the IdP and Service Provider, will have its own entity id.
However, the spec does not use the entity id in any url.

docs/specs/saml.md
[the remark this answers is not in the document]
tung2744: However, we don't know how other service providers construct the request. If we always return error if we see an unsupported field, maybe a lot of service providers cannot be used. Therefore I think ignoring the unsupported fields is more appropriate.

docs/specs/saml.md
louischan-oursky: > I think it need to at least have fields required by a X.509 cert, such as "Not Before", "Not After"...

We can use the JWT claim `exp` and `iat`, `nbf` in the JWK for this. https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4 Then we should be able to generate a X.509 certificate that is stable and not changing.

louischan-oursky: I realized we have been using the claim `created_at` to store the creation time of a JWK.

tung2744: However, it seems jwk are not allowed to have such fields?
https://datatracker.ietf.org/doc/html/rfc7517#section-4
https://datatracker.ietf.org/doc/html/rfc7518#section-6.3.2

tung2744: And actually, storing the exact cert also has a benefit that this avoid us to change the the cert accidentally (Such as when we supported a new algo in future, or changing the subject name of the cert for some reason, we don't want to affect the old certs).

docs/specs/saml.md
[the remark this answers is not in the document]
tung2744: After some thought, it seems we need to support `true` as well:
Say a  SP can first get a email input from user, and send it to the IdP as the subject, and set `AllowCreate=true` which means "please help me create a user with this email".
Let me modify the spec to say this should be supported. (So authflow signup should support using SAML request to signup as well)

tung2744:
Added one commit to restrict the options of name_id_attribute_pointer.
Now it can only be 
- `/sub`
- `/username`
- `/email`
- `/phone_number`

tung2744:
TODO:
- [ ] Add configuration to control the valid time of SAML Assertions

docs/specs/saml.md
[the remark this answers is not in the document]
tung2744: Although usually SPs only have one url, but it is possible to have multiple urls according to the spec. So will change to a list and ensure there is at least one item.

tung2744:
The entity id in urls should be base64url encoded, because it is a uri.

tung2744:
It seems that some SPs have their own defined entity id, so it is better to let them change the entity id in the portal.
And as entity id is modifiable, we want a internal id to reference a service provider in our config.

So I will change the config to:
```yaml
saml:
  service_providers:
    - id: 537b82d32da780b5 # generated id for internal reference
      entity_id: urn:some-external-sp.example.com
```

The secret `saml.service_providers.signing` will also be modified.

```yaml
- data:
    - service_provider_id: 537b82d32da780b5
      certs:
        - pem: "-----BEGIN CERTIFICATE-----\nMIIH/TCCBeWgAwIBAgIQaBYE3/..."
        - pem: "-----BEGIN CERTIFICATE-----\nMIIH/TCCBeWgAwIBAgIQaBYE3/..."
  key: saml.service_providers.signing
```

And the endpoints which requires a entity id in the url, will be changed to use the internal id.

`/saml2/metadata/:service_provider_id`
`/saml2/login/:service_provider_id`
`/saml2/logout/:service_provider_id`

louischan-oursky:
How does other implementation handle `/saml2/metadata/:service_provider_id`? Do they use Entity ID?

tung2744:
@louischan-oursky For auth0, they uses the application id in the url (In oidc, it is the client id. In saml, I think it is something same as the internal id here).
They have no config for the entity id. I guess they just do not use the `Issuer` to resolve to a application, but use the internal id in the path instead.

louischan-oursky:
Test how other implementation handle Issuer or Entity ID

- Test with omit (Error)
- Test with "" (Error)
- Test with "foobar" (Error)
- Test with "urn:foobar" (Maybe OK)
- Test with IdP Entity ID (Maybe OK)

tung2744:
Tested and all of the above works.. I think we can conclude that issuer is not checked.

tung2744:
- [ ] Rename `cert` to `certificate`, `certs` to `certificates`
- [ ] Remove support of SHA1 digest

docs/specs/saml.md
```diff
@@ -0,0 +1,701 @@
...
+saml:
+  service_providers:
+    - id: ENTITY_ID
+      name_id_format: urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified
+      name_id_attribute_pointer: /username
+```
+
+- `name_id_attribute_pointer`: The JSON pointer pointing to a user's standard attribute or custom attribute which will be used as the NameID. The login will fail if the attribute does not exist or it is empty. The default is `/sub`. Only the following values are accepted:
+  - `/sub`: The user id. This is the default.
+  - `/username`: The username.
```
tung2744, line 286: - [ ] This should be `/preferred_username`

tung2744:
- [ ] Rename all `name_id` to `nameid`.

docs/specs/saml.md
```diff
@@ -0,0 +1,701 @@
...
+```yaml
+saml:
+  debug_mode_enabled: true
+  signing:
+    key_id: key01
+    signature_method: RSAwithSHA256
+    digest_method: SHA256
+  service_providers:
+    - id: ENTITY_ID
+      audience: https://example.com
```
tung2744, line 657: - [ ] `recipient` should also be configurable, as it might be different from acs_url if the sp is behind a proxy gateway.

docs/specs/saml.md
```diff
@@ -0,0 +1,701 @@
...
+  - To facilitate cert rotations, two certs are allowed to be uploaded. When there are two certs, the requests and responses will be considered valid if its signature can be validated by any one of the uploaded certs.
+
+### <a id="6_1"></a> Configs
+
+```yaml
+saml:
+  signing:
+    key_id: key01
+    signature_method: RSAwithSHA256
+    digest_method: SHA256
```
tung2744, line 531: The library we are using ("github.com/russellhaering/goxmldsig") does not support configuring digest method and signature method separately. Maybe we just remove `digest_method` and use `signature_method` to control both.

- [ ] Remove `digest_method`

louischan-oursky: Maybe replace `signature_method` and `digest_method` with `signature_algorithm`. Signature algorithm should be a more common word for this thing. Reference https://datatracker.ietf.org/doc/html/rfc7517#section-4.4

For the values, does the SAMLCore spec define any signature algorithm constants? If not, maybe we can use the values from JWA. https://www.rfc-editor.org/rfc/rfc7518.html#section-3.1 I think we support `RS256` only.

tung2744: We can rename the config key to signature_algorithm.
The value `RSAwithSHA256` is from https://www.w3.org/TR/xmldsig-core/ section 6.

louischan-oursky: It seems that the identifier is `http://www.w3.org/2001/04/xmldsig-more#rsa-sha256`, rather than `RSAwithSHA256`

louischan-oursky: https://datatracker.ietf.org/doc/html/rfc6931

tung2744: `RSAwithSHA256` is the name mentioned in section 6.1. The URI is indeed `http://www.w3.org/2001/04/xmldsig-more#rsa-sha256`. We can change to use the URI instead.

docs/specs/saml.md
```diff
@@ -0,0 +1,701 @@
...
+    ag->>-ua: [4] Redirect to https://example.com/acs with <Response>
+    ua->>+sp: [5] Send <Response> to https://example.com/acs
+    Note right of sp: Authenticated as Alice
+    sp->>-ua: [6] Response with protected resources https://example.com/flights
+
+```
+
+- The login endpoint `https://example.authgear.cloud/saml2/login/ENTITY_ID` is from the metadata.
+- The Assertion Consumer Service (ACS) URL `https://example.com/acs` is from `<AuthnRequest>`.
+- `Audience` of the SAML assertions will be set to the `Issuer` of the `<AuthnRequest>`.
```
tung2744, line 61: - [  ] Audience` of the SAML assertions will be set to the `AssertionConsumerURL` of the `<AuthnRequest>`, if given.

docs/specs/saml.md
```diff
@@ -0,0 +1,701 @@
...
+- `Destination` of the `<Response>` will be set to the `AssertionConsumerServiceURL` of the `<AuthnRequest>`.
+- In step [4], Authgear creates an IdP Session with the User Agent using cookies. `SessionIndex` in the `<Response>` of step [5] is the ID of the IdP Session.
+- In step [5], the SP should memorize the authentication state. For example, by creating a session using cookies.
+
```
tung2744, line 65: - [ ] `Recipient` of the SAML assertions will be set to the `AssertionConsumerServiceURL` of the `<AuthnRequest>`, if given.

docs/specs/saml.md
```diff
@@ -0,0 +1,701 @@
...
+Here is an example of configurations about Web Browser SSO:
+
+```yaml
+saml:
+  service_providers:
+    - id: ENTITY_ID
+      audience: https://example.com
+      destination: https://example.com/acs
+      acs_url: https://example.com/acs
```
tung2744, line 103: - [ ] Add `assertion_valid_duration` to control the assertion valid duration. Default `"20m"`.

docs/specs/saml.md
tung2744: https://github.com/authgear/authgear-server/pull/4630#discussion_r1724934439

The logic will follow this comment.

docs/specs/saml.md
```diff
@@ -0,0 +1,705 @@
...
+## <a id="10"></a> An example with all configurable options
+
+```yaml
+saml:
+  debug_mode_enabled: true
+  signing:
+    key_id: key01
+    signature_method: http://www.w3.org/2001/04/xmldsig-more#rsa-sha256
+  service_providers:
+    - id: EXAMPLE_ID
```
tung2744, line 656: - [ ] Remove `id`
- [ ] Add `client_id`

docs/specs/saml.md
```diff
@@ -0,0 +1,705 @@
...
+  - Optional. If not provided, default to `urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Post`. See [Bindings](#5) section for supported bindings.
+
+- `<AttributeConsumingServiceIndex>`
+
+  - Not supported. Will be ignored if provided.
+
+- `<ProviderName>`
+  - Not supported. Will be ignored if provided.
+
+### <a id="1_5"></a> Support Pre-Selecting User by SAML Subject in Authentication Flow
```
tung2744, line 175: - [ ] Document new supported login_hint
- [ ] Document how subject is mapped to login_hint
- [ ] Document the behavior if IsPassive=true and Subject is given

tung2744:
https://linear.app/authgear/issue/DEV-2061/[samlslo]-propagate-logout-to-other-participants

- [ ] Only saml2/logout and /logout endpoints support SLO

louischan-oursky:
@tung2744 Is this ready for merge?

tung2744:
I think yes at the current stage. We can update the spec later when we implement remaining saml features.

louischan-oursky (approved):

PR #4520, opened 2024-07-31
Implement ldap configuration

yyorkchan:


pkg/util/validation/formats.go
```diff
@@ -209,6 +216,137 @@ func (FormatHTTPOriginSpec) CheckFormat(value interface{}) error {
...
+		}
+	}
+	if err != nil {
+		return err
+	}
+	// check if the filter is correct
+	result := buf.String()
+	result = strings.TrimSpace(result)
+
+	_, err = ldap.CompileFilter(result)
```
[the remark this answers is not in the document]
cychiuae: The message is not readable and it expose too much implementation detail

pkg/lib/config/testdata/parse_secret_tests.yaml
```diff
@@ -188,4 +188,24 @@ config:
...
+---
+name: ldap/missing-required-secret-key
+error: |-
+  invalid secrets:
+  /secrets/0/data: required
+    map[actual:<nil> expected:[name] missing:[name]]
+config:
+  secrets:
+    - key: ldap
+      data: {}
```
cychiuae: Try to be specific in the commit message.

> Update secret test yaml 

What are you updating? Everyone knows you are updating secret test yaml but for what.

pkg/lib/config/config.go
```diff
@@ -89,6 +89,8 @@ type AppConfig struct {
 	Captcha       *CaptchaConfig       `json:"captcha,omitempty"`
 	BotProtection *BotProtectionConfig `json:"bot_protection,omitempty"`
 
+	LDAP *LDAPConfig `json:"ldap,omitempty"`
```
cychiuae, line 92: I dont think this is a fix.

pkg/lib/config/secret.go
```diff
@@ -215,6 +215,28 @@ func (c *SecretConfig) validateBotProtectionSecrets(ctx *validation.Context, bot
...
 }
 
+func (c *SecretConfig) validateLDAPServerUserSecrets(ctx *validation.Context, ldapServerConfig []LDAPServerConfig) {
+	c.validateRequire(ctx, LDAPServerUserCredentialsKey, "LDAP server user credentials")
+	_, data, _ := c.LookupDataWithIndex(LDAPServerUserCredentialsKey)
+	ldapServerUserCredentials, ok := data.(*LDAPServerUserCredentials)
+	if ok {
+		for _, server := range ldapServerConfig {
+			matched := false
+			for _, user := range ldapServerUserCredentials.Items {
```
cychiuae, line 225: It is not a `user`. Please rename it.

pkg/lib/config/secret_ldap.go
```diff
@@ -0,0 +1,50 @@
...
+type LDAPServerUserCredentials struct {
+	Items []LDAPServerUserCredentialsItem `json:"items,omitempty"`
+}
+
+var _ = SecretConfigSchema.Add("LDAPServerUserCredentialsItem", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"name": { "type": "string" },
```
cychiuae, line 22: Do we need to add `minLength: 1` checking to name?

pkg/lib/config/testdata/parse_secret_tests.yaml
```diff
@@ -188,4 +188,66 @@ config:
...
+  invalid secrets:
+  /secrets/0/data/items/0: dependentRequired
+    map[Required:map[dn:map[actual:[dn name] expected:[password] missing:[password]]]]
+config:
+  secrets:
+    - key: ldap
+      data:
+        items:
+          - name: test
+            dn: dc=test
```
cychiuae, line 253: End this file will a newline

pkg/lib/config/testdata/secret_config_validate_tests.yaml
```diff
@@ -374,4 +374,99 @@ secret_config:
...
+          search_filter_template: (mail={{.Username}})
+          user_id_attribute_oid: "0.9.2342.19200300.100.1.1"
+secret_config:
+  secrets:
+    - key: ldap
+      data:
+        items:
+        - name: default
+          dn: "cn=admin,dc=localhost"
+          password: "password"
```
cychiuae, line 472: End this file will a newline

pkg/lib/config/ldap.go
```diff
@@ -0,0 +1,97 @@
...
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"servers": {
+			"type": "array",
+			"items": {
+				"$ref": "#/$defs/LDAPServerConfig"
+			},
+			"minItems": 1
```
louischan-oursky, line 13: Remove this. If you reference OAuth providers, there is no `minItems` there.

pkg/lib/config/ldap.go
```diff
@@ -0,0 +1,97 @@
...
+
+type LDAPServerConfig struct {
+	Name                             string   `json:"name,omitempty"`
+	URL                              *url.URL `json:"url,omitempty"`
+	BaseDN                           string   `json:"base_dn,omitempty"`
+	SearchFilterTemplate             string   `json:"search_filter_template,omitempty"`
+	UserUniqueIdentifierAttributeOID string   `json:"user_id_attribute_oid,omitempty"`
+}
+
+var _ = SecretConfigSchema.Add("LDAPServerUserCredentials", `
```
louischan-oursky: Move secret related things in `secret_ldap.go`. Take `bot_protection.go` and `secret_bot_protection.go` as references.

pkg/lib/config/ldap.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+var _ = SecretConfigSchema.Add("LDAPServerUserCredentialsItem", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"name": { "type": "string" },
+		"dn": { "type": "string", "format": "ldap_dn" },
+		"password": { "type": "string" }
```
louischan-oursky: `dn` and `password` should have `minLength: 1`

pkg/lib/config/ldap.go
```diff
@@ -0,0 +1,97 @@
...
+	SearchFilterTemplate             string   `json:"search_filter_template,omitempty"`
+	UserUniqueIdentifierAttributeOID string   `json:"user_id_attribute_oid,omitempty"`
+}
+
+var _ = SecretConfigSchema.Add("LDAPServerUserCredentials", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"items": { "type": "array", "items": { "$ref": "#/$defs/LDAPServerUserCredentialsItem" } }
```
louischan-oursky: Add programmatic validation to ensure if a server is defined, then it has a corresponding item in the secrets. You can also reference bot protection on how to do this.

pkg/util/validation/formats.go
```diff
@@ -209,6 +216,135 @@ func (FormatHTTPOriginSpec) CheckFormat(value interface{}) error {
...
+	if err != nil {
+		return err
+	}
+	var buf bytes.Buffer
+	// make a string array of phone, username, email
+	testcases := []string{"+85298765432", "username", "test@test.com"}
+	for _, testcase := range testcases {
+		err = tmpl.Execute(&buf, map[string]string{"Username": testcase})
+		if err == nil && buf.String() != "" {
+			break
```
louischan-oursky, line 301: The search filter template MUST pass all three test cases in order to be considered as valid. So the loop should look like

```golang
testcases := []string{"+85298765432", "username", "test@test.com"}
for _, testcase := range testcases {
  var buf bytes.Buffer // Note that buf is moved into the loop. You do not need it outside the loop.
  err := tmpl.Execute(&buf, map[string]string{"Username": testcase})
  if err != nil {
    return err
  }
  filterString := buf.String()
  _, err := ldap.CompileFilter(filterString)
  if err != nil {
    return err
  }
}

// If no early return, then the filter template is considered as valid.
return nil
```

pkg/util/validation/formats.go
```diff
@@ -209,6 +216,135 @@ func (FormatHTTPOriginSpec) CheckFormat(value interface{}) error {
...
+
+type FormatLDAPOID struct{}
+
+func (FormatLDAPOID) CheckFormat(value interface{}) error {
+	str, ok := value.(string)
+	if !ok {
+		return nil
+	}
+
+	// numericoid = number 1*( DOT number )
```
louischan-oursky, line 327: Include the relevant production rules as comment here.

```go
// numericoid = number 1*( DOT number )
// number  = DIGIT / ( LDIGIT 1*DIGIT )
// DIGIT   = %x30 / LDIGIT       ; "0"-"9"
// LDIGIT  = %x31-39             ; "1"-"9"
// https://datatracker.ietf.org/doc/html/rfc4512#section-1.4
```

Then you will see what you implement here is not strictly correct. You did not handle number with leading zeros.

Please add test cases for leading zeros.

pkg/lib/config/testdata/parse_secret_tests.yaml
```diff
@@ -188,4 +188,24 @@ config:
...
+        type: hcaptcha # not supported
+---
+name: ldap/valid
+error: null
+config:
+  secrets:
+    - key: ldap
+      data:
+        name: test
+        username: user
```
louischan-oursky: I think the field is called `dn` not `username`, how come it is valid?

Also, please add more test cases, like minLength, dependentRequired. The baseline is, you test every schema you write.

pkg/util/validation/formats.go
```diff
@@ -287,7 +285,9 @@ func (FormatLDAPSearchFilterTemplate) CheckFormat(value interface{}) error {
 		return nil
 	}
 
-	tmpl, err := template.New("search_filter").Parse(str)
+	str = strings.TrimSpace(str)
+
+	tmpl, err := template.New("search_filter").Parse(strings.Trim(str, "\n"))
```
louischan-oursky: I think you do not need to trim space here, you just need to trim space on the rendered result.

pkg/lib/config/secret.go
```diff
@@ -215,6 +215,28 @@ func (c *SecretConfig) validateBotProtectionSecrets(ctx *validation.Context, bot
...
+			}
+			if !matched {
+				ctx.EmitErrorMessage(fmt.Sprintf("LDAP server user credentials for '%s' is required", server.Name))
+			} else {
+				// keys are validated by the jsonschema
+			}
+		}
+	}
+}
+
```
louischan-oursky, line 239: Add test cases for this.

louischan-oursky (changes requested):

pkg/util/validation/formats.go
```diff
@@ -325,11 +318,15 @@ func (FormatLDAPOID) CheckFormat(value interface{}) error {
...
+	// number  = DIGIT / ( LDIGIT 1*DIGIT )
+	// DIGIT   = %x30 / LDIGIT       ; "0"-"9"
+	// LDIGIT  = %x31-39             ; "1"-"9"
+	// https://datatracker.ietf.org/doc/html/rfc4512#section-1.4
 	if len(str) == 0 {
 		return errors.New("expect non-empty OID")
 	}
 
-	matched, err := regexp.MatchString(`^\d+(\.\d+)*$`, str)
+	matched, err := regexp.MatchString(`^[1-9](\.\d+)*$`, str)
```
louischan-oursky, line 329: This is still incorrect. For example, `1.01` is valid against this regex, but invalid as an oid.

pkg/util/validation/formats_test.go
```diff
@@ -189,7 +192,9 @@ func TestFormatLDAPOID(t *testing.T) {
 		So(f(""), ShouldBeError, "expect non-empty OID")
 		So(f(".1.1"), ShouldBeError, "invalid OID")
 		So(f("1.1."), ShouldBeError, "invalid OID")
+		So(f("0.1.3"), ShouldBeError, "invalid OID")
```
louischan-oursky, line 195: `0.1.3` is actually a valid oid. This testcase is incorrect.

pkg/lib/config/testdata/secret_config_validate_tests.yaml
```diff
@@ -398,8 +398,10 @@ app_config:
           search_filter_template: |
             {{if eq .Username "test@test.com"}}
                 (mail={{.Username}})
+            {{else}}
+                (uid={{.Username}})
             {{end}}
-          user_id_attribute_oid: "0.9.2342.19200300.100.1.1"
+          user_id_attribute_oid: "1.9.2342.19200300.100.1.1"
```
louischan-oursky, line 404: Revert this change

pkg/lib/config/testdata/secret_config_validate_tests.yaml
```diff
@@ -431,8 +433,10 @@ app_config:
...
```
louischan-oursky, line 439: Revert this line

pkg/lib/config/testdata/secret_config_validate_tests.yaml
```diff
@@ -461,7 +465,7 @@ app_config:
           url: "ldap://localhost:389"
           base_dn: "dc=localhost"
           search_filter_template: (mail={{.Username}})
-          user_id_attribute_oid: "0.9.2342.19200300.100.1.1"
+          user_id_attribute_oid: "1.9.2342.19200300.100.1.1"
```
louischan-oursky, line 468: Revert this line

louischan-oursky (changes requested):

yyorkchan:
Thanks @louischan-oursky

pkg/lib/config/testdata/parse_secret_tests.yaml
```diff
@@ -188,4 +188,96 @@ config:
...
+    - key: ldap
+      data:
+        items:
+          - dn: dc=test
+            password: secret
+---
+name: ldap/missing-dn-when-password-present
+error: |-
+  invalid secrets:
+  /secrets/0/data/items/0: dependentRequired
```
louischan-oursky, line 232: When we use dependentRequired, the error shape changed a bit. This is unexpected. Let's use allOf and if then to implement the condition.


```
{
  "allOf": [
    {
      "if": {
         "required": ["dn"]
       },
       "then": {
          "required": ["password"]
       },
    },
    {
      "if": {
         "required": ["password"]
       },
       "then": {
          "required": ["dn"]
       },
    },
  ]
}

```

pkg/lib/config/testdata/parse_secret_tests.yaml
```diff
@@ -188,4 +188,96 @@ config:
...
+    map[Required:map[password:map[actual:[name password] expected:[dn] missing:[dn]]]]
```
louischan-oursky, line 233: Note that the map has `Required`, which is not compatible with other error.

pkg/util/validation/formats_test.go
```diff
@@ -119,6 +119,87 @@ func TestFormatHTTPOriginSpec(t *testing.T) {
...
+		So(f("1.1"), ShouldBeNil)
+		So(f("1.1.1"), ShouldBeNil)
+		So(f("0.1.3"), ShouldBeNil)
+		So(f("10.13.10"), ShouldBeNil)
+		So(f("0.9.2342.19200300.100.1.1"), ShouldBeNil)
+		So(f(""), ShouldBeError, "expect non-empty OID")
+		So(f("1"), ShouldBeError, "invalid OID")
+		So(f(".1.1"), ShouldBeError, "invalid OID")
+		So(f("1.1."), ShouldBeError, "invalid OID")
+		So(f("1.1.1.a.2"), ShouldBeError, "invalid OID")
```
louischan-oursky, line 198: Add a test case to show your regex can handle `0.01`

louischan-oursky (changes requested):

pkg/lib/config/ldap.go
```diff
@@ -23,21 +23,21 @@ var _ = Schema.Add("LDAPServerConfig", `
...
 	"additionalProperties": false,
-	"required": ["name", "url", "base_dn", "search_filter_template", "user_id_attribute_oid"],
+	"required": ["name", "url", "base_dn", "search_filter_template", "user_id_attribute_name"],
 	"properties": {
 		"name": { "type": "string", "minLength": 1 },
 		"url": { "type": "string", "format": "ldap_url" },
 		"base_dn": { "type": "string", "format": "ldap_dn" },
 		"search_filter_template": { "type": "string", "format": "ldap_search_filter_template" },
-		"user_id_attribute_oid": { "type": "string", "format": "ldap_oid" }
+		"user_id_attribute_name": { "type": "string", "format": "ldap_attribute" }
```
louischan-oursky, line 32: The format is called `ldap_attribute_name`

pkg/util/validation/formats.go
```diff
@@ -309,29 +309,29 @@ func (FormatLDAPSearchFilterTemplate) CheckFormat(value interface{}) error {
...
+		return errors.New("expect non-empty attribute")
 	}
 
-	matched, err := regexp.MatchString(`^(?:[0-9]|[1-9][0-9]+)(?:\.(?:[0-9]|[1-9][0-9]+))+$`, str)
+	// An attribute description option is represented by the ABNF:
+	// options = *( SEMI option )
+	// option = 1*keychar
+	// keychar = ALPHA / DIGIT / HYPHEN
+	// According to https://datatracker.ietf.org/doc/html/rfc4512#section-2.5
+	matched, err := regexp.MatchString(`^[a-zA-Z\d-]+$`, str)
```
louischan-oursky, line 329: But attribute name starting with a hyphen is uncommon. Let's use production rule of `keystring` here. Update the comment to reflect this implementation choice.

pkg/util/validation/formats_test.go
```diff
@@ -182,20 +182,22 @@ func TestFormatLDAPSearchFilterTemplate(t *testing.T) {
...
+		So(f("ou"), ShouldBeNil)
+		So(f("dc"), ShouldBeNil)
+		So(f("hyphen-"), ShouldBeNil)
+		So(f("123"), ShouldBeNil)
+		So(f("abcdEFG123-"), ShouldBeNil)
+		So(f(""), ShouldBeError, "expect non-empty attribute")
+		So(f("should have no space"), ShouldBeError, "invalid attribute")
+		So(f("abcd!@#$%^&*()"), ShouldBeError, "invalid attribute")
+		So(f("dc="), ShouldBeError, "invalid attribute")
+		So(f("dc=example,dc=com"), ShouldBeError, "invalid attribute")
```
louischan-oursky, line 200: Add a test to show we forbid attribute name starting with a digit or a hyphen.

louischan-oursky (changes requested):

louischan-oursky (approved):

PR #4487, opened 2024-07-23
Change default otp valid time to 5 minutes

yyorkchan:
Demo under no yaml settings for code_valid_period:
https://github.com/user-attachments/assets/b0ce2023-a141-4dbf-9954-0c4cbe7b7676

PR #4523, opened 2024-08-01
Fix VerifyWithSpec is not updated to include Form

louischan-oursky:
In #4487, we require VerifyOptions to have Form if verifying OOB OTP. But we only updated VerifyOneWithSpec.

ref DEV-1680

### DIFF
--- a/pkg/lib/interaction/nodes/authn_oob.go
+++ b/pkg/lib/interaction/nodes/authn_oob.go
@@ -6,6 +6,7 @@ import (
 	"github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/lib/authn"
 	"github.com/authgear/authgear-server/pkg/lib/authn/authenticator"
+	"github.com/authgear/authgear-server/pkg/lib/authn/otp"
 	"github.com/authgear/authgear-server/pkg/lib/facade"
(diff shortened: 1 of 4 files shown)


pkg/lib/interaction/nodes/authn_oob.go
```diff
@@ -36,6 +37,7 @@ func (e *EdgeAuthenticationOOB) Instantiate(ctx *interaction.Context, graph *int
 			Code: input.GetOOBOTP(),
 		},
 	}, &facade.VerifyOptions{
+		Form: otp.FormCode,
```
tung2744, line 40: I'm not sure if it is always code here, as we support using link in oob authentication too.

louischan-oursky: Those login link are handled in pkg/lib/linteraction/nodes/authn_login_link.go

louischan-oursky: So this node is really for code only.

tung2744:
I actually still thinks we should not need to know the `form` of OTP before we are able to verify it. Therefore, `verifyWithSpec` should not take a `Form` as a parameter. (And `VerifyOTP` should not require a Form too).
But I think I will still merge this fix as a quick fix first.

tung2744 (approved):

PR #4526, opened 2024-08-01
Implement ldap client

yyorkchan:


pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,109 @@
...
+	result := buf.String()
+	result = strings.TrimSpace(result)
+
+	_, err = ldap.CompileFilter(result)
+	if err != nil {
+		return "", err
+	}
+
+	return result, nil
+}
```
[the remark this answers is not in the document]
cychiuae:
```go
type LDAPSearchFilterTemplate string

func (LDAPSearchFilterTemplate) Execute(info map[string]interface{}) (string, error) {
   // the execution logic.
} 

// format.go

func (LDAPSearchFilterTemplateFormat() CheckFormat(value interface{}) error {
  templateString := value.(string)
  searchTemplate = LDAPSearchFilterTemplate(templateString)
  testcases := []string{}
  for _, testcase := range testcases {
     _, err := searchTemplate.Execute(testcase)
   .....
  }
  ......
}

// here
func (c *LDAPClient) SomeMethod(username string) {
  searchTemplate = LDAPSearchFilterTemplate(c.templateString)
  searchTemplate.Execute(username)
}

```

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (c *LDAPClient) AuthenticateUser(username string, password string) (*ldap.Entry, error) {
+	searchFilter, err := parseSearchFilter(c.Config.SearchFilterTemplate, username)
+	if err != nil {
+		return nil, err
+	}
+
+	searchRequest := ldap.NewSearchRequest(
+		c.Config.BaseDN,
+		ldap.ScopeWholeSubtree, ldap.DerefAlways, 2, 0, false,
```
cychiuae, line 87: I think we need to add comment to remind us why the forth param is 2

Time limit should not be infinite.

and why false

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,109 @@
...
+
+	sr, err := c.conn.Search(searchRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(sr.Entries) != 1 {
+		return nil, nil
+	}
+
```
cychiuae, line 101: I think here should check `len(sr.Entries)` `=0` or more than 1. We should throw error in both cases

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,109 @@
...
+	"errors"
+	"html/template"
+	"net/url"
+	"strings"
+
+	"github.com/authgear/authgear-server/pkg/lib/config"
+	"github.com/go-ldap/ldap/v3"
+)
+
+type UserProfile *ldap.Entry
```
cychiuae, line 14: Let's remove this type

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"bytes"
+	"errors"
+	"html/template"
+	"net/url"
+	"strings"
+
+	"github.com/authgear/authgear-server/pkg/lib/config"
+	"github.com/go-ldap/ldap/v3"
```
cychiuae, line 11: Please run `make fmt`

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+type UserProfile *ldap.Entry
+
+type LDAPClient struct {
```
cychiuae, line 16: Rename this Client

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,109 @@
...
+
+	trimedFilter := strings.TrimSpace(filter)
+
+	tmpl, err := template.New("search_filter").Parse(strings.Trim(trimedFilter, "\n"))
+	if err != nil {
+		return "", err
+	}
+	var buf bytes.Buffer
+	// make a string array of phone, username, email
+	err = tmpl.Execute(&buf, map[string]string{"Username": username})
```
cychiuae, line 62: I found that this implementation allows user to make filter injection. Please escape the username before execute the template

pkg/lib/ldap/client.go
yyorkchan: SearchRequest.typesOnly: An indicator as to whether Search results are to contain both
   attribute descriptions and values, or just attribute descriptions.
   Setting this field to TRUE causes only attribute descriptions (and
   not values) to be returned.  Setting this field to FALSE causes both
   attribute descriptions and values to be returned.
Hence we set to false for getting both attribute descriptions and values.

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/go-ldap/ldap/v3"
+
+	"github.com/authgear/authgear-server/pkg/api"
+	"github.com/authgear/authgear-server/pkg/lib/config"
+	"github.com/authgear/authgear-server/pkg/util/ldaputil"
+)
+
+type Client struct {
+	Config config.LDAPServerConfig
+	conn   *ldap.Conn
```
louischan-oursky, line 16: First of all, a Client should be stateless. It does not contain a conn.

A common pattern in programming with resources is to return a handle. https://en.wikipedia.org/wiki/Handle_(computing) In this case, you return an interface to represent the handle (that is, the connection)

In this case, we expect this client to do one simple thing, which is to authenticate an end-user, so it should expose only one method.

```golang
type Client struct {
  Config config.LDAPServerConfig
}

func (c *Client) AuthenticateEndUser(username string, password string) (*ldap.Entry, error) {
  // Call the private method to make a new connection.
  conn, err := c.makeConnection()
  if err != nil {
    return nil, err
  }
  // Use defer statement to close the connection.
  defer conn.Close()

  // Call the private method to bind the search user.
  err := c.bind()
  if err != nil {
    return nil, err
  }
  
  entry, err := c.authenticateEndUser(username, password)
  if err != nil {
    return nil, err
  }
  
  return entry, nil
}
```

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (c *Client) Connect() error {
+	url, err := url.Parse(c.Config.URL)
+	if err != nil {
+		return err
+	}
+
+	if url.Port() == "" {
+		if url.Scheme == "ldaps" {
+			c.Config.URL = c.Config.URL + ":636"
```
louischan-oursky, line 33: You modify something someone pass to you. **Never do this**.

Instead, you should modify something you create in this function, that is, `url`.
Another problem, you name the variable `url`, which clashes with the package `url`. That would make you unable to reference to `url` (the package). So, please name it `u`.

To summaries, What you need to do here is 

```
u, err := url.Parse(c.Config.URL)
if err != nil {
  return err
}

if u.Port() == "" {
  switch u.Scheme {
  case "ldaps":
    u.Host = u.Host + ":636"
  case "ldap":
    u.Host = u.Host + ":389"
  }
}

ldapURLString := u.String()
conn, err := ldap.DialURL(ldapURLString)
```

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (c *Client) AuthenticateUser(username string, password string) (*ldap.Entry, error) {
+	searchFilter, err := ldaputil.ParseFilter(c.Config.SearchFilterTemplate, username)
+	if err != nil {
+		return nil, err
+	}
+
+	searchRequest := ldap.NewSearchRequest(
+		c.Config.BaseDN,
+		ldap.ScopeWholeSubtree, ldap.DerefAlways, 2, 10, false,
```
louischan-oursky, line 80: Document the reason we are using these values.

If you read this code 10 days later, can you tell me what `2` and `10` mean?

pkg/util/ldaputil/parseFilter.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"bytes"
+	"strings"
+	"text/template"
+
+	"github.com/go-ldap/ldap/v3"
+)
+
+func ParseFilter(filter string, username string) (string, error) {
```
louischan-oursky, line 11: First of all, who uses this function?

Second, the testcase commit should be merged into this commit. An implementation with a testcase is considered as atomic commit. Do not separate them.

louischan-oursky (changes requested):

pkg/lib/ldap/client.go
yyorkchan: Check if unique attribute exist in entry

pkg/util/ldaputil/parseFilter_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestParseFilter(t *testing.T) {
+	Convey("Parse Filter", t, func() {
+		filterTemplate := "(&(objectclass=persion)(uid={{.Username}}))"
+		username := "hi)(email=*"
+		filter, _ := ParseFilter(filterTemplate, username)
+		So(filter, ShouldNotEqual, "(&(objectclass=persion)(uid=hi)(email=*))")
```
louischan-oursky, line 14: Add ShouldEqual to what.

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,106 @@
...
+		searchFilter,
+		[]string{},
```
louischan-oursky, line 96: Specify `[]string{"*"}` and include https://datatracker.ietf.org/doc/html/rfc4511#section-4.5.1.8 as comment.

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,106 @@
...
+		nil,
```
louischan-oursky, line 97:
```
// sizeLimit is 2 because 2 is sufficient for us to tell if the username can match more then 1 entry.
const sizeLimit = 2
// timeoutInSeconds is the default timeout when performing the search request.
// We may want to make this configurable in the future.
const timeoutInSeconds = 10
```

louischan-oursky, line 97:
```
// We do not need pass controls.
var controls []ldap.Control = nil
```

pkg/lib/ldap/client.go
```diff
@@ -113,5 +117,10 @@ func (c *Client) AuthenticateUser(username string, password string) (*ldap.Entry
 		return nil, err
 	}
 
+	uniqueIdentifierValue := entry.GetAttributeValue(c.Config.UserUniqueIdentifierAttribute)
```
louischan-oursky, line 120: UserUniqueIdentifierAttribute -> UserIDAttributeName

pkg/lib/ldap/client.go
```diff
@@ -124,3 +124,52 @@ func (c *Client) AuthenticateUser(username string, password string) (*ldap.Entry
...
+		if err != nil {
+			return api.ErrLDAPInvalidFilterTemplate
+		}
+		SearchRequest := ldap.NewSearchRequest(
+			c.Config.BaseDN,
+			ldap.ScopeWholeSubtree, ldap.DerefAlways, 2, 10, false,
+			searchFilter,
+			[]string{},
+			nil,
+		)
```
louischan-oursky, line 155: Extract a private method to make a search request.

pkg/api/errors.go
```diff
@@ -35,3 +35,13 @@ var ErrClaimNotVerifiable = NewInvariantViolated("ClaimNotVerifiable", "claim no
 var ChangePasswordFailed = apierrors.Invalid.WithReason("ChangePasswordFailed")
 var ErrNoPassword = ChangePasswordFailed.NewWithCause("the user does not have a password", apierrors.StringCause("NoPassword"))
 var ErrPasswordReused = ChangePasswordFailed.NewWithCause("password reused", apierrors.StringCause("PasswordReused"))
+
+var LDAPConnectionTestFailed = apierrors.ServiceUnavailable.WithReason("LDAPConnectionTestFailed")
+var ErrLDAPCannotConnect = LDAPConnectionTestFailed.NewWithCause("failed to connect", apierrors.StringCause("FailedToConnect"))
+var ErrLDAPInvalidBindUser = LDAPConnectionTestFailed.NewWithCause("failed to bind search user", apierrors.StringCause("FailedToBindSearchUser"))
+var ErrLDAPInvalidBaseDN = LDAPConnectionTestFailed.NewWithCause("invalid base DN", apierrors.StringCause("InvalidBaseDN"))
+var ErrLDAPInvalidFilterTemplate = LDAPConnectionTestFailed.NewWithCause("invalid filter template", apierrors.StringCause("InvalidSearchFilterTemplate"))
```
louischan-oursky, line 43: Remove ErrLDAPInvalidBaseDN

louischan-oursky, line 43: Remove ErrLDAPInvalidFilterTemplate

pkg/api/errors.go
```diff
@@ -35,3 +35,13 @@ var ErrClaimNotVerifiable = NewInvariantViolated("ClaimNotVerifiable", "claim no
...
+var ErrLDAPInvalidUniqueAttribute = LDAPConnectionTestFailed.NewWithCause("invalid ID attribute", apierrors.StringCause("InvalidUserIDAttributeName"))
```
louischan-oursky, line 44: Remove ErrLDAPInvalidUniqueAttribute

louischan-oursky (changes requested):

louischan-oursky (approved):

PR #4527, opened 2024-08-01
Handle hiding brand logo container for inline preview

elise-ng:
Ref DEV-1673

I personally prefer the pure css solution before the refactoring (`&>img[src=""]`) but let's go with the class name direction for now

https://github.com/user-attachments/assets/55722a59-e17b-41ae-8be6-309e4808ef6e

### DIFF
--- a/authui/src/authflowv2/components/brand-logo.css
+++ b/authui/src/authflowv2/components/brand-logo.css
@@ -5,7 +5,7 @@
   }
 
   .brand-logo__container {
-    @apply hidden;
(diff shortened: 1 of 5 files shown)


elise-ng:
@louischan-oursky

louischan-oursky:
@elise-ng I agree with your point. I prefer pure CSS as well. I added a commit to switch this whole thing to pure CSS.

louischan-oursky (approved):

PR #4529, opened 2024-08-02
Set phone input border radius along with main input

elise-ng:
Ref DEV-1697

https://github.com/user-attachments/assets/da610dba-ce10-4c19-91ac-010454856314


elise-ng:
@louischan-oursky

louischan-oursky (approved):

PR #4532, opened 2024-08-05
Fix settings action cannot be used after called makeAuthenticatedURL

tung2744:
ref DEV-1711


louischan-oursky:
I think we better upgrade `make generate` to run `make fmt` to avoid this build error.

```
.PHONY: generate
generate:
	go generate ./pkg/... ./cmd/... ./e2e/...
	$(MAKE) fmt
```

louischan-oursky (changes requested):
Mark this Request changes.

tung2744:
Fixed, thanks!

louischan-oursky (approved):

PR #4533, opened 2024-08-05
[DEPS #4550] Add bot protection to authflow signup verify step

pkong-ds:
## Blocked by
- #4508
- #4550

## What's in this PR?

DEV-1608 requires captcha in signup > verify, which is not supported yet. It is added in this PR

### DIFF
--- a/pkg/lib/config/authentication_flow_bot_protection.go
+++ b/pkg/lib/config/authentication_flow_bot_protection.go
@@ -6,42 +6,11 @@ var _ = Schema.Add("AuthenticationFlowBotProtection", `
 	"additionalProperties": false,
 	"required": ["mode"],
 	"properties": {
-		"mode": { "$ref": "#/$defs/BotProtectionRiskMode" },
-		"provider": { "$ref": "#/$defs/AuthenticationFlowBotProtectionProvider" }
-	},
-	"allOf": [
-		{
-			"if": {
-				"required": ["mode"],
-				"properties": {
-					"mode": {
-						"enum": ["always"]
-					}
-				}
-			},
-			"then": {
-				"required": ["provider"]
-			}
-		}
-	]
-}
-`)
-
-var _ = Schema.Add("AuthenticationFlowBotProtectionProvider", `
-{
-	"type": "object",
-	"additionalProperties": false,
-	"required": ["type"],
-	"properties": {
-		"type": { "type": "string", "enum": ["cloudflare", "recaptchav2"] }
+		"mode": { "$ref": "#/$defs/BotProtectionRiskMode" }
 	}
 }
 `)
 
-type AuthenticationFlowBotProtectionProvider struct {
-	Type BotProtectionProviderType `json:"type,omitempty"`
-}
 type AuthenticationFlowBotProtection struct {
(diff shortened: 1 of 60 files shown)


pkong-ds:
@louischan-oursky Encountered below blocker, can help? 🙏 

## Context
See https://github.com/authgear/authgear-server/pull/4533/commits/e8f54d1cb91f71bcd7e5bf3c0fb1bc30a70073c3, I tried to require `bot_protection` input in `signup > verify` step

## Problem
Similar problem with before: After my changes, the authflow tree will be stuck at a step that requires bot protection.

### Postman Screenshot Video - original problem

Note how an extra response `data:{}` showed.

https://github.com/user-attachments/assets/6570a4ea-d0cb-4102-855c-4bbdaa56cfe0

## Attempt
Previously we work-around this by going one upper layer of intent, and _also requires bot protection input_ there (e.g. if `intent_signup_identify_email` requires bot protection, `intent_signup_identify` will also requires bot protection)

Thus, in my case, I tried to modify `IntentSignupFlowSteps` to requires bot protection if current step is `IntentSignupFlowStepVerify`.

But then it yield another weired result instead

### Postman Screenshot Video - problem after my attempt

Note the step is stuck at `identify` now.

https://github.com/user-attachments/assets/7f53eb43-9d90-482b-8e72-e998f29010ff

## Non-problem
The flow works as expected after I pass bot protection input.

pkong-ds:
As per offline discussion with louis, we plan to

- Not allow `bot_protection` in verify step _directly_ ;
- Instead, in `signup > identify`, check if there is any `verify` sub-step
   - If exists, add `bot_protection` to `signup > identify`
   - else (no verify sub-step`, do not do anything

louischan-oursky (changes requested):
Please rebase.

pkong-ds:
Rebased, thanks!

pkg/lib/config/authentication_flow_bot_protection.go
```diff
@@ -41,7 +41,41 @@ var _ = Schema.Add("AuthenticationFlowBotProtectionProvider", `
...
+func (p *AuthenticationFlowBotProtectionProvider) Equal(p2 *AuthenticationFlowBotProtectionProvider) bool {
+	if p == nil && p2 == nil {
+		return true
+	}
+	if p == nil || p2 == nil {
+		return false
+	}
+
+	return p.Type == p2.Type
+}
```
louischan-oursky, line 54: 1. Modify the spec to remove `bot_protection.provider` in branch.
2. Conform to the spec.

louischan-oursky (changes requested):

pkg/lib/config/authentication_flow_bot_protection.go
pkong-ds: Modified spec in https://github.com/louischan-oursky/authgear-server/commit/36bfc95b33e603b28536df77be0bd48c9a56e0ac

pkong-ds: > conform to spec

Will depends this on #4550

pkong-ds:
Updated, thanks!

louischan-oursky (approved):

PR #4536, opened 2024-08-06
wip: Sign up with ldap

cychiuae:


pkg/lib/ldap/factory.go
```diff
@@ -0,0 +1,40 @@
...
+	var serverConfig *config.LDAPServerConfig
+	for _, s := range f.Config.Servers {
+		if s.Name == serverName {
+			serverConfig = s
+			break
+		}
+	}
+
+	if serverConfig == nil {
+		return nil, api.ErrLDAPServerNotFound
```
louischan-oursky, line 23: I think we will not encounter this error. The JSON schema is supposed to validate the server_name matches one of the defined server. Please update the JSON schema if it is not the case.

So no need to return an API error here. You can just panic.

pkg/lib/ldap/factory.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/authgear/authgear-server/pkg/api"
+	"github.com/authgear/authgear-server/pkg/lib/config"
+)
+
+type ClientFactory struct {
+	Config       *config.LDAPConfig
+	SecretConfig *config.LDAPServerUserCredentials
+}
+
+func (f *ClientFactory) Authenticate(serverName string, username string, password string) (*Entry, error) {
```
louischan-oursky, line 13: Since it is factory, it should expose one one method.

```
func (f *ClientFactory) MakeClient(serverName string) *ldap.Client
```

pkg/lib/authenticationflow/declarative/utils_ldap.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/authgear/authgear-server/pkg/api"
+	"github.com/authgear/authgear-server/pkg/api/model"
+	authflow "github.com/authgear/authgear-server/pkg/lib/authenticationflow"
+	"github.com/authgear/authgear-server/pkg/lib/authn/identity"
+	"github.com/authgear/authgear-server/pkg/lib/config"
+	"github.com/authgear/authgear-server/pkg/lib/ldap"
+)
+
+func createIdentitySpecFromLDAPEntry(deps *authflow.Dependencies, serverName string, entry *ldap.Entry) (*identity.Spec, error) {
+	var ldapServerConfig *config.LDAPServerConfig
```
louischan-oursky, line 13: Add a method in LDAPConfig to find a LDAPServerConfig. Just like OAuthSSOConfig.GetProviderConfig.

```
LDAPConfig.GetServerConfig(serverName string) (*LDAPServerConfig, bool)
```

And then LDAPFactory should take LDAPServerConfig instead of server_name. So it does not need to find twice.

pkg/lib/ldap/entry.go
```diff
@@ -2,6 +2,21 @@ package ldap
 
 import "github.com/go-ldap/ldap/v3"
 
+const (
+	AttributeNameEmail    = "mail"
+	AttributeNameMobile   = "mobile"
+	AttributeNameUsername = "uid"
```
louischan-oursky, line 18: Any references on these attributes? Are they well-known? Please include the references as comments.

louischan-oursky (changes requested):

pkg/lib/authenticationflow/declarative/input_take_ldap.go
```diff
@@ -0,0 +1,77 @@
...
+func (i *InputSchemaTakeLDAP) MakeInput(rawMessage json.RawMessage) (authflow.Input, error) {
+	var input InputTakeLDAP
+	err := i.SchemaBuilder().ToSimpleSchema().Validator().ParseJSONRawMessage(rawMessage, &input)
+	if err != nil {
+		return nil, err
+	}
+	return &input, nil
+}
+
+type InputTakeLDAP struct {
```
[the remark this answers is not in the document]
cychiuae: Actually we need this because `InputSchemaTakeLDAP.MakeInput` will return a `InputTakeInput`.

pkg/lib/ldap/factory.go
cychiuae: Here i just take reference to existing `OAuthProviderFactory`

cychiuae: And I will not make such assumption here because this will assumed the method will only be used in authflow api

cychiuae:
@louischan-oursky Updated thanks!

louischan-oursky (approved):

PR #4540, opened 2024-08-06
Implement ldap service

yyorkchan:


pkg/lib/ldap/service.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+	return matchedConfigs
+}
+
+func (s *Service) Authenticate(username string, password string) (*ldap.Entry, error) {
+	matchedConfigs := s.matchServerConfig()
+	for _, matchedConfig := range matchedConfigs {
+		client := NewClient(matchedConfig.ServerConfig, matchedConfig.SecretConfig)
+		entry, err := client.AuthenticateUser(username, password)
```
cychiuae, line 47: We should only proceed if err is user not found error

pkg/lib/ldap/client.go
```diff
@@ -0,0 +1,178 @@
...
+
+	ldapURLString := u.String()
+
+	conn, err := ldap.DialURL(ldapURLString)
+	if err != nil {
+		return nil, err
+	}
+
+	if u.Scheme == "ldap" {
+		_ = conn.StartTLS(nil)
```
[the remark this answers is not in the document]
cychiuae: https://github.com/go-ldap/ldap/blob/dbdc485259442f987d83e604cd4f5859cfc1be58/v3/conn.go#L389

pkg/lib/ldap/deps.go
```diff
@@ -0,0 +1,10 @@
+package ldap
+
+import (
+	"github.com/google/wire"
+)
+
+var DependencySet = wire.NewSet(
+	wire.Struct(new(Client), "*"),
```
cychiuae, line 8: We don't need to list Client in the dependency set coz it will internally used by Service only

cychiuae:
Please rebase your PR

cychiuae:
Seems I misunderstood the requirement and this PR is no longer need. 

@yyorkchan However, you still need to help fix the problem of startTLS fails will lead to connection close.

PR #4544, opened 2024-08-07
Fix ldap startTLS does not reconnect after fail

yyorkchan:


louischan-oursky (approved):
@yyorkchan I added 2 commits to document things.

PR #4546, opened 2024-08-07
Allow conditional captcha by flow in portal

pkong-ds:
## Preview

https://github.com/user-attachments/assets/8033e212-f2d3-4f6a-8a5c-c0b29b4ba196

image

image


pkong-ds:
@tung2744 can help review this? Thanks~

pkong-ds:
Seems `npm run prettier` in portal somehow failing 😢 lemme check check 🙏

portal/src/graphql/portal/BotProtectionConfigurationScreen.tsx
```diff
@@ -106,14 +211,8 @@ function constructConfig(
...
-    config.bot_protection = {
-      enabled: currentState.enabled,
-      provider: {
-        type: currentState.providerType,
-        site_key:
-          currentState.providerConfigs[currentState.providerType]?.siteKey,
-      },
-    };
+    config.bot_protection.requirements ??= {};
+    config.bot_protection = constructBotProtectionConfig(currentState);
```
tung2744, line 215: It seems the above ??= in L212-214 has no effect because the final object is from `constructBotProtectionConfig(currentState)`

portal/src/graphql/portal/BotProtectionConfigurationScreen.tsx
```diff
@@ -181,7 +280,8 @@ function ProviderCard(props: ProviderCardProps) {
   } = useSystemConfig();
 
   return (
-    <div
+    <button
+      type="button"
```
tung2744, line 284: Maybe not related to this pr, but I think we can replace these inline styles with css classes.
Define a class which only applies to the element when the button is in disabled state, so we don't need to use the `style` attribute.

portal/src/graphql/portal/BotProtectionConfigurationScreen.tsx
```diff
@@ -551,6 +635,326 @@ const BotProtectionConfigurationContent: React.VFC<BotProtectionConfigurationCon
...
+        }))
+      },
+      [setRequirements]
+    )
+    const onRenderFlowConfigLabel = useCallback(
+      (item?: RequirementConfigListItem, _index?: number, _column?: IColumn) => {
+        if (item == null) {
+          return null;
+        }
+        // FIXME: vertical center align text
```
tung2744, line 721: Do we need to fix this FIXME?

pkong-ds: yes, missed it, will fix 🙏

pkong-ds:
@tung2744 updated, thanks!

louischan-oursky:
The cache was added by @IniZio

IniZio:
Yes the cached flag added only recently, can remove it if prettier cache is buggy 🙏

tung2744 (approved):

PR #4549, opened 2024-08-07
Support MFA grace period in authflow

IniZio:
ref DEV-1744 DEV-1621 DEV-1622

**Password**

https://github.com/user-attachments/assets/9c19f6ac-08f2-4514-a701-a08b6128e55b

**TOTP**

https://github.com/user-attachments/assets/e335931e-eb5e-4899-a43c-9163eb6feaa4

**SMS**

https://github.com/user-attachments/assets/b177da36-8314-4a61-9880-6281d4ae34f0

**Email**

https://github.com/user-attachments/assets/bc370c4a-aa2d-4093-88dc-1ba6765f9bec

### DIFF
--- a/pkg/auth/handler/webapp/authflowv2/routes.go
+++ b/pkg/auth/handler/webapp/authflowv2/routes.go
@@ -322,7 +322,7 @@ func (n *AuthflowV2Navigator) navigateStepIdentify(s *webapp.AuthflowScreenWithF
 	}
 }
 func (n *AuthflowV2Navigator) navigateLoginStepAuthenticate(s *webapp.AuthflowScreenWithFlowResponse, r *http.Request, webSessionID string, result *webapp.Result) {
-	switch data := s.BranchStateTokenFlowResponse.Action.Data.(type) {
+	switch data := s.StateTokenFlowResponse.Action.Data.(type) {
 	case declarative.StepAuthenticateData:
 		options := data.Options
 		index := *s.Screen.TakenBranchIndex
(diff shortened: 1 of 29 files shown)


pkong-ds:
Oh no 😞 these steps will need captcha too

IniZio:
@louischan-oursky closed #4498 and include in this one instead 🙏

tung2744:
Please resolve conflicts

pkg/auth/handler/webapp/authflowv2/routes.go
```diff
@@ -322,7 +322,7 @@ func (n *AuthflowV2Navigator) navigateStepIdentify(s *webapp.AuthflowScreenWithF
 	}
 }
 func (n *AuthflowV2Navigator) navigateLoginStepAuthenticate(s *webapp.AuthflowScreenWithFlowResponse, r *http.Request, webSessionID string, result *webapp.Result) {
-	switch data := s.BranchStateTokenFlowResponse.Action.Data.(type) {
+	switch data := s.StateTokenFlowResponse.Action.Data.(type) {
```
IniZio, line 325: I see latest main uses `BranchStateTokenFlowResponse`, think should use `StateTokenFlowResponse` since `BranchStateTokenFlowResponse` is for showing available options only?

IniZio:
@tung2744 Rebased but unsure with authui navigation change in https://github.com/authgear/authgear-server/pull/4549#discussion_r1713313216 :adore2x:

Tried all authenticator options work both in signup and login

pkg/auth/handler/webapp/authflowv2/routes.go
tung2744: It seems actually it is two case.
1. When you take the branch without actually sending an input: Use BranchStateTokenFlowResponse
2. When you take the branch which requires a actual input: Use `StateTokenFlowResponse`

Maybe need @louischan-oursky

tung2744: It seems using `StateTokenFlowResponse` here is correct.
When looking at `takeBranchResultSimple`, `StateTokenFlowResponse` is same as `BranchStateTokenFlowResponse` in case 1.

louischan-oursky:
Replaced by https://github.com/authgear/authgear-server/pull/4555

PR #4551, opened 2024-08-07
Do not cache `prettier:format` in portal

pkong-ds:
## What's in this PR?
It fixes below problem. Now at step 3, the mal-formatted files will be formatted
## Problem

1. run `npm run prettier`
2. mal-formatted files are cached in `./node_modules/.cache/prettier/.prettier-cache` 
    ref https://prettier.io/docs/en/cli.html#--cache
3. run `npm run prettier:format`

At 3, the mal-formatted files will NOT be formatted, which is unexpected

### DIFF
--- a/authui/package-lock.json
+++ b/authui/package-lock.json
@@ -43,7 +43,7 @@
         "parcel": "2.10.2",
         "parcel-reporter-static-files-copy": "1.5.3",
         "postcss": "8.4.31",
-        "prettier": "3.1.0",
+        "prettier": "2.8.8",
         "stylelint": "14.16.1",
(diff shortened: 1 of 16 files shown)


louischan-oursky:
@pkong-ds @IniZio This is a prettier bug, and it was fixed in 2.8.x so we will keep --cache and update to 2.8.x instead. I have modified this PR to do so.

louischan-oursky (approved):

pkong-ds:
Thanks. 

The bug - https://github.com/prettier/prettier/issues/13015
The fix - https://github.com/prettier/prettier/pull/13016

PR #4522, opened 2024-08-01
Support never and tuple matcher in e2e test

IniZio:
- Add never type to e2e matcher
- Fix tuple type and nil handling in e2e matcher

PR #4557, opened 2024-08-08
Fix import api failed to handle email username with upper cases

pkong-ds:
## Blocked by
- #4522 
- #4568 

## What's in this PR?
Fix a bug where 
if
- `authgear.yaml` has `username/email.case_sensitive=false`

then
- import username/email with uppercase with fail
 
## Question
On 2nd thought, maybe the expected behavior is to skip these `invalid uppercase` imports and show error message?


pkong-ds:

louischan-oursky (changes requested):
1. Rename StandardAttributesList() to NonIdentityAwareStandardAttributesList()
2. Fix the failing test.

pkong-ds:
@louischan-oursky local CI pass now, thanks!

louischan-oursky (approved):

PR #4558, opened 2024-08-09
Fix flaky password generator test

IniZio:
ref DEV-1762

### DIFF
--- a/pkg/lib/authn/authenticator/password/generator.go
+++ b/pkg/lib/authn/authenticator/password/generator.go
@@ -153,9 +153,6 @@ func (g *Generator) generateOnce() (string, error) {
 	g.Rand.Shuffle(len(passwordBytes), func(i int, j int) {
 		passwordBytes[i], passwordBytes[j] = passwordBytes[j], passwordBytes[i]
 	})
-	if err != nil {
-		return "", err
-	}
 
 	return string(passwordBytes), nil
 }
--- a/pkg/lib/authn/authenticator/password/generator.go
+++ b/pkg/lib/authn/authenticator/password/generator.go
@@ -1,12 +1,11 @@
 package password
 
 import (
-	"crypto/rand"
 	"io"
+	"math/rand"
 	"strings"
 
 	"github.com/authgear/authgear-server/pkg/lib/config"
-	utilrand "github.com/authgear/authgear-server/pkg/util/rand"
 )
 
 // Character list for each category.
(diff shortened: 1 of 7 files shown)


pkg/lib/authn/authenticator/password/generator_test.go
```diff
@@ -161,14 +161,25 @@ func TestMinGuessableLevelRequirement(t *testing.T) {
 	})
 }
 
+type ExcludedKeywordRandSource struct {
+}
+
+func (s *ExcludedKeywordRandSource) RandomBytes(n int) ([]byte, error) {
+	return []byte{byte(rand.Intn(9))}, nil
```
louischan-oursky, line 168: - You should still respect the argument `n`.
- Use `math/rand/v2` instead.
- I think you mean `rand.Intn(8)`. Because you cast the int to a byte. A byte can be 0,1,2,3,4,5,6,7; 8 values in total. rand.Intn(9) may return 8, which is overflow.

So it should look like

```
output := make([]byte, n)
for i := 0; i < n; ++i {
  b := byte(rand.Intn(8))
  output[i] = b
}
return output, nil
```

louischan-oursky (changes requested):

pkg/lib/authn/authenticator/password/generator_test.go
IniZio: Fixed

pkg/lib/authn/authenticator/password/generator.go
```diff
@@ -77,55 +76,37 @@ func (s characterSet) Append(w io.Writer) error {
...
 		}
 
 		if err := g.checkPassword(password); err == nil {
-			return password, nil
+			return password, i, nil
 		}
 	}
 
-	return "", ErrPasswordGenerateFailed
+	return "", MaxTrials, ErrPasswordGenerateFailed
```
louischan-oursky, line 106: -1

pkg/lib/authn/authenticator/password/generator.go
```diff
@@ -77,55 +76,37 @@ func (s characterSet) Append(w io.Writer) error {
...
+	return password, err
+}
+
+func (g *Generator) generate() (string, int, error) {
 	for i := 0; i < MaxTrials; i++ {
-		password, err := g.generate()
+		password, err := g.generateOnce()
 		if err != nil {
-			return "", err
+			return "", i, err
```
louischan-oursky, line 98: -1

pkg/lib/authn/authenticator/password/generator_test.go
```diff
@@ -161,14 +151,30 @@ func TestMinGuessableLevelRequirement(t *testing.T) {
 	})
 }
 
+// Create a rand that extends Rand and generates 1 on the first call and 0 on the second call
```
louischan-oursky, line 154: Correct the comment.

louischan-oursky (changes requested):

louischan-oursky:
@IniZio It still failed.

IniZio:
Need rename `Intn` to `IntN`, should pass now 🙇

louischan-oursky (approved):

Issue #3233, opened 2023-07-20
Improve performance of list user query


fungc-io:
Re: https://github.com/authgear/authgear-server/issues/3176

Now only the top 10 results are shown
We would like to increase the number with pagination of 100 entries per page

![image](https://github.com/authgear/authgear-server/assets/15207470/0f8efa30-bba3-47fb-a521-d5c3172bbc43)

TODO:

- [ ] Refactor user queries so that it doesn't create N+1 queries on loading users. https://github.com/authgear/authgear-server/blob/2023C01/pkg/lib/authn/user/queries.go#L97 

---

- [x] Rewrite user.Queries
    - Reduce number of queries
        - [x] identity.Service should implement new method for batch querying identities
            - ListByUserIDs(userIDs []string)
        - [x] authenticator.Service should implement new method for batch querying authenticators
            - List(userIDs []string)
        - [x] verification.Service should be re-implemented to batch querying verification statuses
            - Rewrite GetVerificationStatuses(is []*identity.Info) to accept identity from multiple users
        - [x] stdattrs.ServiceNoEvent should be re-implemented to reduce number of queries
            - Rewrite DeriveStandardAttributes to accept multiple userIDs, and issue less queries
    - [x] Rewrite GetMany to use the new methods above
    - [x] Rewrite Get to use GetMany with one item

fungc-io:
@tung2744 please help fill in the technical details of this issue
Thanks! 🙏

tung2744:
This problem has significant impact on performance when the database latency increases. Therefore might need to fix it before supporting multi region database.

tung2744:
Added some technical details, see if you have any opinion? @louischan-oursky

louischan-oursky:
Looks good to me.

tung2744:
Although this improved the performance, the signup flow is still slow if the database latency is high. I think I would like to spend more time to investigate it.

Issue #3176, opened 2023-06-28
[Portal, Search] Error when click pagination on search result


joyzoursky:
## Background
- Environment: Staging
- Testing app: demo-enterprise
- Test account: fung+qa@oursky.com

## Description
When search in User Management and click pagination, it shows error `Error: invalid character 'o' looking for beginning of value`, not sure if it also happens on other search result pagination pages.

## Steps to reproduce
1. Go to User Management on portal
2. Search with some text with many result
3. Scroll down and click to the next page
4. See error message

## Expected result
Should be able to show the next page of result normally.

## Video

User info is masked:

https://github.com/authgear/authgear-server/assets/12539486/e46f631b-c5ad-45e0-a9ef-893f4599ef6e

tung2744:
@louischan-oursky This problem occurs because we used [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after) of elasticsearch. In other places, we assume the cursor is something like `offset:10`. When using elasticsearch, it is an object returned from the elasticsearch api.
This bug cannot be fixed, because if fact our ui is incompatible with the search_after parameter. We need to obtain the cursor from the `sort` field of the elasticsearch result, meaning that if we want to get a cursor pointing to page 4, we must be on page 3. However, the current ui allow us to jump to page 4 from page 1. In this case, we have no way to obtain the cursor.

By searching `CursorToSearchAfter` in code, it seems only user search is using `search_after`. What do you think if we simply switch to `from` param so that we can pass any offset to it?

tung2744:
@fungc-io After discussed with louis, we think that letting user to select the page when searching user seems not very useful. If there are too many results, the user can just type more to reduce the results. Therefore we suggest to remove the pagination for search result in this screen and only display the top 100 items. Note that If the user didn't type anything in the search bar the pagination can be keep.
We can also display some text like "Search results are limited to 100 rows, please modify your keyword if you didn't find what you want." at the bottom to let user know the search result is not fully displayed.
What do you think?

fungc-io:
I think it's not unlikely that a user want to search more than 100 items..
e.g. What if I want to export all gmail.com users, can it be done using Admin API if not the portal?

louischan-oursky:
> What if I want to export all gmail.com users, can it be done using Admin API if not the portal?

[...]

PR #4559, opened 2024-08-09
Make user screen page size as 50

pkong-ds:
- Seems previously the page size is intentionally set to 10 due to performance problem in [Issue #3176](https://github.com/authgear/authgear-server/issues/3176) 

- Then, the performance problem is fixed in [Issue #3233](https://github.com/authgear/authgear-server/issues/3233)

I tried increasing page size, seems no significant load speed difference (locally)

Maybe need to test on staging before/after, see if 50-page-size cause performance issue

### DIFF
--- a/portal/src/graphql/adminapi/UsersScreen.tsx
+++ b/portal/src/graphql/adminapi/UsersScreen.tsx
@@ -37,8 +37,8 @@ import {
 
 const pageSize = 10;
 // We have performance problem on the users query
-// limit to 10 items for now
-const searchResultSize = 10;
+// limit to 50 items for now
+const searchResultSize = 50;
 
 function useRemoteData(options: {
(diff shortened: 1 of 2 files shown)


pkong-ds:
@tung2744 since you were handling #3176 and #3233, maybe you can help review this? Thanks!

tung2744:
IIRC the performance problem is not fixed completely yet. But 50 should have no big problem.

louischan-oursky:
Keep the page size to 10. Only change search page size to 50

fungc-io:
Ok @pkong-ds @louischan-oursky , the video lgtm

tung2744:
A test failed but it seems not related to the pr. I will still merge it.

tung2744 (approved):

PR #4561, opened 2024-08-09
Remove `ON CONFLICT DO NOTHING` clause in e2e custom sqls

pkong-ds:
I found the issue was not

> error not thrown

Instead, it is

> database did not throw error when inserting duplicate primary key records

Which is caused by `ON CONFLICT DO NOTHING` clause in custom sql.

I will remove all `ON CONFLICT DO NOTHING` clauses as a fix 🙏 Since UUID() will be introduced in [DEV-1765](https://linear.app/authgear/issue/DEV-1765/add-generateuuid-to-e2e-template)

### DIFF
--- a/e2e/pkg/testrunner/testcase.go
+++ b/e2e/pkg/testrunner/testcase.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
 	"strings"
@@ -65,8 +66,9 @@ func (tc *TestCase) Run(t *testing.T) {
 		return
 	}
 
-	ok := tc.executeBeforeAll(t, cmd)
-	if !ok {
+	err = tc.executeBeforeAll(cmd)
+	if err != nil {
+		t.Errorf("failed to execute before hooks: %v", err)
 		return
 	}
 
@@ -78,6 +80,7 @@ func (tc *TestCase) Run(t *testing.T) {
 
 	var stepResults []StepResult
 	var state string
+	var ok bool
 
 	for i, step := range tc.Steps {
 		if step.Name == "" {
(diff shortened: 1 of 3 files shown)


e2e/pkg/testrunner/testcase.go
```diff
@@ -95,28 +98,29 @@ func (tc *TestCase) Run(t *testing.T) {
...
 				t.Errorf("failed to execute custom SQL: %v", err)
-				return false
+				return
 			}
 		default:
-			t.Errorf("unknown before hook type: %s", beforeHook.Type)
-			return false
+			errStr := fmt.Sprintf("unknown before hook type: %s", beforeHook.Type)
+			t.Errorf(errStr)
+			return errors.New(errStr)
```
louischan-oursky, line 119: err := fmt.Errorf("unknown before hook type: %v", beforeHook.Type)
t.Errorf("%v", err)
return err

e2e/pkg/testrunner/testcase.go
```diff
@@ -65,8 +66,9 @@ func (tc *TestCase) Run(t *testing.T) {
 		return
 	}
 
-	ok := tc.executeBeforeAll(t, cmd)
-	if !ok {
+	err = tc.executeBeforeAll(t, cmd)
+	if err != nil {
+		t.Errorf("failed to execute before hooks: %v", err)
```
louischan-oursky, line 71: Remove this line.

louischan-oursky (changes requested):

e2e/pkg/testrunner/testcase.go
```diff
@@ -95,28 +98,29 @@ func (tc *TestCase) Run(t *testing.T) {
 }
 
 // Execute before hooks to prepare fixtures
-func (tc *TestCase) executeBeforeAll(t *testing.T, cmd *End2EndCmd) (ok bool) {
+func (tc *TestCase) executeBeforeAll(t *testing.T, cmd *End2EndCmd) (err error) {
```
louischan-oursky, line 101:
```
func (tc *TestCase) executeBeforeAll(cmd *End2EndCmd) (err error) {
  if err != nil {
    return fmt.Errorf("asdasd: %w", err)
  }
}
```

louischan-oursky (changes requested):

pkong-ds:
@louischan-oursky @IniZio updated, thanks!

louischan-oursky (approved):

PR #4577, opened 2024-08-12
Support OIDC core scopes

louischan-oursky:
ref DEV-366

@fungc-io Here is the screenshot of the updated consent screen. The image is annotated with which scope corresponds to which description. However, you actually cannot test this. The OIDC client application you can create in portal is actually a first party OIDC application. A first party application always skip the consent screen.

SCR-20240812-rwwu


fungc-io:
Ok, how about authUI v2?

louischan-oursky:
> Ok, how about authUI v2?

Regardless of which UI, the consent screen is never shown.

fungc-io:
i see

pkg/lib/oauth/oidc/id_token.go
```diff
@@ -281,30 +260,31 @@ func (ti *IDTokenIssuer) GetUserInfo(userID string, clientLike *oauth.ClientLike
...
+				if ScopeAllowsClaim(scope, k) {
+					isAllowed = true
+					break
+				}
+			}
+
+			if isAllowed {
+				out[k] = v
 			}
 		}
```
tung2744, line 285: What about custom attributes in third party app?

louischan-oursky: Currently custom attributes are not visible in third party app. I think we need a new feature to make them visible.

The developer should configure which third party application can request the scope to see a custom attribute.

tung2744 (approved):

PR #4580, opened 2024-08-13
SAML: Add name id configs

tung2744:
ref DEV-1729
requires https://github.com/authgear/authgear-server/pull/4573

### DIFF
--- a/pkg/lib/config/saml.go
+++ b/pkg/lib/config/saml.go
@@ -19,7 +19,9 @@ var _ = Schema.Add("SAMLServiceProviderConfig", `
 	"type": "object",
 	"additionalProperties": false,
 	"properties": {
-		"id": { "type": "string" }
+		"id": { "type": "string" },
+		"name_id_format": { "$ref": "#/$defs/SAMLNameIDFormat" },
+		"name_id_attribute_pointer": { "$ref": "#/$defs/SAMLNameIDAttributePointer" }
 	},
 	"required": ["id"]
 }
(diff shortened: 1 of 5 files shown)


louischan-oursky:
Please rebase.

louischan-oursky (approved):

pkg/lib/config/saml.go
```diff
@@ -19,7 +19,9 @@ var _ = Schema.Add("SAMLServiceProviderConfig", `
 	"type": "object",
 	"additionalProperties": false,
 	"properties": {
-		"id": { "type": "string" }
+		"id": { "type": "string" },
+		"name_id_format": { "$ref": "#/$defs/SAMLNameIDFormat" },
```
louischan-oursky, line 23: NameID should be `nameid`

louischan-oursky (changes requested):

pkg/lib/config/saml.go
tung2744: The PascelCase NameID should be name_id in snake_case?

tung2744:
Sorry I forgot to update the metadata, marking as WIP

pkg/lib/config/saml.go
louischan-oursky: Since it is `rn:oasis:names:tc:SAML:1.1:nameid-format:unspecified`, I think `nameid` is one word.

tung2744: I see. Ok let me change it.

tung2744:
Updated, thanks!

tung2744:
I've marked here https://github.com/authgear/authgear-server/pull/4513#issuecomment-2288013004

pkg/lib/saml/error.go
```diff
@@ -0,0 +1,5 @@
+package saml
+
+import "errors"
+
+var ErrServiceProviderNotFound = errors.New("service provider not found")
```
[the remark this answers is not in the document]
tung2744: Actually in saml endpoints, we need to send the error to the acs url with a `<Response>`. So it is still incorrect to return a api error which returns a json in the current request.
I plan to handle this part in https://linear.app/authgear/issue/DEV-1727/%5Bsamlbrowsersso%5D-return-a-response-with-redirect-binding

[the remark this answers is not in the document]
louischan-oursky: Even with this metadata endpoint? Why not return a `<Response>` directly in this metadata endpoint?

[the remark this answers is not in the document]
tung2744: In metadata endpoint, ErrServiceProviderNotFound was catched and treated as not found error (404).
Usually, metadata endpoint should not have any error. The only error here is 404. Returning any other thing is incorrect.

louischan-oursky (approved):

PR #4584, opened 2024-08-13
Fix old UI crash by LDAP server

yyorkchan:
Now it does not crash
image


yyorkchan:
@louischan-oursky Ready Thanks

cychiuae (commented):
Thx

PR #4585, opened 2024-08-13
Use full redirect to visit /finish

louischan-oursky:
ref DEV-1793

### DIFF
--- a/pkg/auth/webapp/session_authflow.go
+++ b/pkg/auth/webapp/session_authflow.go
@@ -774,6 +774,19 @@ func (s *AuthflowScreenWithFlowResponse) Advance(p string, result *Result) {
 	result.RedirectURI = u.String()
 }
 
+// RedirectToFinish is a fix for https://linear.app/authgear/issue/DEV-1793/investigate-sign-in-directly-with-httpsaccountsportalauthgearcom-crash
+// We need Turbo to visit /finish with a full browser redirect,
+// so CSP and connect-src will not kick in.
+func (s *AuthflowScreenWithFlowResponse) RedirectToFinish(route string, result *Result) {
+	q := url.Values{}
+	q.Set(AuthflowQueryKey, s.Screen.StateToken.XStep)
+	u, _ := url.Parse(route)
+	u.RawQuery = q.Encode()
+
+	result.NavigationAction = "redirect"
+	result.RedirectURI = u.String()
+}
+
 func (s *AuthflowScreenWithFlowResponse) makeFallbackToSMSFromWhatsappRetryHandler(
 	inputFactory func(channel model.AuthenticatorOOBChannel) map[string]interface{},
 	channels []model.AuthenticatorOOBChannel,
--- a/pkg/auth/webapp/session_authflow.go
+++ b/pkg/auth/webapp/session_authflow.go
@@ -764,26 +764,26 @@ func (s *AuthflowScreenWithFlowResponse) Navigate(navigator Navigator, r *http.R
 	navigator.Navigate(s, r, webSessionID, result)
 }
 
-func (s *AuthflowScreenWithFlowResponse) Advance(p string, result *Result) {
-	q := url.Values{}
-	q.Set(AuthflowQueryKey, s.Screen.StateToken.XStep)
-	u, _ := url.Parse(p)
-	u.RawQuery = q.Encode()
-
+// Advance is for advancing to another page to drive the authflow.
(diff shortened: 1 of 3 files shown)


pkg/auth/webapp/session_authflow.go
```diff
@@ -774,6 +774,19 @@ func (s *AuthflowScreenWithFlowResponse) Advance(p string, result *Result) {
...
+// so CSP and connect-src will not kick in.
+func (s *AuthflowScreenWithFlowResponse) RedirectToFinish(route string, result *Result) {
+	q := url.Values{}
+	q.Set(AuthflowQueryKey, s.Screen.StateToken.XStep)
+	u, _ := url.Parse(route)
+	u.RawQuery = q.Encode()
+
+	result.NavigationAction = "redirect"
+	result.RedirectURI = u.String()
+}
```
tung2744, line 788: Sorry I still don't quite understand the cause. Why using `advance` will cause the problem but `redirect` will not?

pkg/auth/webapp/session_authflow.go
```diff
@@ -774,6 +774,19 @@ func (s *AuthflowScreenWithFlowResponse) Advance(p string, result *Result) {
...
 }
 
+// RedirectToFinish is a fix for https://linear.app/authgear/issue/DEV-1793/investigate-sign-in-directly-with-httpsaccountsportalauthgearcom-crash
+// We need Turbo to visit /finish with a full browser redirect,
+// so CSP and connect-src will not kick in.
+func (s *AuthflowScreenWithFlowResponse) RedirectToFinish(route string, result *Result) {
+	q := url.Values{}
+	q.Set(AuthflowQueryKey, s.Screen.StateToken.XStep)
+	u, _ := url.Parse(route)
+	u.RawQuery = q.Encode()
```
tung2744, line 784: And as this part shares the code with `Advance` and they are expected to be same, I think we better make a function to do it.

pkg/auth/webapp/session_authflow.go
louischan-oursky: If it is `advance`, then Turbo will visit `/finish` with XHR. The response of the XHR is 302 with a Location. According to MDN, CSP `connect-src` affects XHR. If Location is not allowed by `connect-src`, the redirect is blocked.

louischan-oursky: If it is `redirect`, then our script is programmed to visit the url with `window.location.href = THE_URL`. And typical HTTP redirect is not affected by CSP.

tung2744: I understand the issue after I read the notes you've written inside the issue.
So it is not `/authflow/v2/finish` itself have problem, but the redirect_uri which finally being redirected to was blocked.

louischan-oursky: Yes.

pkg/auth/webapp/session_authflow.go
```diff
@@ -774,6 +774,19 @@ func (s *AuthflowScreenWithFlowResponse) Advance(p string, result *Result) {
 	result.RedirectURI = u.String()
 }
 
+// RedirectToFinish is a fix for https://linear.app/authgear/issue/DEV-1793/investigate-sign-in-directly-with-httpsaccountsportalauthgearcom-crash
+// We need Turbo to visit /finish with a full browser redirect,
+// so CSP and connect-src will not kick in.
+func (s *AuthflowScreenWithFlowResponse) RedirectToFinish(route string, result *Result) {
```
tung2744, line 780: Should we remove `route` parameter from this function? Because it is always `/authflow/v2/finish`.

pkg/auth/webapp/session_authflow.go
louischan-oursky: OK. Let me change that.

pkg/auth/webapp/session_authflow.go
louischan-oursky: No, we have v1 finish and v2 finish.

tung2744: ok I see

louischan-oursky:
@tung2744 It seems that even I have configured authgear/authgear-server and louischan-oursky/authgear-server to have NPM_AUDIT_LEVEL, the build in this PR did not pick it up.

tung2744:
As this pr does not change any dependencies I think we can ignore the error in npm audit and merge it

tung2744 (approved):

PR #4588, opened 2024-08-14
Add test cases for admin api create user & change password

IniZio:
ref DEV-1646


e2e/var/authgear.yaml
```diff
@@ -130,11 +130,6 @@ test_mode:
     rules:
     - fixed_code: "111111"
       regex: .*
-  email:
```
louischan-oursky, line 133: I think you now need to suppress SMS, because you turn off DEV mode.

go.mod
```diff
@@ -99,6 +99,12 @@ require (
 	github.com/go-jose/go-jose/v3 v3.0.3
 )
 
+require (
+	github.com/asaskevich/EventBus v0.0.0-20200907212545-49d423059eef // indirect
+	github.com/emersion/go-sasl v0.0.0-20200509203442-7bfe0ed36a21 // indirect
+	github.com/emersion/go-smtp v0.21.3 // indirect
+)
+
```
louischan-oursky, line 107: These dependencies are used in e2e only. I think we should make `./e2e` a separate go module, like `./custombuild`.

Then this PR should make no changes to `<project_root>/go.mod`

e2e/pkg/testrunner/testcase.go
```diff
@@ -52,20 +50,20 @@ func (tc *TestCase) FullName() string {
...
 
 	// Create project per test case
-	err := cmd.CreateConfigSource()
+	appID, err := cmd.CreateConfigSource()
 	if err != nil {
 		t.Errorf("failed to create config source: %v", err)
 		return
 	}
 
+	cmd.AppID = appID
```
louischan-oursky, line 65: Why?

It looks to me that this should look like

```
cmd, err := NewEnd2EndCmd(NewEnd2EndCmdOptions{
  TestCase: *tc,
   Test: t,
})
if err != nil {
  t.Errorf("failed to create e2e cmd: %v", err)
  return
}

// And then you can use cmd.AppID
```

e2e/tests/admin_api/create_user_test.go
```diff
@@ -0,0 +1,150 @@
...
+	cmd.AppID = appID
+
+	userEmail := fmt.Sprintf("%s@example.com", appID)
+
+	client := e2eclient.NewClient(
+		ctx,
+		"localhost:4000",
+		"localhost:4002",
+		httputil.HTTPHost(fmt.Sprintf("%s.portal.localhost:4000", appID)),
+	)
```
louischan-oursky, line 40: Too much details to set up a test case. I expect

```
cmd, err := NewEnd2EndCmd(...)

cmd.Client.GraphQLAPI(...)
cmd.CheckEmail(...)

rows, err := cmd.QuerySQLSelectRaw(...) // rows is of type []map[string]interface{}
```

e2e/pkg/e2eclient/client.go
```diff
@@ -242,7 +242,9 @@ type GraphQLAPIRequest struct {
 
 type GraphQLResponse struct {
 	Data   json.RawMessage `json:"data"`
-	Errors []interface{}   `json:"errors"`
+	Errors []struct {
+		Message string `json:"message"`
```
louischan-oursky, line 246: You test against the public API. Message is not part of the public API. You assert `reason`.

louischan-oursky (changes requested):

e2e/Makefile
```diff
@@ -1,3 +1,10 @@
+.PHONY: ci
+ci:
+	@# Check if any `focus: true` is present in the test files
+	@# This is to prevent accidentally running only a subset of tests
+	@! grep -r -l -E 'focus\s*:\strue' tests || (echo "Remove 'focus: true' from test files above." && exit 1)
+	make run
+
```
pkong-ds, line 7: 👍 🥇 🙇

louischan-oursky:
@IniZio Build failed.

louischan-oursky (approved):

PR #4593, opened 2024-08-14
Support search user with ldap identity

yyorkchan:
Demo:
https://github.com/user-attachments/assets/af424908-b71f-4d95-884a-8f70f7bb0494

### DIFF
--- a/pkg/lib/elasticsearch/service.go
+++ b/pkg/lib/elasticsearch/service.go
@@ -13,8 +13,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/authgear/authgear-server/pkg/api/model"
-	identityloginid "github.com/authgear/authgear-server/pkg/lib/authn/identity/loginid"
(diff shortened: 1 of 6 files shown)


pkg/lib/elasticsearch/service.go
```diff
@@ -165,6 +150,24 @@ func (s *Service) getSource(userID string) (*model.ElasticsearchUserSource, acti
...
+		case model.IdentityTypeBiometric:
+			// No additional fields
+		case model.IdentityTypePasskey:
+			// No additional fields
+		case model.IdentityTypeSIWE:
+			// No additional fields
+		case model.IdentityTypeLDAP:
+			// No additional fields
+		default:
+			panic("identity: unknown identity type: " + identityInfo.Type)
```
cychiuae, line 169: I think the convention is prefix the panic message with the package name. Here is not `identity: `

louischan-oursky (approved):

PR #4601, opened 2024-08-15
Switch to use oursky fork of google-github-actions

louischan-oursky:
To avoid build error in oursky/authgear-server


tung2744:
Or we should whitelist `google-github-actions` so that we don't have to maintain it?

louischan-oursky:
Rxxx does not allow me to do so.

tung2744 (approved):

PR #4604, opened 2024-08-15
Fix logo alignment when add/remove logo

cychiuae:
https://github.com/user-attachments/assets/6bfe49ec-dce7-49f9-8ed1-16043792ac13


louischan-oursky:
I encountered another bug when testing this.

- Remove all logos and save
- Refresh
- Switch to English, upload light logo
- Switch to Chinese, upload light logo, but the logo is not uploaded.

cychiuae:
The problem you found seems only happens in Chrome.

louischan-oursky:
Maybe you add `key={lang}` to the component to make it remount?

louischan-oursky (approved):

PR #4618, opened 2024-08-19
Fix unexpected error when setting up passkey

aekt:
Ref: DEV-1783

### DIFF
--- a/pkg/lib/authn/identity/passkey/provider.go
+++ b/pkg/lib/authn/identity/passkey/provider.go
@@ -1,6 +1,7 @@
 package passkey
 
 import (
+	"fmt"
 	"sort"
 
 	"github.com/authgear/authgear-server/pkg/api/model"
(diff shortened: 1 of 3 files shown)


pkg/lib/authn/identity/passkey/provider.go
```diff
@@ -35,8 +34,8 @@ func (p *Provider) Get(userID, id string) (*identity.Passkey, error) {
 	return p.Store.Get(userID, id)
 }
 
-func (p *Provider) GetByAssertionResponse(assertionResponse []byte) (*identity.Passkey, error) {
-	credentialID, err := p.Passkey.GetCredentialIDFromAssertionResponse(assertionResponse)
+func (p *Provider) GetBySpec(spec *identity.PasskeySpec) (*identity.Passkey, error) {
+	_, credentialID, _, err := p.Passkey.PeekAttestationResponse(spec.AttestationResponse)
```
louischan-oursky, line 38: No, this is incorrect. Before the change, GetByAssertionResponse is called when the caller has a assertion response. If we rename it to GetBySpec, then it is expected that the caller can call GetBySpec with either a an attestation response, or a assertion response. We need to handle both case.

pkg/lib/feature/passkey/service.go
```diff
@@ -58,27 +58,6 @@ func (s *Service) ConsumeAttestationResponse(attestationResponse []byte) (err er
 	return
 }
 
-func (s *Service) GetCredentialIDFromAssertionResponse(assertionResponse []byte) (credentialID string, err error) {
```
louischan-oursky, line 61: This should be kept. You need it.

louischan-oursky (changes requested):

louischan-oursky (approved):

PR #4619, opened 2024-08-19
Add `/edit-config` to portal for direct edit on `authgear.yaml`

pkong-ds:
ref DEV-1419

## Preview 1 - Update success

https://github.com/user-attachments/assets/b6ee6aa2-ad54-4830-a615-21c9f5ce3109

## Preview 2 - Update error

https://github.com/user-attachments/assets/851c40db-98fb-49dd-a8fb-382385678031

### DIFF
--- a/pkg/portal/graphql/app.go
+++ b/pkg/portal/graphql/app.go
@@ -214,9 +214,7 @@ var nodeApp = node(
 					if argPaths, ok := p.Args["paths"]; ok {
 						for _, path := range argPaths.([]interface{}) {
 							path := path.(string)
-							if path == configsource.AuthgearYAML {
-								return nil, errors.New("direct access on authgear.yaml is disallowed")
-							}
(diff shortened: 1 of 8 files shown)


louischan-oursky (changes requested):
I discovered 1 problem. When I finish editing the config, and I switch to other screens, the config shown on other screens are stale. It is because App is not refetched after we have edited the config directly.

You can fix this by use the useConfig query (I dont remember the exact name), and call refetch after the config have been saved. This means `const form` and `const state` may no longer redundant.

pkong-ds:

pkong-ds:
hmm, `npm run test` pass locally & I did not change any password-related code 🤔 Can help retry?
https://github.com/authgear/authgear-server/actions/runs/10465349585/job/28980452492?pr=4619

louischan-oursky (approved):

PR #4620, opened 2024-08-19
Use std attrs normalizer to process ldap attributes

yyorkchan:


pkg/lib/authn/identity/ldap/provider.go
```diff
@@ -3,17 +3,26 @@ package ldap
...
+	"github.com/authgear/authgear-server/pkg/lib/ldap"
 	"github.com/authgear/authgear-server/pkg/util/clock"
 	"github.com/authgear/authgear-server/pkg/util/uuid"
 )
 
 var _ service.LDAPIdentityProvider = &Provider{}
 
+type StandardAttributesNormalizer interface {
+	Normalize(stdattrs.T) error
+}
```
cychiuae, line 20: We should not define `StandardAttributesNormalizer` interface here **again**.

pkg/lib/authn/identity/ldap/provider.go
```diff
@@ -100,3 +109,35 @@ func sortIdentities(is []*identity.LDAP) {
...
+		Type: model.IdentityTypeLDAP,
+		LDAP: &identity.LDAPSpec{
+			ServerName:           serverConfig.Name,
+			UserIDAttributeName:  userIDAttributeName,
+			UserIDAttributeValue: userIDAttributeValue,
+			Claims:               claims,
+			RawEntryJSON:         entry.ToJSON(),
+		},
+	}, nil
+}
```
cychiuae, line 143: Don't pollute ldap identity provider

cychiuae: From what I know, identity provider is dealing with identity

pkg/lib/authn/identity/ldap/provider.go
```diff
@@ -100,3 +109,35 @@ func sortIdentities(is []*identity.LDAP) {
...
+		claims[string(model.ClaimPhoneNumber)] = v
+	}
+	if v := entry.GetAttributeValue(ldap.AttributeUID.Name); v != "" {
+		claims[string(model.ClaimPreferredUsername)] = v
+	}
+
+	err := p.StandardAttributesNormalizer.Normalize(claims)
+	if err != nil {
+		claims = map[string]interface{}{}
+	}
```
cychiuae, line 131: I am not sure what we would want to handle here. Since ldap attributes are not guarantee, e.g. arbitrary value can be put in `email`. Would we want to normalise the attribute one by one or by all-or-nothing policy?

cychiuae: @louischan-oursky

pkg/lib/authn/identity/ldap/provider.go
louischan-oursky: Actually it is recommended to define it here again.

Local interfaces allows gomock to generate the mock, which make us able to write unit tests.

pkg/lib/authn/identity/ldap/provider.go
```diff
@@ -100,3 +109,35 @@ func sortIdentities(is []*identity.LDAP) {
 		return is[i].CreatedAt.Before(is[j].CreatedAt)
 	})
 }
+
+func (p *Provider) CreateNormalizedIdentitySpecFromLDAPEntry(serverConfig *config.LDAPServerConfig, entry *ldap.Entry) (*identity.Spec, error) {
```
louischan-oursky, line 113: Rename this to `MakeSpecFromEntry(serverConfig *config.LDAPServerConfig, entry *ldap.Entry) (*identity.Spec, error)`

pkg/lib/authn/identity/ldap/provider.go
louischan-oursky: You return `err` in this case. By returning error, we will not store anything malformed into the database. And we also not silence the error. Currently, this line silences the error, which is super subtle.

pkg/lib/authn/identity/ldap/provider.go
louischan-oursky: Actually I think this is a good place to put this code. We need a place to convert `*ldap.Entry` to a `*identity.Spec`. Since the conversion involves Authgear-specific logic (when we have attribute mapping), LDAP client is not a good candidate to host this code.

louischan-oursky (changes requested):

yyorkchan:
Updated Thanks @louischan-oursky

louischan-oursky (approved):

PR #4622, opened 2024-08-19
Add expire to access-events

louischan-oursky:
ref DEV-1852


pkg/lib/oauth/resolver.go
```diff
@@ -243,6 +243,7 @@ func (re *Resolver) accessOfflineGrant(offlineGrant *OfflineGrant, accessEvent a
 	if err != nil {
 		return nil, err
 	}
+	offlineGrant.ExpireAtForResolvedSession = expiry
```
tung2744, line 246: Looks like it could be easy for us to forget updating `ExpireAtForResolvedSession`.
Should we update `ExpireAtForResolvedSession` in `ComputeOfflineGrantExpiry`?

For idp session maybe we can do it in `computeSessionStorageExpiry`?

louischan-oursky:
@tung2744 Added 11 commits to refactor.

tung2744 (approved):

PR #4626, opened 2024-08-20
Add getUsersByStandardAttribute admin API

mickmak-busicat:
tag @louischan-oursky 

Preview:

https://github.com/user-attachments/assets/f4f88edb-330d-4285-9366-c815fbbc19cb

### DIFF
--- a/e2e/tests/admin_api/get_users_by_standard_attribute_test.go
+++ b/e2e/tests/admin_api/get_users_by_standard_attribute_test.go
@@ -1,0 +1,280 @@
+package adminapi
+
+import (
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+
+	"github.com/authgear/authgear-server/e2e/pkg/e2eclient"
+	"github.com/authgear/authgear-server/e2e/pkg/testrunner"
+)
+
+func TestGetUsersByStandardAttributeByEmail(t *testing.T) {
+	Convey("GetUsersByStandardAttributeByEmai", t, func() {
+		cmd, err := testrunner.NewEnd2EndCmd(testrunner.NewEnd2EndCmdOptions{
+			TestCase: &testrunner.TestCase{
+				Path: "admin_api/get_users_by_standard_attribute_test.go",
+			},
+			Test: t,
+		})
+		So(err, ShouldBeNil)
+
+		// Prepare testing user
+		err = cmd.ExecuteSQLInsertUpdateFile("../../tests/admin_api/get_users_admin_api_user.sql")
+		So(err, ShouldBeNil)
+
+		rawResp, err := cmd.Client.GraphQLAPIRaw(nil, nil, cmd.AppID, e2eclient.GraphQLAPIRequest{
+			Query: `
+				query {
+					getUsersByStandardAttribute(attributeName:"email", attributeValue: "e2e_admin_api_get_users@example.com") {
+							createdAt
+							standardAttributes
+							lastLoginAt
+					}
+				}
+	
+			`,
+		})
+		So(err, ShouldBeNil)
+		So(rawResp, ShouldEqualJSON, `
+			{
+				"data": {
+					"getUsersByStandardAttribute": [
+						{
+							"createdAt": "2024-08-27T07:51:42.040683Z",
+							"lastLoginAt": null,
+							"standardAttributes": {
+								"email": "e2e_admin_api_get_users@example.com",
+								"email_verified": false,
+								"updated_at": 1724745102
+							}
+						}
+					]
+				}
+			}
+		`)
+	})
+}
+func TestGetUsersByStandardAttributeByPhone(t *testing.T) {
+	Convey("GetUsersByStandardAttributeByPhone", t, func() {
+		cmd, err := testrunner.NewEnd2EndCmd(testrunner.NewEnd2EndCmdOptions{
+			TestCase: &testrunner.TestCase{
+				Path: "admin_api/get_users_by_standard_attribute_test.go",
+			},
+			Test: t,
+		})
+		So(err, ShouldBeNil)
+
+		// Prepare testing user
+		err = cmd.ExecuteSQLInsertUpdateFile("../../tests/admin_api/get_users_admin_api_user.sql")
+		So(err, ShouldBeNil)
+
+		rawResp, err := cmd.Client.GraphQLAPIRaw(nil, nil, cmd.AppID, e2eclient.GraphQLAPIRequest{
+			Query: `
+				query {
+					getUsersByStandardAttribute(attributeName:"phone_number", attributeValue: "+85261236544") {
+							createdAt
+							standardAttributes
+							lastLoginAt
+					}
+				}
+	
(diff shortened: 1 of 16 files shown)


pkg/admin/graphql/query.go
```diff
@@ -313,5 +314,33 @@ var query = graphql.NewObject(graphql.ObjectConfig{
 				return graphqlutil.NewConnectionFromResult(lazyItems, result)
 			},
 		},
+		"getUsersByStandardAttribute": &graphql.Field{
+			Description: "Get users by standardAttribute, attributeName must be email, phone_number or preferred_username.",
+			Type:        graphql.NewList(nodeUser),
+			Args: relay.NewConnectionArgs(graphql.FieldConfigArgument{
```
louischan-oursky, line 321: 1. Do not use relay.NewConnectionArgs. That will result in adding some extra arguments we do not need. You can see those extra arguments schema.graphql.

pkg/admin/graphql/query.go
```diff
@@ -313,5 +314,33 @@ var query = graphql.NewObject(graphql.ObjectConfig{
...
+				"attributeName": &graphql.ArgumentConfig{
+					Type: graphql.String,
```
louischan-oursky, line 323: It should be NonNull. The type in the generated schema should be `String!`, not `String`.

pkg/admin/graphql/query.go
```diff
@@ -333,7 +334,7 @@ var query = graphql.NewObject(graphql.ObjectConfig{
 
 				userIDs, err := gqlCtx.UserFacade.GetUsersByStandardAttribute(attributeName, attributeValue)
 				if err != nil {
-					return nil, err
+					return nil, api.ErrGetUsersInvalidArgument.New(err.Error())
```
louischan-oursky, line 337: Do not use api.ErrGetUsersInvalidArgument here. It is because the error could be something else, like I/O error. Just `return nil, err` here.

pkg/lib/facade/coordinator.go
```diff
@@ -1361,7 +1364,24 @@ func (c *Coordinator) MFAListRecoveryCodes(userID string) ([]*mfa.RecoveryCode,
...
+	var loginIDKeyType model.LoginIDKeyType
+	switch attributeName {
+	case stdattrs.Email:
+		loginIDKeyType = model.LoginIDKeyTypeEmail
+	case stdattrs.PhoneNumber:
+		loginIDKeyType = model.LoginIDKeyTypePhone
+	case stdattrs.PreferredUsername:
+		loginIDKeyType = model.LoginIDKeyTypeUsername
+	default:
+		return nil, fmt.Errorf("attributeName must be email, phone_number or preferred_username.")
```
louischan-oursky, line 1376: `return nil, api.ErrGetUsersInvalidArgument.New("attributeName must be email, phone_number or preferred_username.")`

We should do this here because it is the place where we know attributeName is invalid.

pkg/lib/facade/coordinator.go
```diff
@@ -1361,7 +1364,24 @@ func (c *Coordinator) MFAListRecoveryCodes(userID string) ([]*mfa.RecoveryCode,
...
+	}
+
+	normalized, _, err := c.Identities.Normalize(loginIDKeyType, attributeValue)
+	if err != nil {
+		return nil, fmt.Errorf("Invalid attributeValue.")
```
louischan-oursky, line 1381: Since Normalize does not do IO, the error it returns must be error about normalization. So we can write

```
return nil, api.ErrGetUsersInvalidArgument.New("invalid attributeValue")
```

e2e/tests/admin_api/get_users_by_standard_attribute_test.go
```diff
@@ -0,0 +1,290 @@
...
+	var doc map[string]any
+	if err := json.Unmarshal(resp.Data, &doc); err != nil {
+		panic(err)
+	}
+
+	result := doc["getUsersByStandardAttribute"].([]interface{})
+
+	if len(result) == 0 {
+		t.Fatal("result should not be empty")
+	}
```
louischan-oursky, line 56: I prefer we use convey in this test too.

Let me come to you and show you.

e2e/tests/admin_api/get_users_by_standard_attribute_test.go
```diff
@@ -0,0 +1,290 @@
...
+		t.Fatalf("expect error, got nil")
+	}
+
+	for _, err := range resp.Errors {
+		if err.Extensions.Reason == "GetUsersInvalidArgument" {
+			return
+		}
+	}
+
+	t.Fatalf("expect error GetUsersInvalidArgument, got %v", resp.Errors)
```
louischan-oursky, line 244: We should use ShouldEqualJSON to match the actual response.

louischan-oursky (changes requested):

pkg/lib/facade/coordinator.go
mickmak-busicat: moved api invalid error inside coordinator

pkg/lib/facade/coordinator.go
mickmak-busicat: moved api invalid error inside coordinator

pkg/admin/graphql/query.go
mickmak-busicat: moved api invalid error inside coordinator

pkg/admin/graphql/query.go
mickmak-busicat: fixed query args type

pkg/admin/graphql/query.go
mickmak-busicat: fixed query args type

louischan-oursky:
Conflicted and build failed.

e2e/tests/admin_api/get_users_by_standard_attribute_test.go
mickmak-busicat: updated to use Convey

e2e/tests/admin_api/get_users_by_standard_attribute_test.go
mickmak-busicat: updated to use `So` to assert value

mickmak-busicat:
@louischan-oursky rebased 🙏
but ci was failed in portal test which should be no change in this pr, i think it is fixed by rebasing?

louischan-oursky:
Thank you. Next time, do not fix up the commits. Appending commits is easier for reviewing. Of course, fixing up will make the commit history looks much better, but we can always do this after the review and just before merging.

louischan-oursky:
I added 2 commits to fix the build.

louischan-oursky:
Added one more commit to a crash.

The previous code assumes the identity is an Login ID identity `v.LoginID.UserID`, but it is not necessarily the case. And we can just use `v.UserID`.

louischan-oursky (approved):

mickmak-busicat:
Thanks, will do next time

PR #4628, opened 2024-08-20
Implement missing milestones on inspect device token

pkong-ds:
## Preview
Pending

### DIFF
--- a/e2e/.env
+++ b/e2e/.env
@@ -3,7 +3,11 @@ LOG_LEVEL=debug
 TRUST_PROXY=true
 AUTHGEAR_APP_ID=accounts
 AUTHGEAR_CLIENT_ID=portal
-AUTHGEAR_ENDPOINT=http://accounts.portal.localhost:3100
+# Only portal reads this envvar.
+# But we do not test portal in e2e test,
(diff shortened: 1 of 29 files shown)


louischan-oursky:
Maybe we can add e2e tests for this.

pkong-ds:
@louischan-oursky The cookie jar somehow not working. Problems I encountered

1. cookie set in jar, but request unable to get cookie, even if same URL
2. Unable to use `http.Client` cookies feature like `auto-add cookie to request` and `auto-set cookie on receive Response with "Set-Coookie" header`

pkong-ds:
@louischan-oursky thanks a lot for the instruction on `implementing my own Jar`

Found out the problem was `127.0.0.1` host did not match domain `portal.locahost`, which threw `cookiejar: illegal cookie domain attribute` error. 

I fixed it by
- forking official cookie jar to e2e test pkgs
- https://github.com/authgear/authgear-server/pull/4628/commits/6dd3cf1eb99d4a13d2ce562148f0b230832e4754

🙏

louischan-oursky (changes requested):

louischan-oursky (approved):

PR #4630, opened 2024-08-20
SAML: Finish SSO flow

tung2744:
ref DEV-1727

Same as before, you can use this site to test:
https://sptest.iamshowcase.com/authnrequest

If you want to test `IsPassive`, you need the tools here:
https://www.samltool.com/encode.php

Write a saml request manually, and use the encoder above to generate the `SAMLRequest` query parameter, and send it to `http://localhost:3000/saml2/login/{your_sp_id}?SAMLRequest={your_encoded_request}`.

Here is an example request:

```xml
<samlp:AuthnRequest
  xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
  xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion"
  ForceAuthn="false"
  IsPassive="true"
  ID="a1b83a399445569c760a9f733cce58d7aec680316"
  IssueInstant="2024-08-20T11:10:25Z"
  Destination="http://localhost:3000/saml2/login/sp1"
  AssertionConsumerServiceURL="https://sptest.iamshowcase.com/acs"
  ProtocolBinding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"
  Version="2.0"
  ><saml:Issuer>IAMShowcase</saml:Issuer></samlp:AuthnRequest
>
```

### DIFF
--- a/pkg/auth/handler/saml/login_finish.go
+++ b/pkg/auth/handler/saml/login_finish.go
@@ -3,8 +3,14 @@ package saml
 import (
 	"net/http"
 
+	"github.com/authgear/authgear-server/pkg/lib/authn/authenticationinfo"
+	"github.com/authgear/authgear-server/pkg/lib/saml/samlprotocol"
+	"github.com/authgear/authgear-server/pkg/lib/saml/samlprotocol/samlprotocolhttp"
+	"github.com/authgear/authgear-server/pkg/lib/saml/samlsession"
 	"github.com/authgear/authgear-server/pkg/util/httproute"
+	"github.com/authgear/authgear-server/pkg/util/httputil"
 	"github.com/authgear/authgear-server/pkg/util/log"
+	"github.com/authgear/authgear-server/pkg/util/panicutil"
 )
 
 func ConfigureLoginFinishRoute(route httproute.Route) httproute.Route {
@@ -20,9 +26,71 @@ func NewLoginFinishHandlerLogger(lf *log.Factory) *LoginFinishHandlerLogger {
 }
 
 type LoginFinishHandler struct {
-	Logger *LoginFinishHandlerLogger
+	Logger                     *LoginFinishHandlerLogger
+	SAMLService                HandlerSAMLService
+	SAMLSessionService         SAMLSessionService
+	AuthenticationInfoResolver SAMLAuthenticationInfoResolver
+	AuthenticationInfoService  SAMLAuthenticationInfoService
 }
 
 func (h *LoginFinishHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	authInfoID, ok := h.AuthenticationInfoResolver.GetAuthenticationInfoID(r)
+	if !ok {
+		h.Logger.Warningln("authentication info id is missing")
+		http.NotFound(rw, r)
+		return
+	}
 
+	authInfo, err := h.AuthenticationInfoService.Get(authInfoID)
+	if err != nil {
+		// It is unexpected that we've set the code in query, but turns out it does not exist.
+		// We have no idea how to return the error response to user yet,
+		// so we can only panic.
+		panic(err)
+	}
+
+	samlSession, err := h.SAMLSessionService.Get(authInfo.SAMLSessionID)
+	if err != nil {
+		// It is unexpected that we've set the session id in auth info, but turns out it does not exist.
+		// We have no idea how to return the error response to user yet,
+		// so we can only panic.
+		panic(err)
+	}
+
+	result := h.handleLoginResult(authInfo, samlSession)
+	result.WriteResponse(rw, r)
+}
+
+func (h *LoginFinishHandler) handleLoginResult(
+	authInfo *authenticationinfo.Entry,
+	samlSession *samlsession.SAMLSession,
+) (result httputil.Result) {
+	defer func() {
+		if e := recover(); e != nil {
+			e := panicutil.MakeError(e)
+			h.Logger.WithError(e).Error("panic")
+			// TODO(saml): Return a error
+			panic(e)
+		}
+	}()
+
+	authnRequest := samlSession.Entry.AuthnRequest()
+	authenticatedUserID := authInfo.T.UserID
+
+	resp, err := h.SAMLService.IssueSuccessResponse(
+		samlSession.Entry.ServiceProviderID,
+		authenticatedUserID,
+		authnRequest,
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	return &samlprotocolhttp.SAMLResult{
+		CallbackURL: samlSession.Entry.CallbackURL,
+		// TODO(saml): Respect the binding protocol set in request
+		Binding:    samlprotocol.SAMLBindingHTTPPost,
+		Response:   resp,
+		RelayState: samlSession.Entry.RelayState,
+	}
 }
--- a/pkg/auth/handler/saml/login_finish.go
+++ b/pkg/auth/handler/saml/login_finish.go
@@ -77,7 +77,7 @@ func (h *LoginFinishHandler) handleLoginResult(
 			result = samlprotocolhttp.NewSAMLErrorResult(e,
 				samlprotocolhttp.SAMLResult{
 					CallbackURL: callbackURL,
-					Response:    samlprotocol.NewInternalServerErrorResponse(now),
+					Response:    samlprotocol.NewInternalServerErrorResponse(now, h.SAMLService.IdpEntityID()),
 					RelayState:  relayState,
 				},
 			)
@@ -88,6 +88,7 @@ func (h *LoginFinishHandler) handleLoginResult(
 	authenticatedUserID := authInfo.T.UserID
 
 	resp, err := h.SAMLService.IssueSuccessResponse(
+		callbackURL,
 		samlSession.Entry.ServiceProviderID,
 		authenticatedUserID,
 		authnRequest,
--- a/pkg/auth/handler/saml/login_finish.go
+++ b/pkg/auth/handler/saml/login_finish.go
@@ -1,0 +1,28 @@
+package saml
+
+import (
+	"net/http"
+
+	"github.com/authgear/authgear-server/pkg/util/httproute"
+	"github.com/authgear/authgear-server/pkg/util/log"
+)
+
+func ConfigureLoginFinishRoute(route httproute.Route) httproute.Route {
+	return route.
+		WithMethods("GET").
+		WithPathPattern("/saml2/login_finish")
+}
+
+type LoginFinishHandlerLogger struct{ *log.Logger }
+
+func NewLoginFinishHandlerLogger(lf *log.Factory) *LoginFinishHandlerLogger {
+	return &LoginFinishHandlerLogger{lf.New("saml-login-finish-handler")}
+}
+
+type LoginFinishHandler struct {
+	Logger *LoginFinishHandlerLogger
+}
+
+func (h *LoginFinishHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+
+}
--- a/pkg/auth/handler/saml/login_finish.go
+++ b/pkg/auth/handler/saml/login_finish.go
@@ -1,20 +1,12 @@
 package saml
 
 import (
-	"errors"
 	"net/http"
 
-	"github.com/authgear/authgear-server/pkg/lib/authn/authenticationinfo"
-	"github.com/authgear/authgear-server/pkg/lib/infra/db/appdb"
-	"github.com/authgear/authgear-server/pkg/lib/saml/samlerror"
-	"github.com/authgear/authgear-server/pkg/lib/saml/samlprotocol"
 	"github.com/authgear/authgear-server/pkg/lib/saml/samlprotocol/samlprotocolhttp"
-	"github.com/authgear/authgear-server/pkg/lib/saml/samlsession"
 	"github.com/authgear/authgear-server/pkg/util/clock"
 	"github.com/authgear/authgear-server/pkg/util/httproute"
-	"github.com/authgear/authgear-server/pkg/util/httputil"
 	"github.com/authgear/authgear-server/pkg/util/log"
-	"github.com/authgear/authgear-server/pkg/util/panicutil"
 )
 
 func ConfigureLoginFinishRoute(route httproute.Route) httproute.Route {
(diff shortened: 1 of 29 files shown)


pkg/auth/handler/saml/login_finish.go
```diff
@@ -20,9 +26,71 @@ func NewLoginFinishHandlerLogger(lf *log.Factory) *LoginFinishHandlerLogger {
...
+	SAMLSessionService         SAMLSessionService
+	AuthenticationInfoResolver SAMLAuthenticationInfoResolver
+	AuthenticationInfoService  SAMLAuthenticationInfoService
 }
 
 func (h *LoginFinishHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	authInfoID, ok := h.AuthenticationInfoResolver.GetAuthenticationInfoID(r)
+	if !ok {
+		h.Logger.Warningln("authentication info id is missing")
+		http.NotFound(rw, r)
```
louischan-oursky, line 38: Returning 404 does not seem appropriate here. I think we do not expect people to invoke this endpoint directly. Maybe `http.Error(w, "invoking this endpoint directly is not supported", http.StatusBadRequest)`?

pkg/auth/handler/saml/login_finish.go
```diff
@@ -20,9 +26,71 @@ func NewLoginFinishHandlerLogger(lf *log.Factory) *LoginFinishHandlerLogger {
...
+	samlSession *samlsession.SAMLSession,
+) (result httputil.Result) {
+	defer func() {
+		if e := recover(); e != nil {
+			e := panicutil.MakeError(e)
+			h.Logger.WithError(e).Error("panic")
+			// TODO(saml): Return a error
+			panic(e)
+		}
+	}()
```
louischan-oursky: It is better not to recover() here. Instead

```
var err error
defer func() {
  if err != nil {
    // Return a SAML error to ACS URL
  }
}()
```

pkg/auth/handler/saml/login_finish.go
```diff
@@ -20,9 +26,71 @@ func NewLoginFinishHandlerLogger(lf *log.Factory) *LoginFinishHandlerLogger {
...
+	authnRequest := samlSession.Entry.AuthnRequest()
+	authenticatedUserID := authInfo.T.UserID
+
+	resp, err := h.SAMLService.IssueSuccessResponse(
+		samlSession.Entry.ServiceProviderID,
+		authenticatedUserID,
+		authnRequest,
+	)
+	if err != nil {
+		panic(err)
```
louischan-oursky: And do not panic here.

pkg/lib/saml/samlprotocol/const.go
```diff
@@ -29,3 +29,15 @@ const (
...
+const (
+	SAMLIssertFormatEntity = "urn:oasis:names:tc:SAML:2.0:nameid-format:entity"
+)
+
+const (
+	SAMLAttrnameFormatBasic = "urn:oasis:names:tc:SAML:2.0:attrname-format:basic"
+)
+
+const (
+	SAMLAttrTypeString = "xs:string"
```
louischan-oursky, line 42: `xs:` seems to be a XML namespace, which depends on how the namespace is named at the root of XML document. Does crewsaml exposes this constant so that we do not need to repeat it here? If not, we better document this hack, and saying upgrading crewsaml may break our code.

pkg/lib/config/saml.go
```diff
@@ -134,6 +138,15 @@ var _ = Schema.Add("SAMLSigningSignatureMethod", `
 
 type SAMLSigningSignatureMethod string
 
+func (m SAMLSigningSignatureMethod) ToDsigSignatureMethod() string {
+	switch m {
+	case SAMLSigningSignatureMethodRSASHA256:
+		return dsig.RSASHA256SignatureMethod
```
louischan-oursky, line 157: This library already uses the correct constant. https://github.com/russellhaering/goxmldsig/blob/main/xml_constants.go#L46C2-L46C26

And I discovered they call it `signature method` instead of `signature algorithm`. So in our config, we should make it like

```
signature_method: "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"
```

---


Rename it back to `signature_method`, and correct the enum values.

pkg/lib/saml/service.go
```diff
@@ -304,7 +308,65 @@ func (s *Service) IssueSuccessResponse(
...
+
+	// Sign the response
+	responseEl := response.Element()
+	signedResponseEl, err := signingContext.SignEnveloped(responseEl)
+	if err != nil {
+		return nil, err
+	}
+
+	responseSigEl := signedResponseEl.ChildElements()[len(signedResponseEl.ChildElements())-1]
+	response.Signature = responseSigEl
```
louischan-oursky, line 373: This code is worth being hosted in its own method. Consider adding some tests for this method.

```
signature, err := s.SignResponse(response)
if err != nil {
  return nil, err
}
response.Signature = signature.

func (s *Service) SignResponse(response) (Signature, error)
```

pkg/lib/saml/service.go
```diff
@@ -304,7 +308,65 @@ func (s *Service) IssueSuccessResponse(
...
+	err := activeCert.Key.Raw(&rsaPrivateKey)
+	if err != nil {
+		panic(err)
+	}
+
+	keyPair := tls.Certificate{
+		Certificate: [][]byte{activeCert.Certificate.Data()},
+		PrivateKey:  &rsaPrivateKey,
+		Leaf:        activeCert.Certificate.X509Certificate(),
+	}
```
louischan-oursky, line 452: Instead of using TLSCertKeyStore, we can borrow MemoryX509KeyStore from https://github.com/russellhaering/goxmldsig/blob/v1.4.0/keystore.go#L31

Using TLS here does not seem appropriate, we are not doing anything related to TLS here.

pkg/lib/saml/samlprotocol/samlprotocolhttp/result.go
```diff
@@ -36,7 +36,8 @@ func (s *SAMLResult) IsInternalError() bool {
 
 type SAMLErrorResult struct {
 	SAMLResult
-	Cause error
+	Cause        error
+	IsUnexpected bool
```
louischan-oursky, line 40: I think we should use `errors.As` to check if it is a SAML error. If it is, then it is an expected error, otherwise, it is unexpected. So remove this field.

pkg/lib/saml/service.go
```diff
@@ -227,18 +227,23 @@ func (s *Service) IssueSuccessResponse(
...
 	recipient := callbackURL
+	if sp.Recipient != "" {
+		recipient = sp.Recipient
+	}
 
-	// TODO(saml): Use configured audience if configured
 	audience := callbackURL
+	if sp.Audience != "" {
+		audience = sp.Audience
+	}
```
louischan-oursky, line 254: Should the field in the request override the config? Can you leave a reference to the spec here to document what we should do here?

louischan-oursky (changes requested):

pkg/lib/saml/service.go
tung2744: https://github.com/authgear/authgear-server/pull/4513#discussion_r1723097475

Config should override the ACSURL, this is to allow different values in destination, recipient and audience.

pkg/lib/saml/samlprotocol/samlprotocolhttp/result.go
tung2744: As discussed before, errors from service layer (i.e. SAMLError) does not determine the the error in handler layer.
The same saml error could have different meaning in different handler.
For example, if we got `service_provider_not_found` in login or metadata, it is expected error, but if we got it in login_finish, it is unexpected error (because the service provider id is from session).
So only the handler should translate any error to expected to unexpected error.

pkg/lib/saml/service.go
louischan-oursky: Which section of the spec specifies this. Include the section here as reference for future reader.

pkg/lib/saml/service.go
tung2744: I copied the implementation from https://github.com/crewjam/saml/blob/main/identity_provider.go#L1073-L1078

Maybe let me try to use MemoryX509KeyStore and see if it works.

pkg/lib/saml/samlprotocol/samlprotocolhttp/result.go
louischan-oursky: Then I suggest we replace `NewSAMLErrorResult` with `NewExpectedSAMLErrorResult` and `NewUnExpectedSAMLErrorResult`. The positional argument of `true, false` is very easy to miss.

pkg/lib/config/saml.go
tung2744: ok. Just to note:
There are two thing in xmldsig, the signing method and digest method. Our config now controls both.

pkg/lib/saml/samlprotocol/const.go
tung2744: `xs:string` is used to represent the type string in xml schema. Which I don't think is related to the crewsaml implementation.

However I don't know which spec exactly defined this, I only know it is commonly used.
The only related spec I found is https://www.w3.org/TR/xmlschema-2/

pkg/auth/handler/saml/login_finish.go
tung2744: Any internal server error should be converted to an error response with status `StatusResponder`. So I add recover here.

pkg/auth/handler/saml/login_finish.go
tung2744: It is an unexpected error here so I panic here.

pkg/lib/saml/samlprotocol/const.go
louischan-oursky: Copy https://github.com/crewjam/saml/blob/main/schema.go#L502

pkg/lib/saml/service.go
louischan-oursky: Use NewDefaultSigningContext, and call SetSignatureMethod

pkg/lib/config/testdata/config_tests.yaml
```diff
@@ -1184,7 +1184,10 @@ config:
         nameid_format: urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress
         nameid_attribute_pointer: /preferred_username
         acs_urls:
-          - http://localhost:3000/saml-test
+          - http://localhost:3000/acs
+        destination: http://localhost:3000/destination
```
louischan-oursky, line 1188:
```
acs_urls:
- asc_url: http://localhost:3000/acs
- asc_url: http://localhost:3000/acs
   destination: http://localhost:3000/destination
```

pkg/lib/saml/service.go
louischan-oursky: Document this behavior is borrowed from another implementation.

pkg/lib/saml/service.go
louischan-oursky: Replace `SignEnveloped` with `ConstructSignature(response, true)`. Do not need to manipulate the child.

pkg/lib/saml/service.go
```diff
@@ -196,3 +210,256 @@ func (s *Service) ValidateAuthnRequest(serviceProviderId string, authnRequest *s
...
+						}},
+					},
+				},
+			},
+		},
+	}
+
+	response.Assertion = assertion
+
+	// Sign the assertion
```
louischan-oursky, line 348:
```
assertionSig, err := s.SignElement(response.Assertion)
if err != nil {
  return nil, err
}
response.Assertion.Sigatnure = assertionSig

responseSig, err := s.SignElement(response)
if err != nil {
  return nil, err
}
response.Sigatnure = responseSig
```

louischan-oursky (changes requested):

pkg/lib/saml/service.go
tung2744: I found that `MemoryX509KeyStore` cannot be constructed. I will keep using tls keystore here.

tung2744: Or maybe I create a struct which implements x509KeyStore

louischan-oursky: > maybe I create a struct which implements x509KeyStore

Yes. You cannot use `MemoryX509KeyStore` from the library, its fields are all private. But you can copy it very easily. It is just a few lines of code.

pkg/lib/config/testdata/config_tests.yaml
tung2744: @louischan-oursky 
After a second thought, if we say allowing multiple acs_url is for allowing the response to be sent through proxies, it seems doesn't quite make sense if a single service provider expect different `destination` value when using different acs_url, as the response should finally reach the same server. So I prefer to keep the current config structure. 

```
  ACSUrl                         Destination
https://proxy1.example.com --> https://example.authgear.cloud
https://proxy2.example.com --> https://example.authgear.cloud
```

louischan-oursky: If we continue this argument, we can say if the developer wants to have a different ASC_URL, they can create another SP.

louischan-oursky: I remember we talked about another way of supporting a SP. That is by setting a URL to the metadata of a SP. Therefore, I think this SP config should mirror the structure of the metadata of a SP. So if the metadata of a SP supports multiple occurence of ACS_URL, then this should be an array. What do you think?

tung2744: Yes. So acs_url is an array, but destination is not?

louischan-oursky: 1. For `recipient`, it is ASC_URL by default. If the config has set `recipient`, use config.
2. For `destination`, it is ASC_URL by default. If the config has set `destination`, use config.
3. For `audience`
  - Always include ACS_URL as one audience.
  - If AuthnRequest has audience, check if it is ASC_URL, or `audience` in config. If violated, reject the request.
  - If config has `audience`, output it.

louischan-oursky: Please put this into our spec as well.

tung2744:
The signature validation succeed if I put the raw base64 encoded string into the tool.
Not sure why decoded xml does not work.

pkg/lib/config/testdata/config_tests.yaml
tung2744: I found that the library we are using does not allow us to put multiple audience into it, we may need to define our own response struct.

tung2744: @louischan-oursky I would like to copy the whole `schema.go` from `crewjamsaml` into our codebase, so we can modify it easily, what do you think?

louischan-oursky: Feel free to do it.

louischan-oursky (approved):

PR #4635, opened 2024-08-21
Fix change password in settings page leads to crash

cychiuae:


louischan-oursky (approved):
@cychiuae I added a few commits to improve things.

cychiuae:
THanks!

PR #4643, opened 2024-08-22
Make sure createIdentity will create OOB OTP authenticator

louischan-oursky:
ref DEV-1880

### DIFF
--- a/e2e/tests/admin_api/create_identity_test.go
+++ b/e2e/tests/admin_api/create_identity_test.go
@@ -1,0 +1,359 @@
+package adminapi
+
+import (
+	"encoding/json"
+	"fmt"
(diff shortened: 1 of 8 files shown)


louischan-oursky:
@tung2744 Please review this. This is quite urgent.

tung2744 (approved):
Do we know what causes the regression?

louischan-oursky:
The old createIdentity is based on interaction. createIdentity will only create the OOBOTP authenticator when the user does not have any primary authenticator. But the created user has a password, so the OOB OTP authenticator is not created.

I rewrite createIdentity and make it behavior more predictable. If OOB OTP is enabled, then the authenticator is always created along with the identity.

PR #4646, opened 2024-08-23
Add endpoint direct access configs in authgear.yaml

toml5566:
Ref DEV-1820

Add default_brand_page_redirect_uri in authgear.yaml UIConfig


louischan-oursky:
Let's open a PR for the whole feature. This change is too small.

PR #4647, opened 2024-08-23
Fix Auth UI v2 font styles

cychiuae:


cychiuae:
@louischan-oursky @tung2744 @IniZio @pkong-ds

authui/src/authflowv2/components/phone-input.css
```diff
@@ -63,7 +63,7 @@
 
   .phone-input__country-flag {
     @apply inline-block align-middle ltr:mr-1 rtl:ml-1;
-    height: 1em;
-    width: 1em;
+    height: 1rem;
+    width: 1rem;
```
cychiuae, line 67: Change to use w-6 and h-6

authui/src/authflowv2/components/select.css
```diff
@@ -75,8 +75,8 @@
     --select__empty-font-weight: var(--typography-body-large-font-weight);
 
     @media (min-width: theme("screens.tablet")) {
-      --select__options-height: 18em;
-      --select__dropdown-width: 21.5625em;
+      --select__options-height: 18rem;
```
cychiuae, line 78: Rename to max height

authui/src/authflowv2/components/select.css
```diff
@@ -75,8 +75,8 @@
...
+      --select__dropdown-width: 21.5625rem;
```
cychiuae, line 79: 22rem looks better

PR #4640, opened 2024-08-22
SettingV2 page home

yyorkchan:
Demo:

https://github.com/user-attachments/assets/d3e9db93-1f05-4173-ad12-e1e9398cdabd

PR #4662, opened 2024-08-27
[DEPS #4640] Implement authui v2 navbar

cychiuae:
based on #4640


cychiuae:
@yyorkchan @Kuixz Please take a look at the last commit i made

louischan-oursky (approved):

resources/authgear/templates/en/web/authflowv2/__navbar.html
```diff
@@ -0,0 +1,17 @@
...
+  <a
+    class="navbar__item inline-flex items-center justify-center"
+    title="{{ .BackTitle }}"
+    href="{{ .BackHref }}"
+    data-turbo-action="replace"
+  >
+    <i class="navbar__item_icon material-icons">
+      arrow_back_ios
+    </i>
+    <span class="inline-block ml-3">{{ .BackTitle }}</span>
```
louischan-oursky, line 14: We need to use `ltr:ml-3 rtl:mr-3`. Or you use gap

@yyorkchan @cychiuae We need to pay attention to the left/right usage. If you use explicit left/right class, then you probably need to write `ltr:` and `rtl:`

louischan-oursky (changes requested):

louischan-oursky (approved):

PR #4664, opened 2024-08-27
Fix pipes nested in templates not traversed in validateHTMLTemplate

pkong-ds:
cc @kiootic or was it intentional not to traverse into `TemplateNode.Pipe`?


louischan-oursky:
@pkong-ds Please add a comment to ignore the lint error.

pkong-ds:
Updated! `make lint` now pass locally :pray

louischan-oursky (approved):

PR #4678, opened 2024-09-02
Fix custom select dropdown

cychiuae:
The original problem can be reproduced by configuring Authgear to only allow use phone number as login id and allow user to select many countries (I enables all countries in my demo). Basically make the login page to only contain the phone input without other options

### Before

https://github.com/user-attachments/assets/705d6244-f4ba-4918-afd1-001d4573b158

### After

https://github.com/user-attachments/assets/a579ec63-8c69-46ab-9b8b-ae32d2989d34

### DIFF
--- a/authui/src/authflowv2/customSelect.ts
+++ b/authui/src/authflowv2/customSelect.ts
@@ -13,6 +13,7 @@ export class CustomSelectController extends Controller {
   static targets = [
     "input",
     "trigger",
+    "dropdownContainer",
     "dropdown",
     "search",
     "clearSearch",
@@ -29,6 +30,7 @@ export class CustomSelectController extends Controller {
   declare readonly inputTarget: HTMLInputElement;
   declare readonly triggerTarget: HTMLButtonElement;
   declare readonly dropdownTarget: HTMLElement;
+  declare readonly dropdownContainerTarget: HTMLElement;
   declare readonly searchTarget: HTMLInputElement;
   declare readonly clearSearchTarget: HTMLElement;
   declare readonly optionsTarget: HTMLElement;
@@ -66,7 +68,7 @@ export class CustomSelectController extends Controller {
   }
 
   get isOpen(): boolean {
-    return !this.dropdownTarget.classList.contains("hidden");
+    return !this.dropdownContainerTarget.classList.contains("hidden");
   }
 
   get value() {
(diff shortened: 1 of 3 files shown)


resources/authgear/templates/en/web/authflowv2/__phone_input.html
```diff
@@ -43,70 +43,75 @@
...
-        <li class="select__empty">
-          {{ template "v2.component.phone-input.default.no-results-found" }}
-        </li>
-      </template>
+        <template data-custom-select-target="emptyTemplate">
+          <li class="select__empty">
+            {{ template "v2.component.phone-input.default.no-results-found" }}
+          </li>
+        </template>
+      </div>
```
cychiuae, line 114: Here I just wrap the original dropdown with a `div`. You can hide whitespace change to view the real change

authui/src/authflowv2/customSelect.ts
```diff
@@ -147,11 +149,10 @@ export class CustomSelectController extends Controller {
   }
 
   close() {
-    if (this.dropdownTarget.classList.contains("hidden")) return;
+    if (this.dropdownContainerTarget.classList.contains("hidden")) return;
 
-    this.dropdownTarget.classList.add("hidden");
+    this.dropdownContainerTarget.classList.add("hidden");
     this.triggerTarget.setAttribute("aria-expanded", "false");
-    this.triggerTarget.focus();
```
cychiuae, line 154: @IniZio 

Any reason for us to focus the trigger target when closed the dropdown?

cychiuae:
cc @louischan-oursky @pkong-ds

authui/src/authflowv2/customSelect.ts
IniZio: Wanted to refocus the trigger after user just selected an item in the dropdown.

cychiuae: I know but what is it for? The reason I tried to remove this focus action is because when I click to phone input box to close the dropdown, it immediate blurs the phone input because here called to focus the trigger.

IniZio: But html select should have similar behaviour? 

If you have an html select and input in same form, clicking the input with select opened will not focus the input immediately.

louischan-oursky (changes requested):
Per offline discussion, we will drop the commit of "fixing focus"

cychiuae:
@louischan-oursky Updated

louischan-oursky (approved):

PR #4680, opened 2024-09-02
Fix ldap translation key

pkong-ds:
- Fixed `v2-ldap-branching` not renamed to new format
- Renamed ldap keys


louischan-oursky:
@pkong-ds I added a commit to fix.

louischan-oursky (approved):

PR #4683, opened 2024-09-03
Fix password generator doesnt have default character set when requirement is enabled

mickmak-busicat:
Preview:

## Password generate in portal

https://github.com/user-attachments/assets/ae196e04-d19d-4b07-be09-b611c2c576ce

## Auto generate password

https://github.com/user-attachments/assets/d6cfe584-a9c3-4c2b-a518-b84f7365e464

### DIFF
--- a/pkg/lib/authn/authenticator/password/generator.go
+++ b/pkg/lib/authn/authenticator/password/generator.go
@@ -170,37 +170,12 @@ func (g *Generator) generateOnce() (string, error) {
 func prepareCharacterSet(policy *config.PasswordPolicyConfig) (string, error) {
 	set := map[characterSet]struct{}{}
 
-	if policy.AlphabetRequired {
-		set[characterSetAlphabet] = struct{}{}
-	}
-	if policy.LowercaseRequired {
-		set[characterSetLowercase] = struct{}{}
-	}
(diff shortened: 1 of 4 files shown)


pkg/lib/authn/authenticator/password/generator.go
```diff
@@ -170,28 +170,14 @@ func (g *Generator) generateOnce() (string, error) {
...
 	if policy.SymbolRequired {
 		set[characterSetSymbol] = struct{}{}
 	}
 
-	// Default to alphanumeric if no character set is required.
-	if len(set) == 0 {
-		set[characterSetAlphabet] = struct{}{}
-		set[characterSetDigit] = struct{}{}
-	}
-
```
louischan-oursky, line 194: Like in js, line 181 to line 190 are redundant. Remove them.

mickmak-busicat:
@louischan-oursky check when you have time🙏

louischan-oursky (changes requested):

pkg/lib/authn/authenticator/password/generator.go
mickmak-busicat: fixed up the change

louischan-oursky (approved):

PR #4684, opened 2024-09-03
Update portal date display format

mickmak-busicat:
@louischan-oursky as no preset from luxon can display exactly in this format `22 Jul, 2024, 4:37 PM (UTC+01:00)`
so used format string to update the format, let me know if there is any concern

Preview:

Screenshot 2024-09-03 at 11 42 06 AM

Screenshot 2024-09-03 at 11 59 56 AM

Screenshot 2024-09-03 at 12 00 34 PM


portal/src/util/formatDatetime.ts
```diff
@@ -18,5 +17,5 @@ export function formatDatetime(
   }
 
   datetime = datetime.setLocale(locale);
-  return datetime.toLocaleString(format);
+  return datetime.toFormat("ff ('UTC'ZZ)");
```
louischan-oursky, line 20: Per offline discussion, we will still use toLocaleString, but with a custom format options.

Please include this link above the option object as comment for future reference. https://tc39.es/ecma402/#sec-datetimeformat-abstracts

louischan-oursky (changes requested):

portal/src/util/formatDatetime.ts
mickmak-busicat: updated, thanks

louischan-oursky (approved):

PR #4687, opened 2024-09-03
Stricten user import schema

pkong-ds:
ref DEV-1482

Some directions I thought of, but only implemented the checked ones
- [x] add `minLength: 1` to required string fields
- [ ] add regex to some expected fields, like email, phone number, birthdate
- [ ] add enum to expected fields, like zoneinfo, locale

ref https://docs.authgear.com/how-to-guide/user-management/import-users-using-user-import-api#input-format

@louischan-oursky see if you expected more strict rules? thanks 🙏

### DIFF
--- a/pkg/lib/userimport/model.go
+++ b/pkg/lib/userimport/model.go
@@ -24,8 +24,7 @@ var RecordSchemaForIdentifierPhoneNumber *validation.SimpleSchema
 var RecordSchemaForIdentifierPreferredUsername *validation.SimpleSchema
 
 func init() {
-	nullString := validation.SchemaBuilder{}.
-		Types(validation.TypeString, validation.TypeNull)
+	reusedSchemaBuilders := reuseSchemaBuilders()
 	str := validation.SchemaBuilder{}.
 		Type(validation.TypeString)
 
@@ -34,9 +33,6 @@ func init() {
 	makeBase := func() validation.SchemaBuilder {
 		boolean := validation.SchemaBuilder{}.
 			Type(validation.TypeBoolean)
-
-		nullObject := validation.SchemaBuilder{}.
-			Types(validation.TypeObject, validation.TypeNull)
 
 		customAttributes := validation.SchemaBuilder{}.
 			Type(validation.TypeObject)
@@ -65,8 +61,8 @@ func init() {
 			Type(validation.TypeObject).
 			AdditionalPropertiesFalse()
 		mfa.Properties().
-			Property("email", nullString).
-			Property("phone_number", nullString).
+			Property("email", reusedSchemaBuilders.Email.Nullable()).
+			Property("phone_number", reusedSchemaBuilders.PhoneNumber.Nullable()).
 			Property("password", password).
 			Property("totp", totp)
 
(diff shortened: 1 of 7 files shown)


louischan-oursky (changes requested):
In pkg/lib/authn/stdattrs/validate.go, there are the actual schema that the std attr module will use to validate the standard attributes. We can borrow the schemas there.

We must pay attention to one thing. validation.SchemaBuilder is just a map, so we must not mutate the schema in stdattrs. I am saying this because the schema used in userimport is slightly different from the schemas in stdattrs. In particular, the schema in userimport allows null.

So my suggestion is

1. Make the SchemaBuilder* in stdattrs private. To avoid external modules import them and mutate them accidentally.
2. Expose methods to make new SchemaBuilder in stdattrs. So that the logic still live in stdattrs.
3. Call the newly exposed methods in userimport to validate more strict.

pkong-ds:

pkg/util/validation/schema_builder.go
```diff
@@ -150,3 +150,19 @@ func (b SchemaBuilder) ToSimpleSchema() *SimpleSchema {
 	}
 	return NewSimpleSchema(string(bytes))
 }
+
+// This function allow copying the schema builder to a reference
+// Expected usage is to avoid mutating original schema builder
+func (b SchemaBuilder) Copy() SchemaBuilder {
```
louischan-oursky, line 156: Clone()

pkg/util/validation/schema_builder.go
```diff
@@ -150,3 +150,19 @@ func (b SchemaBuilder) ToSimpleSchema() *SimpleSchema {
...
+	newB := make(SchemaBuilder)
+	for k, v := range b {
+		vb, ok := v.(SchemaBuilder)
+		if ok {
+			newB[k] = vb.Copy()
```
louischan-oursky, line 161: Clone()

pkg/util/validation/schema_builder.go
```diff
@@ -166,3 +166,21 @@ func (b SchemaBuilder) Copy() SchemaBuilder {
 
 	return newB
 }
+
+func (b SchemaBuilder) Nullable() SchemaBuilder {
+	existingTypes := b["type"]
```
louischan-oursky, line 171: value, ok := b["type"]

pkg/util/validation/schema_builder.go
```diff
@@ -166,3 +166,21 @@ func (b SchemaBuilder) Copy() SchemaBuilder {
...
+
+	if existingTypes == nil {
```
louischan-oursky, line 173: if !ok

pkg/util/validation/schema_builder.go
```diff
@@ -166,3 +166,21 @@ func (b SchemaBuilder) Copy() SchemaBuilder {
 
 	return newB
 }
+
+func (b SchemaBuilder) Nullable() SchemaBuilder {
```
louischan-oursky, line 170: AddTypeNull()

louischan-oursky, line 170: Move near to `Type()` and `Types()`

pkg/util/validation/schema_builder.go
```diff
@@ -166,3 +166,21 @@ func (b SchemaBuilder) Copy() SchemaBuilder {
...
+	existingTypes := b["type"]
+
+	if existingTypes == nil {
+		b.Type(TypeNull)
```
louischan-oursky, line 174: This new method should look like `AddRequired()`

louischan-oursky (changes requested):

pkong-ds:
Updated, thanks!

pkg/lib/userimport/model.go
```diff
@@ -138,7 +138,7 @@ var nonIdentityAwareStandardAttributeKeys []string = []string{
 	"address",
 }
 
-func mapGetNullable[M ~map[string]interface{}, T constraints.Ordered | ~bool](m M, key string) (*T, bool) {
+func mapGetAddTypeNull[M ~map[string]interface{}, T constraints.Ordered | ~bool](m M, key string) (*T, bool) {
```
louischan-oursky, line 141: Revert this change.

louischan-oursky (changes requested):

pkong-ds:
@louischan-oursky sor, fixed up~

louischan-oursky (approved):

PR #4689, opened 2024-09-03
Introduce error pointer in authUI v2

pkong-ds:


pkong-ds:
Hi @louischan-oursky please check if this PR is what's expected

## implemetation
- define per-page error pointer in `base_error.go`
- get error by key in `$.ErrorJSON`
- [TODO] in error.html, instead of parsing `$.Error`, it should requires the errror as a param

pkong-ds:
Add 2 controllers

controllerA - construct validation errors by path
controllerB - allow page to declare rule-based non-validation errors

Reference portal fore

louischan-oursky:
I close this first.

PR #4691, opened 2024-09-03
Switch to Vite as JS bundler for AuthUI

andychow326:
ref DEV-1978


louischan-oursky (changes requested):
1. We must ensure the build is deterministic. How can we ensure this in CI? We can borrow React Strict mode. We intentionally build twice, and compare if the two builds are reproducible (or deterministic).

Makefile
```diff
@@ -162,7 +162,8 @@ authui:
 	npm run --silent --prefix ./authui typecheck
 	npm run --silent --prefix ./authui format
 	npm run --silent --prefix ./authui build
-	rm resources/authgear/generated/build*.html
+	rm -rf resources/authgear/generated/src
```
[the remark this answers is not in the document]
andychow326: The purpose of such command is to delete redundant html files (after bundling) which we don't need after we build the authui. :adore2x:

authui/src/imagepicker.ts
```diff
@@ -11,6 +11,8 @@ import {
   hideErrorMessage,
 } from "./messageBar";
 import { Controller } from "@hotwired/stimulus";
+import Cropper from "cropperjs";
+import "cropperjs/dist/cropper.min.css";
```
louischan-oursky, line 15: Put this in `authgear.ts`

authui/check-bundle-consistency.sh
```diff
@@ -0,0 +1,57 @@
...
+{
+  for i in $(seq 1 $N)
+  do
+    echo "[INFO] Build no.$i"
+
+    # Clear vite cache before building bundles
+    rm -rf node_modules/.vite > /dev/null 2>&1
+    npm run build > /dev/null 2>&1
+
+    jq --sort-keys . $MANIFEST_LOCATION > $i.json
```
louischan-oursky, line 23: You cannot use jq. It may not be installed on the CI machine.

You can use `python3 -m json.tool --sort-keys --no-ensure-ascii --indent 2 <input >output`

louischan-oursky, line 23: You do not write things in our project. You write to a temporary directory.

Since you write to a temporary directory, you do not need to clean it up. We expect it will be cleaned up by somebody else.

To create a temporary directory, you use `mktemp -d`. Please `man mktemp` to see its usage.

louischan-oursky (changes requested):

authui/check-bundle-consistency.sh
andychow326: ok!

andychow326: ok!

louischan-oursky (approved):

PR #4697, opened 2024-09-04
Revamp social connection flow

aekt:
ref DEV-1910, DEV-1911


louischan-oursky (approved):
I am not involved in the design, so I am not sure if this is expected. Right now, when I click edit, I am immediately redirected to reauth, which I think is quite surprising. I think a better UI should be let to see the edit screen with the secret masked. Only when I click reveal secret, I am redirected to reauth.

Otherwise, the PR looks good to me

@aekt @fungc-io

PR #4698, opened 2024-09-04
SAML: Add client_id to service provider

tung2744:
ref DEV-2002

Steps to update existing configs:
1. Deploy the code in this pr
2. Add `client_id` in existing config. Old endpoints should still work, but metadata will be updated.
3. Switch to new endpoints.
4. Remove `id` from configs.
5. Remove `id` from code.

### DIFF
--- a/pkg/lib/config/config.go
+++ b/pkg/lib/config/config.go
@@ -357,6 +357,22 @@ func (c *AppConfig) validateSAML(ctx *validation.Context) {
 				"actual":   0,
 			})
 		}
+
+		for idx, sp := range c.SAML.ServiceProviders {
+			if sp.ClientID != "" {
+				found := false
+				for _, oauthClient := range c.OAuth.Clients {
+					if sp.ClientID == oauthClient.ClientID {
+						found = true
+						break
+					}
+				}
+				if !found {
+					ctx.Child("saml", "service_providers", strconv.Itoa(idx), "client_id").
+						EmitErrorMessage("client_id does not exist in /oauth/clients")
+				}
+			}
+		}
 	}
 }
 
--- a/pkg/lib/config/saml.go
+++ b/pkg/lib/config/saml.go
@@ -37,7 +37,15 @@ var _ = Schema.Add("SAMLServiceProviderConfig", `
 		"audience": { "type": "string", "format": "uri" },
 		"assertion_valid_duration":  { "$ref": "#/$defs/DurationString" }
 	},
-	"required": ["acs_urls"]
+	"required": ["acs_urls"],
+	"anyOf": [
+		{
+			"required": ["id"]
+		},
+		{
+			"required": ["client_id"]
+		}
+	]
 }
 `)
 
--- a/pkg/lib/config/saml.go
+++ b/pkg/lib/config/saml.go
@@ -47,8 +47,11 @@ type SAMLConfig struct {
 }
 
 func (c *SAMLConfig) ResolveProvider(id string) (*SAMLServiceProviderConfig, bool) {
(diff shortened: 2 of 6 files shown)


pkg/lib/config/config.go
```diff
@@ -357,6 +357,22 @@ func (c *AppConfig) validateSAML(ctx *validation.Context) {
...
+						found = true
+						break
+					}
+				}
+				if !found {
+					ctx.Child("saml", "service_providers", strconv.Itoa(idx), "client_id").
+						EmitErrorMessage("client_id does not exist in /oauth/clients")
+				}
+			}
+		}
```
louischan-oursky, line 375: Add tests for this validation logic.

pkg/lib/config/saml.go
```diff
@@ -37,7 +37,15 @@ var _ = Schema.Add("SAMLServiceProviderConfig", `
...
-	"required": ["acs_urls"]
+	"required": ["acs_urls"],
+	"anyOf": [
+		{
+			"required": ["id"]
+		},
+		{
+			"required": ["client_id"]
+		}
+	]
```
louischan-oursky, line 48: Add tests for this validation.

louischan-oursky (changes requested):

tung2744:
Added, thanks!

pkg/lib/config/testdata/config_tests.yaml
```diff
@@ -1200,3 +1200,39 @@ config:
...
+  id: test
+  http:
+    public_origin: http://test
+  saml:
+    signing:
+      key_id: asdfasdf
+    service_providers:
+      - nameid_format: urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress
+        acs_urls:
+          - http://localhost:3000/acs
```
louischan-oursky, line 1221: Add two more clients here to show the validation pass case. That is
1. A SP with `id` should pass
2. A SP with `client_id` should pass

pkg/lib/config/testdata/config_tests.yaml
```diff
@@ -1200,3 +1200,39 @@ config:
...
+  http:
+    public_origin: http://test
+  saml:
+    signing:
+      key_id: asdfasdf
+    service_providers:
+      - client_id: notexist
+        nameid_format: urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress
+        acs_urls:
+          - http://localhost:3000/acs
```
louischan-oursky, line 1238: Add one more client here to show the validation pass case. That is
1. A SP with a client id pointing to an existing client.

Then one test case is enough to demo the validation is correct.

louischan-oursky (changes requested):

pkg/lib/config/testdata/config_tests.yaml
tung2744: Added two items

pkg/lib/config/testdata/config_tests.yaml
tung2744: > Then one test case is enough to demo the validation is correct.

Not sure what you mean here. I added a valid item in the above test case which point to a valid client.

tung2744: But I think we still need another test case to show there is an error if `client_id` is not a valid client.

louischan-oursky: Your test`saml-service-provider-invalid-client-id` did this already.

louischan-oursky (approved):

PR #4700, opened 2024-09-04
Do not send posthog.resetGroups error to sentry

pkong-ds:
ref DEV-1767

## Preview - not sent to sentry

https://github.com/user-attachments/assets/694793b1-fdbe-470f-8135-873f14387a55

### DIFF
--- a/portal/src/ReactApp.tsx
+++ b/portal/src/ReactApp.tsx
@@ -7,7 +7,11 @@ import React, {
   useCallback,
   useMemo,
 } from "react";
-import { init as sentryInit } from "@sentry/react";
+import {
+  Exception as SentryException,
+  Event as SentryEvent,
+  init as sentryInit,
+} from "@sentry/react";
 import { BrowserRouter, Routes, Route, Navigate } from "react-router-dom";
 import {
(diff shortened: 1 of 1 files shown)


portal/src/ReactApp.tsx
```diff
@@ -78,11 +82,34 @@ async function loadSystemConfig(): Promise<SystemConfig> {
...
+    exception:
+      event.exception != null
+        ? {
+            values: event.exception.values?.filter(
+              (value) => !isPosthogResetGroupsException(value)
+            ),
+          }
+        : undefined,
+  };
+}
```
louischan-oursky, line 105: From the doc https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-before-send It says return null to discard. Did I misunderstand anything?

louischan-oursky (changes requested):

portal/src/ReactApp.tsx
pkong-ds: I was mistaken that one event can have other exceptions. Fixed

louischan-oursky (approved):

PR #4705, opened 2024-09-05
Introduce template functions to handle validation errors and non-validation errors

pkong-ds:
ref DEV-1763

## Blocked by
- #4729 

~@louischan-oursky My attempt to implement what we agreed on yesterday, please check if ok!~


pkg/util/template/funcerror.go
```diff
@@ -0,0 +1,64 @@
...
+		return apiErr
+	}
+}
+
+// This function takes any error and returns a validation error
+//
+// i.e. err.Reason == `ValidationFailed`
+//
+// otherwise it returns nil
+func GetValidationError(apiErr *apierrors.APIError, inputFieldName string) (validationErr *apierrors.APIError) {
```
louischan-oursky, line 29: This is not correct. What I told you is to return a validation error with details specific to the given field. So this function should
1. Clone apiErr
2. Filter out any details that is not related to inputFieldName.
3. Return the new apiErr

louischan-oursky (changes requested):

pkg/util/template/funcerror.go
pkong-ds: > Filter out any details that is not related to inputFieldName.

Say given below error and inputFieldName `password_input`, by `filter out any details`, should we remove all other items in `missing`?

## Input apiErr

```
{
	"code": 400,
	"info": {
		"causes": [
			{
				"details": {
					"actual": [
						"gorilla.csrf.Token",
						"x_step"
					],
					"expected": [
						"x_confirm_password",
						"x_password"
					],
					"missing": [
						"x_confirm_password",
						"x_password"
					]
				},
				"kind": "required",
				"location": ""
			}
		]
	},
	"message": "invalid value",
	"name": "Invalid",
	"reason": "ValidationFailed"
}
```

## output

```
{
	"code": 400,
	"info": {
		"causes": [
			{
				"details": {
					"actual": [
						"gorilla.csrf.Token",
						"x_step"
					],
					"expected": [
						"x_confirm_password",
						"x_password"
					],
					"missing": [
						"x_password"
					]
				},
				"kind": "required",
				"location": ""
			}
		]
	},
	"message": "invalid value",
	"name": "Invalid",
	"reason": "ValidationFailed"
}
```

louischan-oursky: Good question. This is a special case. The keyword `required` reports the error in location one level up. So you're right in this case.

Please also add tests to test this filter function,

pkong-ds: 2 cases
- Normal case - `location` is non-empty, can just use
- special case - `location` is empty, need to loop through `missing`

```
`/login_id` -> phone
`/x_login_id` -> phone
```

### Proposed api

```
{{ getValidationError apiErr "login_id" "x_login_id" }}
```

pkong-ds:
@louischan-oursky Tried a more declarative api, please help check if okay 🙏

pkg/util/template/funcerror_test.go
```diff
@@ -0,0 +1,72 @@
...
+				"kind": "required",
+				"location": ""
+			}
+		]
+	},
+	"message": "invalid value",
+	"name": "Invalid",
+	"reason": "ValidationFailed"
+}
+`)
```
louischan-oursky, line 51: Given

```
{
    "code": 400,
    "info": {
        "causes": [
            {
                "details": {
                    "actual": [
                        "gorilla.csrf.Token",
                        "x_step"
                    ],
                    "expected": [
                        "x_confirm_password",
                        "x_password"
                    ],
                    "missing": [
                        "x_confirm_password",
                        "x_password"
                    ]
                },
                "kind": "required",
                "location": ""
            },
            {
              "kind": "minLength",
              "location": "/x_password"
            },
            {
              "kind": "minLength",
              "location": "/x_confirm_password"
            }
        ]
    },
    "message": "invalid value",
    "name": "Invalid",
    "reason": "ValidationFailed"
}
```

```
resolveError $.RawError (dict
  newPasswordField (dict
    "by_location" (list "x_password")
  )
)
```

it should return

```
{
    "code": 400,
    "info": {
        "causes": [
            {
                "details": {
                    "actual": [
                        "gorilla.csrf.Token",
                        "x_step"
                    ],
                    "expected": [
                        "x_password"
                    ],
                    "missing": [
                        "x_password"
                    ]
                },
                "kind": "required",
                "location": ""
            },
            {
              "kind": "minLength",
              "location": "/x_password"
            }
        ]
    },
    "message": "invalid value",
    "name": "Invalid",
    "reason": "ValidationFailed"
}

```

---

Problems in your proposed API

1. Things should be optional. If I do not care about reason, I do not need to specify `by_reason`
2. You still did not "slice" the errors. See my above expected output.

louischan-oursky (changes requested):

pkg/util/template/funcerror_test.go
pkong-ds: okok! Lemme update it

pkong-ds:

- Added slice (trim) validation errors
- Allow optional fields (even empty dict)

pkg/api/apierrors/error.go
```diff
@@ -65,6 +65,19 @@ func (e *APIError) HasCause(kind string) bool {
 	return false
 }
 
+func (e *APIError) Clone() *APIError {
+	info := make(map[string]interface{})
+	for k, v := range e.Info {
+		info[k] = v
+	}
```
louischan-oursky, line 72: The clone is not deep.

pkg/util/template/funcerror.go
```diff
@@ -0,0 +1,217 @@
...
+func getValidationErrorByLocation(apiErr *apierrors.APIError, locations []string) *apierrors.APIError {
+	out := apiErr.Clone()
+	causes, ok := out.Info["causes"]
+	if !ok {
+		panic("apiErr.Info['causes'] is not set")
+	}
+	causeArr, ok := causes.([]interface{})
+	if !ok {
+		panic("apiErr.Info['causes'] is not an array")
+	}
```
louischan-oursky, line 105: No panic in this function. Just return nil if something is unexpected.

pkg/util/template/funcerror.go
```diff
@@ -0,0 +1,217 @@
...
+func trimValidationErrorCauses(causes []map[string]interface{}, locations []string) []map[string]interface{} {
+	trimmedCauses := slice.Map(causes, func(cause map[string]interface{}) map[string]interface{} {
+		causeLoc, ok := cause["location"]
+		if !ok {
+			panic("cause['location'] is not set")
+		}
+		causeLoc, ok = causeLoc.(string)
+		if !ok {
+			panic("cause['location'] is not string")
+		}
```
louischan-oursky, line 135: Actually, you can do both with `location, ok := case["location"].(string)`

pkg/util/template/funcerror.go
```diff
@@ -0,0 +1,217 @@
...
+	kind, ok := cause["kind"]
+	if !ok {
+		panic("cause['kind'] is not set")
+	}
+	kind, ok = kind.(string)
+	if !ok {
+		panic("cause['kind'] is not string")
+	}
+	if kind != "required" {
+		panic(fmt.Errorf("unexpected cause without location - expected: required, got: %v", kind))
```
louischan-oursky, line 172: Again, do not panic. We are handling error here. Panic in error handling will cause the original error to lose.

Instead we should handle error in a resilient way. I mention we only handle the case that we know, for unknown case, we just do nothing. See what I mean?

- We first check `kind`, instead of `location`.
- If `kind` is `required` and location is `""`, then this is a situation we know how to handle.
- Otherwise, if location is not `""`, then this is another situation we know how to handle.
- Otherwise, we do not panic and return gracefully.

louischan-oursky (changes requested):

pkong-ds:

Seems we agree on the API of `resolveError` already. Lemme start applying the changes across entire authUI 

Question - is authUI v1 needed? I plan to apply on v2 only 🙏

pkong-ds:
Used `resolveError` and new `__render_error.html` in below pages

<details>

<summary> Expand me </summary>


- `new_password_field.html`
  - `reset_password.html`
  - `change_password.html`
  - `create_password.html`
- `__password_field.html`
  - `enter_password.html`
  - `ldap_login.html`

</details>

I guess this PR remaining actionables

1. Ensure all changed screens errors work normally; that is, to attach screenshot videos here (or add tests in https://github.com/oursky/e2e-test-authgear? not sure if the effort is of similar amount)
2. Change other screens that requires refactoring
    - quick search seems remaining OAuthError that has hacky error usage

Actionables not expected in this PR

1. Replace all `__error.html` with `__render_error.html`; that is, pass error explicitly to render

Let's sync tmr 🙏

resources/authgear/templates/en/web/authflowv2/__render_error.html
```diff
@@ -0,0 +1,306 @@
+{{/*
```
louischan-oursky, line 1: Remove this. We do not want to maintain __error.html and __render_error.html

resources/authgear/templates/en/web/authflowv2/change_password.html
```diff
@@ -1,9 +1,29 @@
...
+
+{{ $has_unknown_err := not (isNil $unknown_err )}}
 {{ $unknown_error_message := "" }}
-{{ if $.HasUnknownError }}
-  {{ $unknown_error_message = (include "authflowv2/__error.html" .) }}
+{{ if $has_unknown_err }}
+  {{ $unknown_error_message = (include "authflowv2/__render_error.html" (dict
+    "Ctx" $
+    "Error" $unknown_err
+  )) }}
```
louischan-oursky, line 26: You actually do not need __render_error.html, you can just write

```
(include "authflowv2/__error.html" (merge (dict "Error" thenewerror) .))
```

louischan-oursky (changes requested):

pkong-ds:
Updated, thanks

pkong-ds:
As per offline discussion, This PR 

### Will
✅ make commit history cleaner 
⏳ look for other problematic pattern (see [Comment on problematic patten](https://github.com/authgear/authgear-server/pull/4705/commits/e1b71887efab2e66cb925cf1d055db9c0466b391), wip)

### Won't
⛔ refactor `$oauth_error` handling in `login.html` and `signup.html`

resources/authgear/templates/en/web/authflowv2/ldap_login.html
```diff
@@ -64,8 +92,7 @@ <h1 class="screen-title">
       "Type" "old-password"
       "AutoFocus" $.ShouldFocusInput
       "Classname" "w-full"
-      "HasError" $.PasswordInputError.HasError
-      "HasErrorMessage" $.PasswordInputError.HasErrorMessage
+      "Error" $pw_err
```
louischan-oursky, line 95: I take a closer at this, and I found after your change, the behavior is different. Have you ever tried the original behavior?

louischan-oursky (changes requested):

resources/authgear/templates/en/web/authflowv2/ldap_login.html
pkong-ds: Right, missed that `InvalidCredentials` should be 

1. `both username and password input` red border
2. no error messages for input fields
3.  show alert message by `Invalid credentials`

Lemme fix that 🙏

louischan-oursky: So as I said, you do not need to change the interface of __password_field.html

pkong-ds: >  do not need to change the interface of __password_field.html
Disagree on this. I think we need `.Error` to be passed into `__password_field.html` for below

```
<p class="input__error-message" data-text-field-target="errorMessage">
      {{ template "authflowv2/__error.html" (merge (dict "Error" $.Error) $.Ctx) }}
    </p>
```

pkong-ds: So I suggest below diff

```diff
- "HasError`
+ "Error"
   "HasErrorMessage" // keep

```

louischan-oursky: OK. Remove .HasErrorMessage and add .ErrorMessage

Then in __password_field.html, render .ErrorMessage directly.

louischan-oursky: I want it to support two things

- A flag (or a prop) to make it in red
- A prop to pass it an error message to render

So this becomes

- HasError
- ErrorMessage

pkong-ds: Sounds great. Lemme implement that.

Will do similarly for `__new_password_field.html` 🙏 Thanks daic

pkong-ds: Sor @louischan-oursky , on second thought, why not pass the entire Error over HasError flag?

So it becomes

- Error
- ErrorMessage

louischan-oursky: > Have you ever tried the original behavior?

Did you?

If you have, then you will see in a case, the field is just in red, without any error message.

pkong-ds: > Did you?

Yes I did. I understand that we want 2 separate controls on `field red` and `has error message`

I am suggesting `Error` over `HasError` because that way the consumer `__password_field.html` has control over the entire Error object, instead of a processed flag.

It's rather trivial, but just wanna point it out

louischan-oursky: The rationale for __password_field.html to consume an entire Error object is __password_field.html needs to do something with .Error. But it does not need to. It just need to display an given error message.

Or you found such a case?

pkong-ds: > Or you found such a case?

Nope. Maybe let's just pass it when it's needed (which is unlikely).

Will implement below interface 

- HasError
- ErrorMessage

for new_password_field, will be

```
"HasNewPasswordError" $has_pw_err
    "NewPasswordErrorMessage" $pw_error_message
    "HasConfirmPasswordError" $confirm_pw_err
    "ConfirmPasswordErrorMessage" $confirm_pw_error_message
```

pkong-ds:
Updated, thanks! Changes since [Comment on problematic patten](https://github.com/authgear/authgear-server/pull/4705/commits/e7a8541afcfc79603f19c151e699f78a400b68b5)

Will squash commits after finalizing

pkong-ds:
## change_password.html 



## create_password.html 



## enter_oob_otp.html 



## enter_password.html 



## enter_recovery_code.html 



## enter_totp.html 



## forgot_password_otp.html 



## Idap_login.html 

https://github.com/user-attachments/assets/8c08ae3b-ba89-4190-9f1b-d430a3107aeb


## reset_password.html 



## setup_oob_otp.html 



## setup_totp_verify.html

pkg/util/template/funcerror.go
```diff
@@ -93,17 +93,16 @@ func getError(apiErr *apierrors.APIError, getErrInput *GetErrorInput) *apierrors
 	if apiErr == nil {
 		return nil
 	}
+	apiErrByReason := getErrorByReason(apiErr, getErrInput.ByReason)
+	if apiErrByReason != nil {
+		return apiErrByReason
+	}
```
louischan-oursky, line 99: I thought by_location and by_reason are independent. But in your implementation they are not.

resources/authgear/templates/en/web/authflowv2/__password_field.html
```diff
@@ -23,9 +33,9 @@
     "PasswordRules" .PasswordRules
     "InputAttrs" $input_attrs
   ) }}
-  {{ if $has_err }}
+  {{ if $.HasError }}
```
louischan-oursky, line 36: Checks ErrorMessage instead of HasError. So technically, it can display an error message without being highlighted in red.

resources/authgear/templates/en/web/authflowv2/__new_password_field.html
```diff
@@ -36,13 +36,12 @@
         `
       )
     }}
-    {{ $has_pw_input_err_msg := $has_new_pw_err }}
     {{ $pw_input_error_class := "input__error-message"}}
-    {{ if not $has_pw_input_err_msg }}
+    {{ if (eq $.NewPasswordErrorMessage "") }}
```
louischan-oursky, line 40: `{{ if $.NewPasswordErrorMessage }}`

resources/authgear/templates/en/web/authflowv2/__new_password_field.html
```diff
@@ -61,20 +60,15 @@
...
       )
     }}
-    {{/* confirm password has error msg only if 
-      - confirm password input has error; AND
-      - new password input has no error
-     */}}
-    {{ $has_confirm_password_input_err_msg := (and $has_confirm_pw_err (not $has_new_pw_err))}}
     {{ $confirm_input_error_class := "input__error-message"}}
-    {{ if (not $has_confirm_password_input_err_msg) }}
+    {{ if (eq $.ConfirmPasswordErrorMessage "") }}
```
louischan-oursky, line 64: `{{ if $.ConfirmPasswordErrorMessage }}`

louischan-oursky (changes requested):

pkg/util/template/funcerror.go
pkong-ds: > I thought by_location and by_reason are independent. 

Should they be independent? 

What if the caller wants to match `ValidationFailed` without providing any location, as seen in `setup_oob_otp.html`? 🙏

resources/authgear/templates/en/web/authflowv2/__new_password_field.html
pkong-ds: Will update as `{{ if (not $.NewPasswordErrorMessage) }}`
Quoting [golang doc](https://pkg.go.dev/text/template#hdr-Actions) here for reference 🙏 

```
{{if pipeline}} T1 {{end}}
	If the value of the pipeline is empty, no output is generated;
	otherwise, T1 is executed. The empty values are false, 0, any
	nil pointer or interface value, and any array, slice, map, or
	string of length zero.
	Dot is unaffected.
```

pkong-ds:
Updated, thanks!

louischan-oursky (approved):

pkg/util/template/funcerror.go
pkong-ds: ok, thanks daic!

PR #4708, opened 2024-09-06
SAML: Reject unsupported bindings

tung2744:
ref DEV-1737


louischan-oursky (changes requested):
The test failed.

tung2744:
Fixed, thanks!

louischan-oursky (approved):

PR #4710, opened 2024-09-06
Migrate Authgear Portal to use Vite as bundler

andychow326:
ref DEV-2010


louischan-oursky (changes requested):
1. HMR does not work on imported .json files.
2. Add a script to check reproducible build.

andychow326:
> 1. HMR does not work on imported .json files.

@louischan-oursky The HMR does work on imported .json files but it does not work on our SystemConfig#translation. Seems it also did not work for a while.....

This issue comes from 2 reasons:

1. System Config is not reloaded when HMR triggered
    - The translation file (`portal/locale-data/en.json`) is loaded from https://github.com/authgear/authgear-server/blob/2f2bf042db730306b4aed7f058f76cb3cdbd0a20/portal/src/system-config.ts#L220 as the `defaultSystemConfig`
    - The System Config is loaded conditionally (start/error) https://github.com/authgear/authgear-server/blob/2f2bf042db730306b4aed7f058f76cb3cdbd0a20/portal/src/ReactApp.tsx#L332

2. Missing version update event condition by `@oursky/react-messageformat`
    - Should also update the version when `this.props.messageByID !== prevProps.messageByID` https://github.com/oursky/react-messageformat/blob/4e4c6e53205d5d2b68eac750a8353732dff03794/src/react.tsx#L62

Seems to fix the above is a bit complicated and maybe we can just leave those for now? :adore2x:

> 2. Add a script to check reproducible build.

Added :adore2x:

louischan-oursky (approved):

PR #4717, opened 2024-09-09
Specify User Export API

louischan-oursky:
ref DEV-1884

Also tag @mickmak-busicat @christinachee

### DIFF
--- a/docs/specs/user-export.md
+++ b/docs/specs/user-export.md
@@ -76,6 +76,7 @@ If `csv.fields` is unspecified, the default is derived with the following rules:
   {"pointer": "/identities"},
   {"pointer": "/mfa/emails"},
   {"pointer": "/mfa/phone_numbers"},
+  {"pointer": "/mfa/totps"},
   {"pointer": "/biometric_count"},
   {"pointer": "/passkey_count"}
 ]
@@ -483,6 +484,12 @@ Here is an example of the record
   "mfa": {
     "emails": ["louischan@oursky.com"],
     "phone_numbers": ["+85298765432"]
+    "totps" [
+      {
+        "secret": "the-secret",
+        "uri": "otpauth://totp...."
+      }
+    ]
   },
 
   "biometric_count": 0,
@@ -508,6 +515,7 @@ Here is an example of the record
   - `ldap`
 - `mfa.emails`: The list of MFA emails the user has. It is an empty array if the user has none.
 - `mfa.phone_numbers`: The list of MFA phone numbers the user has. It is an empty array if the user has none.
+- `mfa.totps`: The list of MFA TOTP authenticator the user has. It is an empty array if the user has none.
 - `biometric_count`: The number of biometric login the user has.
 - `passkey_count`: The number of passkey the user has.
 
--- a/docs/specs/user-export.md
+++ b/docs/specs/user-export.md
@@ -559,7 +559,7 @@ Here is an example of the record
 
 > Future work: Support exporting the password hash.
 
-## Ceveats
+## Caveats
 
 Since we do not export the password hash,
 The exported JSON record cannot be imported into another Authgear project directly.
--- a/docs/specs/user-export.md
+++ b/docs/specs/user-export.md
@@ -94,7 +94,7 @@ The response body of a just created export looks like
 }
 ```
 
-The response body of completed export looks like
+The response body of a completed export looks like
 
 ```
 {
@@ -108,6 +108,26 @@ The response body of completed export looks like
   "download_url": "https://some-signed-url?with-a=signature"
 }
 ```
+
+The response body of a failed export looks like
+
+```
+{
+  "id": "some_opaque_string",
+  "status": "completed",
+  "created_at": "2024-01-01T00:00:00.000Z",
+  "failed_at": "2024-01-01T00:01:00.000Z",
+  "request": {
+    "format": "ndjson"
+  },
+  "error": {
+    "message": "blahblahblah",
+    "reason": "SomeReason"
+  }
+}
+```
+
+- `error`: The API error object we have been using in all other API.
 
 ## The export file
 
--- a/docs/specs/user-export.md
+++ b/docs/specs/user-export.md
@@ -1,0 +1,284 @@
+# User Export
+
+User Export allows the developer to bulk export all users from Authgear to a file.
+
+## About User Export
+
+- It is not a synchronous operation. The export is created and runs in the background. The developer can query the status of it.
+
+## Create an export
+
+The endpoint is `POST /_api/admin/users/export`.
+
+- The endpoint requires Admin API JWT token to access.
+- This endpoint is added to Admin API server.
+- It is not part of the GraphQL API.
+- At most 1 running export at any given moment for a project.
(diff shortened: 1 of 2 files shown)


docs/specs/user-export.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Create an export
+
+The endpoint is `POST /_api/admin/users/export`.
+
+- The endpoint requires Admin API JWT token to access.
+- This endpoint is added to Admin API server.
+- It is not part of the GraphQL API.
+- At most 1 running export at any given moment for a project.
+- The rate limit of creating a export is 24 per 24h. This means failed requests do not count.
```
christinachee, line 17: Is this rolling 24 hours or refresh at every 00:00 UTC?

louischan-oursky: It is rolling. And it will be implemented using our existing rate limit mechanism.

docs/specs/user-export.md
```diff
@@ -0,0 +1,284 @@
...
+  }
+}
+```
+
+- `format`: Required. It must be `ndjson` or `csv`.
+  - `ndjson`: The output is a ndjson file. See https://github.com/ndjson/ndjson-spec
+  - `csv`: The output is a CSV file. See https://datatracker.ietf.org/doc/html/rfc4180
+- `csv.fields`: Required when `format` is `csv`. It must be an non-empty array.
+  - `csv.fields.json_pointer`: Required. Select which field in the record to output. It must be a JSON pointer. See https://datatracker.ietf.org/doc/html/rfc6901 and [The record format](#the-record-format)
+  - `csv.fields.field_name`: Optional. If it is not given, then the field name is derived from `json_pointer` by taking the last reference token. For example, if the JSON pointers are `/sub` and `/address/formatted`, then the field names are `sub` and `formatted` respectively. If the field names are not unique, it is an error when the export is initiated. An error is immediately returned in this case, the import job is not created. See [The error response of Create an export](#the-error-response-of-create-an-export)
```
chpapa, line 41: In the example given, `/address/formatted` will automatically convert to `formatted` sounds quite uncomfortable, that meant `field_name` are always required.

Alternatively:

1. Instead of taking the last reference token, use a conversion of removing the first `/` and converting`/` to `-` instead? E.g. `/address/formatted` to `address-formatted`
2. For all default fields, give a reasonable default; And only use last reference token for fields without default.

The goal should be `csv.fields.field_name` is really "Optional".

docs/specs/user-export.md
```diff
@@ -0,0 +1,284 @@
...
+The name of the export file is `myapp-userexport_deadbeef-20240909104651Z.ndjson`.
+
+### The metadata of the export file
+
+- `Content-Disposition`: `attachment; filename=FILENAME`
+- `Content-Type`: `application/x-ndjson` for `ndjson`, `text/csv` for `csv`.
+
+Since the export file is accessed with a signed URL of a short validity, setting `Cache-Control` is not really helpful.
+
+### The housekeep of the export file
```
chpapa, line 134: What about on-perm deployment? Or there is a way on-perm deployment can disable this feature?

(So we might need another error code for disabled feature?)

docs/specs/user-export.md
```diff
@@ -0,0 +1,284 @@
...
+      }
+    ]
+  }
+}
+```
+
+- `format`: Required. It must be `ndjson` or `csv`.
+  - `ndjson`: The output is a ndjson file. See https://github.com/ndjson/ndjson-spec
+  - `csv`: The output is a CSV file. See https://datatracker.ietf.org/doc/html/rfc4180
+- `csv.fields`: Required when `format` is `csv`. It must be an non-empty array.
```
chpapa, line 39: Shall we make it optional as well? So when it's optional, use one of the approaches in the below's naming way to export all fields?

docs/specs/user-export.md
louischan-oursky: I intended to support cloud storage only. So on-perm deployment also requires a cloud storage bucket.

mickmak-busicat: do we need a new configure page for the setting up the cloud storage info of a project?

chpapa: In that case let's add a option to disable this feature, and the error code on the API when it's disabled?

So that for some on-perm deployment which we don't have cloud storage bucket, we can at least disable it instead of let it "hang there"...

docs/specs/user-export.md
tung2744, line 41: We might also want to document the behavior if the field is a item in an array.

docs/specs/user-export.md
tung2744: Agree that we should have a default format for csv too.

tung2744:
How to configure the bucket used to store the files?

docs/specs/user-export.md
tung2744: Do we allow controlling it in the config?

docs/specs/user-export.md
```diff
@@ -0,0 +1,284 @@
...
+  }
+}
+```
+
+The response body of completed export looks like
+
+```
+{
+  "id": "some_opaque_string",
+  "status": "completed",
```
tung2744, line 102: I think there should be a `failed` status

mickmak-busicat: will we want to store the fail reason as well in the response?

louischan-oursky:
I have added several commits to address the comments.

docs/specs/user-export.md
```diff
@@ -0,0 +1,553 @@
...
+
+For example, given `pointer` is `/roles/0`,
+
+- Then `parts` is `["roles", "0"]`.
+- The join result is `roles.0`.
+- The field name is the join result.
+
+Regardless of whether the field names are given or derived, they must be unique.
+If the field names are not unique, it is an error when the export is created.
+An error is immediately returned in this case, the import is not created.
```
mickmak-busicat, line 172: i think "import" is typo?

docs/specs/user-export.md
```diff
@@ -0,0 +1,553 @@
...
+See [The response body](#the-response-body).
+
+### The error response of Create an export
+
+> we have a global middleware of handling Admin API authentication,
+> that middleware returns 403 without a JSON body when authentication failed.
+
+|Description|Name|Reason|Info|
+|---|---|---|---|
+|When user import is disabled|`InternalError`|`UserExportDisabled`|-|
```
mickmak-busicat, line 186: same

docs/specs/user-export.md
```diff
@@ -0,0 +1,553 @@
...
+See [The response body](#the-response-body).
+
+### The error response of Get the status of an export
+
+> we have a global middleware of handling Admin API authentication,
+> that middleware returns 403 without a JSON body when authentication failed.
+
+|Description|Name|Reason|Info|
+|---|---|---|---|
+|When user import is disabled|`InternalError`|`UserExportDisabled`|-|
```
mickmak-busicat, line 213: same

louischan-oursky:
@tung2744 This is ready for merge.

tung2744 (approved):

PR #4723, opened 2024-09-10
Hard code special cases to allow in translation key linter

pkong-ds:
ref DEV-2017


devtools/gotemplatelinter/translation_key_rule_special_cases.go
```diff
@@ -0,0 +1,51 @@
...
+	"dialog-close-btn",
+	"authflowv2/__error_account",
+	"page-navbar",
+	"__forgot_password_alternative",
+	"__settings_profile_item",
+	"__settings_profile_date_item",
+	"__settings_profile_address_item",
+	"__settings_profile_locale_item",
+	"__settings_gender_edit_custom_gender_input",
+}
```
louischan-oursky, line 26: Remove these.

devtools/gotemplatelinter/translation_key_rule_special_cases.go
```diff
@@ -0,0 +1,51 @@
...
+
+var allowedVariableKeys []string = []string{
+	"$label",
+	"$labelKey",
+}
```
louischan-oursky, line 31: Remove these.

louischan-oursky (changes requested):

devtools/gotemplatelinter/translation_key_rule_special_cases.go
pkong-ds: Picked 2, please suggest if otherwise~

pkong-ds:
Updated, thanks!

devtools/gotemplatelinter/translation_key_rule_special_cases.go
louischan-oursky: I mean 3.

devtools/gotemplatelinter/translation_key_rule_special_cases.go
louischan-oursky: I mean 3 here as well.

pkong-ds:
Updated with 3, thanks!

PR #4731, opened 2024-09-11
Remove non-existing $.HasUnknownError in wechat.html

pkong-ds:
Found this weird behavior when working on PR #4705 

- `wechat.html` is referencing `$.HasUnknownError`
- If you do a global search `HasUnknownError`, it is only returned by SSR in password pages and ldap_login
   - [`wechat.go`](https://github.com/authgear/authgear-server/blob/main/pkg/auth/handler/webapp/authflowv2/wechat.go#L54-L55) only use [base view model, which does not have HasUnknownError](https://github.com/authgear/authgear-server/blob/main/pkg/auth/handler/webapp/viewmodels/base.go)
   - i.e. `$.HasUnknownError` is always nil in`wechat.html`

Not sure if expected, maybe need to introduce back the wechat unknown error somehow?

### DIFF
--- a/resources/authgear/templates/en/web/authflowv2/wechat.html
+++ b/resources/authgear/templates/en/web/authflowv2/wechat.html
@@ -1,10 +1,6 @@
 {{ template "authflowv2/__page_frame.html" . }}
 
 {{ define "page-content" }}
-{{ $unknown_error_message := "" }}
-{{ if $.HasUnknownError }}
-  {{ $unknown_error_message = (include "authflowv2/__error.html" .) }}
-{{ end }}
 
 {{ if $.IsNativePlatform }}
 {{ if $.WechatRedirectURI }}
@@ -21,7 +17,7 @@
       (dict
         "Type" "error"
         "Classname" "mt-4"
-        "Message" $unknown_error_message
+        "Message" (include "authflowv2/__error.html" .)
       )
     }}
   </header>


resources/authgear/templates/en/web/authflowv2/wechat.html
```diff
@@ -17,13 +13,6 @@ <h1 class="screen-title">
     <h2 class="screen-description">
       {{ template "v2.page.wechat-auth.default.app-description" }}
     </h2>
-    {{ template "authflowv2/__alert_message.html"
-      (dict
-        "Type" "error"
-        "Classname" "mt-4"
-        "Message" $unknown_error_message
-      )
-    }}
```
louischan-oursky, line 26: The baseline is every page should has `__alert_message.html` to show error. So the fix should be

- Keep `__alert_message.html`

```
    {{ template "authflowv2/__alert_message.html"
      (dict
        "Type" "error"
        "Classname" "mt-4"
        "Message" (include "authflowv2/__error.html" .)
      )
    }}
```

louischan-oursky (changes requested):

resources/authgear/templates/en/web/authflowv2/wechat.html
pkong-ds: Right, Lemme change that 🙏 

Is there doc for `how to setup oauth locally`? namely wechat and google 

Otherwise will just follow official authgear doc like this one https://docs.authgear.com/how-to-guide/how-to-setup-sso-integrations/social-login-providers/wechat

louischan-oursky: You cannot set up Wechat easily because it requires approval from Wechat to create an application.

louischan-oursky: For google, you just follow the official doc.

pkong-ds: > You cannot set up Wechat easily

Oh.. thanks for the heads up. Lemme still apply for future.

 I guess I just blindly change the error and rely on you to help me test it this time 🙇

pkong-ds:
Updated, thanks!

louischan-oursky (approved):

PR #4734, opened 2024-09-12
Fix forgot password error display

pkong-ds:
ref DEV-2058

Maybe should enquire fung on `what error to show under input field` vs `what error to show in alert message`

### DIFF
--- a/resources/authgear/templates/en/web/authflowv2/forgot_password.html
+++ b/resources/authgear/templates/en/web/authflowv2/forgot_password.html
@@ -37,12 +37,6 @@
             {{ template "v2.page.forgot-password.default.email-input-description" }}
           {{ end }}
         </p>
-        {{ template "authflowv2/__alert_message.html"
-          (dict
-            "Type" "error"
-            "Classname" "mt-4"
-          )
-        }}
       </header>
       {{ if or $show_email_input $show_phone_input }}
         <form 


resources/authgear/templates/en/web/authflowv2/forgot_password.html
```diff
@@ -37,12 +37,6 @@ <h1 class="screen-title">{{ template "v2.page.forgot-password.default.title" }}<
             {{ template "v2.page.forgot-password.default.email-input-description" }}
           {{ end }}
         </p>
-        {{ template "authflowv2/__alert_message.html"
-          (dict
-            "Type" "error"
-            "Classname" "mt-4"
-          )
-        }}
```
louischan-oursky, line 45: Again, every page should have __alert_message.html, why do you keep wanting to remove it?

louischan-oursky (changes requested):

resources/authgear/templates/en/web/authflowv2/forgot_password.html
pkong-ds: Oh, the `__alert_message.html` in this file does not do anything right now. Note `Message` param is missing.

As mentioned in DEV-2058

> Currently, all errors is put under input field.

So this commit is just removing unused code

pkong-ds: I think a proper fix would be to distinguish what error to show in input/ what error to show in alert message. But might need fung input on this?

louischan-oursky: I see.

louischan-oursky (approved):

PR #4741, opened 2024-09-12
SAML/SLO: Update config & metadata

tung2744:
ref DEV-2059

### DIFF
--- a/pkg/lib/config/saml.go
+++ b/pkg/lib/config/saml.go
@@ -1,6 +1,10 @@
 package config
 
 import (
+	crewjamsaml "github.com/crewjam/saml"
+
+	"github.com/authgear/authgear-server/pkg/lib/saml/samlprotocol"
+
 	"github.com/iawaknahc/jsonschema/pkg/jsonpointer"
 )
 
@@ -35,7 +39,16 @@ var _ = Schema.Add("SAMLServiceProviderConfig", `
 		"destination": { "type": "string", "format": "uri" },
 		"recipient": { "type": "string", "format": "uri" },
 		"audience": { "type": "string", "format": "uri" },
-		"assertion_valid_duration":  { "$ref": "#/$defs/DurationString" }
+		"assertion_valid_duration":  { "$ref": "#/$defs/DurationString" },
+		"slo_enabled": { "type": "boolean" },
+		"slo_callback_url": { "type": "string", "format": "uri" },
+		"slo_binding": {
+			"type": "string",
+			"enum": [
+				"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect",
+				"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Post"
+			]
+		}
 	},
 	"required": ["acs_urls"],
 	"anyOf": [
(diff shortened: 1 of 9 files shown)


pkg/lib/config/saml.go
```diff
@@ -76,12 +104,15 @@ var _ = Schema.Add("SAMLNameIDFormat", `
 }
 `)
 
-type SAMLNameIDFormat string
-
-const (
-	SAMLNameIDFormatUnspecified  SAMLNameIDFormat = "urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"
-	SAMLNameIDFormatEmailAddress SAMLNameIDFormat = "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress"
-)
+var _ = Schema.Add("SAMLBinding", `
```
louischan-oursky, line 107: This is unused. I guess you want to use it, right?

pkg/lib/config/saml.go
```diff
@@ -104,29 +135,36 @@ func (p SAMLNameIDAttributePointer) MustGetJSONPointer() jsonpointer.T {
...
+	if c.NameIDFormat == samlprotocol.SAMLNameIDFormatUnspecified && c.NameIDAttributePointer == "" {
 		c.NameIDAttributePointer = "/sub"
 	}
 
 	if c.AssertionValidDuration == "" {
 		c.AssertionValidDuration = DurationString("20m")
 	}
+
+	if c.SLOBinding == "" {
+		c.SLOBinding = crewjamsaml.HTTPPostBinding
```
louischan-oursky, line 166: Use samlprotocol.SAMLBindingHTTPPost. Remove the import of `crewjamsaml` in this file.

pkg/lib/saml/samlprotocol/const.go
```diff
@@ -16,6 +16,11 @@ var SSOSupportedBindings []SAMLBinding = []SAMLBinding{
 	SAMLBindingHTTPPost,
 }
 
+var SLOSupportedBindings []SAMLBinding = []SAMLBinding{
+	SAMLBindingHTTPRedirect,
+	SAMLBindingHTTPPost,
+}
```
louischan-oursky, line 22: do we actually going to both? I am asking this because in ACS, we support HTTP-Post only.

louischan-oursky (changes requested):

pkg/lib/saml/samlprotocol/const.go
tung2744: I think yes, as logout response is much shorter so I guess using redirect binding is possible.

pkg/lib/config/saml.go
tung2744: Oh it is a mistake

pkg/lib/saml/samlprotocol/const.go
tung2744: Sorry this is not related to my previous comment.
This supported bindings is about what binding we accept in our slo service to receive a logout request. And yes redirect and post is accepted.
Like in SSO service, we also accept both bindings to receive a AuthnRequest.

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #4745, opened 2024-09-13
[DEV-2066] Report language tag when render templates

chankiyu22:
refs DEV-2066

### DIFF
--- a/pkg/util/template/effective_resource.go
+++ b/pkg/util/template/effective_resource.go
@@ -1,0 +1,18 @@
+package template
+
+import (
+	htmltemplate "html/template"
+	texttemplate "text/template"
+)
+
+type HTMLTemplateEffectiveResource struct {
+	Data        []byte
+	LanguageTag string
+	Template    *htmltemplate.Template
+}
+
+type TextTemplateEffectiveResource struct {
+	Data        []byte
+	LanguageTag string
+	Template    *texttemplate.Template
+}
--- a/pkg/util/template/engine.go
+++ b/pkg/util/template/engine.go
@@ -15,6 +15,11 @@ import (
 	"github.com/authgear/authgear-server/pkg/util/messageformat"
 )
 
+type RenderResult struct {
+	String      string
+	LanguageTag string
+}
+
 type EngineTemplateResolver interface {
 	ResolveHTML(desc *HTML, preferredLanguages []string) (*HTMLTemplateEffectiveResource, error)
 	ResolveMessageHTML(desc *MessageHTML, preferredLanguages []string) (*HTMLTemplateEffectiveResource, error)
@@ -49,7 +54,7 @@ func (e *Engine) Translation(preferredLanguages []string) (*TranslationMap, erro
 	return &TranslationMap{items: items}, nil
 }
 
-func (e *Engine) Render(resource Resource, preferredLanguages []string, data interface{}) (string, error) {
+func (e *Engine) Render(resource Resource, preferredLanguages []string, data interface{}) (*RenderResult, error) {
 	switch desc := resource.(type) {
 	case *HTML:
 		return e.renderHTML(desc, preferredLanguages, data)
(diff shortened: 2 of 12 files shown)


louischan-oursky:
I think you need to rebase? I didn't see the new changes.

chankiyu22:
@louischan-oursky Rebased.

pkg/util/template/resolver.go
```diff
@@ -30,7 +30,7 @@ func (r *Resolver) ResolveHTML(desc *HTML, preferredLanguages []string) (*htmlte
 		return nil, err
 	}
 
-	return resrc.(*htmltemplate.Template), nil
+	return resrc.(*HTMLTemplatesEffectiveResource).Template, nil
```
louischan-oursky: This commit has to be rewritten. We have to keep return []byte because all existing consumers expect []byte. You can only change EffectiveResource to return something other than *text.Template or *html.Template.

To do so, you need to audit all existing places that perform the cast. `.(\*.*Template)`

pkg/util/template/engine.go
```diff
@@ -15,6 +15,11 @@ import (
 	"github.com/authgear/authgear-server/pkg/util/messageformat"
 )
 
+type StringWithLanguageTag struct {
```
louischan-oursky, line 18: Name this RenderResult

```
type RenderResult struct {
  LanguageTag string
  String string
}
```

Note that in your original code, `languageTag` is a private type, so the outsider cannot see LanguageTag actually. You need to make it a String.

louischan-oursky (changes requested):

pkg/util/template/resolver.go
chankiyu22:
```sh
$ grep -rni '\.(*.\+\.Template)' pkg
pkg/util/template/resolver.go:33:       return resrc.(*htmltemplate.Template), nil
pkg/util/template/resolver.go:46:       return resrc.(*htmltemplate.Template), nil
pkg/util/template/resolver.go:59:       return resrc.(*texttemplate.Template), nil
pkg/util/template/resolver.go:72:       return resrc.(*texttemplate.Template), nil
pkg/util/template/resource_test.go:42:                  tpl := result.(*htmltemplate.Template)
```

chankiyu22:
@louischan-oursky Updated.

pkg/util/template/effective_resource.go
```diff
@@ -0,0 +1,82 @@
+package template
+
+import (
+	"fmt"
+	htmltemplate "html/template"
+	texttemplate "text/template"
+)
+
+type EffectiveResourceView interface {
```
louischan-oursky, line 9: Remove this.

pkg/util/template/effective_resource.go
```diff
@@ -0,0 +1,82 @@
...
+	fn()
+}
+
+type EffectiveResource struct {
```
louischan-oursky, line 13: Remove this.

pkg/util/template/effective_resource.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (v EffectiveResource) fn() {}
+
+var _ EffectiveResourceView = &EffectiveResource{}
+
+// An intermediate view to be resolved to html and text view
+type TemplatesEffectiveResourceIntermediateView interface {
```
louischan-oursky, line 21: Remove this.

pkg/util/template/effective_resource.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+type TemplatesEffectiveIntermediateResource struct {
+	Data        []byte
+	LanguageTag languageTag
+}
+
+var _ TemplatesEffectiveResourceIntermediateView = &TemplatesEffectiveIntermediateResource{}
+
+type HTMLTemplatesEffectiveResourceView interface {
```
louischan-oursky, line 31: Remove this.

pkg/util/template/effective_resource.go
```diff
@@ -0,0 +1,82 @@
...
+	EffectiveResourceView
+	TemplatesEffectiveResourceIntermediateView
+}
+
+type HTMLTemplatesEffectiveResource struct {
+	EffectiveResource
+	*TemplatesEffectiveIntermediateResource
+	Template *htmltemplate.Template
+}
```
louischan-oursky, line 40: Simplify this struct to 

```
type HTMLTemplateEffectiveResource struct {
  Data []byte
  LanguageTag String
  Template *htmltemplate.Template
}
```

Note that **there is no s** in `Template`

Apply this to TextTemplateEffectiveResource

Note that **there is no s** in `Template`

louischan-oursky (changes requested):

chankiyu22:
@louischan-oursky updated.

louischan-oursky (approved):

PR #4749, opened 2024-09-16
SAML/SLO: Return a response in logout endpoint

tung2744:
ref DEV-2060

TODO:
- [x] Put the signature of the response to query if using redirect binding


pkg/auth/handler/saml/logout.go
```diff
@@ -0,0 +1,238 @@
...
+	r *http.Request,
+	sp *config.SAMLServiceProviderConfig,
+	callbackURL string,
+) (relayState string, result samlprotocolhttp.SAMLResult) {
+	now := h.Logger.Time.UTC()
+	var parseResult samlbinding.SAMLBindingParseResult
+	var logoutRequest *samlprotocol.LogoutRequest
+
+	// Get data with corresponding binding
+	var err error
```
louischan-oursky, line 79: 1. Prefer more than one methods returning (A, error). The error could be a error result that can be written directly.
2. One method to see the flow e.g. handleSLORequest

```
func handleSLORequest() {
  request, err := parse(r)
}
```


```
var err error

defer func() {
  if err != nil {
    // ...
  }
}()



switch {
case A
  if err != nil {
    return
  }
case B
}
```

louischan-oursky (changes requested):
1. Delete samlerror. Move the contents to samlprotocol/error_blahblahblah.go
2. Delete samlprotocolhttp. Move the contents to samlbinding.
3. Remove IsInternalError()

tung2744:
Updated, thanks!

pkg/auth/handler/saml/logout.go
```diff
@@ -147,28 +164,34 @@ func (h *LogoutHandler) handleSLORequest(
 			break
```
louischan-oursky, line 164: No break

pkg/auth/handler/saml/logout.go
```diff
@@ -0,0 +1,257 @@
...
+		if parseErr != nil {
+			return nil, nil, "", parseErr
+		}
+		logoutRequest, err = samlprotocol.ParseLogoutRequest([]byte(r.SAMLRequestXML))
+		if err != nil {
+			err = &samlprotocol.ParseRequestFailedError{
+				Reason: "malformed LogoutRequest",
+				Cause:  err,
+			}
+			break
```
louischan-oursky, line 109: No break

pkg/auth/handler/saml/logout.go
```diff
@@ -0,0 +1,257 @@
...
```
louischan-oursky, line 125: No break

pkg/auth/handler/saml/logout.go
```diff
@@ -0,0 +1,257 @@
...
+			parseResult.SigAlg,
+			parseResult.RelayState,
+			parseResult.Signature)
+		if err != nil {
+			break
+		}
+	case *samlbinding.SAMLBindingHTTPPostParseResult:
+		err = h.SAMLService.VerifyEmbeddedSignature(sp, parseResult.SAMLRequestXML)
+		if err != nil {
+			break
```
louischan-oursky, line 169: No break

louischan-oursky (changes requested):

louischan-oursky (approved):

tung2744:
I found that I forgot to remove the samlerror package, although no code is using them. I will remove them in next pr.

PR #4756, opened 2024-09-20
Use prepared statements for SQL queries

kiootic:
ref DEV-1740

### DIFF
--- a/pkg/lib/infra/db/handle.go
+++ b/pkg/lib/infra/db/handle.go
@@ -1,7 +1,5 @@
 package db
 
-import "github.com/jmoiron/sqlx"
-
 // Handle allows a function to be run within a transaction.
 type Handle interface {
 	// WithTx runs do within a transaction.
@@ -13,5 +11,5 @@ type Handle interface {
 	ReadOnly(do func() error) (err error)
 
 	// conn allows internal access to the ongoing transaction.
-	conn() (sqlx.ExtContext, error)
+	conn() (*txConn, error)
 }
--- a/pkg/lib/infra/db/hook_handle.go
+++ b/pkg/lib/infra/db/hook_handle.go
@@ -153,7 +153,7 @@ func (h *HookHandle) beginTx() (*txConn, error) {
 	return &txConn{
 		db:        db,
 		tx:        tx,
-		logger:    db.logger,
+		logger:    h.Logger,
 		doPrepare: h.ConnectionOptions.UsePreparedStatements,
 	}, nil
 }
--- a/pkg/lib/infra/db/hook_handle.go
+++ b/pkg/lib/infra/db/hook_handle.go
@@ -5,8 +5,6 @@ import (
 	"database/sql"
 	"fmt"
 
-	"github.com/jmoiron/sqlx"
-
 	"github.com/authgear/authgear-server/pkg/util/errorutil"
 	"github.com/authgear/authgear-server/pkg/util/log"
 )
@@ -17,7 +15,7 @@ type HookHandle struct {
 	ConnectionOptions ConnectionOptions
 	Logger            *log.Logger
 
-	tx    *sqlx.Tx
+	tx    *txConn
 	hooks []TransactionHook
 }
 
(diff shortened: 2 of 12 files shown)


pkg/lib/infra/db/handle.go
```diff
@@ -13,5 +15,10 @@ type Handle interface {
 	ReadOnly(do func() error) (err error)
 
 	// conn allows internal access to the ongoing transaction.
-	conn() (sqlx.ExtContext, error)
+	conn() (*txConn, error)
+}
+
+type ConnContext interface {
```
louischan-oursky, line 21: Is this used?

pkg/lib/infra/db/hook_handle.go
```diff
@@ -139,20 +139,20 @@ func (h *HookHandle) ReadOnly(do func() error) (err error) {
 	return
 }
 
-func (h *HookHandle) beginTx() (*sqlx.Tx, error) {
+func (h *HookHandle) beginTx() (*PoolDB, *sqlx.Tx, error) {
```
louischan-oursky, line 142: I suggest we change beginTx to return `(*txConn, error)`

pkg/lib/infra/db/pool.go
```diff
@@ -53,23 +80,28 @@ func (p *Pool) Close() (err error) {
 
 	p.closed = true
 	for _, db := range p.cache {
-		if closeErr := db.Close(); closeErr != nil {
+		if closeErr := db.DB.Close(); closeErr != nil {
```
louischan-oursky, line 83: Call PoolDB.Close()

pkg/lib/infra/db/pool.go
```diff
@@ -1,25 +1,52 @@
...
 	"sync"
 
 	"github.com/jmoiron/sqlx"
 )
 
+type PoolDB struct {
+	DB       *sqlx.DB
+	stmtLock *sync.RWMutex
+	stmts    map[string]*sqlx.Stmt
+}
```
louischan-oursky, line 15: Implement Close() for PoolDB. It should Close all cached prepared statements, and then close the DB.

louischan-oursky (changes requested):
I tested this locally with

1. Add some logs in db/pool.go to check if the stmt is cache hit or cache miss.
2. Sign in to populate some stmts.
3. Shut down the db
4. Refresh the settings page, so connection denied.
5. Start the db again
6. Refresh the settings page. Surprisingly, the stmts are still cache hit.

---

We lack a way to invalidate the stmts in case they went wrong. I think we at least should invalidate the stmts in case of IO error. What do you think?

pkg/lib/infra/db/handle.go
kiootic: This is actually the interface implemented by `txConn`; but since `conn()` is not public, it ends up unused. Let me remove this.

pkg/lib/infra/db/tx_conn.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (t *txConn) QueryContext(ctx context.Context, query string, args ...interface{}) (*sql.Rows, error) {
+	if !t.doPrepare {
+		return t.tx.QueryContext(ctx, query, args...)
+	}
+
+	stmt, err := t.prepare(ctx, query)
+	if err != nil {
+		return t.tx.QueryContext(ctx, query, args...)
```
louischan-oursky, line 32: We do not know the prepared statements keep failing to prepare itself.

pkg/lib/infra/db/tx_conn.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return stmt.QueryContext(ctx, args...)
```
louischan-oursky, line 34: If we log here, then we know the SQL is executed via a prepared statement.


```
t.Logger.Debug
```

louischan-oursky (changes requested):

louischan-oursky:
I added a commit to a fix nil crash.

PR #4757, opened 2024-09-23
Fix audit log filter bar horizontal spacing

pkong-ds:
ref DEV-2068

## Preview (red border for measure purpose)

image
image
image

### DIFF
--- a/portal/src/components/audit-log/AuditLogFilterBar.module.css
+++ b/portal/src/components/audit-log/AuditLogFilterBar.module.css
@@ -14,7 +14,7 @@
 }
 
 .filterContainer {
-  @apply flex items-center;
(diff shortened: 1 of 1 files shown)


portal/src/components/audit-log/AuditLogFilterBar.module.css
```diff
@@ -40,6 +39,7 @@
 }
 
 .filterActionContainer {
+  @apply ml-4 tablet:ml-0;
```
louischan-oursky, line 42: Avoid using l, use gap.

louischan-oursky (changes requested):

pkong-ds:
Updated, thanks

PR #4760, opened 2024-09-23
Fix undefined app ID in Create New Project Screen

andychow326:
ref DEV-2097

image


pkong-ds:
Thanks 🙏 Why did this happen? Quick google on `importing json in js` seems both `import * as` and `import` are both supported

tung2744:
Maybe it is related to the recent switch to vite

andychow326:
> Thanks 🙏 Do you know why this happened? Quick google on `importing json in js` seems both `import * as` and `import` are both supported

In the recent vite migration, https://www.typescriptlang.org/tsconfig/#esModuleInterop is being turned on to fix es6 modules importing problems. After that, we need to use default import if possible :sosad:

louischan-oursky (approved):

PR #4771, opened 2024-09-25
Fix non-first signup identify screen still show "Welcome"

pkong-ds:
ref DEV-1420

## What's in this PR?

Previously the copywriting in signup page did not handle `custom authflow with multiple identify steps`, where it always showed "welcome"

This PR capture whether current screen is in the middle of the flow (i.e. flow already started before current screen) by checking `if x_step query param is in the URL`

## Preview

https://github.com/user-attachments/assets/127094dd-f75d-45a8-8369-f6be88b189b2


pkg/auth/handler/webapp/viewmodels/base.go
```diff
@@ -239,9 +240,12 @@ func (m *BaseViewModeler) ViewModel(r *http.Request, rw http.ResponseWriter) Bas
 		bpLang = intl.ResolveRecaptchaV2(resolvedLanguageTag)
 	}
 
+	requestURI := r.URL.RequestURI()
+	hasXStep := strings.Contains(requestURI, webapp.AuthflowQueryKey)
```
louischan-oursky, line 244: This may result in false positive (but that's is unlikely though). 

I would write

```
HasXStep: r.URL.Query().Has(webapp.AuthflowQueryKey),
```

resources/authgear/templates/en/translation.json
```diff
@@ -1181,6 +1181,7 @@
   "v2.page.signup.default.subtitle": "Sign up for {AppName}",
   "v2.page.signup.default.switch-to-login": "Already have an account? <a href=\"{href}\">Log in</a>",
   "v2.page.signup.default.title": "Welcome",
+  "v2.page.signup.non-first-identify.title": "Continue Sign Up",
```
louischan-oursky, line 1184: `v2.page.signup.continue.title`

resources/authgear/templates/en/web/authflowv2/signup.html
```diff
@@ -39,7 +39,7 @@
   {{ end }}
   {{ $appName := (translate "app.name" nil) }}
   {{ $clientName := or $.ClientName "null" }}
-
+  {{ $authflowStarted := $.HasXStep }}
```
louischan-oursky, line 42:
```
{{ $isContinue := $.HasXStep }}
```

louischan-oursky (changes requested):
According to the issue description, the subtitle needs changes as well, but this PR did not do that. Please do so.

pkong-ds:
Updated, thanks!

louischan-oursky (approved):

PR #4773, opened 2024-09-25
Support user export to CSV format

mickmak-busicat:
ref dev-2056

Preview:

## CSV field validation error:
Screenshot 2024-09-26 at 12 21 44 AM
Screenshot 2024-09-26 at 12 22 21 AM

## Sample CSV (default field set):
[accounts-userexport_RHY34ZR76GVNSC6PC83DPA0FSWMD4P1F-20240925074610Z.csv](https://github.com/user-attachments/files/17134774/accounts-userexport_RHY34ZR76GVNSC6PC83DPA0FSWMD4P1F-20240925074610Z.csv)

## Sample CSV with specified fields:
Screenshot 2024-09-26 at 12 24 02 AM

### DIFF
--- a/pkg/lib/userexport/model.go
+++ b/pkg/lib/userexport/model.go
@@ -58,44 +58,40 @@ var RequestSchema = validation.NewSimpleSchema(`
 }
 `)
 
-const DefaultCSVExportField = `
-{
-  "fields": [
-		{"pointer": "/sub"},
-		{"pointer": "/preferred_username"},
-		{"pointer": "/email"},
-		{"pointer": "/phone_number"},
-		{"pointer": "/email_verified"},
-		{"pointer": "/phone_number_verified"},
-		{"pointer": "/name"},
-		{"pointer": "/given_name"},
-		{"pointer": "/middle_name"},
-		{"pointer": "/nickname"},
-		{"pointer": "/profile"},
-		{"pointer": "/picture"},
-		{"pointer": "/website"},
-		{"pointer": "/gender"},
-		{"pointer": "/birthdate"},
-		{"pointer": "/zoneinfo"},
-		{"pointer": "/locale"},
-		{"pointer": "/address/formatted"},
-		{"pointer": "/address/street_address"},
-		{"pointer": "/address/locality"},
-		{"pointer": "/address/region"},
-		{"pointer": "/address/postal_code"},
-		{"pointer": "/address/country"},
-		{"pointer": "/roles"},
-		{"pointer": "/groups"},
-		{"pointer": "/disabled"},
-		{"pointer": "/identities"},
-		{"pointer": "/mfa/emails"},
-		{"pointer": "/mfa/phone_numbers"},
-		{"pointer": "/mfa/totps"},
-		{"pointer": "/biometric_count"},
-		{"pointer": "/passkey_count"}
-	]
+var defaultCSVExportFields = []*FieldPointer{
+	&FieldPointer{Pointer: "/sub"},
+	&FieldPointer{Pointer: "/preferred_username"},
+	&FieldPointer{Pointer: "/email"},
+	&FieldPointer{Pointer: "/phone_number"},
+	&FieldPointer{Pointer: "/email_verified"},
+	&FieldPointer{Pointer: "/phone_number_verified"},
+	&FieldPointer{Pointer: "/name"},
+	&FieldPointer{Pointer: "/given_name"},
+	&FieldPointer{Pointer: "/middle_name"},
+	&FieldPointer{Pointer: "/nickname"},
+	&FieldPointer{Pointer: "/profile"},
+	&FieldPointer{Pointer: "/picture"},
+	&FieldPointer{Pointer: "/website"},
+	&FieldPointer{Pointer: "/gender"},
+	&FieldPointer{Pointer: "/birthdate"},
+	&FieldPointer{Pointer: "/zoneinfo"},
+	&FieldPointer{Pointer: "/locale"},
+	&FieldPointer{Pointer: "/address/formatted"},
+	&FieldPointer{Pointer: "/address/street_address"},
+	&FieldPointer{Pointer: "/address/locality"},
+	&FieldPointer{Pointer: "/address/region"},
+	&FieldPointer{Pointer: "/address/postal_code"},
+	&FieldPointer{Pointer: "/address/country"},
+	&FieldPointer{Pointer: "/roles"},
+	&FieldPointer{Pointer: "/groups"},
+	&FieldPointer{Pointer: "/disabled"},
+	&FieldPointer{Pointer: "/identities"},
+	&FieldPointer{Pointer: "/mfa/emails"},
+	&FieldPointer{Pointer: "/mfa/phone_numbers"},
+	&FieldPointer{Pointer: "/mfa/totps"},
+	&FieldPointer{Pointer: "/biometric_count"},
+	&FieldPointer{Pointer: "/passkey_count"},
 }
-`
 
 type FieldPointer struct {
 	Pointer   string `json:"pointer,omitempty"`
--- a/pkg/lib/userexport/model.go
+++ b/pkg/lib/userexport/model.go
@@ -2,7 +2,12 @@ package userexport
 
 import (
 	"encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
 	"time"
+
+	"github.com/iawaknahc/jsonschema/pkg/jsonpointer"
 
 	"github.com/authgear/authgear-server/pkg/api/apierrors"
 	"github.com/authgear/authgear-server/pkg/api/model"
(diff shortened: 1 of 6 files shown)


pkg/lib/userexport/model.go
```diff
@@ -24,14 +24,89 @@ var RequestSchema = validation.NewSimpleSchema(`
...
+		{"pointer": "/mfa/totps"},
+		{"pointer": "/biometric_count"},
+		{"pointer": "/passkey_count"}
+	]
+}
+`
+
+type FieldPointer struct {
+	Pointer   string `json:"pointer"`
+	FieldName string `json:"field_name"`
```
louischan-oursky, line 101: Add `omitempty`

pkg/lib/userexport/model.go
```diff
@@ -24,14 +24,89 @@ var RequestSchema = validation.NewSimpleSchema(`
...
+}
+
+type CSVField struct {
+	Fields []*FieldPointer `json:"fields"`
```
louischan-oursky, line 105: Add `omitempty`

pkg/lib/userexport/model.go
```diff
@@ -24,14 +24,89 @@ var RequestSchema = validation.NewSimpleSchema(`
...
+		{"pointer": "/mfa/phone_numbers"},
+		{"pointer": "/mfa/totps"},
+		{"pointer": "/biometric_count"},
+		{"pointer": "/passkey_count"}
+	]
+}
+`
+
+type FieldPointer struct {
+	Pointer   string `json:"pointer"`
```
louischan-oursky, line 100: Add `omitempty`

pkg/lib/userexport/model.go
```diff
@@ -24,14 +24,89 @@ var RequestSchema = validation.NewSimpleSchema(`
...
+}
+
+type CSVField struct {
+	Fields []*FieldPointer `json:"fields"`
+}
+
 type Request struct {
-	Format string `json:"format"`
+	Format string   `json:"format"`
+	CSV    CSVField `json:"csv"`
```
louischan-oursky, line 110: Add `omitempty` and change the type to `*CSVField`

In general, we do not serialize `null` unless it has a meaning.

pkg/lib/userexport/model.go
```diff
@@ -24,14 +24,89 @@ var RequestSchema = validation.NewSimpleSchema(`
...
+}
+
+type CSVHeaderFieldNames struct {
```
louischan-oursky, line 113: Is this used? If not, remove it.

pkg/lib/userexport/model_test.go
```diff
@@ -26,7 +26,7 @@ func TestRequest(t *testing.T) {
 		serialized, err := json.Marshal(request)
 		So(err, ShouldBeNil)
 
-		So(string(serialized), ShouldEqualJSON, requestBody)
+		So(string(serialized), ShouldEqualJSON, `{"csv":{"fields":null},"format":"ndjson"}`)
```
louischan-oursky, line 29: This test is good. It shows a bug. If we serialize the request, then `"csv"` should be omitted if it is unspecified originally.

pkg/lib/userexport/model_test.go
```diff
@@ -65,5 +65,40 @@ func TestRequest(t *testing.T) {
...
   map[actual:unknown format expected:[ndjson csv]]`)
+
+		test(`
+{
+	"format": "csv",
+	"csv": {
+		"fields": []
+	}
+}
+		`, "")
```
louischan-oursky, line 76: This test is good. But the result is wrong. If fields is specified, then it must be an array with at least one item. 

https://github.com/authgear/authgear-server/pull/4717/files#diff-5d0c564f0dbed14ca1655b1752d72835c8115dbe937c0a20f830584f1ff8edefR59

> If this is specified, then it must be an non-empty list.

So your fix is add `minItems: 1`

pkg/lib/userexport/model_test.go
```diff
@@ -65,5 +65,40 @@ func TestRequest(t *testing.T) {
...
+		test(`
+{
+	"format": "csv",
+	"csv": {
+		"fields": [{ "pointer": "/sub", "field_name": "user_id" }]
+	}
+}
+		`, "")
```
louischan-oursky, line 102: Add a test 

```
{
	"format": "csv",
	"csv": {
		"fields": [{ "pointer": "/foobar" }]
	}
}
```

The error should show /foobar is not a valid value of an enum.

The schema is not strict enough. You need to specify all possible pointers as enum.

pkg/lib/userexport/model.go
```diff
@@ -24,14 +24,89 @@ var RequestSchema = validation.NewSimpleSchema(`
...
+		{"pointer": "/disabled"},
+		{"pointer": "/identities"},
+		{"pointer": "/mfa/emails"},
+		{"pointer": "/mfa/phone_numbers"},
+		{"pointer": "/mfa/totps"},
+		{"pointer": "/biometric_count"},
+		{"pointer": "/passkey_count"}
+	]
+}
+`
```
louischan-oursky, line 97: I think you need to refactor this code a bit.

Here are what we want to do

1. We want the schema to include all possible pointers as an enum.
2. We want to have a list of default fields when `fields` is unspecified.
3. And we want to use a SINGLE list, to avoid out-of-sync issue.

One way you can do that is to build `RequestSchema` with `SchemaBuilder` in `func init()`

pkg/lib/userexport/model.go
```diff
@@ -226,3 +231,77 @@ func NewResponseFromTask(task *redisqueue.Task) (*Response, error) {
 
 	return response, nil
 }
+
+func jsonPointerToCSVFieldName(pointer string) (fieldName string) {
+	splitFn := func(c rune) bool {
+		return c == '/'
+	}
+	fields := strings.FieldsFunc(pointer, splitFn)
+	return strings.Join(fields, ".")
```
louischan-oursky, line 240: You can just parse the pointer with `jsonpointer.Parse`, it gives you a slice of strings. Then you call `strings.Join(referenceTokens, ".")`. Write test for this function.

pkg/lib/userexport/model.go
```diff
@@ -226,3 +231,77 @@ func NewResponseFromTask(task *redisqueue.Task) (*Response, error) {
...
+		fields = append(fields, fieldName)
+	}
+
+	headerFields = &CSVHeaderFieldNames{
+		FieldNames: fields,
+	}
+
+	if isDuplicated {
+		headerFieldsString, _ := json.Marshal(headerFields)
+		return nil, ErrUserExportDuplicateField.New(fmt.Sprintf("%s", headerFieldsString))
```
louischan-oursky, line 269: https://github.com/authgear/authgear-server/pull/4717/files#diff-5d0c564f0dbed14ca1655b1752d72835c8115dbe937c0a20f830584f1ff8edefR190

The spec says you need to populate the error info `field_names`
The error message should be static, use this `field names are not unique`

pkg/lib/userexport/model.go
```diff
@@ -226,3 +231,77 @@ func NewResponseFromTask(task *redisqueue.Task) (*Response, error) {
...
+	case map[string]interface{}:
+		valueJson, _ := json.Marshal(v)
+		fieldValue = string(valueJson)
+	case float64:
+		fieldValue = strconv.FormatFloat(v, 'f', -1, 64)
+	case nil:
+		fieldValue = ""
+	default:
+		fieldValue = v.(string)
+	}
```
louischan-oursky, line 304: This switch does not handle pointer to the supported types.

louischan-oursky (changes requested):

pkg/lib/userexport/model.go
mickmak-busicat: Added panic to default case and separated string case

pkg/lib/userexport/model.go
mickmak-busicat: updated duplicate field error so not used, removed

pkg/lib/userexport/model.go
mickmak-busicat: updated to this message, the duplicated field moved to error info

pkg/lib/userexport/model.go
mickmak-busicat: updated

pkg/lib/userexport/model.go
mickmak-busicat: added

pkg/lib/userexport/model.go
mickmak-busicat: added

pkg/lib/userexport/model.go
mickmak-busicat: added

pkg/lib/userexport/model.go
mickmak-busicat: added

pkg/lib/userexport/model_test.go
mickmak-busicat: added this condition

pkg/lib/userexport/model_test.go
mickmak-busicat: added omitempty to csv field

pkg/lib/userexport/model_test.go
mickmak-busicat: as discussed, some traverse pointer like "/identities/0/claims/email" cannot pass enum validation
so skipped this implementation

pkg/lib/userexport/model.go
mickmak-busicat: as discussed, some traverse pointer like "/identities/0/claims/email" cannot pass enum validation
so skipped this implementation

mickmak-busicat:
@louischan-oursky fixing other comment 🙏

pkg/lib/userexport/model.go
```diff
@@ -24,14 +29,86 @@ var RequestSchema = validation.NewSimpleSchema(`
...
+					}
+				}
+			}
 		}
 	},
 	"required": ["format"]
 }
 `)
 
+const DefaultCSVExportField = `
```
louischan-oursky, line 61: Please make these changes

1. Change this variable to private.
2. Instead of making this a JSON document to be deserialized each time, initialize this to be a slice of the struct.

louischan-oursky (changes requested):

pkg/lib/userexport/model.go
mickmak-busicat: updated to use slice of struct

PR #4777, opened 2024-09-26
Use substring match to filter posthog type error

pkong-ds:
ref DEV-1767

## Notes
Actually not sure how to reproduce [this](https://authgear.sentry.io/issues/5897586517/?alert_rule_id=15285121&alert_timestamp=1727150257207&alert_type=email&environment=production&notification_uuid=904c3b4b-2df3-4f8d-912c-49c1e67af8d3&project=4507492100014080&referrer=alert_email)

Decided to use a substring match to also catch error like this.

Question - maybe I should just filter all errors with `posthog` in it? 🤔 Since we plan to handle it properly in [this issue](https://linear.app/authgear/issue/DEV-2131/integrate-gtm-posthog-osano-properly) anyway.

cc @tung2744

### DIFF
--- a/portal/src/ReactApp.tsx
+++ b/portal/src/ReactApp.tsx
@@ -83,10 +83,7 @@ async function loadSystemConfig(): Promise<SystemConfig> {
 }
 
 function isPosthogResetGroupsException(ex: SentryException) {
-  return (
-    ex.type === "TypeError" &&
-    ex.value === "posthog.resetGroups is not a function"
-  );
+  return ex.type === "TypeError" && ex.value?.includes("posthog.resetGroups");
 }
 function isPosthogResetGroupsEvent(event: SentryEvent) {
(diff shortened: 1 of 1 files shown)


portal/src/ReactApp.tsx
```diff
@@ -85,7 +85,7 @@ async function loadSystemConfig(): Promise<SystemConfig> {
 function isPosthogResetGroupsException(ex: SentryException) {
   return (
     ex.type === "TypeError" &&
-    ex.value === "posthog.resetGroups is not a function"
+    ex.value?.includes("posthog.resetGroups is not a function")
```
tung2744, line 88: Maybe includes `posthog.resetGroups`?

pkong-ds: Updated 🙏

tung2744 (approved):

PR #4778, opened 2024-09-26
Support Redis 6.0 explicitly

louischan-oursky:
ref DEV-2129

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -15,8 +15,8 @@ services:
       - "5432:5432"
 
   redis:
-    # redis:6.0 crashed randomly when I run k6 load test.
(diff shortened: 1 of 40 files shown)


louischan-oursky:
Manually test regression for

1. rate limit
2. lockout
3. passkey

Because they involve some code changes instead of just changing the type of `conn`.

Found a regression of lockout, but it is not related to my changes. That was also fixed in this PR.

tung2744:
> Found a regression of lockout, but it is not related to my changes. That was also fixed in this PR.

Which commit is it?

louischan-oursky:
Sorry, forgot to push.

tung2744 (approved):

PR #4779, opened 2024-09-27
Add quota to user import/export job

kiootic:
ref DEV-2104

### DIFF
--- a/pkg/lib/usage/limit.go
+++ b/pkg/lib/usage/limit.go
@@ -10,6 +10,8 @@ import (
 	"github.com/authgear/authgear-server/pkg/lib/infra/redis/appredis"
 	"github.com/authgear/authgear-server/pkg/util/clock"
 	"github.com/authgear/authgear-server/pkg/util/log"
+
+	goredis "github.com/go-redis/redis/v8"
 )
 
 type Logger struct{ *log.Logger }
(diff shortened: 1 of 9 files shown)


pkg/lib/usage/limit.go
```diff
@@ -38,35 +78,31 @@ func (l *Limiter) getResetTime(c *config.UsageLimitConfig) time.Time {
...
-		usage, err := conn.IncrBy(ctx, key, 1).Result()
-		if err != nil {
-			return err
-		}
-
-		tokens = int64(quota) - usage
-
-		resetTime := l.getResetTime(config)
-		// Ignore error
-		_, _ = conn.PExpireAt(ctx, key, resetTime).Result()
```
louischan-oursky, line 61: We used to use PEXPIREAT instead of EXPIREAT. Did we make a mistake here?

louischan-oursky (changes requested):

pkg/lib/usage/limit.go
kiootic: PEXPIREAT uses millisecond precision and EXPIREAT uses second precision. Since the reset time is always truncated to days, the behavior is same.

louischan-oursky: I get it. You call `t.Unix()` before passing the timestamp to the Lua script.

louischan-oursky (approved):

PR #4782, opened 2024-09-30
Handle admin_api reset password in authflow v2

pkong-ds:
ref DEV-2048

## What's in this PR?
This PR
- generates `admin_reset_password` flow as an account_recovery flow by default
- uses said flow when the account_recovery_code has `FlowReference.Name == "admin_reset_password"`

## Preview

https://github.com/user-attachments/assets/c50df1e8-1456-4506-a3cb-0180aae3f569

### DIFF
--- a/pkg/auth/handler/webapp/authflowv2/reset_password.go
+++ b/pkg/auth/handler/webapp/authflowv2/reset_password.go
@@ -43,7 +43,7 @@ type AuthflowV2ResetPasswordHandler struct {
 	Renderer      handlerwebapp.Renderer
 }
 
-func (h *AuthflowV2ResetPasswordHandler) GetData(w http.ResponseWriter, r *http.Request, screen *webapp.AuthflowScreenWithFlowResponse) (map[string]interface{}, error) {
+func (h *AuthflowV2ResetPasswordHandler) GetAuthflowData(w http.ResponseWriter, r *http.Request, screen *webapp.AuthflowScreenWithFlowResponse) (map[string]interface{}, error) {
 	data := make(map[string]interface{})
 
 	baseViewModel := h.BaseViewModel.ViewModelForAuthFlow(r, w)
@@ -64,7 +64,7 @@ func (h *AuthflowV2ResetPasswordHandler) GetData(w http.ResponseWriter, r *http.
 	return data, nil
 }
 
-func (h *AuthflowV2ResetPasswordHandler) GetErrorData(w http.ResponseWriter, r *http.Request, err error) (map[string]interface{}, error) {
+func (h *AuthflowV2ResetPasswordHandler) GetAuthflowErrorData(w http.ResponseWriter, r *http.Request, err error) (map[string]interface{}, error) {
 	data := make(map[string]interface{})
 
 	baseViewModel := h.BaseViewModel.ViewModelForAuthFlow(r, w)
@@ -75,10 +75,22 @@ func (h *AuthflowV2ResetPasswordHandler) GetErrorData(w http.ResponseWriter, r *
 }
 
 func (h *AuthflowV2ResetPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if isURLFromAdminAPI(r) {
+		h.serveHTTPNonAuthflow(w, r)
+	} else {
+		h.serveHTTPAuthflow(w, r)
+	}
+}
+
+func (h *AuthflowV2ResetPasswordHandler) serveHTTPNonAuthflow(w http.ResponseWriter, r *http.Request) {
+	// TODO: implement non authflow reset password handler
+}
+
+func (h *AuthflowV2ResetPasswordHandler) serveHTTPAuthflow(w http.ResponseWriter, r *http.Request) {
 	var handlers handlerwebapp.AuthflowControllerHandlers
 	handlers.Get(func(s *webapp.Session, screen *webapp.AuthflowScreenWithFlowResponse) error {
 
-		data, err := h.GetData(w, r, screen)
+		data, err := h.GetAuthflowData(w, r, screen)
 		if err != nil {
 			return err
 		}
@@ -115,7 +127,7 @@ func (h *AuthflowV2ResetPasswordHandler) ServeHTTP(w http.ResponseWriter, r *htt
 		if !apierrors.IsKind(err, forgotpassword.PasswordResetFailed) {
 			return err
 		}
-		data, err := h.GetErrorData(w, r, err)
+		data, err := h.GetAuthflowErrorData(w, r, err)
 		if err != nil {
 			return err
 		}
@@ -132,4 +144,10 @@ func (h *AuthflowV2ResetPasswordHandler) ServeHTTP(w http.ResponseWriter, r *htt
 	} else {
 		h.Controller.HandleStep(w, r, &handlers)
 	}
+}
+
+var fromAdminAPIQueryKey = "x_from_admin_api"
+
+func isURLFromAdminAPI(r *http.Request) bool {
+	return r.URL.Query().Get(fromAdminAPIQueryKey) == "true"
 }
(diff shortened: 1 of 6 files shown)


tung2744:
As discussed in huddle,
1.  Inject a new query param in the reset password link generated from admin api
2. In reset password screen, run another set of logic to reset the password instead of executing authflow if the new query exist.

pkong-ds:
> Inject a new query param in the reset password link generated from admin api
    In reset password screen, run another set of logic to reset the password instead of executing authflow if the new query exist.

Updated as this, but seems `http.Redirect` not working 😢 lemme continue on wed 🙏

pkg/auth/handler/webapp/authflowv2/reset_password.go
```diff
@@ -75,10 +110,87 @@ func (h *AuthflowV2ResetPasswordHandler) GetErrorData(w http.ResponseWriter, r *
...
+			return err
+		}
+		newPassword := r.Form.Get("x_password")
+		confirmPassword := r.Form.Get("x_confirm_password")
+		err = pwd.ConfirmPassword(newPassword, confirmPassword)
+		if err != nil {
+			return err
+		}
+
+		err = h.Database.WithTx(func() error {
```
pkong-ds, line 162: @tung2744 Would love your view on this 🙏 In this `admin-api only reset password screen`, I encountered `hook-handle: transaction not started` error and fixed it by injecting a database handle to this webapp handler. 

Do you see any potential problems with this? Mine: Maybe too many concurrent database txns if many user reset password through admin api at once -> huge load to db?

Maybe I should inject the db handle to a higher level service? 🤔 

Thanks :adore:

tung2744: I think starting the transaction at handler level is correct. And I think it is already the highest level.

pkong-ds: I see i see thanks!

pkg/auth/handler/webapp/authflowv2/reset_password.go
```diff
@@ -75,10 +110,87 @@ func (h *AuthflowV2ResetPasswordHandler) GetErrorData(w http.ResponseWriter, r *
...
+			var err error
+			if code == "" {
+				err = newCodeRequiredErr()
+			} else {
+				err = handler(w, r)
+			}
+			if err != nil {
+				if apierrors.IsAPIError(err) {
+					// Still uses AuthflowController to render error because of same logic
+					h.Controller.ErrorRenderer.RenderError(w, r, err)
```
tung2744, line 133: Inject `ErrorRenderer` to the handle directly and use it, instead of access it through authflow controller.

pkg/auth/handler/webapp/authflowv2/reset_password.go
```diff
@@ -75,10 +110,87 @@ func (h *AuthflowV2ResetPasswordHandler) GetErrorData(w http.ResponseWriter, r *
...
+	if isURLFromAdminAPI(r) {
+		h.serveHTTPNonAuthflow(w, r)
+	} else {
+		h.serveHTTPAuthflow(w, r)
+	}
+}
+
+func (h *AuthflowV2ResetPasswordHandler) serveHTTPNonAuthflow(w http.ResponseWriter, r *http.Request) {
+	code := r.URL.Query().Get("code")
+	makeHTTPHandler := func(handler func(w http.ResponseWriter, r *http.Request) error) http.Handler {
```
tung2744, line 122: Use the webapp controller instead of writing a new handler pattern. You can check pkg/auth/handler/webapp/authflowv2/select_account.go

pkg/auth/handler/webapp/authflowv2/reset_password.go
```diff
@@ -133,3 +245,15 @@ func (h *AuthflowV2ResetPasswordHandler) ServeHTTP(w http.ResponseWriter, r *htt
...
+
+var fromAdminAPIQueryKey = "x_from_admin_api"
+
+func isURLFromAdminAPI(r *http.Request) bool {
+	return r.URL.Query().Get(fromAdminAPIQueryKey) == "true"
+}
+
+func newCodeRequiredErr() *apierrors.APIError {
+	apiErr := apierrors.AsAPIError(fmt.Errorf("code required in admin_api reset password page"))
+	apiErr.Reason = "CodeRequiredInAdminAPIResetPasswordPage"
```
tung2744, line 257: Should reuse the existing `PasswordResetFailed` error.

pkg/auth/handler/webapp/authflowv2/reset_password_success.go
```diff
@@ -44,7 +45,61 @@ func (h *AuthflowV2ResetPasswordSuccessHandler) GetData(w http.ResponseWriter, r
...
+	switch r.Method {
+	case "GET":
+		getHandler.ServeHTTP(w, r)
+	default:
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+}
+
+func (h *AuthflowV2ResetPasswordSuccessHandler) serveHTTPAuthflow(w http.ResponseWriter, r *http.Request) {
```
tung2744, line 102: Originally, it is already using `HandleWithoutFlow` which does not assume it is in an authflow, so I think we don't need to modify this handler.

pkg/auth/handler/webapp/authflowv2/reset_password.go
pkong-ds: I see. Seems no need if we use `webapp controller` just like `pkg/auth/handler/webapp/authflowv2/select_account.go`

pkg/lib/deps/deps_provider.go
```diff
@@ -88,6 +90,7 @@ var AppRootDeps = wire.NewSet(
 	wire.Bind(new(event.Database), new(*appdb.Handle)),
 	wire.Bind(new(workflow.ServiceDatabase), new(*appdb.Handle)),
 	wire.Bind(new(authenticationflow.ServiceDatabase), new(*appdb.Handle)),
+	wire.Bind(new(handlerwebappauthflowv2.ResetPasswordHandlerDatabase), new(*appdb.Handle)),
```
tung2744, line 93: Try to move this to auth/deps.go to avoid the lint error?

tung2744:
> EDIT: found out it's our own importlinter 🤔 do you know what was it for?

It is for maintaining a certain import structure in the project. See Makefile lint for the rules.
And lib is not allowed to import the auth package.

pkg/lib/deps/deps_provider.go
pkong-ds: Thanks, it worked 🙇 curious - does it matter (in terms of code structure) where we declare these wire DI?

If not - why not just put all wire bindings in one holy file? Since not these `deps` file are too big to be readable anyway

pkong-ds: > for maintaining a certain import structure

Oh, thanksss 🙏

pkg/lib/authn/otp/sender.go
```diff
@@ -105,6 +106,9 @@ func (s *MessageSender) setupTemplateContext(msg *PreparedMessage, opts SendOpti
 			linkURL = s.Endpoints.ResetPasswordEndpointURL()
 			query := linkURL.Query()
 			query.Set("code", opts.OTP)
+			if opts.IsAdminAPIResetPassword {
+				query.Set("x_from_admin_api", "true")
```
tung2744, line 110: Make this a constant, and use it in webapp

pkong-ds: Updated, thanks!

tung2744 (approved):
LGTM. See if @louischan-oursky would like to take a look, or I will merge it tmr day end.

louischan-oursky:
LGTM.

louischan-oursky (approved):

PR #4791, opened 2024-10-03
Run user import job as multiple tasks

kiootic:
ref DEV-2105

### DIFF
--- a/pkg/redisqueue/consumer.go
+++ b/pkg/redisqueue/consumer.go
@@ -221,8 +221,46 @@ func (c *Consumer) work(ctx context.Context) {
 		}
 	}
 
+	// Reserve before dequeue.
+	// If we do it in reverse order, we could dequeue the job, but rate limited.
+	// And the server shuts down, the job will be gone forever.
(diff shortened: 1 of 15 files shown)


pkg/redisqueue/consumer.go
```diff
@@ -221,26 +221,46 @@ func (c *Consumer) work(ctx context.Context) {
...
 		return
 	}
 
-	c.dequeueBackoff.Reset()
+	c.logger.
+		WithField("tat", reservation.GetTimeToAct()).
+		WithField("bucket_key", c.limitBucket.Key()).
+		Debug("consume reservation")
 
-	if r := c.limiter.Reserve(c.limitBucket); r.Error() != nil {
```
kiootic, line 235: @louischan-oursky It's intentional to reserve after dequeue, otherwise every worker would eventually proceed to dequeue & run tasks simultaneously given enough time passed while queue is empty, due to the nature of rate limit token refill.

louischan-oursky (approved):

PR #4793, opened 2024-10-03
SAML: Add e2e tests for SSO flow

tung2744:
ref DEV-1830


e2e/tests/saml/sso.test.yaml
```diff
@@ -1,9 +1,43 @@
 ---
 name: SAML SSO
 focus: true
```
louischan-oursky: Remember to remove focus: true

pkg/auth/handler/webapp/csrf_middleware.go
```diff
@@ -36,9 +36,13 @@ type CSRFMiddleware struct {
 	BaseViewModel           *viewmodels.BaseViewModeler
 	Renderer                Renderer
 	UIImplementationService CSRFMiddlewareUIImplementationService
+	EnvironmentConfig       *config.EnvironmentConfig
 }
 
 func (m *CSRFMiddleware) Handle(next http.Handler) http.Handler {
+	if m.EnvironmentConfig.End2EndCSRFProtectionDisabled {
+		return next
+	}
```
louischan-oursky, line 45: Is it possible that we parse the HTML to get the CSRF token in the HTML so that we do not need to introduce this flag?

louischan-oursky (changes requested):

pkg/auth/handler/webapp/csrf_middleware.go
tung2744: I think it is possible, but I think csrf in e2e is not very useful so I simply disable it to reduce noises in the tests.

louischan-oursky: It would be useful if we have e2e tests (headless browser-based ones) on authui. Anyway, let's keep this for now.

louischan-oursky (approved):

PR #4799, opened 2024-10-07
Add yaml-language-server to contribution docs

pkong-ds:
## Notes
Or should we actually host these schemas through this repo?

## Preview
image

## What's in this PR?

Allow local editing of `authgear.yaml` and `authgear.secrets.yaml` have said features

> 
    YAML validation:
        Detects whether the entire file is valid yaml
    Validation:
        Detects errors such as:
            Node is not found
            Node has an invalid key node type
            Node has an invalid type
            Node is not a valid child node
        Detects warnings such as:
            Node is an additional property of parent
    Auto completion:
        Auto completes on all commands
        Scalar nodes autocomplete to schema's defaults if they exist
    Hover support:
        Hovering over a node shows description if available
    Document outlining:
        Shows a complete document outline of all nodes in the document

ref https://github.com/redhat-developer/yaml-language-server?tab=readme-ov-file#features

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,6 +98,7 @@ This project uses asdf, and there is a .tool-versions file at the project root.
    `authgear.yaml` must contain the following contents for the portal to work.
 
    ```yaml
+   # yaml-language-server: $schema=../tmp/app-config.schema.json # get this by running `make export-schemas`
    id: accounts # Make sure the ID matches AUTHGEAR_APP_ID environment variable.
    http:
       # Make sure this matches the host used to access main Authgear server.


CONTRIBUTING.md
```diff
@@ -98,6 +98,7 @@ This project uses asdf, and there is a .tool-versions file at the project root.
    `authgear.yaml` must contain the following contents for the portal to work.
 
    ```yaml
+   # yaml-language-server: $schema=../tmp/app-config.schema.json # get this by running `make export-schemas`
```
louischan-oursky, line 101: This is not very useful. A URL pointing to github raw file of the schema would be better. But we currently do not track the JSON schema (because it is placed in a git-ignored directory).

pkong-ds: agree, what about below?

## Proposed

1. put at `authgear-server` root
2. put at `authgear-server/config-schemas`

```
authgear-yaml.schema.json # or keep app-config.schema.json
authgear-secrets-yaml.schema.json # or keep secrets-config.schema.json
```

louischan-oursky: It was not done because the JSON schema was written for validation only. It was not intended to be used by a language server to facilitate autocompletion.

So the first thing is to test whether the schema at its current shape is good enough. I do not have time to verify this.

pkong-ds: I see. Let's not spend effort on this now then.

Will close this PR

Issue #3626, opened 2024-01-16
Authflow V2 coding guidelines


louischan-oursky:
## We have 2 levels of theming.

- The first level is defined in `base.css`.
- The second level is defined in the component CSS.
- Take this as example https://github.com/authgear/authgear-server/pull/4662
- `--navbar__title-font-family` is the second level.
- `--typography-title-large__font-family` is the first level.

## We borrow BEM naming convention to name the CSS variables.

- Take this as example https://github.com/authgear/authgear-server/pull/4662
- The component file called `navbar.css`, so the component name is `navbar`.
- The CSS variables of this component must start with `--navbar_`.
- You define a CSS variable for each CUSTOMISABLE property. When you are in doubt, you prefer not introducing a CSS variable. We can always add it later.
- `--navbar__some-meaning-css-property`
- `--navbar__some-meaning-css-property--variant`

## HTML semantics is important. 

- Use appropriate elements. https://github.com/authgear/authgear-server/pull/3614#discussion_r1451976049 https://github.com/authgear/authgear-server/pull/3613#discussion_r1452898325
- Remove unnecessary divs. https://github.com/authgear/authgear-server/pull/3613#discussion_r1452898198

## Use `gap` for spacing.

See https://github.com/authgear/authgear-server/pull/3613#discussion_r1452897671

## Use `ltr:` and `rtl:` when you use explicit left/right.

See https://github.com/authgear/authgear-server/pull/4662#discussion_r1735509674

louischan-oursky:
Follow up issues

- https://github.com/authgear/authgear-server/issues/3628
- https://github.com/authgear/authgear-server/issues/3630

louischan-oursky:
@tung2744 @IniZio @Ngkaokis Added Point 11.

louischan-oursky:
Added Point 12.

fungc-io:
@louischan is this documented? can I close it?

louischan-oursky:
Let me close this. We can refer to this issue when we need it in the future.

rickmak:
We should move the agreed content to `CONTRIBUTING.md`.

PR #4804, opened 2024-10-07
[Settings v2] Use history back for settings v2 pages navbar

IniZio:
ref DEV-2094

Initially wanted to hide back button when there is no prior history. However `window.history.length` actually always return all backword + forward navigations. 

We cannot reliably check if current page is the last item in the stack


louischan-oursky:
Build failed.

resources/authgear/templates/en/web/authflowv2/__navbar.html
```diff
@@ -2,13 +2,13 @@
 <nav class="navbar">
   <h1 class="navbar__title">{{ $.Title }}</h1>
 
-  <a
-    class="navbar__item inline-flex items-center justify-center gap-x-3"
+  <button
+    class="navbar__item"
```
louischan-oursky, line 6: When javascript is disabled, this button has no effect. So when javascript is disabled, it should be hidden. The idiomatic way to do this is to add `hidden` to the class, so initially it is hidden. When the controller connects, you run `this.element.classList.remove("hidden")` to make the button visible.

louischan-oursky (changes requested):

resources/authgear/templates/en/web/authflowv2/__navbar.html
louischan-oursky: Added this to https://github.com/authgear/authgear-server/issues/3626#:~:text=Always%20think%20about%20when%20the%20case%20when%20javascript%20is%20disabled

IniZio: Fixed up

louischan-oursky (approved):

PR #4808, opened 2024-10-08
Handle JSON syntax error

louischan-oursky:
ref DEV-2152

### DIFF
--- a/pkg/util/validation/validate.go
+++ b/pkg/util/validation/validate.go
@@ -3,6 +3,7 @@ package validation
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


pkg/util/validation/validate.go
```diff
@@ -62,6 +73,16 @@ func (v *SchemaValidator) Validate(r io.Reader) error {
...
+		// It is observed that json.NewDecoder.Decode and json.Unmarshal
+		// returns different error with the input is empty.
+		// json.Unmarshal returns *json.SyntaxError while
+		// json.NewDecoder.Decode returns io.EOF
+		// https://go.dev/play/p/WHEtDYzJKTo
+		// So we convert the io.EOF here.
+		if errors.Is(err, io.EOF) {
+			return errJSONSyntaxErrorOfEmptyInput
+		}
+
```
tung2744, line 85: Why we don't simply use `newInvalidJSON`?

louischan-oursky: newInvalidJSON takes *json.SyntaxError. We do not have a *json.SyntaxError here.

tung2744 (approved):

PR #4792, opened 2024-10-03
[Settings v2] Manage MFA - Without Recovery Code

IniZio:
ref DEV-1964
ref DEV-1967
ref DEV-1971
ref DEV-1961

CI: https://github.com/IniZio/authgear-server/pull/1

Dependencies:
- [x] https://github.com/authgear/authgear-server/pull/4784
- [ ] https://github.com/authgear/authgear-server/pull/4786
- [ ] https://github.com/authgear/authgear-server/pull/4790

Demo:

**TOTP**

https://github.com/user-attachments/assets/b4c46e9d-1fd4-414c-a0b7-6b3e9627c382

**Email**

https://github.com/user-attachments/assets/bf1d6ed9-0c2a-47e0-bb3f-d48e1b5a5b00

**OTP**

https://github.com/user-attachments/assets/20333273-8f2d-4802-ba2c-5c297ea30b9c

**Password**

https://github.com/user-attachments/assets/1c37aeaf-cf92-486b-8ca1-6f846c1f1f18

**Change Password**

https://github.com/user-attachments/assets/dc92b22f-ca4d-44d9-b276-73842b66de38

PR #4813, opened 2024-10-10
[Settings v2] Manage Social Accounts

IniZio:
ref DEV-1860
ref DEV-2165

[CI](https://github.com/IniZio/authgear-server/tree/DEV-1860-settings-oauth-1)

Dependencies
- [x] https://github.com/authgear/authgear-server/pull/4792

https://github.com/user-attachments/assets/bd534467-8e58-4374-8523-0adcb4b57c86


pkg/auth/handler/webapp/sso_callback.go
```diff
@@ -38,6 +41,22 @@ func (h *SSOCallbackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
+	s := session.GetSession(r.Context())
+
+	if state.AccountManagementToken != "" {
+		_, err = h.AccountManagement.FinishAddingIdentityOAuth(s, &accountmanagement.FinishAddingIdentityOAuthInput{
+			Token: state.AccountManagementToken,
+			Query: r.URL.Query().Encode(),
+		})
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
```
louischan-oursky, line 52: We should handle the error properly.

1. The error could be an API error. For example, linking to a OAuth account that is already associated with an existing Authgear user. This should be handled by showing the error in the original route where the add operation is triggered.
2. The error could be an OAuth error produced by the OAuth provider, such as complaining the client secret is invalid. Let's see how is this error handled in v1.
3. The error could be an I/O error. This error can be panic to let the panic middleware to handle.

louischan-oursky (changes requested):

pkg/auth/handler/webapp/sso_callback.go
IniZio: Updated to use authflow like error handling

*API error e.g. conflict identity*
Screen Shot 2024-10-14 at 19 42 45

*Unknown error e.g. secret error*
Screen Shot 2024-10-14 at 19 42 38

resources/authgear/templates/en/web/authflowv2/__error.html
```diff
@@ -156,8 +156,23 @@
         <span>{{ include "v2.error.developer-email-required" nil }}</span>
       {{ end }}
     {{ else if eq .Error.reason "AccountManagementDuplicatedIdentity" }}
-      {{ $dict := (dict "IdentityTypeIncoming" "other") }}
-      <span>{{ include "v2.error.account-conflict" $dict }}</span>
+      {{ $info := $.Error.info }}
+      {{ $dict :=
+        (dict
+          "FlowType" $info.FlowType
```
louischan-oursky, line 162: There is no FlowType. Remove this.

pkg/auth/handler/webapp/sso_callback.go
```diff
@@ -44,16 +47,25 @@ func (h *SSOCallbackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
 		_, err = h.AccountManagement.FinishAddingIdentityOAuth(s, &accountmanagement.FinishAddingIdentityOAuthInput{
 			Token: state.AccountManagementToken,
 			Query: r.URL.Query().Encode(),
 		})
 		if err != nil {
-			http.Error(w, err.Error(), http.StatusBadRequest)
+			if identity.IsErrDuplicatedIdentity(err) {
+				err = accountmanagement.NewErrAccountManagementDuplicatedIdentity(err)
+			}
```
louischan-oursky, line 62: Why do we need to translate error here? Shouldn't this be done by AccountManagementService?

louischan-oursky (changes requested):

louischan-oursky (changes requested):
SCR-20241015-nluo
SCR-20241015-nlvk

This account has two identities, one email login ID, one google login. But it does not allow to remove the google login

IniZio:
@louischan-oursky Updated to check other identity types as well 🙏

louischan-oursky (approved):

PR #4820, opened 2024-10-10
[Settings v2] Fix new timespans not formatted in settings

IniZio:
ref DEV-2166

CI: https://github.com/IniZio/authgear-server/pull/8

### DIFF
--- a/pkg/auth/handler/webapp/authflowv2/settings_biometric.go
+++ b/pkg/auth/handler/webapp/authflowv2/settings_biometric.go
@@ -53,27 +53,21 @@ func (h *AuthflowV2SettingsBiometricHandler) GetData(r *http.Request, rw http.Re
 	baseViewModel := h.BaseViewModel.ViewModel(r, rw)
 	viewmodels.Embed(data, baseViewModel)
 
-	// BiometricViewModel
-	var biometricIdentityInfos []*identity.Biometric
-	err := h.Database.WithTx(func() (err error) {
-		// SettingsViewModel
-		settingsViewModel, err := h.SettingsViewModel.ViewModel(*userID)
-		if err != nil {
-			return err
-		}
-		viewmodels.Embed(data, *settingsViewModel)
-
-		biometricIdentityInfos, err = h.BiometricProvider.List(*userID)
-		if err != nil {
-			return err
-		}
-		return nil
-	})
+	// SettingsViewModel
+	settingsViewModel, err := h.SettingsViewModel.ViewModel(*userID)
 	if err != nil {
 		return nil, err
 	}
+	viewmodels.Embed(data, *settingsViewModel)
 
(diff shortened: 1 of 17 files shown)


louischan-oursky:
Please rebase.

louischan-oursky (changes requested):
1. We want to keep the "API" unchanged. That means we should not have any changes to the translation string. The translation strings should continue to work.
2. Instead, we should fix the controller to make it able to detect new date spans to be formatted. This can be done with mutation observer.
3. Instead of fixing format-date-relative, we want to introduce a new Controller that lives in authflowv2/ (so that it is linted by eslint), that is based on format-date-relative, with mutation observer, and respect to user timezone.

pkg/auth/handler/webapp/authflowv2/settings_biometric.go
```diff
@@ -90,7 +98,11 @@ func (h *AuthflowV2SettingsBiometricHandler) ServeHTTP(w http.ResponseWriter, r
 	defer ctrl.ServeWithoutDBTx()
 
 	ctrl.Get(func() error {
-		data, err := h.GetData(r, w)
+		var data map[string]interface{}
+		err := h.Database.WithTx(func() error {
+			data, err = h.GetData(r, w)
+			return err
+		})
```
louischan-oursky, line 105: I suggest being consistant, and use WithTx here, instead of in GetData. Please change all other occurrences.

louischan-oursky (changes requested):

resources/authgear/templates/en/web/authflowv2/__base_page_frame.html
```diff
@@ -27,7 +27,7 @@
   data-controller="prevent-double-tap lockout restore-form loading authflow-passkey-error bot-protection"
   data-restore-form-json-value="{{ $.FormJSON }}"
   data-action="dblclick->prevent-double-tap#action"
-  data-lockout-format-date-relative-outlet="[data-controller='format-date-relative']"
+  data-lockout-format-date-outlet="[data-controller='format-date']"
```
louischan-oursky, line 30: We cannot simply change this. If you look at the source code of Lockout Controller, you see it imports the old controller directly, so it has a hard dependency on the old controller.

I suggest we make the lockout controller NOT to depend on other controller. Maybe we can just simply remove the dependency.

louischan-oursky (changes requested):

resources/authgear/templates/en/web/authflowv2/__base_page_frame.html
IniZio: Tried lockout controller behaves normally without the outlet.

However found another issue where the lockout overlay does not work without js. Will create another issue to follow up

louischan-oursky (approved):

PR #4825, opened 2024-10-15
[Settings v2] Fix country, string, number and dropdown fields in custom attribute

IniZio:
ref DEV-1914
ref DEV-1985

[CI](https://github.com/IniZio/authgear-server/tree/DEV-1914)


louischan-oursky (changes requested):
1. Some data type shows `Not Provided` even it does have value set. Number of zeros are shown as `Not provided`.
SCR-20241016-lxfq
2. string cannot be saved. The save button is always disabled.
SCR-20241016-lxln

louischan-oursky:
Per discussion, the changes of https://github.com/authgear/authgear-server/pull/4829 will be included in this PR too. And the fix to "Not Provided" will be included in this PR.

resources/authgear/templates/en/web/authflowv2/settings_profile.html
```diff
@@ -173,9 +173,12 @@
     {{ $url = $.EditURL }}
   {{ end }}
 
-  {{ $content := (translate "v2.page.settings-profile.default.attribute-value-unspecified-label" nil) }}
-  {{ if $.Content }}
-    {{ $content = $.Content }}
+  {{ $content := $.Content }}
+  {{ if (eq $.Content nil) }}
```
louischan-oursky, line 177: I remember we use `isNil` to check `nil`

louischan-oursky: f1b0f7edee5bc74666b42a10e0f0e03c228553ed

IniZio: Updated 🙏

PR #4827, opened 2024-10-15
SAML Portal: Support query and mutate saml secrets

tung2744:
ref DEV-1706

TODO:

- [x] Update tests


pkg/lib/config/secret_update_instruction.go
```diff
@@ -569,9 +577,66 @@ func (i *BotProtectionProviderCredentialsUpdateInstruction) unset(currentConfig
...
+	}
+
+	if credentials == nil {
+		credentials = &SAMLIdpSigningMaterials{
+			Certificates: []*SAMLIdpSigningCertificate{},
+		}
+	}
+
+	// FIXME: The common name is not important, but we might want to put the app id here.
+	newCert, err := GenerateSAMLIdpSigningCertificate("urn:authgear.com")
```
louischan-oursky: What value did we use previously?

pkg/lib/config/configsource/resources.go
```diff
@@ -538,6 +538,10 @@ func (d AuthgearSecretYAMLDescriptor) UpdateResource(ctx context.Context, _ []re
 		// Since the key neither use for sig nor enc
 		GenerateClientSecretOctetKeyFunc: secrets.GenerateOctetKey,
 		GenerateAdminAPIAuthKeyFunc:      secrets.GenerateRSAKey,
+		GenerateSAMLIdpSigningCertificate: func() (*config.SAMLIdpSigningCertificate, error) {
+			// FIXME: The common name is not important, but we might want to put the app id here.
```
louischan-oursky, line 542: What value does other competitor put in here?

pkg/lib/config/secret_update_instruction.go
tung2744: For existing projects, I created the certs manually with a value like `urn:project_id.authgear.com`

pkg/lib/config/configsource/resources.go
tung2744: It is something like an id.

louischan-oursky: I suggest using `APP_HOST_SUFFIX` (which is something like `.authgear.cloud`) to derive the common name here.

tung2744: Updated to use the saml entity id here.

louischan-oursky (approved):

PR #4835, opened 2024-10-16
Rename store_grant_mock.go to store_grant_mock_test.go

louischan-oursky:
`make fmt` is programmed to ignore `_mock_test.go`. So this generated file should be named `store_grant_mock_test.go`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,6 @@ ensure-important-modules-up-to-date:
 generate:
 	go generate ./pkg/... ./cmd/...
 	cd ./e2e && go generate ./...
-	$(MAKE) fmt
 
 .PHONY: test
 test:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -45,7 +45,6 @@ ensure-important-modules-up-to-date:
 generate:
 	go generate ./pkg/... ./cmd/...
 	cd ./e2e && go generate ./...
-	$(MAKE) fmt
```
tung2744, line 48: I remember the reason we run fmt after go generate is fmt will reorder the imports in a different way with go generate.

louischan-oursky: Yes, it is because the fixed file is not ignored by `make fmt`. Now we have fixed the file, `make fmt` will no longer try to format generated files, so we do not need to run `make fmt`

tung2744 (approved):

PR #4844, opened 2024-10-17
SAML/Portal: Implement Service Provider settings

tung2744:
ref DEV-1706

Known issue:
- Cannot delete client if it has saml certificates, will fix in later prs.

PR #4849, opened 2024-10-21
SAML/Portal: Implement Idp related section and screens

tung2744:
ref DEV-1706

requires https://github.com/authgear/authgear-server/pull/4844


tung2744:
Updated saml certificate screen so that no save is needed:
https://linear.app/authgear/issue/DEV-1706/saml-setup-in-portal#comment-e85b3694

You might also want to pay attention to this commit:
https://github.com/authgear/authgear-server/pull/4849/commits/8042ea516f24c50bb1767b9088869be09f2764ae

In this commit, I introduced `useEvent` and `useAsyncSetState` so that `save()` in form always reference the correct state even we save immediately after setState.

tung2744:
Pushed one more commit so that we don't need those additional hooks mentioned above.

louischan-oursky:
SCR-20241023-ljre

When the form state is like the above, pressing save will not succeed. The error is shown at the top, but the message is empty.

A suggestion: Maybe we should disable the "Check message signature" if there is no certificate. And we change the warning to `To enable "Check message signature", please provide at least one certificate`, then we do not need to handle this error.

tung2744:
Updated according to the comment above.
Screenshot 2024-10-23 at 2 16 56 PM
@fungc-io See if this is ok?

louischan-oursky (approved):

PR #4858, opened 2024-10-24
Prevent routes from appearing in search engines

tung2744:
ref DEV-290


pkg/util/httputil/noindex.go
```diff
@@ -0,0 +1,12 @@
+package httputil
+
+import (
+	"net/http"
+)
+
+func NoIndex(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("X-Robots-Tag", "noindex")
```
louischan-oursky, line 9: `noindex, nofollow`

resources/authgear/templates/en/web/__html_head.html
```diff
@@ -3,6 +3,7 @@
 <meta charset="UTF-8">
 <title>{{ template "app.name" }}</title>
 <link rel="shortcut icon" href="{{ call $.StaticAssetURL "favicon" }}">
+<meta name="robots" content="noindex">
```
louischan-oursky, line 6: `noindex, nofollow`

resources/authgear/templates/en/web/authflowv2/__html_head.html
```diff
@@ -3,6 +3,7 @@
 <meta charset="UTF-8">
 <title>{{ include "app.name" nil }}</title>
 <link rel="shortcut icon" href="{{ call $.StaticAssetURL "favicon" }}">
+<meta name="robots" content="noindex">
```
louischan-oursky, line 6: `noindex, nofollow`

pkg/util/httputil/noindex.go
```diff
@@ -0,0 +1,12 @@
+package httputil
+
+import (
+	"net/http"
+)
+
+func NoIndex(next http.Handler) http.Handler {
```
louischan-oursky, line 7: Name this `XRobotsTag`

pkg/lib/robots/robots_txt.go
```diff
@@ -0,0 +1,23 @@
+package robots
```
louischan-oursky, line 1: I suggest we put this in httputil. And make this handler like `HealthCheckHandler`.

pkg/lib/robots/deps.go
```diff
@@ -0,0 +1,9 @@
+package robots
+
+import (
+	"github.com/google/wire"
+)
+
+var DependencySet = wire.NewSet(
+	wire.Struct(new(Handler), "*"),
+)
```
louischan-oursky, line 9: No need to use wire here.

pkg/lib/robots/robots_txt.go
```diff
@@ -0,0 +1,23 @@
...
+`, "\n")
+
+type Handler struct{}
+
+func (h *Handler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	body := []byte(robotsTxt)
+	rw.Header().Set("Content-Type", "text/plain")
+	rw.Header().Set("Content-Length", strconv.Itoa(len(body)))
+	rw.WriteHeader(http.StatusOK)
+	_, _ = rw.Write(body)
```
louischan-oursky, line 22: Add a test for this handler.

pkg/lib/robots/robots_txt.go
```diff
@@ -0,0 +1,23 @@
...
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+// Disallow all crawlers
+var robotsTxt = strings.Trim(`
+User-agent: *
+Disallow: /
+`, "\n")
```
louischan-oursky, line 13: I suggest we remove `strings.Trim` and make the string content match the actual content we will return.

The body should be `User-agent: *\nDisallow: /\n` The trailing newline should be kept.

louischan-oursky (changes requested):

pkg/util/httputil/noindex.go
tung2744: It seems `none` is better.
https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#nofollow

pkg/lib/robots/robots_txt.go
tung2744: What do you want to test? The response body?

pkg/util/httputil/noindex.go
louischan-oursky: Not sure if Bing knows `none`.

pkg/lib/robots/robots_txt.go
louischan-oursky: Yes. The response body.

pkg/auth/routes.go
```diff
@@ -45,12 +45,18 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource) *h
 		PathPattern: "/healthz",
 	}, p.RootHandler(newHealthzHandler))
 
+	router.Add(httproute.Route{
+		Methods:     []string{"GET"},
+		PathPattern: "/robots.txt",
+	}, p.RootHandler(newRobotsTXTHandler))
+
```
louischan-oursky, line 54: Please also add this handler to
- `pkg/admin/routes.go`
- `pkg/resolver/routes.go`
- `pkg/images/routes.go`
- `pkg/portal/routes.go`

louischan-oursky (changes requested):

pkg/lib/robots/robots_txt.go
tung2744: It looks like testing the framework :sosoad:

pkg/util/httputil/noindex.go
tung2744: https://www.bing.com/webmasters/help/which-robots-metatags-does-bing-support-5198d240
Ok maybe we use noindex, nofollow to be safe

pkg/lib/robots/robots_txt.go
louischan-oursky: Originally the body was generated by strings.Trim, so I said we need to test it. If the body is a literal string, then the test is not needed.

pkg/auth/routes.go
tung2744: Updated. However, only the main server will serve at the root, so search engines always find robots.txt at the root so they should always find main server's robots.txt.

tung2744: Removed according to discussion in slack: We need search engines to crawl them at least once so that they are removed from the index.

tung2744:
Updated, thanks!

louischan-oursky:
Please also add the meta tag to `portal/src/index.html`

louischan-oursky (approved):

PR #4866, opened 2024-10-29
Deduplicate emails in settings profile

tung2744:
ref DEV-1860


pkg/auth/handler/webapp/viewmodels/settings_profile.go
```diff
@@ -92,7 +93,7 @@ type SettingsProfileViewModeler struct {
 
 // nolint: gocognit
 func (m *SettingsProfileViewModeler) ViewModel(userID string) (*SettingsProfileViewModel, error) {
-	var emails []string
+	var emails setutil.Set[string]
```
louischan-oursky, line 96: I guess we need to apply the same logic to phone numbers and usernames?

louischan-oursky (changes requested):

pkg/auth/handler/webapp/viewmodels/settings_profile.go
tung2744: Updated, thanks!

louischan-oursky (approved):

PR #4870, opened 2024-10-29
Set up pgaudit locally

louischan-oursky:
ref DEV-1547


tung2744:
Do we still need this?

louischan-oursky:
Not harmful to have this setup locally. So yes.

tung2744 (approved):

PR #4872, opened 2024-10-30
Optimize navigations in settings in some special cases

tung2744:
ref DEV-2248

- Added a query `q_from` to store the originating url, so that the back button can back to there.

### DIFF
--- a/pkg/auth/webapp/result.go
+++ b/pkg/auth/webapp/result.go
@@ -45,7 +45,9 @@ func (r *Result) WriteResponse(w http.ResponseWriter, req *http.Request) {
 		}
 	}
 
-	// Propagate q_from
+	// Propagate q_back_url
+	// When user POST something to an endpoint, and that endpoint returned a redirect response,
+	// from the user's perspective the user is reaching the redirected uri from the page he POST the request.
 	if req.URL.Query().Get(QueryBackURL) != "" {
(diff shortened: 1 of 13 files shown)


pkg/auth/webapp/result.go
```diff
@@ -45,6 +45,11 @@ func (r *Result) WriteResponse(w http.ResponseWriter, req *http.Request) {
 		}
 	}
 
+	// Propagate q_from
+	if req.URL.Query().Get(QueryFrom) != "" {
+		q.Set(QueryFrom, req.URL.Query().Get(QueryFrom))
+	}
```
louischan-oursky, line 51: Wouldn't this make `q_from` like `x_from`?

louischan-oursky (changes requested):

pkg/auth/webapp/result.go
tung2744: We only propagate q_from in an redirect response, but not in any page.
Rationale: When user POST something to an endpoint, and that endpoint returned an redirect response. In the user's perspective the user is reaching the redirected uri from the page he POST the request.

tung2744: Added a comment to explain it

louischan-oursky:
Looks good to me. Please fix the build.

louischan-oursky (approved):

PR #4873, opened 2024-10-30
Trim whitespaces in email and username checker

tung2744:
ref DEV-2253

- We did validate emails, but mail.ParseAddress is success when the email containing a space.

### DIFF
--- a/pkg/lib/authn/identity/loginid/checker.go
+++ b/pkg/lib/authn/identity/loginid/checker.go
@@ -28,10 +28,10 @@ func (c *Checker) validateOne(ctx *validation.Context, loginID identity.LoginIDS
 	allowed := false
 	for _, keyConfig := range c.Config.Keys {
 		if keyConfig.Key == loginID.Key {
-			if len(loginID.Value) > *keyConfig.MaxLength {
+			if len(loginID.Value.TrimSpace()) > *keyConfig.MaxLength {
 				ctx.EmitError("maxLength", map[string]interface{}{
 					"expected": *keyConfig.MaxLength,
-					"actual":   len(loginID.Value),
+					"actual":   len(loginID.Value.TrimSpace()),
 				})
 				return
 			}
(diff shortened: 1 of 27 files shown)


tung2744:
Marked as do not merge because there are some users in production having whitespaces and will conflict with existing user if we remove those white spaces.

pkg/lib/authn/identity/loginid/checker.go
```diff
@@ -15,10 +17,17 @@ type Checker struct {
 	TypeCheckerFactory *TypeCheckerFactory
 }
 
-func (c *Checker) ValidateOne(loginID identity.LoginIDSpec, options CheckerOptions) error {
+func (c *Checker) ValidateOne(loginID identity.LoginIDSpec, options CheckerOptions) (identity.LoginIDSpec, error) {
```
louischan-oursky, line 20: Having side effects (changing spec.Value) is not good in a function named ValidateOne. Validation normally should be pure. I have the following suggestions

1. Forbid constructing LoginIDSpec directly. Must use NewLoginIDSpec. NewLoginIDSpec calls strings.TrimSpace.
2. Changing the type of LoginIDSpec.Value to UserInputString. This is a safer option, as the compiler will complain on every existing call sites. Existing places have to be modified to `spec.Value.TrimSpace()` to get a safe string to use.

```
type UserInputString struct {
  UnsafeString string
}

func (s UserInputString) TrimSpace() string {
  return strings.TrimSpace(s.UnsafeString)
}

type LoginIDSpec struct {
  Key string
  Type model.LoginIDKey
  Value UserInputString 
}
```

Finally, I think UserInputString can be placed in `pkg/util/userinpututil/string.go`

louischan-oursky (changes requested):

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #4879, opened 2024-11-01
Fix settings missing oauth identities in email / phone list

tung2744:
ref DEV-1860


resources/authgear/templates/en/web/authflowv2/settings_identity_list_email.html
```diff
@@ -61,15 +71,15 @@
     {{ end }}
   {{ end }}
 
-  {{ if and ($.PrimaryEmail) (gt (len $.EmailIdentities) (1)) }}
+  {{ if and ($.PrimaryEmail) (gt (len $.AllEmails) (1)) }}
     <div class="mt-5">
-      <div class="settings-title">
+      <div class="settings-title text-start">
```
tung2744, line 76: 1. I fixed the alignment here, but I am not sure if it is correct to hard code `text-start` here, or it should respect the auth ui content alignment. I hard-coded `text-start` here because it seems other places in settings also do in this way.
2. The font size is a bit different with the design. I am not sure if I should introduce new components so I didn't change it.

louischan-oursky (changes requested):
It still does not handle LDAP identities, which could also have email.

tung2744:
It seems the figma doesn't handle ldap identities, maybe need @fungc-io

tung2744:
TODO:
- [ ] Handle phone_number and preferred_username from oauth identities

louischan-oursky (approved):

PR #4880, opened 2024-11-05
Return a correct response on not found routes

tung2744:
ref DEV-2231

### DIFF
--- a/pkg/auth/handler/webapp/authflowv2/not_found.go
+++ b/pkg/auth/handler/webapp/authflowv2/not_found.go
@@ -27,20 +27,10 @@ func (h *AuthflowV2NotFoundHandler) GetData(r *http.Request, w http.ResponseWrit
 }
 
 func (h *AuthflowV2NotFoundHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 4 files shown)


pkg/auth/handler/webapp/authflowv2/not_found.go
```diff
@@ -27,20 +27,10 @@ func (h *AuthflowV2NotFoundHandler) GetData(r *http.Request, w http.ResponseWrit
...
-	ctrl.Get(func() error {
-		data, err := h.GetData(r, w)
-		if err != nil {
-			return err
-		}
-
-		h.Renderer.RenderHTML(w, r, TemplateWebNotFoundHTML, data)
-		return nil
-	})
+	h.Renderer.RenderHTML(w, r, TemplateWebNotFoundHTML, data)
```
louischan-oursky, line 35: - Should call RenderHTMLStatus.
- Introduce RenderHTMLStatus(w, r, status, tpl, data)
- Change RenderHTML to use RenderHTMLStatus

louischan-oursky (changes requested):

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #4883, opened 2024-11-05
Do not allow using device token to login if no secondary authenticator exist

tung2744:
ref DEV-2263


louischan-oursky (changes requested):
Please add a e2e test to demonstrate we have fixed the bug.

I think the test could be

1. Set up the project to be MFA required.
1. Sign in and save device token.
1. Sign in again and see MFA is skipped.
1. Call Admin API to remove the secondary authenticator.
1. Sign in again, see no authenticator error.

tung2744:
Currently it is impossible to call admin api in a normal e2e test. I think I will delete it by SQL instead.

louischan-oursky (approved):
Thank you. The test looks very good.

PR #4884, opened 2024-11-08
Restrict the usage of http.Client for otel

louischan-oursky:
ref DEV-2281


louischan-oursky:
I added a commit to fix ./custombuild

tung2744 (approved):

PR #4889, opened 2024-11-13
Refactor messaging mechanism for otel

louischan-oursky:
ref DEV-2255

Examples of the new events

## whatsapp.suppressed
SCR-20241119-nqzg

## whatsapp.error
SCR-20241119-nrjd

## sms.error
SCR-20241119-nrfd

## email.suppressed
SCR-20241119-nrlv

### DIFF
--- a/pkg/lib/messaging/sender.go
+++ b/pkg/lib/messaging/sender.go
@@ -40,7 +40,8 @@ type SMSSender interface {
 }
 
 type WhatsappSender interface {
-	SendAuthenticationOTP(Ctx context.Context, opts *whatsapp.SendAuthenticationOTPOptions) error
+	ResolveSendAuthenticationOTPOptions(ctx context.Context, opts *whatsapp.SendAuthenticationOTPOptions) (*whatsapp.ResolvedSendAuthenticationOTPOptions, error)
+	SendAuthenticationOTP(ctx context.Context, opts *whatsapp.ResolvedSendAuthenticationOTPOptions) error
 }
 
 type Sender struct {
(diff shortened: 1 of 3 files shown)


pkg/lib/messaging/sender.go
```diff
@@ -249,26 +248,16 @@ func (s *Sender) SendWhatsappImmediately(ctx context.Context, msgType translatio
...
-func (s *Sender) devModeSendWhatsapp(opts *whatsapp.SendTemplateOptions) {
-	data, _ := json.MarshalIndent(opts.Components, "", "  ")
-
+func (s *Sender) devModeSendWhatsapp(opts *whatsapp.SendAuthenticationOTPOptions) {
 	s.Logger.
 		WithField("recipient", opts.To).
-		WithField("template_name", opts.TemplateName).
-		WithField("language", opts.Language).
-		WithField("components", string(data)).
-		WithField("namespace", opts.Namespace).
```
tung2744: I guess we want to log language, template name, namespace here, because it could be useful when debugging whatsapp otp.

louischan-oursky: Those information is only available in the whatsapp/Service. In messaging/Sender, the available information is just the information contained in whatsapp/SendAuthenticationOTPOptions.

The previous behavior is a bit hacky. If dev mode is enabled, we synthesize those information. See https://github.com/authgear/authgear-server/blob/2024-11-18.0/pkg/lib/infra/whatsapp/service.go#L86

In this PR, I centralized the handling of dev mode, test mode, as well as the audit log in messaging/Sender. But it is impossible for messaging/Sender to know the extra whatsapp information you mentioned. Those information is only available if Whatsapp is really configured.

tung2744: But I still think devMode should log any imformation that affects the resulting whatsapp message, just like sending is logging the whole email, but not the otp code & receiver only.

It is is difficult to do it in `messaging/Sender`, maybe we can create a method in `WhatsappSender` to do the logging?

> Those information is only available if Whatsapp is really configured.

I guess this is expected?

louischan-oursky: OK. Let me separate the send operation into two steps.

louischan-oursky:
@fungc-io Included screenshots of the new events.

fungc-io:
lgtm, thanks!

tung2744 (approved):

PR #4892, opened 2024-11-15
Fix suppress idp session is not working correctly

tung2744:
ref DEV-2246

### DIFF
--- a/pkg/auth/handler/webapp/authflowv2/select_account.go
+++ b/pkg/auth/handler/webapp/authflowv2/select_account.go
@@ -119,15 +119,6 @@ func (h *AuthflowV2SelectAccountHandler) ServeHTTP(w http.ResponseWriter, r *htt
 	oauthProviderAlias := ""
 	var loginHint *oauth.LoginHint
 
-	// When x_suppress_idp_session_cookie is true, ignore IDP session cookie.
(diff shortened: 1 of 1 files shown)


pkg/auth/handler/webapp/authflowv2/select_account.go
```diff
@@ -146,6 +137,15 @@ func (h *AuthflowV2SelectAccountHandler) ServeHTTP(w http.ResponseWriter, r *htt
...
 
+		// When x_suppress_idp_session_cookie is true, ignore IDP session cookie.
+		if suppressIDPSessionCookie {
+			session = nil
+		}
+		// Ignore any session that is not allow to be used here
+		if !oauth.ContainsAllScopes(oauth.SessionScopes(session), []string{oauth.PreAuthenticatedURLScope}) {
+			session = nil
+		}
+
```
louischan-oursky, line 148: How does this move fix the bug?

tung2744: The bug was caused by moving the code into `BeforeHandle`, but I forgot to move these code. As a result `suppressIDPSessionCookie` is not assigned at the time `if suppressIDPSessionCookie ` is run.

PR #4894, opened 2024-11-18
Revert the old implementation of DEV-1468

louischan-oursky:
ref DEV-1468


louischan-oursky:
It is because we will not simply re-apply the changes. The PR has to be partly rewritten according to https://linear.app/authgear/issue/DEV-1468/audit-log-for-password-changed#comment-f364601c

tung2744 (approved):

PR #4898, opened 2024-11-20
Fix prepared statements never cached

louischan-oursky:
ref DEV-2299


tung2744:
Should we merged this first as a quick fix before we really fixes the problem?

PR #4899, opened 2024-11-20
[WIP] Fix prepared statements never cached

louischan-oursky:
ref DEV-2299


louischan-oursky:
I figured why e2e got stuck. When the e2e test script call /healthz to determine whether the server is ready. In /healthz, QueryRowWith is used to run a simple query to test the connection to the database is healthy.

But the usage of QueryRowWith is not correct there. The result of QueryRowWith, which is a *sql.Row, actually must be used by calling Scan. Inside Scan, the *sql.Rows embedded in it will be Close properly.

Why this problem is not observed before this change? It is observed that if the *sql.Row is produced by a *sql.Tx, when the *sql.Tx rolls back or commits, the unclosed *sql.Rows will be closed together. However, when we enable prepared statement, we use a connection-bound *sql.Stmt to produce a *sql.Row (or *sql.Rows). In this case, when the *sql.Tx rolls back or commits, the unclosed *sql.Row or *sql.Rows is not closed together. And *sql.Conn.Close is documented to wait for any active query to close before it can close itself. So we observed the forever hang.

When *sql.StmtContext is used to produced a transaction-bound *sql.Stmt, the *sql.Row or *sql.Rows produced by the transaction-bound *sql.Stmt also share the close-together behavior. This explains why it has been working fine.

louischan-oursky:
Added one more commit to remove PoolDB.

PR #4905, opened 2024-11-22
Fix incorrect task import output

tung2744:
ref DEV-2309

### DIFF
--- a/pkg/lib/userimport/model.go
+++ b/pkg/lib/userimport/model.go
@@ -459,7 +459,7 @@ type Summary struct {
 
 type Detail struct {
 	Index    int                   `json:"index"`
-	Record   Record                `json:"record,omitempty"`
+	Record   interface{}           `json:"record,omitempty"`
 	Outcome  Outcome               `json:"outcome,omitempty"`
 	UserID   string                `json:"user_id,omitempty"`
 	Warnings []Warning             `json:"warnings,omitempty"`
--- a/pkg/lib/userimport/model.go
+++ b/pkg/lib/userimport/model.go
@@ -170,6 +170,21 @@ func mapGetNonNull[M ~map[string]interface{}, T constraints.Ordered | ~bool](m M
 	return v, true
 }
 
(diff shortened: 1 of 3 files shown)


pkg/lib/userimport/service.go
```diff
@@ -173,6 +173,9 @@ func (s *UserImportService) importRecordInConn(
 	if record != nil {
 		record.Redact()
 		detail.Record = record
+	} else {
+		// record is nil means parsing is failed, return the raw input
+		detail.Record = rawMessage
```
louischan-oursky, line 178: It is no good to include the raw message because it could contain un-redacted information. Drop this commit. Having the correct index should be enough to pinpoint the problematic row.

louischan-oursky (changes requested):

pkg/lib/userimport/service.go
tung2744: As our api has a body size limit, importers could split an import into multiple jobs. If we only return an index, it requires additional calculation / mapping to understand the error. It will be much easier to understand the error if we include the input object in the result. If your concern is un-redacted password / mfa data, I prefer to redact the raw input and still include it in the result.

tung2744: I've added a commit to redact password and mfa from user input, see if it is ok?

pkg/lib/userimport/model.go
```diff
@@ -459,7 +498,7 @@ type Summary struct {
 
 type Detail struct {
 	Index    int                   `json:"index"`
-	Record   Record                `json:"record,omitempty"`
+	Record   interface{}           `json:"record,omitempty"`
```
louischan-oursky, line 501: Change this type back to `Record`.

louischan-oursky (changes requested):

pkg/lib/userimport/model.go
tung2744: Fixed, thanks!

louischan-oursky (approved):

PR #4906, opened 2024-11-22
Collect metrics

louischan-oursky:
ref DEV-2259


pkg/lib/deps/middleware_request.go
```diff
@@ -53,6 +55,13 @@ func (m *RequestMiddleware) Handle(next http.Handler) http.Handler {
 
 		ap := m.RootProvider.NewAppProvider(r.Context(), appCtx)
 		r = r.WithContext(withProvider(r.Context(), ap))
+
+		{
+			key := otelauthgear.AttributeKeyProjectID
+			val := key.String(string(appCtx.Config.AppConfig.ID))
+			r = r.WithContext(context.WithValue(r.Context(), key, val))
+		}
+
```
tung2744, line 64: Should this be done inside `NewAppProvider()`?

louischan-oursky: I did not do that because NewAppProvider or withProvider was only used here. So by just looking at this file, we know exactly what was put into context.

tung2744 (approved):

PR #4913, opened 2024-11-29
Fix FileServer does not recognize Rollup default asset name

louischan-oursky:
ref DEV-2323


tung2744 (approved):

louischan-oursky:
Found a bug in the new implementation.

PR #4914, opened 2024-11-29
Fix /oauth-redirect is mistakenly handled as 404

louischan-oursky:
ref DEV-2323

### DIFF
--- a/authui/packages/vite-plugin-authgear-authui/src/index.js
+++ b/authui/packages/vite-plugin-authgear-authui/src/index.js
@@ -24,9 +24,33 @@ function removeLeadingSlash(s) {
  * @return {string}
  */
 function nameWithoutHash(filePath) {
-  // The hash is placed between the base name and the extension
-  // e.g. abc.[hash].js or abc.def.[hash].css
-  return filePath.replace(/\.[0-9a-z]{8}\./, ".");
+  const originalFilePath = filePath;
+
+  const exts = [];
+
+  while (true) {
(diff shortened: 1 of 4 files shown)


louischan-oursky:
@tung2744 I discovered another problem. Do not merge yet.

louischan-oursky:
Added one more commit to correct authui vite configuration.

authui/packages/vite-plugin-authgear-authui/src/index.js
```diff
@@ -342,11 +366,6 @@ function buildPlugin({ input }) {
                 }
                 return null;
               },
-              hashCharacters: "hex",
-              assetFileNames: () => "[name].[hash][extname]",
-              chunkFileNames: () => "[name].[hash].js",
-              entryFileNames: () => "[name].[hash].js",
-              sourcemapFileNames: () => "[name].[chunkhash].map.js",
```
tung2744, line 349: Another thought: prefix the hash with some custom prefix like `aghash-[hash]` which we will check in the regex?

tung2744 (approved):

authui/packages/vite-plugin-authgear-authui/src/index.js
louischan-oursky: Actually I want to unset `assetsDir: ""` to use the original behavior of Vite. Then the generated built assets should reside in a separate directory. We do not need the regex at all.

PR #4915, opened 2024-12-02
Add message counts for custom and free plan

tung2744:
ref DEV-2322
ref DEV-2284
requires https://github.com/authgear/authgear-server/pull/4911


tung2744:
Added two commits for some ui adjustments

pkg/lib/usage/globaldb_store.go
```diff
@@ -205,3 +205,50 @@ func (s *GlobalDBStore) FetchUsageRecords(
...
+		"name",
+		"period",
+		"start_time",
+		"end_time",
+		"count",
+		"stripe_timestamp",
+	).
+		From(s.SQLBuilder.TableName("_portal_usage_record")).
+		Where(
+			"app_id = ? AND name = ? AND period = ? AND start_time >= ? AND end_time <= ?",
```
louischan-oursky, line 229: The index we have only cover `(app_id, name, period, start_time)`. If we use `end_time` in the query, the index may not be used.

However, `end_time` is actually determined by `period` and `start_time`. So I think we can rewrite the query so that it only reference the indexed columns.

louischan-oursky (changes requested):

pkg/lib/usage/globaldb_store.go
tung2744: @louischan-oursky I've confirmed the index can be correctly used:
```
EXPLAIN ANALYZE SELECT * FROM _portal_usage_record WHERE app_id = 'my-app' AND name = 'active-user' AND period = '1 day' AND start_time >= NOW() AND end_time <= NOW();
```

| QUERY PLAN |
| --- |
| Index Scan using \_portal\_usage\_record\_app\_id\_name\_period\_start\_time\_key on \_portal\_usage\_record  \(cost=0.15..8.18 rows=1 width=188\) \(actual time=0.136..0.137 rows=0 loops=1\) |
|   Index Cond: \(\(app\_id = 'my\-app'::text\) AND \(name = 'active\-user'::text\) AND \(period = '1 day'::text\) AND \(start\_time \>= now\(\)\)\) |
|   Filter: \(end\_time \<= now\(\)\) |
| Planning Time: 0.732 ms |
| Execution Time: 0.244 ms |

louischan-oursky: The `start_time >= t` condition can still include rows that is to be filtered by the filter. But we actually can rewrite the condition to `start_time >= t0 AND start_time < t1`

tung2744: Do you mean you want to skip the step `Filter: (end_time <= now())`?

tung2744: Sorry wait, it seems I've calculated the end time wrongly for monthly period.

pkg/lib/usage/globaldb_store.go
```diff
@@ -214,6 +215,19 @@ func (s *GlobalDBStore) FetchUsageRecordsInRange(
...
+	switch period {
+	case periodical.Hourly:
+		endStartTime = toEndTime.Add(-time.Hour)
+	case periodical.Daily:
+		endStartTime = toEndTime.Add(-time.Hour * 24)
+	case periodical.Weekly:
+		endStartTime = toEndTime.Add(-time.Hour * 24 * 7)
+	case periodical.Monthly:
+		endStartTime = timeutil.FirstDayOfTheMonth(toEndTime).Add(-time.Hour * 24)
+	}
```
louischan-oursky, line 230: I still prefer `start_time >= ? AND start_time < ?`. This is more aligned with the definition of the column. `start_time` is inclusive and `end_time` is exclusive. To convert the inclusive `toEndTime`, we should add 1 `period` to it.

tung2744: I  guess you mean `toEndTime` of the function should be exclusive. However this makes it difficult to use, you need to add a little moment to `toEndTime` (e.g. 1 seconds) to query the rows you want.

pkg/lib/usage/globaldb_store.go
tung2744: Fixed in the latest commit.

pkg/lib/usage/globaldb_store.go
louischan-oursky: No, `toEndTime` can be inclusive. I mean how to convert the inclusive Go value `toEndTime` to a value suitable for use in SQL.

tung2744: Actually I found that the new implementation is not very intuitive which makes me prefer the previous implementation. I also thinks that the performance gain for this change is not very important.

tung2744: > No, toEndTime can be inclusive. I mean how to convert the inclusive Go value toEndTime to a value suitable for use in SQL.

If you mean calculate `endStartTime` by minus 1 period from `toEndTime`, it is already the current implementation. However, we cannot do `start_time >= ? AND start_time < ?`. We need `start_time <= ?` so that all needed rows are returned, because `end_time` minus 1 period is exactly the `start_time`.

tung2744: Removed conversion of toEndTime in latest commit and used it to filter start_time exclusively.

louischan-oursky (approved):

PR #4916, opened 2024-12-02
Use non-empty assetsDir

louischan-oursky:
ref DEV-2323

It turns out we have 3 usages of FileServer.

1. To serve portal, which is built by Vite.
2. To serve authui generated assets, which is also built by Vite.
3. to Serve authui user-generated assets, like custom CSS, app logo image.

For case 3, it is still using its own hash name scheme (pkg/util/pathutil), but the FileServer it uses has been updated to use a non-empty AssetsDir.


tung2744:
`ensure-reproducible-build` failed

authui/packages/vite-plugin-authgear-authui/src/index.js
```diff
@@ -280,8 +273,8 @@ function buildPlugin({ input }) {
 
           for (const node of head.childNodes) {
             if (node.tagName === "LINK") {
-              const hashedName = removeLeadingSlash(node.getAttribute("href"));
-              const key = nameWithoutHash(hashedName);
+              const href = node.getAttribute("href");
+              const key = nameWithoutHash(path.basename(href));
```
tung2744, line 277: path.basename only keep the last part of the path, e.g. `/img/icon-abcd.svg` -> `icon-abcd.svg`, not sure if this is expected.

louischan-oursky: For existing usage pattern, yes, it is expected. It is because the existing usage of `GeneratedStaticAssetURL` expects basename only.

tung2744: 🤔 Then maybe there will be some problem if there are multiple files with the same name in different directories?

louischan-oursky:
Added a commit to work around the false positive of reproducible build

authui/packages/vite-plugin-authgear-authui/src/index.js
tung2744: So there is indeed some problem if there are multiple files with the same filename in different directories, because it causes key collision in the manifest json.
It works for now as we don't have files with the same name, not sure if it would be a concern in the future. Maybe we can ignore it for now.

louischan-oursky: We should migrate to use `manifest: true` and `.vite/manifest.json` instead of writing our own `manifest.json`. But that's more complicated so I leave it for now.

tung2744 (approved):

PR #4922, opened 2024-12-04
Stop using otelhttp due to memory leak

louischan-oursky:


louischan-oursky:
@tung2744 We need this merged first

tung2744 (approved):

PR #4924, opened 2024-12-05
Soft remove SIWE

louischan-oursky:
ref DEV-2300


pkg/lib/config/environment.go
```diff
@@ -18,7 +18,7 @@ type WebAppCDNHost string
 
 type CORSAllowedOrigins string
 
-type NFTIndexerAPIEndpoint string
+type Deprecated_NFTIndexerAPIEndpoint string
```
tung2744, line 21: Why do we need to keep this type?

pkg/lib/config/config.go
```diff
@@ -84,7 +84,7 @@ type AppConfig struct {
 	Verification              *VerificationConfig   `json:"verification,omitempty"`
 	Deprecated_OTP            *OTPLegacyConfig      `json:"otp,omitempty"`
 
-	Web3 *Web3Config `json:"web3,omitempty"`
+	Web3 *Deprecated_Web3Config `json:"web3,omitempty"`
```
tung2744, line 87: Also rename `Web3` to `Deprecated_Web3`?

pkg/lib/config/feature.go
```diff
@@ -44,7 +44,7 @@ type FeatureConfig struct {
 	RateLimits       *RateLimitsFeatureConfig       `json:"rate_limits,omitempty"`
 	Messaging        *MessagingFeatureConfig        `json:"messaging,omitempty"`
 	Collaborator     *CollaboratorFeatureConfig     `json:"collaborator,omitempty"`
-	Web3             *Web3FeatureConfig             `json:"web3,omitempty"`
+	Web3             *Deprecated_Web3FeatureConfig  `json:"web3,omitempty"`
```
tung2744, line 47: Also rename `Web3` to `Deprecated_Web3`?

pkg/lib/config/environment.go
louischan-oursky: You're right. Will remove it.

louischan-oursky:
Added 3 commits to address the comments.

tung2744 (approved):

PR #4928, opened 2024-12-09
Support MinIO

louischan-oursky:
ref DEV-2346


pkg/lib/cloudstorage/minio.go
```diff
@@ -0,0 +1,108 @@
+package cloudstorage
```
tung2744, line 1: I thought we are integrating minio same as S3, because it is S3 compatible 🤔

tung2744 (approved):

pkg/lib/cloudstorage/minio.go
louischan-oursky: Theoretically, yes. But MinIO docs does not cover how to use the AWS SDK with it. So I install the MinIO SDK instead.

PR #4931, opened 2024-12-10
Avoid portal dependencies from reference external CDN

louischan-oursky:
ref DEV-2357


portal/src/public/shared-assets/fabric-icons-0-467ee27f.woff
tung2744, line 1: Do we need to commit the files in shared-assets? 🤔 
Or they should be generated after npm install?

louischan-oursky: The postinstall script was intended to automate the update of these files, not for generating the files after npm install. Having the `public` directory contain we will serve with a simple `git checkout` is easier to understand.

tung2744: I see. But I am worried about the repo size :sosad:

tung2744 (approved):

PR #4937, opened 2024-12-11
Support Authgear Once user provision

louischan-oursky:
ref DEV-2341


Makefile
```diff
@@ -1,16 +1,56 @@
...
+# 1. You use ::= instead of = because = defines a recursively expanded variable.
+#    See https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html
+# 2. You use ::= instead of := because ::= is a POSIX standard.
+#    See https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html
+# 3. You do not use ?= because it is shorthand to define a recursively expanded variable.
+#    See https://www.gnu.org/software/make/manual/html_node/Conditional-Assignment.html
+#    You should use the long form documented in the above link instead.
+# 4. When you override a variable in the command line, as documented in https://www.gnu.org/software/make/manual/html_node/Overriding.html
+#    you specify the variable with ::= instead of = or :=
+#    If you fail to do so, the variable becomes recursively expanded variable accidentally.
```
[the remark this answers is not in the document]
louischan-oursky: No problem in our current usage. I just learnt this today and realized we are not very consistent with the use of the variables. We never intent to use recursively expanded variables.

pkg/portal/service/collaborator.go
```diff
@@ -482,6 +481,18 @@ func (s *CollaboratorService) SendInvitation(
...
+	if err != nil {
+		return nil, err
+	}
+
+	if !inviteeExists {
+		err = s.createAccountForInvitee(ctx, invitedBy, inviteeEmail)
+		if err != nil {
+			return nil, err
+		}
+	}
```
tung2744, line 494: New developers (If any) reading this part may be confused:
Why `createAccountForInvitee` is called but actually no account created?

What about:
```
// In CollaboratorService.SendInvitation
if IsAuthgearOnce {
  // ... checkInviteeExistenceByEmail
  // ... createAccountForInvitee
}

// In another file
//go:build !authgearonce
// +build !authgearonce
const IsAuthgearOnce = false

// In another file
//go:build authgearonce
// +build authgearonce
const IsAuthgearOnce = true
```

louischan-oursky: Agree, let me make the change according to your suggestion.

tung2744 (approved):

PR #4940, opened 2024-12-12
Build image in arm64 & amd64

tung2744:
ref DEV-2362

PR #4943, opened 2024-12-13
Build multi platform image & Reorganize Makefiles

tung2744:
ref DEV-2369
ref DEV-2371
based on https://github.com/authgear/authgear-server/pull/4940


tung2744:
This pr is ready for review

Makefile
```diff
@@ -162,26 +162,25 @@ check-tidy:
...
-push-image:
-	docker manifest inspect $(DOCKER_IMAGE):$(GIT_HASH) > /dev/null; if [ $$? -eq 0 ]; then \
-		echo "$(DOCKER_IMAGE):$(GIT_HASH) exists. Skip push"; \
-	else \
-		docker push $(DOCKER_IMAGE):latest ;\
-		docker push $(DOCKER_IMAGE):$(GIT_HASH) ;\
-		if [ ! -z $(GIT_NAME) ]; then docker push $(DOCKER_IMAGE):$(GIT_NAME); fi ;\
-	fi
+	$(eval DOCKER_IMAGE ::= quay.io/theauthgear/$(IMAGE_NAME))
+	$(eval BUILD_OPTS ::= --tag $(DOCKER_IMAGE))
```
louischan-oursky: To declare target-specific variable, follow this syntax https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html

```makefile
build-image: DOCKER_IMAGE ::= quay.io/theauthgear/$(IMAGE_NAME)
build-image: BUILD_OPTS ::= --tag $(DOCKER_IMAGE)
build-image:
  # The original body of this target goes here.
```

Makefile
```diff
@@ -162,26 +162,25 @@ check-tidy:
...
+ifneq (${GIT_HASH},)
+	$(eval BUILD_OPTS += --tag $(DOCKER_IMAGE):$(GIT_HASH))
+endif
+ifneq (${GIT_NAME},)
+	$(eval BUILD_OPTS += --tag $(DOCKER_IMAGE):$(GIT_NAME))
+endif
+endif # endif TAG_IMAGE
+ifeq ($(PUSH_IMAGE),true)
+	$(eval BUILD_OPTS += --push)
+endif
```
louischan-oursky: These makefile statements should be written without eval

```makefile
build-image:
DOCKER_IMAGE ::= quay.io/theauthgear/$(IMAGE_NAME)
BUILD_OPTS ::= --tag $(DOCKER_IMAGE)
ifeq (${TAG_IMAGE},true)
BUILD_OPTS += --tag $(DOCKER_IMAGE):latest
ifneq (${GIT_HASH},)
BUILD_OPTS += --tag $(DOCKER_IMAGE):$(GIT_HASH)
endif
# ...
build-image:
  # The original body of this target does here.
```

In general, we should use this pattern

```makefile
thetarget:
# makefile statements go here.
# Note that they MUST NOT start with a tab, otherwise they are interpreted as shell commands
thetarget:
  # The shell commands go here.
  # Note that they MUST start with a tab
```

In particular, we CANNOT write

```makefile
thetarget:
FOOBAR ::= 42
    echo $(FOOBAR)
```

It will complain with `Makefile:6: *** recipe commences before first target.  Stop.` because Make thinks the `echo` shell command IS NOT associated with a target.

common.mk
```diff
@@ -47,27 +47,48 @@ start-portal:
...
 	$(eval BUILD_OPTS += --push)
 endif
-ifneq ($(BUILD_PLATFORMS),)
-	$(eval BUILD_OPTS += --platform $(BUILD_PLATFORMS))
+ifeq ($(EXTRA_BUILD_OPTS),true)
+	$(eval BUILD_OPTS += $(EXTRA_BUILD_OPTS))
 endif
 	@# Add --pull so that we are using the latest base image.
 	@# The build context is the parent directory
-	docker build --pull --ssh=default \
```
louischan-oursky: I guess `--ssh=default` is needed in case of custombuild?

.github/workflows/ci.yaml
```diff
@@ -249,7 +249,62 @@ jobs:
...
+      run: |
+        docker logout quay.io
+  authgear-image-arm64:
+    if: ${{ github.repository != 'oursky/authgear-server' }}
+    runs-on: ubuntu-24.04
+    needs: ["authgear-test", "authui-test"]
+    env:
+      TARGET: authgear
+      IMAGE_NAME: quay.io/theauthgear/authgear-server
+      PUSH_IMAGE: "${{ (github.repository == 'authgear/authgear-server' && github.event_name == 'push') && 'true' || 'false' }}"
```
louischan-oursky: We should not push these `-arm64` or `-amd64` images. They are intermediate artifacts that should be combined and pushed at the end.

louischan-oursky (changes requested):

.github/workflows/ci.yaml
tung2744: I want to avoid the complication of using the digests by using a more understandable tag.
And in fact, the layers are always pushed. The difference is only whether that layer has a given tag name or not.
I think having a tag with a architecture suffix should not be a problem.

common.mk
tung2744: Handled by EXTRA_BUILD_OPTS in later commits

Makefile
louischan-oursky: This comment is outdated.

tung2744:

.github/actions/build-custom-image/action.yaml
```diff
@@ -29,10 +31,18 @@ runs:
       TARGET: ${{ inputs.target }}
       IMAGE_NAME: ${{ inputs.image_name }}
       REPO: ${{ inputs.docker_repo }}
+      BUILDER: ${{ inputs.builder }}
```
louischan-oursky, line 34: It does not seem used at all. The `--builder` argument is hard-coded to use `container-builder`

According to https://docs.docker.com/build/builders/#selected-builder we can use `BUILDX_BUILDER` to specify a builder. I think using env is easier to understand than `--builder`

.github/workflows/oursky.yaml
```diff
@@ -55,12 +69,26 @@ jobs:
         printf "$AUTHGEAR_PRIVATE_DEPLOY_KEY" | base64 --decode | ssh-add -
         echo "SSH_AUTH_SOCK=$SSH_AUTH_SOCK" >> "$GITHUB_ENV"
         echo "SSH_AGENT_PID=$SSH_AGENT_PID" >> "$GITHUB_ENV"
+    - name: Setup container builder
+      run: |
+        docker buildx create \
+          --name container-builder \
+          --driver docker-container \
+          --bootstrap --use
```
louischan-oursky, line 79: You can 

```
echo "BUILDX_BUILDER=container-builder" >> "$GITHUB_ENV"
```

to make BUILDX_BUILDER available to subsequent steps.

.github/workflows/oursky.yaml
```diff
@@ -35,15 +35,17 @@ jobs:
       env:
         REPO_PREFIX: ${{ secrets.AUTHGEAR_CUSTOM_BUILD_REPO_PREFIX }}
       run: |
+        MANIFEST_NAME="./$(uuidgen).json"
```
louischan-oursky, line 38: Use 

```
MANIFEST_PATH="$(mktemp).json"
```

1. It is written to the temp directory, thus avoid the bug you encountered in the first place. (Wrote to the same location accidentally)
2. It is now named `MANIFEST_PATH`, meaning that you do not need to `custombuild/MANIFEST_PATH`

common.mk
```diff
@@ -71,6 +71,7 @@ build-image:
 	@# Add --pull so that we are using the latest base image.
 	@# The build context is the parent directory
 	docker build --pull \
+		--provenance=false \
```
louischan-oursky, line 74: Why?

louischan-oursky (changes requested):

louischan-oursky:
Let me try to fix the `--builder` and manifest path for you.

common.mk
tung2744: This is related to the unknown manifest mentioned yesterday. I found out that they are [Provenance attestations](https://docs.docker.com/build/metadata/attestations/slsa-provenance/). So I disabled it in build.

.github/actions/build-custom-image/action.yaml
```diff
@@ -31,16 +31,17 @@ runs:
       REPO: ${{ inputs.docker_repo }}
     shell: bash
     run: |
-      MANIFEST_NAME="./$(uuidgen).json"
+      METADATA_FILE="$(mktemp).json"
```
tung2744, line 34: I guess we need `--suffix` to make a `.json` file, or there will become two files: The file created by `mktemp` and a file beside it named `(mktemp).json`.

louischan-oursky: I did not do that because `mktemp` on macOS does not have `--suffix`. That's a GNU specific option. I guess this is harmless. The name returned by `mktemp` is unique enough.

louischan-oursky (approved):

PR #4944, opened 2024-12-16
Ignore network errors

tung2744:
ref DEV-2274

### DIFF
--- a/authui/src/errors.ts
+++ b/authui/src/errors.ts
@@ -1,4 +1,10 @@
-// Modified from https://github.com/sindresorhus/is-network-error/blob/main/index.js
+// The original work is https://github.com/sindresorhus/is-network-error/blob/main/index.js
+// The modifications are
+// 1. Remove the error message checking for Node.js, since this code does not run on Node.js.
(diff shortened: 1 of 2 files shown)


authui/src/errors.ts
```diff
@@ -0,0 +1,23 @@
+// Modified from https://github.com/sindresorhus/is-network-error/blob/main/index.js
```
louischan-oursky, line 1: I see the original work has an extra checking for Safari 17+ to check if the error has an undefined stack. I guess we need to port that as well.

louischan-oursky (changes requested):

authui/src/errors.ts
tung2744: In my testing that check incorrectly filter out network errors in safari (stack is not undefined), so I removed that check.

louischan-oursky: Fair. Can you include this test as comment to explain this is one of the modification? It helps us to understand why the code is different from the original work.

louischan-oursky: Let me do this for you.

louischan-oursky (approved):

PR #4946, opened 2024-12-17
Handle module load failure with reload button

louischan-oursky:
ref DEV-2367

https://github.com/user-attachments/assets/493ab630-0d99-4be4-bfa0-92e7260ce6f5

### DIFF
--- a/portal/src/ErrorBoundSuspense.tsx
+++ b/portal/src/ErrorBoundSuspense.tsx
@@ -1,0 +1,159 @@
+import React, {
+  Suspense,
+  SuspenseProps,
+  lazy,
+  useCallback,
+  useMemo,
+  useState,
+  useEffect,
+} from "react";
+import { ErrorBoundary, type ErrorBoundaryProps } from "react-error-boundary";
+
+export interface ErrorBoundSuspenseProps {
+  errorBoundaryFallbackComponent: NonNullable<
+    ErrorBoundaryProps["FallbackComponent"]
+  >;
+  suspenseFallback: SuspenseProps["fallback"];
+  factory: () => Promise<{ default: React.ComponentType<any> }>;
+  children: (Component: React.ComponentType<any>) => React.ReactNode;
+}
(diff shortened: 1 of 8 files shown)


portal/src/ErrorBoundSuspense.tsx
```diff
@@ -0,0 +1,159 @@
...
+    }
+  } catch (_e: unknown) {
+    // ignore
+  }
+
+  return null;
+}
+
+function makeFactory(urlToModule: URL) {
+  // Browsers cache modules, including failed attempt.
```
tung2744, line 69: 🤔 It seems we should add "no-cache" to 404 responses of our server?

tung2744 (approved):

portal/src/ErrorBoundSuspense.tsx
louischan-oursky: The stdlib does that already. See https://pkg.go.dev/net/http#ServeContent But browser seem ignore HTTP cache header in this particular case.

louischan-oursky: https://github.com/user-attachments/assets/16a8d895-396e-4938-a17e-6976f43ac331

The source code of the server

```golang
package main

import (
	"fmt"
	"net/http"
	"net/http/httputil"
)

var html = `<!DOCTYPE html>
<html>
<head>
	<title>Hello</title>
</head>
<body>
</body>
</html>
`

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte(html))
	})
	http.HandleFunc("/index.js", func(w http.ResponseWriter, r *http.Request) {
		b, _ := httputil.DumpRequest(r, true)
		fmt.Printf("%v\n", string(b))
		w.Header().Set("Cache-Control", "no-store")
		w.WriteHeader(500)
	})
	http.ListenAndServe(":3000", nil)
}

```

PR #4949, opened 2024-12-18
Ignore network error in portal

tung2744:
ref DEV-2274

### DIFF
--- a/portal/src/ReactApp.tsx
+++ b/portal/src/ReactApp.tsx
@@ -8,6 +8,7 @@ import React, {
 import {
   Exception as SentryException,
   ErrorEvent as SentryErrorEvent,
+  EventHint,
   init as sentryInit,
 } from "@sentry/react";
 import { BrowserRouter, Routes, Route, Navigate } from "react-router-dom";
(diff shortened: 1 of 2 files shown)


portal/src/ReactApp.tsx
```diff
@@ -71,10 +73,13 @@ function isPosthogResetGroupsEvent(event: SentryErrorEvent) {
 }
 
 // DEV-1767: Unknown cause on posthog error, silence for now
-function sentryBeforeSend(event: SentryErrorEvent) {
+function sentryBeforeSend(event: SentryErrorEvent, hint: EventHint) {
   if (isPosthogResetGroupsEvent(event)) {
     return null;
   }
+  if (isNetworkError(hint)) {
```
louischan-oursky, line 80: In authui we use `hint.originalException` but we use `hint` directly here. Why is the difference?

louischan-oursky (changes requested):

portal/src/ReactApp.tsx
tung2744: Sorry it is a mistake, and as the type is `unknown` typecheck cannot catch it

tung2744: Fixed, thanks!

louischan-oursky (approved):

PR #4950, opened 2024-12-18
Expose delete_at in user export

tung2744:
ref DEV-2353

### DIFF
--- a/docs/specs/user-export.md
+++ b/docs/specs/user-export.md
@@ -550,7 +550,7 @@ Here is an example of the record
 - `roles`: The role keys of all roles directly assigned to the user. This DOES NOT include roles implied by the groups.
 - `groups`: The group keys of all groups assigned to the user.
 - `disabled`: Whether the user is disabled.
-- `delete_at`: The timestamp in rfc3339 when the user will be deleted. `null` if deletion is not scheduled.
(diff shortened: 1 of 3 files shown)


docs/specs/user-export.md
```diff
@@ -549,6 +550,7 @@ Here is an example of the record
 - `roles`: The role keys of all roles directly assigned to the user. This DOES NOT include roles implied by the groups.
 - `groups`: The group keys of all groups assigned to the user.
 - `disabled`: Whether the user is disabled.
+- `delete_at`: The timestamp in rfc3339 when the user will be deleted. `null` if deletion is not scheduled.
```
louischan-oursky, line 553: It is not `null`. It is absent when deletion is not scheduled, because we use `omitempty` in the struct tag.

So please change to

> The timestamp in RFC3339 format indicating when the scheduled deletion of the user will happen. If it is absent, the user is not scheduled for deletion.

louischan-oursky (changes requested):

docs/specs/user-export.md
tung2744: Updated, thanks!

louischan-oursky (approved):

louischan-oursky:
I added one more commit to include `delete_at` in the default export fields.

PR #4951, opened 2024-12-19
Fix some <Route> that have children does not have <Route index={true}>

louischan-oursky:
ref DEV-2367


tung2744 (approved):

louischan-oursky:
Wait

louischan-oursky:
I discovered one more problem.

louischan-oursky:
@tung2744 Added some commits to fix the problem of error being entirely ignored in the previous approach. Now we use the error boundary provided by Sentry. It capture the error automatically, but still respect beforeSend. I also update isNetworkError to classify error produced by `import()`. The original work handle `fetch()` only.

tung2744 (approved):

tung2744:
A lint failed

louischan-oursky:
Added a commit to ignore the false positive.

PR #4956, opened 2024-12-20
Some settings v2 ui fixes

tung2744:
ref DEV-2344

I found another issue when removing the duplicated error alert, follow up in another issue:
https://linear.app/authgear/issue/DEV-2390/%5Bsettingsv2%5D-hide-error-alert-for-some-errors


authui/src/authflowv2/components/settings-item.css
```diff
@@ -55,14 +72,19 @@
     --settings-item__forward-arrow-color: var(--color-neutral-200);
 
     --settings-item__border-color: var(--color-neutral-400);
+
+    :root.dark {
+      --settings-item-content__text-color: var(--color-neutral-200);
+    }
```
louischan-oursky, line 79: This is already inside a `:root.dark` block. Do we really need a nested `:root.dark`?

louischan-oursky (changes requested):

authui/src/authflowv2/components/settings-item.css
tung2744: Right it is redundant, removed in latest commit

authui/src/authflowv2/components/settings-item.css
```diff
@@ -61,6 +61,7 @@
       --typography-title-medium__font-weight
     );
     --settings-item-content__text-color: var(--color-neutral-400);
+    --settings-item-content__text-color--highligh: var(--color-link);
```
louischan-oursky, line 64: typo in `highligh` and it should be `highlighted` (adjective)

louischan-oursky (changes requested):
conflicted.

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #4964, opened 2025-01-06
Settings ui updates - Part 2

tung2744:
ref DEV-2344
based on https://github.com/authgear/authgear-server/pull/4956

Maybe you would like to pay attention to this commit: https://github.com/authgear/authgear-server/pull/4964/commits/4c850b81af0b4a63836487a494c6b7c01f67945b
I've disabled the length-zero-no-unit rule in this commit because it removes units from css variables which breaks calculation of calc.


authui/src/authflowv2/components/settings-content.css
```diff
@@ -9,7 +9,9 @@
 
     @media (min-width: theme("screens.tablet")) {
       margin: 0 auto;
-      width: var(--settings-content__width);
+      width: calc(
+        var(--settings-content__width) + var(--widget-content__px, 0px)
```
louischan-oursky, line 13: It is better to define `:root { --widget-content__px: 0px}` in `widget.css`, so in here, we do not need to provide a custom fallback value.

authui/package.json
```diff
@@ -16,7 +16,6 @@
     "stylelint": "stylelint --cache './src/**/*.{css,scss}'",
     "dev": "vite",
     "build": "vite build --outDir '../resources/authgear/generated'",
-    "watch": "vite build --watch --outDir '../resources/authgear/generated'",
```
louischan-oursky, line 19: But simply running `vite` will not output to `'../resources/authgear/generated'`, isn't it?

authui/.stylelintrc.json
```diff
@@ -20,6 +16,7 @@
       {
         "ignoreAtRules": ["mixin", "include", "tailwind", "layer"]
       }
-    ]
+    ],
+    "length-zero-no-unit": null
```
louischan-oursky, line 20: Opened this issue to follow up https://github.com/oursky/stylelint-oursky/issues/23

authui/package.json
tung2744: Yes it won't. The current `dev` commands works by generating a `__generated_asset.html` which contains urls pointing to the vite development server.
For development purpose `dev` should be enough therefore we don't need `watch`. The reason I remove `watch` is it always remove the `.gitkeep` inside `resources/authgear/generated` and it is easy to commit this change accidentally.

tung2744:
Rebased and updated, thanks!

tung2744:
Updated to remove the new css variable and use a component class instead, thanks!

tung2744:
Added commits to remove settings-action-item and replace it with settings-item

louischan-oursky (approved):

PR #4975, opened 2025-01-10
Switch to use sessionType=refresh_token

louischan-oursky:
ref DEV-2326

~~Depend on the bug fix https://github.com/authgear/authgear-sdk-js/pull/336~~ Included in this PR already.


portal/src/graphql/portal/Authenticated.tsx
```diff
@@ -122,11 +173,79 @@ export interface ConfigureAuthgearOptions {
 export async function configureAuthgear(
   options: ConfigureAuthgearOptions
 ): Promise<void> {
+  // eslint-disable-next-line no-console -- Output the session type to console for easier debugging.
+  console.log("authgear: sessionType =", SESSION_TYPE);
```
tung2744: console.debug or console.info can be used to avoid lint error, and also give it a proper log level

louischan-oursky: Currently, we only allow `warn` and `error`. I change this call to `console.info` and keep the `no-console` exception.

tung2744 (approved):

PR #4977, opened 2025-01-13
Mutate and query sms gateway credentials in apis

tung2744:
ref DEV-2301


tung2744:
Included in https://github.com/authgear/authgear-server/pull/4982

PR #4983, opened 2025-01-15
Fix nil pointer exception when usage limit reached

tung2744:
ref DEV-2430

### DIFF
--- a/pkg/lib/messaging/limits.go
+++ b/pkg/lib/messaging/limits.go
@@ -51,7 +51,10 @@ type Limits struct {
 }
 
 func (l *Limits) cancel(ctx context.Context, reservation *Reservation) {
-	l.UsageLimiter.Cancel(ctx, reservation.UsageReservation)
+	// Usage reservation can be nil if usage limit reached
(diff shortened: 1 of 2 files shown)


pkg/lib/messaging/limits.go
```diff
@@ -51,7 +51,10 @@ type Limits struct {
 }
 
 func (l *Limits) cancel(ctx context.Context, reservation *Reservation) {
-	l.UsageLimiter.Cancel(ctx, reservation.UsageReservation)
+	// Usage reservation can be nil if usage limit reached
+	if reservation.UsageReservation != nil {
+		l.UsageLimiter.Cancel(ctx, reservation.UsageReservation)
+	}
```
louischan-oursky, line 57: Instead of fixing here, we should make Cancel() handle nil reservation.

In pkg/lib/ratelimit/limiter.go Cancel() handle nil Reservation too.

louischan-oursky (changes requested):

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #4984, opened 2025-01-16
Display a fatal error if no sms client is available

tung2744:
ref DEV-2301


pkg/lib/infra/sms/sender.go
```diff
@@ -9,16 +9,14 @@ import (
...
+type Sender struct {
 	Logger         Logger
 	ClientResolver *ClientResolver
 }
 
-var _ smsapi.Client = (*Client)(nil)
-
-func (c *Client) Send(ctx context.Context, opts smsapi.SendOptions) error {
+func (c *Sender) Send(ctx context.Context, client smsapi.Client, opts smsapi.SendOptions) error {
 	client, _, err := c.ClientResolver.ResolveClient()
```
louischan-oursky, line 20: I think we should use the client in the argument list.

louischan-oursky (changes requested):

pkg/lib/infra/sms/sender.go
tung2744: Right, fixed thanks!

louischan-oursky (approved):

PR #4986, opened 2025-01-17
Make purpose part of the pool key

louischan-oursky:
ref DEV-2444

### DIFF
--- a/pkg/lib/infra/db/appdb/deps.go
+++ b/pkg/lib/infra/db/appdb/deps.go
@@ -73,14 +73,17 @@ func NewHandle(
 	credentials *config.DatabaseCredentials,
 	lf *log.Factory,
 ) *Handle {
+	info := db.ConnectionInfo{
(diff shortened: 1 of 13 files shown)


pkg/lib/infra/db/appdb/deps.go
```diff
@@ -73,14 +73,17 @@ func NewHandle(
 	credentials *config.DatabaseCredentials,
 	lf *log.Factory,
 ) *Handle {
+	info := db.ConnectionInfo{
+		Purpose:     db.ConnectionPurposeMain,
```
tung2744, line 77: Should global db and app db has different purpose?
The global db is used for configsource in db and always hold one connection for listening to pg pubsub.

louischan-oursky: You're correct. I amended the commit to add `global`, and split `audit` into `audit_read_only` and `audit_read_write`

PR #4988, opened 2025-01-20
Vendor in https://github.com/authgear/graphql-go-relay/tree/f56c3cce7

louischan-oursky:


tung2744:
Why do we do this?

louischan-oursky:
It is because

1. The source repository has not been updated for a long time.
2. This go module is relatively simple to be vendored.
3. The fork we maintain at https://github.com/authgear/graphql-go-relay often has vulnerabilities that we need to fix. It is more convenient if we can vendored this repository, so that we only need to fix vulnerabilities once.

tung2744 (approved):

PR #4990, opened 2025-01-20
Implement test sms

tung2744:
ref DEV-2301

Screenshot 2025-01-20 at 4 35 36 PM


pkg/portal/graphql/sms_mutation.go
```diff
@@ -0,0 +1,104 @@
...
+			Description: "The sms provider configuration.",
+		},
+	},
+})
+
+var smsProviderConfigurationInput = graphql.NewInputObject(graphql.InputObjectConfig{
+	Name: "SMSProviderConfigurationInput",
+	Fields: graphql.InputObjectConfigFieldMap{
+		"twilio": &graphql.InputObjectFieldConfig{
+			Type:        graphql.NewNonNull(graphql.ID),
```
louischan-oursky: The type is incorrect. Should it be SMSProviderConfigurationTwilioInput?

pkg/portal/graphql/sms_mutation.go
```diff
@@ -0,0 +1,104 @@
...
+	Name: "SMSProviderConfigurationTwilioInput",
+	Fields: graphql.InputObjectConfigFieldMap{
+		"accountSID": &graphql.InputObjectFieldConfig{
+			Type: graphql.NewNonNull(graphql.String),
+		},
+		"authToken": &graphql.InputObjectFieldConfig{
+			Type: graphql.String,
+		},
+		"messagingServiceSID": &graphql.InputObjectFieldConfig{
+			Type: graphql.NewNonNull(graphql.String),
```
louischan-oursky, line 59: This should be optional.

pkg/portal/graphql/sms_mutation.go
```diff
@@ -98,6 +98,13 @@ var _ = registerMutationField(
 				return nil, err
 			}
 
+			if providerConfig.Twilio != nil {
+				err = gqlCtx.SMSService.SendByTwilio(ctx, app, to, *providerConfig.Twilio)
```
louischan-oursky, line 135: The SMS service should expose a single method that takes model.SMSProviderConfigurationInput instead, so we do not need to have a long if-then-else block here.

pkg/portal/sms/service.go
```diff
@@ -0,0 +1,46 @@
...
+	})
+
+	translationService := NewTranslationService(app)
+	sender, err := translationService.GetSenderForTestSMS(ctx)
+	if err != nil {
+		return err
+	}
+
+	return twilioClient.Send(ctx, smsapi.SendOptions{
+		Sender: sender,
```
louischan-oursky, line 50: I discover a design problem here. We should allow the developer to configure `sms.default.sender` in this page. And in the send-test-sms mutation input, we should take the sender, instead of deriving the sender from the current project configuration.

The SMTP page has the same problem. See https://linear.app/authgear/issue/DEV-2310/allow-setting-default-sender-in-the-custom-email-provider-page

`sms.default.sender` is not always used. We should document this in the configuration page as well. In particular, when Twilio messaging service SID is specified, the sender is ignored.

cc @fungc-io

pkg/portal/graphql/sms_mutation.go
```diff
@@ -85,13 +85,13 @@ var smsProviderConfigurationDenoInput = graphql.NewInputObject(graphql.InputObje
 })
 
 var _ = registerMutationField(
-	"sendTestSMTPConfigurationEmail",
+	"sendTestSMSConfiguration",
 	&graphql.Field{
-		Description: "Send test STMP configuration email",
+		Description: "Send test SMS configuration",
```
louischan-oursky, line 91: Change this to `Send a SMS to test the configuration`

pkg/portal/graphql/sms_mutation.go
```diff
@@ -0,0 +1,160 @@
...
+			Type:        graphql.NewNonNull(graphql.ID),
+			Description: "App ID to test.",
+		},
+		"to": &graphql.InputObjectFieldConfig{
+			Type:        graphql.NewNonNull(graphql.String),
+			Description: "The recipient phone number.",
+		},
+		"providerConfiguration": &graphql.InputObjectFieldConfig{
+			Type:        graphql.NewNonNull(smsProviderConfigurationInput),
+			Description: "The sms provider configuration.",
```
louischan-oursky, line 26: `The SMS provider configuration.`

pkg/portal/graphql/sms_mutation.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+	},
+})
+
+var smsProviderConfigurationInput = graphql.NewInputObject(graphql.InputObjectConfig{
+	Name: "SMSProviderConfigurationInput",
+	Fields: graphql.InputObjectConfigFieldMap{
+		"twilio": &graphql.InputObjectFieldConfig{
+			Type:        smsProviderConfigurationTwilioInput,
+			Description: "Configuration of Twilio",
```
louischan-oursky, line 36: `Twilio configuration`

pkg/portal/graphql/sms_mutation.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+		"webhook": &graphql.InputObjectFieldConfig{
+			Type:        smsProviderConfigurationWebhookInput,
+			Description: "Configuration of Webhook",
```
louischan-oursky, line 40: `Webhook Configuration`

pkg/portal/graphql/sms_mutation.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+		"deno": &graphql.InputObjectFieldConfig{
+			Type:        smsProviderConfigurationDenoInput,
+			Description: "Configuration of Deno hook",
```
louischan-oursky, line 44: `Deno hook configuration`

louischan-oursky (changes requested):

pkg/portal/sms/service.go
fungc-io: I see, got it

pkg/portal/graphql/sms_mutation.go
tung2744: Fixed in later commits

pkg/portal/sms/service.go
tung2744: I think I will handle this in later pr.

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #4995, opened 2025-01-22
Fix settings action always return error

tung2744:
ref DEV-2458

### DIFF
--- a/pkg/auth/webapp/session_middleware.go
+++ b/pkg/auth/webapp/session_middleware.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"net/url"
 
 	"github.com/authgear/authgear-server/pkg/lib/oauth/oauthsession"
 	"github.com/authgear/authgear-server/pkg/lib/oauth/oidc"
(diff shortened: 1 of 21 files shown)


pkg/auth/webapp/session_middleware.go
```diff
@@ -58,27 +59,42 @@ func (m *SessionMiddleware) Handle(next http.Handler) http.Handler {
...
+			u := url.URL{
+				Path:     r.URL.Path,
+				RawQuery: r.URL.Query().Encode(),
+			}
+			redirect := httputil.ResultRedirect{
+				URL:     u.String(),
+				Cookies: result.Cookies,
+			}
+			redirect.WriteResponse(w, r)
+			return
```
louischan-oursky, line 77: In what condition does this double create happen?

Or this problem is solved by the next commit (that is, do not mount this twice)

louischan-oursky (changes requested):

pkg/auth/webapp/session_middleware.go
tung2744: 1. When we open settings action, the url is something like `/change_password?x_ref=abcd`. Then in this request, we create a new web session.
2. Then, because the original session middleware does not remove the `x_ref` from the url (It remove the query parameter from the URL object in code, but does not trigger a redirect so that the query parameter still exist in the url of the browser), when the webapp POST a form to the current location, the x_ref exist in the URL, so that the POST request created a new webapp session.
3. Even if 2 is fixed by other means (e.g. I tried to fix it by checking the request method), in flows which include multiple steps, the x_ref query will be preserved and brings to the next step's GET request, so a new web session is again created.

I finally fixed it by immediately trigger a redirect which removes the x_ref from url to ensure x_ref is only consumed once.

tung2744: And the double mounted session middleware actually does not cause any problem in my testing, I just discovered that during debugging and therefore removed it.

louischan-oursky: I see. I added a commit to remove unused return values.

louischan-oursky (approved):

PR #5001, opened 2025-01-24
Support add login id setting actions

tung2744:
ref DEV-2411

Can test with https://github.com/authgear/authgear-sdk-flutter/pull/72

### DIFF
--- a/pkg/auth/handler/webapp/controller.go
+++ b/pkg/auth/handler/webapp/controller.go
@@ -2,15 +2,18 @@ package webapp
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/authgear/authgear-server/pkg/api/apierrors"
 	"github.com/authgear/authgear-server/pkg/auth/handler/webapp/viewmodels"
 	"github.com/authgear/authgear-server/pkg/auth/webapp"
+	"github.com/authgear/authgear-server/pkg/lib/authn/authenticationinfo"
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/lib/infra/db/appdb"
 	"github.com/authgear/authgear-server/pkg/lib/infra/redis/appredis"
 	"github.com/authgear/authgear-server/pkg/lib/interaction"
+	"github.com/authgear/authgear-server/pkg/lib/oauth/oauthsession"
 	"github.com/authgear/authgear-server/pkg/lib/session"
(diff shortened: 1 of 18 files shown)


pkg/auth/handler/webapp/controller.go
```diff
@@ -346,3 +374,59 @@ func (c *Controller) InteractionOAuthCallback(ctx context.Context, oauthInput In
...
+
+	entry.T.SettingsActionResult = oauthsession.NewSettingsActionResult()
+	err = c.OAuthSessions.Save(ctx, entry)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Controller) GetSettingsActionResult(ctx context.Context, webSession *webapp.Session) (*webapp.Result, bool, error) {
```
louischan-oursky, line 414: The caller will panic when this function returns `nil, false, nil`. I suggest we move the panic inside this function instead.

Specifically, this function panic if one of the following holds

1. webSession is nil
2. webSession.RedirectURI is empty
3. webSession.Extra["authentication_info_id"] is not a string.
4. webSession.Extra["authentication_info_id"] is an empty string.

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -586,6 +586,9 @@ func (h *AuthorizationHandler) handleSettingsAction(
 ) (httputil.Result, error) {
 	redirectURI, err := h.UIURLBuilder.BuildSettingsActionURL(client, r, oauthSessionEntry)
 	if err != nil {
+		if apierrors.IsKind(err, oidc.ErrInvalidSettingsAction) {
```
louischan-oursky, line 589: I suggest we change ErrInvalidSettingsAction to be a protocol error, so that we will not forget the translation (as there is no translation anymore).

louischan-oursky (changes requested):

pkg/auth/handler/webapp/controller.go
tung2744: > The caller will panic when this function returns nil, false, nil
Why? If the boolean is false, the first return value shouldn't be used so it won't panic. The meaning of such a return result is "we are not in a setting action flow", which is a possible case.
Caller should always check the boolean before using the result.

louischan-oursky: My comment was written before I saw the commit titled `Reduce code duplication by creating a method to finish and redirect together`.

It looks good to me now.

louischan-oursky (approved):

PR #5004, opened 2025-02-03
Handle "name" of Sign in with Apple

louischan-oursky:
ref DEV-2468

### DIFF
--- a/pkg/lib/authn/identity/oauth/provider.go
+++ b/pkg/lib/authn/identity/oauth/provider.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/authgear/authgear-server/pkg/lib/authn/identity"
 	"github.com/authgear/authgear-server/pkg/lib/config"
+	liboauthrelyingparty "github.com/authgear/authgear-server/pkg/lib/oauthrelyingparty"
 	"github.com/authgear/authgear-server/pkg/util/clock"
 	"github.com/authgear/authgear-server/pkg/util/uuid"
 )
@@ -90,8 +91,15 @@ func (p *Provider) WithUpdate(
 	claims map[string]interface{},
 ) *identity.OAuth {
 	newIden := *iden
-	newIden.UserProfile = rawProfile
-	newIden.Claims = claims
+	// For non-Apple provider, we can just update.
+	// For Apple, we need to merge given_name and family_name because
+	// they only available at THE FIRST TIME authorization.
+	if newIden.ProviderID.Type == liboauthrelyingparty.TypeApple {
+		newIden.Apple_MergeRawProfileAndClaims(rawProfile, claims)
+	} else {
+		newIden.UserProfile = rawProfile
+		newIden.Claims = claims
+	}
 
 	return &newIden
 }
--- a/pkg/lib/oauthrelyingparty/apple/provider.go
+++ b/pkg/lib/oauthrelyingparty/apple/provider.go
@@ -2,6 +2,9 @@ package apple
 
 import (
 	"context"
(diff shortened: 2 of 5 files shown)


pkg/lib/oauthrelyingparty/apple/provider.go
```diff
@@ -199,6 +215,41 @@ func (p Apple) GetUserProfile(ctx context.Context, deps oauthrelyingparty.Depend
...
+	//
+	// Instead, the name of the end-user is only available in the response when the end-user authorizes THE FIRST TIME.
+	// https://developer.apple.com/documentation/sign_in_with_apple/incorporating-sign-in-with-apple-into-other-platforms#Handle-the-response
+	//
+	// In THE FIRST TIME authorization, the form post request includes a form field named "user".
+	// This form field is a JSON document, with the following shape.
+	//
+	// { "name": { "firstName": "string", "lastName": "string" }, "email": "string" }
+	//
+	// Since "user" only appears THE FIRST TIME, you MUST revoke the authorization if
```
tung2744, line 229: @louischan-oursky If user only exist in the first time, later if the user sign in again, as the claims of the identity will be updated, the name will be disappeared?

tung2744: In fact, I seems didn't find any code that populate a "name" into the authgear standard user attribute. It seems what we populate is only `email`, `phone_number` and `username`.

louischan-oursky: > If user only exist in the first time, later if the user sign in again, as the claims of the identity will be updated, the name will be disappeared?

Yes, I overlooked this. Our OAuth abstraction is now stateless, it does not have access to the existing profile. Maybe we can only add a special case in identity service to "merge" Apple user profile and standard attributes.

louischan-oursky: > In fact, I seems didn't find any code that populate a "name" into the authgear standard user attribute. It seems what we populate is only email, phone_number and username.

See pkg/lib/feature/stdattrs/service.go:44 `PopulateStandardAttributes`

louischan-oursky: Added a commit to address this problem.

tung2744 (approved):

pkg/lib/authn/identity/oauth/provider.go
```diff
@@ -90,8 +91,15 @@ func (p *Provider) WithUpdate(
...
-	newIden.Claims = claims
+	// For non-Apple provider, we can just update.
+	// For Apple, we need to merge given_name and family_name because
+	// they only available at THE FIRST TIME authorization.
+	if newIden.ProviderID.Type == liboauthrelyingparty.TypeApple {
+		newIden.Apple_MergeRawProfileAndClaims(rawProfile, claims)
+	} else {
+		newIden.UserProfile = rawProfile
+		newIden.Claims = claims
+	}
```
tung2744, line 102: Or instead of having a special code path for apple, we can always "merge" the original claims and the new claims, by setting all existing keys in the new claims in the original claims, and ignore non-existing keys?

louischan-oursky: > setting all existing keys in the new claims in the original claims, and ignore non-existing keys?

I didn't do this because

1. existing claims has key `a`
2. incoming claims DOES NOT have key `a`

Then the resulting claims should not have `a`.

But in case of Apple, we want the reverse, we want `given_name` and `family_name` to be present.

PR #5008, opened 2025-02-05
Use another port for vite hmr

tung2744:
When using vite development server, there are always some http requests load indefinitely. It seems that the default port of hmr of vite development server is `3100` which collides with the port we used for authgear server.


louischan-oursky:
Where can I find the documentation about the default being 3100?

tung2744:
https://vite.dev/config/server-options#server-hmr

The documentation does not mention the default port used. But by inspecting the network request I found it is using 3100.

louischan-oursky (approved):

PR #5009, opened 2025-02-06
Display a redirecting page if user try to visit a completed web session

tung2744:
ref DEV-2484

There are still some remaining problems. See DEV-2484 for details.


tung2744:
I found this breaks the normal settings page because we don't have web session in that case. Let me mark as WIP first.

tung2744:
Added https://github.com/authgear/authgear-server/pull/5009/commits/4369fcb59b8f8f5591a2167be0754b3e8622c31d to fix the problem mentioned above

louischan-oursky (changes requested):

tung2744:
@louischan-oursky Updated, thanks! Also added commits which rewrites change password so that it handle redirect same as other settings action.
Changes start from [15b7761](https://github.com/authgear/authgear-server/pull/5009/commits/15b7761fb055d0d2b3fcf888abd964346130200c)

louischan-oursky:
Please rebase

louischan-oursky (approved):

PR #5012, opened 2025-02-12
Classify sms errors

tung2744:
ref DEV-2454
Test with https://github.com/authgear/authgear-sms-gateway/pull/23


pkg/lib/authn/otp/sender.go
```diff
@@ -41,7 +41,7 @@ type TranslationService interface {
 
 type Sender interface {
 	SendEmailInNewGoroutine(ctx context.Context, msgType translation.MessageType, opts *mail.SendOptions) error
-	SendSMSInNewGoroutine(ctx context.Context, msgType translation.MessageType, opts *sms.SendOptions) error
+	SendSMSImmediately(ctx context.Context, msgType translation.MessageType, opts *sms.SendOptions) error
```
louischan-oursky, line 44: Changing this will have an great impact on the duration of database transaction.

I think we need to first restructure authflow code to make it support running I/O operation without the context of a ongoing database transaction.

pkg/lib/infra/sms/twilio/model.go
```diff
@@ -36,3 +36,16 @@ func ParseSendResponse(jsonData []byte) (*SendResponse, error) {
 	}
 	return &response, nil
 }
+
+type ErrorResponse struct {
+	Code int `json:"code,omitempty"`
```
louischan-oursky, line 41: Other than `code`, does it have any other documented fields? Any reference you can find on their official documentation?

pkg/lib/infra/sms/twilio/client.go
```diff
@@ -98,15 +99,70 @@ func (t *TwilioClient) Send(ctx context.Context, options smsapi.SendOptions) err
...
+	if sendResponse.ErrorCode != nil {
+		return t.makeError(*sendResponse.ErrorCode, dumpedResponse)
 	}
 
-	// Failed case.
-	return &smsapi.SendError{
+	return nil
+}
+
+func (t *TwilioClient) parseAndHandleErrorResponse(
```
louischan-oursky, line 109: This seems unused.

pkg/lib/infra/sms/custom/model.go
```diff
@@ -12,3 +14,19 @@ type SendOptions struct {
 	LanguageTag       string                    `json:"language_tag"`
 	TemplateVariables *smsapi.TemplateVariables `json:"template_variables"`
 }
+
+// See https://github.com/authgear/authgear-sms-gateway/blob/main/pkg/handler/api.go
+type ResponseBody struct {
+	Code           string  `json:"code"`
+	ErrorDetail    *string `json:"error_detail,omitempty"`
```
louischan-oursky, line 21: It is named `error_description` there.

pkg/lib/infra/sms/custom/client.go
```diff
@@ -94,6 +94,10 @@ func (w *SMSWebHook) Call(ctx context.Context, u *url.URL, payload SendOptions)
 
 	responseBody, err := ParseResponseBody(bodyBytes)
 	if err != nil {
+		// This is not something we understand, check the status code to determine if it is a success
+		if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+			return nil
+		}
```
louischan-oursky, line 100: We should check if the error is produced by the `json` package. Please add tests to ensure this does not break legacy gateways.

louischan-oursky (changes requested):

pkg/lib/infra/sms/custom/model.go
tung2744: I am planning to add a new field in the response body, so that the sms gateway can return detail about the error which I we can display it in the ui. For example, the error code returned from the gateway.

pkg/lib/infra/sms/twilio/model.go
tung2744: I cannot find the reference, but in my testing it returns 

```
{
  "code": 21211,
  "message": "Invalid 'To' Phone Number: xxx",
  "more_info": "https://www.twilio.com/docs/errors/21211",
  "status": 400
}
```

tung2744: I found it here, you need to choose "json" to view the error in json.
https://www.twilio.com/docs/usage/twilios-response

Let me add a comment to reference this url.

louischan-oursky: It seems that all fields are optional. If `code` is not present, treat it as unknown error.

tung2744: Yes, this is handled in parseAndHandleErrorResponse, if it is something we do not understand (Including response without a code, or even a html), we simply treat as unknown error.

pkg/lib/authn/otp/sender.go
louischan-oursky: Per offline discussion, we will change the method signature of ReactTo to support delayed one-time function.

@tung2744

1. Keep SendSMSInNewGoroutine (for latte) in 
2. Introduce SendSMSImmediately
3. Split `Send` in pkg/lib/authn/otp/sender.go into `SendImmediately` and `SendInNewGoroutine`.
4. authflow uses SendImmediately, while latte uses SendInNewGoroutine

pkg/lib/authenticationflow/accept.go
```diff
@@ -131,26 +138,24 @@ func accept(ctx context.Context, deps *Dependencies, flows Flows, inputFn func(i
...
 				changed = true
-
-				return &AcceptResult{
-					BotProtectionVerificationResult: &BotProtectionVerificationResult{
-						Outcome: BotProtectionVerificationOutcomeFailed,
-					}}, botprotection.ErrVerificationFailed
+				result.BotProtectionVerificationResult = &BotProtectionVerificationResult{
+					Outcome: BotProtectionVerificationOutcomeFailed,
+				}
+				return result, botprotection.ErrVerificationFailed
```
louischan-oursky, line 150: This function uses named return value, so please use named return value consistently. Mixing them is a source of confusion.

pkg/lib/authenticationflow/accept.go
```diff
@@ -131,26 +138,24 @@ func accept(ctx context.Context, deps *Dependencies, flows Flows, inputFn func(i
...
 				// We still consider the flow has something changes.
 				changed = true
-				return &AcceptResult{
-					BotProtectionVerificationResult: &BotProtectionVerificationResult{
-						Outcome: BotProtectionVerificationOutcomeFailed,
-					}}, botprotection.ErrVerificationServiceUnavailable
+				result.BotProtectionVerificationResult = &BotProtectionVerificationResult{
+					Outcome: BotProtectionVerificationOutcomeFailed,
+				}
+				return result, botprotection.ErrVerificationServiceUnavailable
```
louischan-oursky, line 158: Use named return value consistently.

pkg/lib/authenticationflow/accept.go
```diff
@@ -171,12 +176,15 @@ func accept(ctx context.Context, deps *Dependencies, flows Flows, inputFn func(i
...
+			nextNode = *reactToResult.Node
+			result.DelayedOneTimeFunctions = append(result.DelayedOneTimeFunctions, reactToResult.DelayedOneTimeFunction)
 		default:
-			panic(fmt.Errorf("uxepected type of ReactToResult %t", reactToResult))
+			panic(fmt.Errorf("failed to append node: uxepected type of ReactToResult %t", reactToResult))
 		}
 		err = appendNode(ctx, deps, findInputReactorResult.Flows, nextNode)
 		if err != nil {
-			return
+			return result, err
```
louischan-oursky, line 187: Use named return value consistently.

pkg/lib/messaging/sender.go
```diff
@@ -190,25 +198,41 @@ func (s *Sender) SendSMSImmediately(ctx context.Context, msgType translation.Mes
...
 	}
 
-	err = s.SMSSender.Send(ctx, client, *opts)
-	if err != nil {
-		otelauthgear.IntCounterAddOne(
-			ctx,
-			otelauthgear.CounterSMSRequestCount,
-			otelauthgear.WithStatusError(),
-		)
+	send := func() error {
```
louischan-oursky, line 201: This send function should take `ctx` as the first argument.

pkg/lib/messaging/sender.go
```diff
@@ -190,25 +198,41 @@ func (s *Sender) SendSMSImmediately(ctx context.Context, msgType translation.Mes
...
+		}
+		return nil
+	}
+
+	if isAsync {
+		go func() {
+			// Detach the deadline so that the context is not canceled along with the request.
+			ctx = context.WithoutCancel(ctx)
+			_ = send()
+		}()
```
louischan-oursky, line 230:
```golang
if isAsync {
  // Re-define ctx here so that the defer statement body captures this without-cancel ctx.
  ctx = context.WithoutCancel(ctx)
  // Pass the ctx to the goroutine explicitly.
  go func(ctx context.Context) {
      _ = send(ctx)
  }(ctx)
}
```

louischan-oursky (changes requested):

pkg/lib/messaging/sender.go
tung2744: If we pass ctx to the go func by parameter, why do we need redefine?

pkg/lib/authenticationflow/accept.go
tung2744: Maybe I will change it to unnamed return values. I found it difficult to understand what is the return value when there is a loop

pkg/lib/messaging/sender.go
louischan-oursky: For readability.

```golang
if isAsync {
  ctx = context.WithoutCancel(ctx)
  // Pass the ctx to the goroutine explicitly.
  go func(ctx context.Context) {
      _ = send(ctx)
  }(ctx)
}
```

vs

```golang
if isAsync {
  // Pass the ctx to the goroutine explicitly.
  go func(ctx context.Context) {
      _ = send(ctx)
  }(context.WithoutCancel(ctx))
}
```

I think the former is more readable.

tung2744: But that actually exposed the `ctx` outside the block, which can then be used incorrectly by code under it. I prefer do it thisway

```
if isAsync {
  ctxWithoutCancel := context.WithoutCancel(ctx)
  // Pass the ctx to the goroutine explicitly.
  go func(ctxWithoutCancel context.Context) {
      _ = send(ctxWithoutCancel)
  }(ctxWithoutCancel)
}
```

pkg/lib/authenticationflow/node.go
```diff
@@ -13,6 +13,12 @@ type Node struct {
 	SubFlow *Flow      `json:"flow,omitempty"`
 }
 
+var _ ReactToResult = &Node{}
+
+func (n *Node) reactToResult() {
+	panic("unimplemented")
```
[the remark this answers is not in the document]
tung2744: I've removed it in later commits.

pkg/lib/infra/sms/custom/model.go
tung2744: As discussed it will be replaced by

```
	ProviderName      string `json:"provider_name,omitempty"`
	ProviderErrorCode string `json:"provider_error_code,omitempty"`
```

pkg/lib/infra/sms/custom/client.go
tung2744: Originally we don't even try to read the response and only check the statusCode. If we want to ensure it behave the same as before we should still treat it as success even it is not a json parsing error.

pkg/lib/messaging/sender.go
louischan-oursky: Rename the argument to `ctx`. 

```golang
if isAsync {
  ctxWithoutCancel := context.WithoutCancel(ctx)
  go func(ctx context.Context) {
      _ = send(ctx)
  }(ctxWithoutCancel)
}
```

pkg/lib/authenticationflow/accept.go
tung2744: I found that this function must be using named return value because the defer function will change the return value, so I will change it back to named return values.

tung2744:
I've addressed the comments above, but e2e tests are failing so I'm checking.

tung2744:
This is ready for review, thanks!

tung2744:
TODO: 
- [ ] Update translations

pkg/lib/infra/sms/twilio/client.go
```diff
@@ -168,7 +168,7 @@ func (t *TwilioClient) makeError(
 				"Detail": errorCode,
 			}), err)
 	case 30002:
-		err = errors.Join(smsapi.ErrKindAuthorizationFailed.NewWithInfo(
+		err = errors.Join(smsapi.ErrKindDeliveryRejected.NewWithInfo(
 			"twilio authorization failed", apierrors.Details{
```
louischan-oursky: The message needs update as well. Please also cross-check other providers.

pkg/lib/infra/sms/custom/client.go
```diff
@@ -65,67 +76,101 @@ type SMSWebHook struct {
...
+		return nil
+	}
+
+	jsonText, err := json.Marshal(output)
 	if err != nil {
 		return err
 	}
 
-	return nil
+	responseBody, err := ParseResponseBody(jsonText)
```
louischan-oursky, line 166: If the existing SMS deno hook return `{"code": 1}`, the json parsing will fail, but we should still treat this as success. Please add a test case to show this.

pkg/lib/infra/sms/custom/client_test.go
```diff
@@ -0,0 +1,114 @@
...
+		// And do not care about the response body
+		// We don't want to break this behavior
+
+		var webhook hook.WebHook = &mockWebHook{}
+
+		smsWebHook := &SMSWebHook{
+			WebHook: webhook,
+			Client: &mockWebHookClient{
+				ResponseStatusCode: 200,
+				ResponseBody:       io.NopCloser(strings.NewReader("")),
```
louischan-oursky, line 87: Add a test case in which the response body is `{"status": 1}`.

louischan-oursky (changes requested):

pkg/lib/infra/sms/twilio/client.go
tung2744: Should be fixed in later commits.

pkg/lib/infra/sms/custom/client.go
tung2744: Currently, sms deno hook is not allowed to return anything. So existing deno hooks should return `null`.

louischan-oursky: What do you mean by not allowed? I think the hook can return anything they want, given that they write enough type annotation to bypass type checking.

tung2744: You are right that the hook can return anything if the function return type is set to any. However I think existing hooks are configured by us so there should be no hook written using incorrect return type? In fact I am not sure if there is anyone using the deno sms hook after we have our sms gateway deployed?

louischan-oursky (approved):

PR #5014, opened 2025-02-17
Build a docker image for running ONCE - Part 2

louischan-oursky:


once/docker-entrypoint.sh
```diff
@@ -385,7 +385,65 @@ secrets:
...
+    x_application_type: traditional_webapp
+search:
+  implementation: postgresql
+ui:
+  signup_login_flow_enabled: true
+verification:
+  claims:
+    email:
+      enabled: false
+      required: false
```
tung2744, line 430: Feel like we need a go cmd program to do this, or we might forget to update this config in the future.

once/Dockerfile
```diff
@@ -151,7 +151,7 @@ COPY ./custombuild/ ./custombuild/
 # Let //go:embed to embed the built static files
 COPY --from=authgear-portal-stage-0 /usr/src/app/resources/authgear/ ./resources/authgear/
 COPY --from=authgear-portal-stage-1 /usr/src/app/dist/ ./resources/portal/static/
-RUN make build BIN_NAME=authgear-portal TARGET=portal GIT_HASH=$GIT_HASH
+RUN make build AUTHGEARONCE=1 BIN_NAME=authgear-portal TARGET=portal GIT_HASH=$GIT_HASH
```
tung2744, line 154: I think this should be done by ARG
https://docs.docker.com/reference/dockerfile/#arg

louischan-oursky: Since this is the Dockerfile to build once, AUTHGEARONCE should always be set. If it is a ARG, then it depends on the caller to set it correctly, which is no good.

once/docker-entrypoint.sh
louischan-oursky: We do have such command `authgear init --for-helm-chart`, but it is currently interactive-only, thus not suitable for scripting purpose. We can enhance it to make it detect the stdin is tty, and enter interactive / non-interactive mode automatically. Then we need to turn each question into a flag. If the flag is supplied, then the interactive question is skipped. In non-interactive mode, a missing flag means use the default.

I did not do that because it would take me some time to implement the above.

once/Dockerfile
tung2744: I have this comment because I saw this file is generated from the authgear & portal dockerfile, with some string replacement which looks like easy to break. And modifier of authgear Dockerfile & portal Dockerfile might not know this command is related to the once Dockerfile.

Another suggestion:
Use `ENV AUTHGEARONCE=1` on top of the once Dockerfile, so that the builds automatically have `AUTHGEARONCE=1` set and we don't have to do string replacements.

louischan-oursky: Added two commits to address this.

I review the official doc on Dockerfile again, and it is better to use `ARG` with default value.

tung2744 (approved):

PR #5020, opened 2025-02-20
Allow using isPossibleNumber to validate phone number

tung2744:
ref DEV-1139

https://github.com/user-attachments/assets/cd73d1da-e73d-4708-a0fd-5788796feb9a

Screenshot 2025-02-21 at 3 36 18 PM


pkg/lib/config/configsource/database.go
```diff
@@ -221,13 +222,19 @@ func (d *Database) resolveAppIDByDomain(ctx context.Context, r *http.Request) (s
...
+func (d *Database) ResolveContext(ctx context.Context, appID string) (context.Context, *config.AppContext, error) {
 	value, _ := d.appMap.LoadOrStore(appID, &dbApp{
 		appID: appID,
 		load:  &sync.Once{},
 	})
 	app := value.(*dbApp)
-	return app.Load(ctx, d)
+	appCtx, err := app.Load(ctx, d)
+	if err != nil {
+		return nil, nil, err
```
louischan-oursky: In case of error, we should return the original context. It is safer to return the original context so that the caller can write

```golang
defer func() {
  if err != nil {
    // ctx is not nil here. So doSomethingWithCtx will not be broken.
    doSomethingWithCtx(ctx, ...)
  }
}()
ctx, appCtx, err = ResolveContext()
```

pkg/lib/feature/accountanonymization/runnable.go
```diff
@@ -38,7 +38,8 @@ func (r *Runnable) Run(ctx context.Context) error {
 		return err
 	}
 	for _, appUser := range appUsers {
-		appContext, err := r.AppContextResolver.ResolveContext(ctx, appUser.AppID)
+		var appContext *config.AppContext
+		ctx, appContext, err = r.AppContextResolver.ResolveContext(ctx, appUser.AppID)
```
louischan-oursky: This is problematic. We are passing the project-specific ctx to ResolveContext in non-1st iteration of the loop.

It is better to extract the loop body into a separate function so that it takes a context.Context (the original one from Run()).

pkg/lib/feature/accountdeletion/runnable.go
```diff
@@ -38,7 +38,8 @@ func (r *Runnable) Run(ctx context.Context) error {
 		return err
 	}
 	for _, appUser := range appUsers {
-		appContext, err := r.AppContextResolver.ResolveContext(ctx, appUser.AppID)
+		var appContext *config.AppContext
+		ctx, appContext, err = r.AppContextResolver.ResolveContext(ctx, appUser.AppID)
```
louischan-oursky: We suffer the same project-specific-context-being-pass-to-next-iteration problem here.

pkg/redisqueue/consumer.go
```diff
@@ -166,22 +167,23 @@ func (c *Consumer) dequeue(ctx context.Context) (*redisqueue.Task, *deps.AppProv
...
-
-		appProvider = c.rootProvider.NewAppProvider(ctx, appCtx)
-		appProvider.LoggerFactory.DefaultFields["task_id"] = task.ID
+		appID = queueItem.AppID
 		return nil
 	})
 
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
```
louischan-oursky: Return original context in case of error.

pkg/redisqueue/consumer.go
```diff
@@ -166,22 +167,23 @@ func (c *Consumer) dequeue(ctx context.Context) (*redisqueue.Task, *deps.AppProv
...
 	}
 
-	return &task, appProvider, err
+	ctx, appCtx, err := c.configSourceController.ResolveContext(ctx, appID)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("resolve app context: %w", err)
```
louischan-oursky: Return original context in case of error.

louischan-oursky (changes requested):
1. Please rebase.
2. Most comments are no longer relevant because you change the function signature to avoid context misuse. I observe a potential pattern. If a function returns a tuple, and the tuple contains a context.Context, then it is likely that the context will be incorrect / misused. Maybe we should introduce a /devtools/goanalysis analyzer to forbid this pattern.

tung2744:
@louischan-oursky Rebased, thanks!

louischan-oursky (approved):

PR #5025, opened 2025-02-24
Update authgear_deno_hook version in portal templates

tung2744:
ref DEV-2535

### DIFF
--- a/portal/src/util/deno.ts
+++ b/portal/src/util/deno.ts
@@ -1,3 +1,3 @@
-const authgear_deno_hook_version = "v1.4.0";
+const authgear_deno_hook_version = "v1.5.0";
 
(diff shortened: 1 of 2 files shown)


portal/src/util/deno.ts
```diff
@@ -1,3 +1,3 @@
-const authgear_deno_hook_version = "v1.4.0";
+const authgear_deno_hook_version = "v1.5.0";
```
louischan-oursky, line 1: You also need to update the template in SMSProviderConfigurationScreen to use the new response type.

louischan-oursky (changes requested):

portal/src/util/deno.ts
tung2744: oh right

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #5034, opened 2025-02-26
Allow phone number if any of the possible region is within allow list

tung2744:
ref DEV-1139


pkg/util/phone/parser_test.go
```diff
@@ -235,3 +238,27 @@ func TestParse_IsPossibleNumber_ReturnE164(t *testing.T) {
...
+		}
+
+		var expectedMap map[int][]string
+
+		err = json.Unmarshal(data, &expectedMap)
+		if err != nil {
+			panic(err)
+		}
+
+		So(m, ShouldEqual, expectedMap)
```
louischan-oursky, line 262: Shouldn't we use `ShouldResemble` here? AFAIK, ShouldEqual uses `==` which cannot be used for comparing maps.

louischan-oursky: https://github.com/smarty/assertions/commit/e64f0e2935168e4829711112dc859c21e2e85e27

louischan-oursky (approved):

PR #5037, opened 2025-02-27
Return available MFAs for latte's select MFA step

IniZio:
ref DEV-2488

### DIFF
--- a/pkg/latte/intent_login.go
+++ b/pkg/latte/intent_login.go
@@ -174,10 +174,24 @@ func (i *IntentLogin) OutputData(ctx context.Context, deps *workflow.Dependencie
 		maskedEmails = append(maskedEmails, mail.MaskAddress(identity.LoginID.LoginID))
 	}
 
-	output := map[string]interface{}{}
-	output["masked_emails"] = maskedEmails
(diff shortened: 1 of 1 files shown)


pkg/latte/node_verified_authenticator.go
```diff
@@ -56,7 +56,18 @@ func (*NodeVerifiedAuthenticator) ReactTo(ctx context.Context, deps *workflow.De
...
+	}
+
+	type NodeVerifiedAuthenticatorOutput struct {
+		PasswordEnabled bool `json:"password_enabled"`
+	}
+
+	return NodeVerifiedAuthenticatorOutput{
+		PasswordEnabled: len(pwdAuthenticators) > 0,
+	}, nil
 }
```
tung2744, line 71: Currently, `inputSelectAuthenticatorType` is accepted in `IntentLogin`, thus the data should also be outputted in that intent.
I suggest we add a field `"authenticator_types": ["password", "oob_otp_email"]` which tells the frontend which options are available.

IniZio: Updated to return authenticator types

tung2744 (approved):

PR #5040, opened 2025-03-03
Debug context.Canceled

louischan-oursky:
ref DEV-2534


pkg/api/apierrors/kinds.go
```diff
@@ -46,6 +46,14 @@ func (n Name) HTTPStatus() int {
 type Kind struct {
 	Name   Name   `json:"name"`
 	Reason string `json:"reason"`
+
+	IsSkipLoggingToExternalService bool `json:"-"`
```
tung2744, line 50: In authui, the APIError could be read from cookie, thus deserialized from a serialized APIError. So I think IsSkipLoggingToExternalService could be lost.

louischan-oursky: The existing usage of IsSkipLoggingToExternalService are errors unrelated to authui, so this is not a problem at the moment.

I think we will not serialize this private field. In case in the future, we really need to support IsSkipLoggingToExternalService, then I think we should log the error at the time of writing the cookie.

louischan-oursky:
@tung2744 I am thinking of adding the detect-context-canceled-workaround in this PR too.

louischan-oursky:
I have added a commit to apply the workaround.

tung2744 (approved):

PR #5047, opened 2025-03-06
Upgrade stylelint

tung2744:
ref DEV-2523


portal/.browserslistrc
```diff
@@ -4,8 +4,8 @@
 # As of 2020-08-27, the latest version of Edge is 84.
 # And Edge has adopted a release schedule similar to Chrome and Firefox.
 # Therefore we can write this.
-iOS >= 14
-Safari >= 14
+iOS >= 14.5
+Safari >= 14.1
```
tung2744, line 8: I've increased the supported safari version so that gap in flex can be used. See if any concern.

portal/.stylelintrc.json
```diff
@@ -12,7 +12,8 @@
           "flexbox",
           "multicolumn",
           "css-nesting",
-          "css-overflow"
+          "css-overflow",
+          "css-display-contents"
```
tung2744, line 16: 1. `overflow` is a commonly used property and all browsers supports it. What is not supported is `overflow: clip`. The rule seems does not check the value and report error for all `overflow` so I disabled the rule.
2. `display: contents` Is actually ok to use if we do not use it on a button. So I disabled the rule here. See https://caniuse.com/css-display-contents

tung2744:
I also found that stylelint is not compatible with tailwindcss. For example, if I use `@apply xx` to apply a style, it will never report error.

portal/package.json
```diff
@@ -34,7 +34,7 @@
     "@graphql-codegen/typescript-react-apollo": "4.1.0",
     "@jest/globals": "29.7.0",
     "@oursky/eslint-plugin": "^15.0.1",
-    "@oursky/stylelint-oursky": "3.0.0",
+    "@oursky/stylelint-oursky": "5.1.0",
```
louischan-oursky, line 37: Use caret `^`.

portal/package.json
```diff
@@ -53,7 +53,7 @@
     "jest-environment-jsdom": "29.7.0",
     "node-html-parser": "^6.1.13",
     "prettier": "2.8.8",
-    "stylelint": "14.11.0",
+    "stylelint": "16.15.0",
```
louischan-oursky, line 56: Use caret `^`

louischan-oursky (changes requested):
To align with authui, turn on `ignorePartialSupport`

```
diff --git i/portal/.stylelintrc.json w/portal/.stylelintrc.json
index e9c2241e1..a504908bc 100644
--- i/portal/.stylelintrc.json
+++ w/portal/.stylelintrc.json
@@ -4,16 +4,10 @@
     "plugin/no-unsupported-browser-features": [
       true,
       {
+        "ignorePartialSupport": true,
         "ignore": [
-          "viewport-units",
-          "css-appearance",
-          "css-gradients",
           "css3-cursors",
-          "flexbox",
-          "multicolumn",
-          "css-nesting",
-          "css-overflow",
-          "css-display-contents"
+          "css-nesting"
         ]
       }
     ],

```

tung2744:
Updated, thanks!

louischan-oursky:
I added a few improvements.

louischan-oursky (approved):

PR #5056, opened 2025-03-11
Add sender field in smtp configs

tung2744:
ref DEV-2310

Screenshot 2025-03-11 at 8 58 30 PM


pkg/portal/smtp/service.go
```diff
@@ -74,7 +75,9 @@ func (s *Service) SendTestEmail(ctx context.Context, app *model.App, options Sen
 
 	err = dialer.DialAndSend(message)
 	if err != nil {
-		return
+		return errors.Join(SMTPTestFailed.NewWithInfo("smtp test failed", map[string]interface{}{
+			"Message": err.Error(),
+		}), err)
```
louischan-oursky, line 76: What about just use `err.Error()` as message?

pkg/lib/translation/service.go
```diff
@@ -70,30 +82,58 @@ func (s *Service) GetSenderForTestSMS(ctx context.Context) (sender string, err e
...
 	}
 	if err != nil {
 		return
 	}
 
-	subject, err = t.RenderText(fmt.Sprintf("email.%s.subject", name), variables)
+	subject, err = baseTranslations.RenderText(fmt.Sprintf("email.%s.subject", name), variables)
 	if errors.Is(err, template.ErrNotFound) {
-		subject, err = t.RenderText("email.default.subject", variables)
+		subject, err = baseTranslations.RenderText("email.default.subject", variables)
```
louischan-oursky, line 136: Please add new tests for this changed behavior.

pkg/lib/translation/service.go
```diff
@@ -70,30 +82,58 @@ func (s *Service) GetSenderForTestSMS(ctx context.Context) (sender string, err e
...
+				return
+			}
+		}
 	}
 
-	replyTo, err = t.RenderText(fmt.Sprintf("email.%s.reply-to", name), variables)
+	replyTo, err = baseTranslations.RenderText(fmt.Sprintf("email.%s.reply-to", name), variables)
 	if errors.Is(err, template.ErrNotFound) {
-		replyTo, err = t.RenderText("email.default.reply-to", variables)
+		replyTo, err = baseTranslations.RenderText("email.default.reply-to", variables)
```
louischan-oursky, line 128: We overlooked this. `Reply-To: ` should have the same treatment with `From: `

I also found that Reply-To in default translation is empty, but in infra/mail/sender.go, we did not check Reply-To is empty or not and always include it. I think it does not seem correct.

In summary, From and Reply-To should always be used with the corresponding level of SMTP secret. The normal resolution logic applies to Subject.

louischan-oursky (changes requested):

pkg/portal/smtp/service.go
tung2744: Currently the joined error message will be returned as `message` so I didn't use it.
Added a commit to remove internal errors from the graphql error, so that message can be directly used.

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #5069, opened 2025-03-14
Setup custom radix colors

tung2744:
ref DEV-2586


louischan-oursky:
Please rebase

portal/src/radix-color.css
```diff
@@ -1,6 +1,8 @@
 /* This file defines our own custom color palette */
 /* https://www.figma.com/design/wAIzBnO9un4fYcq5ampR5k/Radix-Themes-(Storybook)?node-id=3795-92063&t=hD90rcrEF95c6rUb-0 */
 
+/* stylelint-disable no-duplicate-selectors -- We always have dupliacted selectors in this file */
```
louischan-oursky, line 4: I think you can just put all declaration in `:root` to avoid this exception.

tung2744: But I want to put the `:root` selector together with `[data-accent-color="xxx"]` selector in pair so we can easily know they need to exist together.

tung2744: Or maybe I can try to use [scss mixin](https://sass-lang.com/documentation/at-rules/mixin/#arguments) to generate the classes.

louischan-oursky: No, we do not use CSS preprocessor in this project.

tung2744: oh right

louischan-oursky: Just use the file-specific exception

louischan-oursky (approved):

PR #5071, opened 2025-03-17
Build ONCE docker image - Part 4

louischan-oursky:
ref DEV-2412

SCR-20250314-qlml

(Note that the accounts project is hidden and the create project button is hidden)

It can be tested with `./once/docker-compose.yaml`

```
services:
  once:
    build:
      context: ../
      dockerfile: ./once/Dockerfile
    volumes:
    - postgres_data:/var/lib/postgresql/data
    - redis_data:/var/lib/redis/data
    - certbot_data:/etc/letsencrypt
    - minio_data:/var/lib/minio/data
    environment:
      POSTGRES_PASSWORD: "1234"
      REDIS_PASSWORD: "1234"
      MINIO_ROOT_PASSWORD: "12345678"
      # Edit your /etc/hosts to include those hosts.
      AUTHGEAR_HTTP_ORIGIN_PORTAL: "http://portal.localhost"
      AUTHGEAR_HTTP_ORIGIN_ACCOUNTS: "http://accounts.localhost"
      AUTHGEAR_HTTP_ORIGIN_PROJECT: "http://project.localhost"
      AUTHGEAR_CERTBOT_RUN_INTERVAL: "86400"
      AUTHGEAR_ONCE_ADMIN_USER_EMAIL: "louischan@oursky.com"
      AUTHGEAR_ONCE_ADMIN_USER_PASSWORD: "12345678"
      #AUTHGEAR_CERTBOT_ENVIRONMENT: "production"
    ports:
    - "80:80"
    - "443:443"
    - "5432:5432"
    - "9001:9001"
    - "8090:8090"

volumes:
  postgres_data:
    driver: local
  redis_data:
    driver: local
  certbot_data:
    driver: local
  minio_data:
    driver: local
```

```
cd once
# To remove any previous data
docker compose down -v
docker compose build
docker compose up
# Visit http://portal.localhost
```

### DIFF
--- a/.github/actions/build-custom-image/action.yaml
+++ b/.github/actions/build-custom-image/action.yaml
@@ -1,6 +1,6 @@
 name: Build custom image
 inputs:
-  target:
+  dockerfile:
     required: true
   image_name:
     required: true
@@ -26,7 +26,7 @@ runs:
       gcloud auth configure-docker "$REPO"
   - name: Build and Push
     env:
-      TARGET: ${{ inputs.target }}
+      DOCKERFILE: ${{ inputs.dockerfile }}
       IMAGE_NAME: ${{ inputs.image_name }}
       REPO: ${{ inputs.docker_repo }}
       BUILD_ARCH: amd64
--- a/.github/actions/build-custom-image/action.yaml
+++ b/.github/actions/build-custom-image/action.yaml
@@ -29,20 +29,17 @@ runs:
       TARGET: ${{ inputs.target }}
       IMAGE_NAME: ${{ inputs.image_name }}
       REPO: ${{ inputs.docker_repo }}
+      BUILD_ARCH: amd64
+      EXTRA_BUILD_OPTS: '--ssh=default'
+      OUTPUT: ${{ format('type=image,name={0},push-by-digest=true,name-canonical=true,push=true', inputs.image_name) }}
     shell: bash
     run: |
-      METADATA_FILE="$(mktemp).json"
+      export METADATA_FILE="$(mktemp).json"
       echo "METADATA_FILE=$METADATA_FILE"
-      make -C custombuild build-image \
-        TARGET=$TARGET \
-        BUILD_ARCH=amd64 \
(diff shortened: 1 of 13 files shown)


.github/actions/build-custom-image/action.yaml
```diff
@@ -26,23 +26,19 @@ runs:
...
+      DOCKERFILE: ${{ inputs.dockerfile }}
       IMAGE_NAME: ${{ inputs.image_name }}
-      REPO: ${{ inputs.docker_repo }}
+      BUILD_ARCH: amd64
+      EXTRA_BUILD_OPTS: '--ssh=default'
+      OUTPUT: ${{ format('type=image,name={0},push-by-digest=true,name-canonical=true,push=true', inputs.image_name) }}
     shell: bash
     run: |
-      METADATA_FILE="$(mktemp).json"
+      export METADATA_FILE="$(mktemp).json"
```
tung2744, line 36: oh why it works originally without export 🤔

tung2744 (approved):

.github/actions/build-custom-image/action.yaml
louischan-oursky: Before the change, write `make something METADATA_FILE="$METADATA_FILE"`, the variable (not exported, but a local variable) is passed explicitly via CLI.

After the change, we export the variable to make it global, thus child process can "see" it.

tung2744: But it is not on the same line with the `make xx` command

tung2744: I think lines in the same `run` step are executed in the same shell, and therefore:
```
A="a"

echo $A
```
Will print "a" even without export.

louischan-oursky:
```
# This declare a unexported shell variable.
A=a

# The unexported shell variable is evaluated and passed as argument to the command echo.
echo $A
```

```
A=a

# The make command DOES NOT see A because it is unexported
make something
```

```
export A=a

# The make command DOES see A because it is exported.
make something
```

PR #5072, opened 2025-03-17
Specify support of using go template to render saml attribute

tung2744:
ref DEV-1722


docs/specs/saml.md
```diff
@@ -439,7 +440,12 @@ The `attributes` object under items of `service_providers` is used to customize
   - `from_user_profile_attribute`: Required. A JSON pointer to a field of the user profile. A value will be read using the JSON pointer from the user profile. And write to the SAML attribute with `Name` specified by `to_saml_attribute`.
   - `to_saml_attribute`: Required. The `Name` of the SAML attribute to write the value.
 
-  2. A hook. It runs a hook to computes the resulting SAML attributes. See the below section [Use a hook to compute SAML attributes](#use-a-hook-to-compute-saml-attributes) for more details about the hook. The object contains the following fields:
+  2. A object uses go template to render a string as the value of a attribute. The object contains the following fields:
```
louischan-oursky, line 443: I realize we haven't implemented this feature, so I am considering rename/restructure the config.

```yaml
mappings:
- to: username
  type: user_profile
  user_profile:
    pointer: /preferred_username
- to: email
  type: go_template
  go_template:
    template: "{{ ... }}"
- to: phone
  type: hook
  hook:
    url: ...
```

It seems that user_profile is just a special case of go_template, since it can be implemented with go template, for example

```
mappings:
- to: email
  type: go_template
  go_template:
    template: "{{ .email }}"
```

So I suggest we remove user_profile from the spec.

---

We also want to specify what happen if the referenced attribute is absent in the user info.

For example, https://go.dev/play/p/Pmztw-Ib0zl It will output `<no value>` which is not something we want. This behavior can be changed by https://pkg.go.dev/text/template#Template.Option But we have this issue https://github.com/golang/go/issues/24963#issuecomment-400086111 To work around this, the template has to be `{{ if .email }}{{ .email }}{{ end }}`. But that's too verbose for such a simple case, and it is definitely not intuitive.

tung2744: > so I am considering rename/restructure the config.

I am ok with the renaming, but I believe we had a discussion before and there is a reason why we don't use the nested pointer structure.
```
user_profile:
    pointer: /preferred_username
```

> It seems that user_profile is just a special case of go_template, since it can be implemented with go template, for example

Not exactly:
1. Some user attributes could be a boolean (e.g. email_verified), or an array of string (e.g. x_roles). In such cases, configuring using a go_template might be what you want.
2. A simple mapping is always much simple and easy to understand, the go_template should be considered an advanced feature which little people will use.

> We also want to specify what happen if the referenced attribute is absent in the user info.

I think we can simply follow the default behavior of go template as we already mention it is a go template in our spec?

tung2744: - If missing, show ""

tung2744:
@louischan-oursky @chpapa @fungc-io Updated according to latest discussion, thanks!

docs/specs/saml.md
```diff
@@ -436,32 +443,109 @@ The `attributes` object under items of `service_providers` is used to customize
...
+Given the following `attributes` config:
+
+```yaml
+attributes:
+  definitions:
+    - name: mail
+      name_format: urn:oasis:names:tc:SAML:2.0:attrname-format:basic
+  mappings:
+    - from:
+        template: "{{.preferred_username}}@example.com"
```
tung2744, line 493: I found that there will be problems if we use html template to render the attribute.
Try this playground:
https://go.dev/play/p/3hd6eSK3d9U

You can see any email with `+` will be escaped in the attribute, which may cause problem.

So I think we need to render the template with `text/template` instead of `html/template`, with one exception that `<no value>` will be rendered as empty string.

And I suggest we modify the config to explicitly mention the type of the template:

```yaml
mappings:
  - from:
      template:
        text: "{{.email}}"
    to:
      saml_attribute: mail
```

And for any html template we support in the future, it should looks like
```yaml
template:
  html: "<p>{{.email}}</p>"
```

louischan-oursky:
```
from:
  text_template:
    template: "{{ .email }}"
  user_profile:
    pointer: /email
  html_template:
    template: "{{ .email }}"
```

```
formatGoTimeDate "2006-01-02" (parseGoTimeDate "02/01/2006" .custom_attributes.joinDate)

```

tung2744: We will modify the config to
```yaml
from:
  text_template:
    template: "{{ .email }}"
```

So that if we need a config of html template in future, we can make it as
```yaml
html_template:
  template: "{{ .email }}"
```

tung2744:
Should we merge this?

louischan-oursky (approved):

louischan-oursky:
Yes, let me merge it.

PR #5077, opened 2025-03-20
Fix initial country is not selected in onboarding survey

tung2744:
ref dev-2568

### DIFF
--- a/portal/src/OnboardingSurveyScreen.tsx
+++ b/portal/src/OnboardingSurveyScreen.tsx
@@ -27,6 +27,7 @@ import { useProvideError } from "./hook/error";
 import { useViewerQuery } from "./graphql/portal/query/viewerQuery";
 import SurveyLayout from "./OnboardingSurveyLayout";
 import styles from "./OnboardingSurveyScreen.module.css";
+import ShowLoading from "./ShowLoading";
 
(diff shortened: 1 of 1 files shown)


portal/src/OnboardingSurveyScreen.tsx
```diff
@@ -600,22 +600,24 @@ function Step3Team(_props: StepProps) {
...
+          ) : null}
         </FormProvider>
       </div>
     </SurveyLayout>
   );
 }
 
 function Step3Personal(_props: StepProps) {
-  const { viewer } = useViewerQuery();
+  const { viewer, loading: viewerLoading } = useViewerQuery();
```
louischan-oursky, line 620: We should wait until this query is fully loaded before showing the whole screen, like we did in other screens.

louischan-oursky (changes requested):

portal/src/OnboardingSurveyScreen.tsx
tung2744: I made it like this because I didn't find a loading state for this screen. Maybe I just add a loading spinner.

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #5082, opened 2025-03-21
Implement SAML attribute mapping

tung2744:
ref DEV-1735

### DIFF
--- a/pkg/lib/config/saml.go
+++ b/pkg/lib/config/saml.go
@@ -40,7 +40,8 @@ var _ = Schema.Add("SAMLServiceProviderConfig", `
 		"slo_enabled": { "type": "boolean" },
 		"slo_callback_url": { "type": "string", "format": "uri" },
 		"slo_binding": { "$ref": "#/$defs/SAMLSLOBinding" },
-		"signature_verification_enabled": { "type": "boolean" }
+		"signature_verification_enabled": { "type": "boolean" },
+		"attributes": { "$ref": "#/$defs/SAMLAttributesConfig" }
 	},
 	"required": ["client_id", "acs_urls"],
 	"allOf": [
@@ -131,6 +132,7 @@ type SAMLServiceProviderConfig struct {
 	SLOCallbackURL               string                        `json:"slo_callback_url,omitempty"`
 	SLOBinding                   samlprotocol.SAMLBinding      `json:"slo_binding,omitempty"`
 	SignatureVerificationEnabled bool                          `json:"signature_verification_enabled,omitempty"`
+	Attributes                   *SAMLAttributesConfig         `json:"attributes,omitempty"`
 }
 
 func (c *SAMLServiceProviderConfig) SetDefaults() {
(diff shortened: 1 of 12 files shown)


pkg/lib/config/saml.go
```diff
@@ -204,3 +206,64 @@ type SAMLSigningDigestMethod string
...
+}
+`)
+
+type SAMLAttributeDefinition struct {
+	Name         string                  `json:"name,omitempty"`
+	NameFormat   SAMLAttributeNameFormat `json:"name_format,omitempty"`
+	FriendlyName string                  `json:"friendly_name,omitempty"`
+}
+
+func (c *SAMLAttributeDefinition) SetDefaults() {
```
louischan-oursky, line 251: I forget whether a struct in a slice would have its SetDefaults() called. Let's add a test to verify this.

Or you can rename NameFormat to NameFormat_WriteOnly and add a GetNameFormat getter.

pkg/lib/saml/service.go
```diff
@@ -137,6 +147,7 @@ func (s *Service) IdpMetadata(serviceProviderId string) (*samlprotocol.Metadata,
 					SingleLogoutServices: sloServices,
 				},
 				SingleSignOnServices: ssoServices,
+				Attributes:           attributes,
```
louischan-oursky, line 151: I did not read the spec. But is there any other XML attribute or XML element to tell whether a attribute is always present?

pkg/lib/saml/service.go
```diff
@@ -892,6 +887,63 @@ func (s Service) recordSessionParticipant(
...
+				samlprotocol.AttributeValue{
+					Type:  samlprotocol.SAMLAttrTypeString,
+					Value: val,
+				},
+			)
+		}
+
+		attrs = append(attrs, samlprotocol.Attribute{
+			Name:       attrDef.Name,
+			NameFormat: string(attrDef.NameFormat),
```
louischan-oursky, line 939: Also include FriendlyName?

pkg/lib/saml/service.go
```diff
@@ -912,3 +964,55 @@ var _ dsig.X509KeyStore = &x509KeyStore{}
...
+	case string:
+		return []string{raw}, nil
+	case float64:
+		return []string{fmt.Sprintf("%v", raw)}, nil
+	case bool:
+		return []string{fmt.Sprintf("%v", raw)}, nil
+	case []any:
+		return formatSlice(raw)
+	case map[string]any:
+		return formatMap(raw)
```
louischan-oursky, line 1014: I think we only need to support primitive types for now.

If the resolved attribute is not a primitive Go data type, then we should fail with an error.

There is a caveat. A JSON number in Go is always deserialized as float64, but an attribute could be an integer. If we just use the deserialized type float64, the output SAML attribute is inaccurate.

pkg/lib/saml/service.go
```diff
@@ -892,6 +887,63 @@ func (s Service) recordSessionParticipant(
...
+	for _, attrDef := range sp.Attributes.Definitions {
+		values, ok := valuesMap[attrDef.Name]
+		if !ok {
+			continue
+		}
+		attrValues := []samlprotocol.AttributeValue{}
+		for _, val := range values {
+			attrValues = append(attrValues,
+				samlprotocol.AttributeValue{
+					Type:  samlprotocol.SAMLAttrTypeString,
```
louischan-oursky, line 931: I think we should support primitive types and output the correct type. For example, if the resolved user profile attribute is a string, then the XML type is string.

We need to reference https://www.w3.org/TR/xmlschema-2/#built-in-primitive-datatypes and specify in the spec that what Go data type map to which XML data type.

louischan-oursky (changes requested):

pkg/lib/config/saml.go
tung2744: I've tested it will be called. Let me add a test for it.

pkg/lib/saml/service.go
tung2744: For attribute,  do you mean SAML attribute, or attribute in an XML element?
If SAML attribute, I think the metadata is telling what attribute this idp supports. However, it is quite common that the metadata does not include such information or it does not match with the actual output in other providers.

pkg/lib/saml/service.go
tung2744: For primitive types, do you mean string, float64, bool?

I think slice should also be supported, attribute values in xml can always be multiple.

So maybe I will remove the support of map.

tung2744: > There is a caveat. A JSON number in Go is always deserialized as float64, but an attribute could be an integer. If we just use the deserialized type float64, the output SAML attribute is inaccurate.

I think we will simply return [decimal](https://www.w3.org/TR/xmlschema-2/#decimal) for any float64.

pkg/lib/saml/service.go
tung2744: I've modified the spec in https://github.com/authgear/authgear-server/pull/5072/commits/c334fe5549a91f82f2c9cfea486fce8479554109

Documented the mappings between JSON data type and SAML attribute type when using simple mapping.

louischan-oursky:
In https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf

line 1219, it says Attribute can contain no AttributeValue
line 1250, it says if the attribute is null, Attribute must contain `<AttributeValue xsi:nil=true/>`

The implemented behavior follows line 1219, but does line 1250 apply?

tung2744:
@louischan-oursky I think it depends on how we interpret `null` or missing keys in the user info json. In the current implementation, both are treated as "missing", which means no `AttributeValue`. So the case in line 1250 never occurs.
We can interpret `null` in JSON as `nil` in SAML attribute, if we prefer to do so.

louischan-oursky (approved):

PR #5085, opened 2025-03-25
Support Whatsapp Cloud API

louischan-oursky:
ref DEV-241


pkg/lib/config/secret_whatsapp.go
```diff
@@ -106,3 +106,208 @@ func (c *WhatsappOnPremisesCredentials) SensitiveStrings() []string {
...
+
+func (c *WhatsappCloudAPICredentials) SensitiveStrings() []string {
+	return []string{
+		c.PhoneNumberID,
+		c.AccessToken,
+	}
+}
+
+// https://developers.facebook.com/docs/whatsapp/business-management-api/message-templates/supported-languages
+var _ = SecretConfigSchema.Add("WhatsappCloudAPITemplateLanguage", `
```
tung2744, line 152: Isn't this list same as the on-prem template languages?

pkg/lib/infra/whatsapp/errors.go
```diff
@@ -42,8 +53,8 @@ func (e *WhatsappAPIError) SkipLogging() bool {
 	case 401:
 		return true
 	default:
-		if e.ParsedResponse != nil {
-			if firstErrorCode, ok := e.ParsedResponse.FirstErrorCode(); ok {
+		if e.OnPremisesResponse != nil {
```
tung2744, line 56: Also handle `CloudAPIResponse` here? (Although it could be unreachable because cloud api does not return such error)

pkg/lib/config/secret_whatsapp.go
louischan-oursky: Yes, you are correct. They are the same. So we are going to update the list of supported languages in On-Premises API.

PR #5101, opened 2025-03-31
Implement otel sql semconv v1.30.0

louischan-oursky:
ref DEV-2585


tung2744:
Do we need to document how and when to generate these code?

louischan-oursky:
No, the code is generated by `//go:generate` gogenerate.go. So `make generate` will generate the code in every invocation already.

tung2744 (approved):

PR #5110, opened 2025-04-02
Handle more twilio errors

tung2744:
ref DEV-2599

### DIFF
--- a/pkg/auth/handler/webapp/authflow_controller.go
+++ b/pkg/auth/handler/webapp/authflow_controller.go
@@ -1008,11 +1008,7 @@ func (c *AuthflowController) makeHTTPHandler(s *webapp.Session, screen *webapp.A
 		}
 
 		if err != nil {
-			if apierrors.IsAPIError(err) {
-				c.renderError(r.Context(), w, r, err)
-			} else {
-				panic(err)
-			}
+			c.renderError(r.Context(), w, r, err)
 		}
 	})
 }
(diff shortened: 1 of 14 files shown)


tung2744:
Added [9195e23](https://github.com/authgear/authgear-server/pull/5110/commits/9195e23447554068caf9b24e49aec3bd8af51b3b) because I found that transient error of sms gateway (Such as timeout of the sms webhook) causing the authflow to show unrecoverable error page.

pkg/auth/handler/webapp/authflow_controller.go
```diff
@@ -1008,11 +1008,7 @@ func (c *AuthflowController) makeHTTPHandler(s *webapp.Session, screen *webapp.A
 		}
 
 		if err != nil {
-			if apierrors.IsAPIError(err) {
-				c.renderError(r.Context(), w, r, err)
-			} else {
-				panic(err)
-			}
+			c.renderError(r.Context(), w, r, err)
```
louischan-oursky, line 1011: The panic middleware will log the error for us. Now we do not panic, I think the error is not logged.

tung2744: You are right. Then I think the log should occur in error renderer.

tung2744:
Added two more commits so we handle the error better.

louischan-oursky:
Test failed.

tung2744:
Added [38406e0](https://github.com/authgear/authgear-server/pull/5110/commits/38406e028cb7895b46c672581aca3e27989f6fcc) to display provider code even the error is unknown.

louischan-oursky (approved):

PR #5125, opened 2025-04-14
Update crewjam/saml to 0.5.0

louischan-oursky:


pkg/lib/saml/samlprotocol/schema.go
```diff
@@ -502,7 +517,7 @@ func (r *Response) Element() *etree.Element {
 	// cannonicalizer, but prefix lists do not appear to be implemented correctly
 	// in some libraries, so the safest action is to always produce XML that is
 	// (a) in canonical form and (b) does not require prefix lists.
-	el.CreateAttr(fmt.Sprintf("xmlns:%s", xmlSchemaNamespace), "http://www.w3.org/2001/XMLSchema")
+	el.CreateAttr("xmlns:xs", "http://www.w3.org/2001/XMLSchema")
```
tung2744, line 520: I think this change is not needed, using `xmlSchemaNamespace` here ensures we use the same namespace name in our code.

pkg/lib/saml/samlprotocol/update_helper.go
```diff
@@ -0,0 +1,15 @@
...
+// So that you can use `git diff` to view the changes.
+// This file is guarded by a build tag so it does not run normally.
+//
+// It is intended you run it with
+//  go generate -tags updatecrewjamsaml ./pkg/lib/saml/samlprotocol
+
+//go:generate curl -sSL https://raw.githubusercontent.com/crewjam/saml/refs/tags/v0.5.0/schema.go -o schema.go
+//go:generate curl -sSL https://raw.githubusercontent.com/crewjam/saml/refs/tags/v0.5.0/duration.go -o duration.go
+//go:generate curl -sSL https://raw.githubusercontent.com/crewjam/saml/refs/tags/v0.5.0/time.go -o time.go
+//go:generate curl -sSL https://raw.githubusercontent.com/crewjam/saml/refs/tags/v0.5.0/metadata.go -o metadata.go
```
tung2744, line 15: I think it is not intended to sync the files.
The reason we version the files in our repo is we will change them, so it might not be same as the original file.

louischan-oursky: The intention is not to sync them, but to review the changes. So this file does not run by default.

pkg/lib/saml/samlprotocol/schema.go
louischan-oursky: Fixed up this line.

PR #5132, opened 2025-04-15
Specify behavior of disabled user login with biometric

tung2744:


tung2744:
1. We will first handle disabled user in biometric login according to this spec.
2. We will then revisit the current user statuses to separate "Scheduled deletion/anonymization" from disabled.

docs/specs/biometric.md
```diff
@@ -73,3 +73,35 @@ identity:
   biometric:
     list_enabled: false
 ```
+
+## Errors
```
louischan-oursky, line 77: Given that the OAuth error shape is quite limited, I am thinking if we want to introduce a way to include our API error into a OAuth response.

https://datatracker.ietf.org/doc/html/rfc6749#section-8.5 specifies how to add additional error codes.

We can add a new error code `authgear_error`. When `error=authgear_error`, the response will have an extra field `authgear_error`, where the value is a Authgear API error JSON object.

Then the new SDKs be taught to decode `error=authgear_error`, throw an instance `ServerError` to handle this.

tung2744: As oauth error could appear in the url, I am not sure if it is appropriate to add field to store the api error json object which could be long.

PR #5134, opened 2025-04-15
Reduce error events sent to sentry

tung2744:
ref DEV-2631


louischan-oursky (changes requested):
1. There is a compilation error
2. Generate translations

tung2744:
Fixed, thanks!

louischan-oursky (approved):

PR #5136, opened 2025-04-16
Always log delivery error immediately

louischan-oursky:
ref DEV-2641

### DIFF
--- a/pkg/lib/messaging/sender.go
+++ b/pkg/lib/messaging/sender.go
@@ -92,11 +92,17 @@ func (s *Sender) SendEmailInNewGoroutine(ctx context.Context, msgType translatio
 	sendInTx := func(ctx context.Context) error {
 		err := s.MailSender.Send(*opts)
 		if err != nil {
+			// Log the send error immediately.
+			s.Logger.WithError(err).WithFields(logrus.Fields{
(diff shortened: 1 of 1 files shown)


tung2744:
If the send is not async, isn't the error should propagate to the caller and handled by it? The problem should be why the caller iss having error but the error isn't logged?

louischan-oursky:
> If the send is not async, isn't the error should propagate to the caller and handled by it?

It is easier to troubleshoot if we always log the delivery log. If we delegate to the caller to handle it, it may or may not be logged.

> The problem should be why the caller iss having error but the error isn't logged?

errors like context.Canceled are not logged to external service. Even if it is logged, it is hard to associate that particular context.Canceled originates from here.

> And the slack message in the original issue seems related to metrics but not logging?

The original issue is that we receive some alerts. The alert is triggered by a counter with `status=error`. The condition of incrementing the counter is `err != nil`. I suspect `err` can be something like context.Conceled, thus the audit log `sms.error` was not written.

louischan-oursky:
And we always log Whatsapp delivery error even before this change. So this change aligns the behavior of SMS and Whatsapp

tung2744 (approved):

PR #5140, opened 2025-04-16
Return relevant errors in Finish authflow in unexpected conditions

tung2744:
ref DEV-2648


pkg/auth/handler/webapp/authflow_controller.go
```diff
@@ -672,6 +672,13 @@ func (c *AuthflowController) AdvanceWithInputs(
 }
 
 func (c *AuthflowController) Finish(ctx context.Context, r *http.Request, s *webapp.Session) (*webapp.Result, error) {
+	if s == nil {
+		panic(fmt.Errorf("unexpected: session is missing in Finish"))
+	}
+	if s.Authflow == nil {
+		return nil, fmt.Errorf("cannot finish authflow: session is not in a flow")
+	}
+
```
louischan-oursky, line 681: I think the reason is that the caller is using HandleWithoutFlow, so s can be nil in some case.

louischan-oursky: I studied the history. The finish handler uses HandleWithoutFlow right from the beginning https://github.com/authgear/authgear-server/blame/b2999993ed035a75acd2f136e6c92b018e4be877/pkg/auth/handler/webapp/authflow_finish_flow.go
https://github.com/authgear/authgear-server/blame/b2999993ed035a75acd2f136e6c92b018e4be877/pkg/auth/handler/webapp/authflowv2/finish_flow.go
But it does not explain why the handler is doing so.

tung2744: But it still cannot explain why it is nil, because usually people enter this route right after they completed the last step of a flow.

tung2744: I believe it is because the authflow is actually completed at the point you navigate to this screen, i.e. you cannot get it by state token. Therefore the handling is different with handling a step.

louischan-oursky: I remember we have changed the behavior of NOT deleting the flow when we finish the flow, so when the user is still logging in, the session should still be there.

Would that be the end-user navigate back? For example the end-user sign in, use the webapp for more than 20 minutes, and then navigate back to a point where /finish is visited again.

louischan-oursky: In that case we should show the completed screen (i.e. the error screen) that we already have.

tung2744: It could be, but I think we first need to distinguish between missing session (s is nil), missing Authflow (s.Authflow is nil), and code error (s.Authflow.AllScreen) is nil. Now we don't know it.

tung2744: I think I can also handle the session not found error in this pr first.

louischan-oursky: s.Authflow.AllScreen can be nil, reading a nil map will result in !ok. In here, we will return authflow.ErrFlowNotFound

tung2744: Added a commit to handle session not found

louischan-oursky: Per our offline discussion

1. Rename HandleWithoutFlow to HandleWithoutScreen. HandleWithoutScreen requires session. If session is not found, the error handling is the same as HandleStep.
2. Add HandleWithoutSession. This is for CSRF instruction handler only.

louischan-oursky (approved):

PR #5142, opened 2025-04-17
Bump golang.org/x/net to 0.38.0

louischan-oursky:


scripts/sh/ensure-important-modules-up-to-date.sh
```diff
@@ -0,0 +1,38 @@
...
+  "golang.org/x/net",
+  "golang.org/x/oauth2",
+  "golang.org/x/sync",
+  "golang.org/x/sys",
+  "golang.org/x/telemetry",
+  "golang.org/x/term",
+  "golang.org/x/text",
+  "golang.org/x/time",
+  "golang.org/x/tools",
+  "golang.org/x/vuln"
```
tung2744, line 19: If I understand it correctly, `ensure-important-modules-up-to-date` is for forcing upgrade to latest version of packages. And packages which needs to be always latest is only `github.com/nyaruka/phonenumbers` because we need the latest db. What is the reason we want to include other packages here?

louischan-oursky: Updated.

tung2744 (approved):

PR #5144, opened 2025-04-22
Disable Whatsapp when it is Authgear ONCE

louischan-oursky:
ref DEV-2427

SCR-20250422-mtmq

### DIFF
--- a/portal/src/graphql/portal/LoginMethodConfigurationScreen.tsx
+++ b/portal/src/graphql/portal/LoginMethodConfigurationScreen.tsx
@@ -40,15 +40,16 @@ import {
   VerificationCriteria,
   VerificationClaimsConfig,
   PasswordPolicyFeatureConfig,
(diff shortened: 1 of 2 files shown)


tung2744:
Do we need to change the default in `AuthenticatorOOBSMSConfig` too?

louischan-oursky:
> Do we need to change the default in AuthenticatorOOBSMSConfig too?

If you mean changing `DEFAULT_PHONE_OTP_MODE`, I think we can change that as well. But not changing it also works, as explained in the commit message.

tung2744:
I mean changing this line:
https://github.com/authgear/authgear-server/blob/main/pkg/lib/config/authenticator.go#L344

But if the generated config must have the field set to sms then it should be ok

louischan-oursky:
Let's just leave that unchanged.

tung2744 (approved):

PR #5147, opened 2025-04-22
Setup project wizard layout

tung2744:
ref DEV-2642


tung2744:
Included in https://github.com/authgear/authgear-server/pull/5150

PR #5149, opened 2025-04-22
Remind configure SMS provider in ONCE

louischan-oursky:
ref DEV-2613

SCR-20250422-qujj
SCR-20250422-quku
SCR-20250422-qumr

SCR-20250423-kvki
SCR-20250423-kwrc


tung2744:
Looks like the red message bar is having too much spaces at the end if the content is not long enough.

louischan-oursky:
Reduced the width in this page.

SCR-20250424-oclm

I did not reduced the width in other pages because the intention of the design is to align the width of sibling elements

SCR-20250424-ocyl

tung2744 (approved):

PR #5157, opened 2025-04-28
Implement WhiteButton

tung2744:
ref DEV-2659

Screenshot 2025-04-28 at 3 14 55 PM

### DIFF
--- a/portal/src/components/v2/WhiteButton/WhiteButton.tsx
+++ b/portal/src/components/v2/WhiteButton/WhiteButton.tsx
@@ -1,0 +1,41 @@
+import React from "react";
+import cn from "classnames";
+import { Button } from "@radix-ui/themes";
+import styles from "./WhiteButton.module.css";
+
+export interface WhiteButtonProps {
+  size: "1" | "2" | "3" | "4";
+  disabled?: boolean;
+  loading?: boolean;
+  text?: React.ReactNode;
+
+  type?: "button" | "reset" | "submit";
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+}
+
+export function WhiteButton({
+  size,
+  disabled,
+  loading,
+  text,
+
+  type = "button",
+  onClick,
+}: WhiteButtonProps): React.ReactElement {
+  return (
+    <Button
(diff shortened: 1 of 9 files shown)


portal/src/components/v2/WhiteButton/WhiteButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+  size: "1" | "2" | "3" | "4";
+  disabled?: boolean;
+  loading?: boolean;
+  text?: React.ReactNode;
+
+  type?: "button" | "reset" | "submit";
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+}
+
+export function WhiteButton({
```
louischan-oursky, line 16: This WhiteButton has hover state, but when it is pressed (i.e. active state), it is visually equal to normal state. I cross checked PrimaryButton, PrimaryButton does not have this issue.

tung2744: The normal color is accent/9, and active color is accent/10. So it is because the generated colors are similar if accent color is #ffffff. Maybe we can let designer to try it in the storybook and see if any feedback.

louischan-oursky: Actually in the Figma file, there is a pressed state in PrimaryButton. But WhiteButton does not have it. I think that's a problem, right?

tung2744: Yes. I think I will remind designer to take a look. But I believe finally the active color will follow radix button active color, which is accent/10.

tung2744: I've checked the actual implementation and the radix design file, and I think they are not 100% match.
The actual implementation have a `background-color: var(--accent-10)`, with a `filter: brightness(1.08)`, while the figma use a opacity to do it.
I think as long as we are clear about the button is implemented with radix button, and all behaviors should follow the button of radix button, it will be enough.

louischan-oursky (approved):

PR #5167, opened 2025-05-02
Project Wizard: Support project wizard in ONCE

tung2744:
ref DEV-2654

Screenshot 2025-05-02 at 6 51 34 PM


tung2744:
@louischan-oursky We discussed about hiding survey in ONCE, but I found that we always have one project in ONCE, therefore the onboarding survey should never be triggered. Did you found the survey will be triggered in some case?

louischan-oursky (approved):

PR #5169, opened 2025-05-06
Add config to switch between v1 and v2 project wizard

tung2744:
ref DEV-2684


louischan-oursky (changes requested):
According to the build error log, we also need to bring back the removed v1 GTM code.

tung2744:
Fixed, thanks!

louischan-oursky (approved):

PR #5172, opened 2025-05-07
Fix portal session middleware cannot access Authgear in ONCE

louischan-oursky:
ref DEV-2689


tung2744:
Do you know the reason why the ip cannot be reached within the container? Because the vm should be able to reach the same ip, I am thinking if it is something we can solve with network config of docker.

louischan-oursky:
I believe this is related to the combination of these things

- The host iptables
- https://docs.docker.com/engine/network/packet-filtering-firewalls/

I am not familiar with them so let's work around the problem by not routing through the public IP address of the host machine.

tung2744 (approved):

PR #5176, opened 2025-05-09
Support ID token in login flow and signup_login flow

louischan-oursky:
ref DEV-2695


e2e/tests/login/id_token/test.yaml
```diff
@@ -0,0 +1,100 @@
+name: Login with ID token
+focus: true
```
pkong-ds, line 2: forgot to remove L2

pkong-ds (changes requested):

e2e/tests/login/id_token/test.yaml
pkong-ds: oh, ci caught that already 👍

louischan-oursky: Thanks for reminding.

louischan-oursky:
I did not check if the Auth UI has any problem handling the new `id_token`. Let me have a check first.

louischan-oursky:
Checked that `/login` and `/signup` ignore `id_token`. No changes are needed.

tung2744 (approved):

PR #5177, opened 2025-05-09
Add preview widget route

tung2744:
ref DEV-2690

### DIFF
--- a/pkg/auth/routes.go
+++ b/pkg/auth/routes.go
@@ -43,7 +43,7 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource) ht
 
 	router.Health(p.RootHandler(newHealthzHandler))
 
-	rootChain := httproute.Chain(
+	noProjectChain := httproute.Chain(
 		p.RootMiddleware(newOtelMiddleware),
 		p.RootMiddleware(newPanicMiddleware),
 		p.RootMiddleware(newBodyLimitMiddleware),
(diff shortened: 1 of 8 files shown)


pkg/auth/routes.go
```diff
@@ -43,14 +43,18 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource) ht
...
 		p.RootMiddleware(newSentryMiddleware),
 		httproute.MiddlewareFunc(httputil.XContentTypeOptionsNosniff),
 		httproute.MiddlewareFunc(httputil.PermissionsPolicyHeader),
 		httproute.MiddlewareFunc(httputil.XRobotsTag),
+	)
+
+	rootChain := httproute.Chain(
+		noProjectChain,
 		MakeWebAppRequestMiddleware(p, configSource, newWebAppRequestMiddleware),
 		p.Middleware(newContextHolderMiddleware),
```
louischan-oursky, line 59: I think ContextHolderMiddleware is needed in noProjectChain too.

louischan-oursky (changes requested):

pkg/auth/routes.go
tung2744: Currently the preview route does not use anything inside it.

louischan-oursky (approved):

pkg/auth/routes.go
tung2744: I found that the current no project routes are missing CSP, I created another issue to configure CSP of the no project route:
https://linear.app/authgear/issue/DEV-2705/configure-csp-for-preview-route

louischan-oursky: The CSP thing should require CSP nonce which depends on ContextHolderMiddleware.

PR #5179, opened 2025-05-13
Reorganize components and fix fonts

tung2744:
ref DES-134
ref DEV-2693


portal/src/index.css
```diff
@@ -26,6 +26,9 @@ html {
   margin: 0;
   padding: 0;
   box-sizing: border-box;
+  /* For some reason, some elements will have font-family: ; set by the user agent stylesheet */
+  /* So we reset it to inherit */
+  font-family: inherit;
```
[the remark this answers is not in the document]
louischan-oursky: The tailwindcss preflight should have done this https://github.com/tailwindlabs/tailwindcss/blob/main/packages/tailwindcss/preflight.css#L244 Or it does not actually work?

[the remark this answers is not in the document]
tung2744: @louischan-oursky 

https://github.com/authgear/authgear-server/blob/main/portal/tailwind.config.js#L3

We does not enable preflight of tailwind css. I guess maybe it has some conflict with the existing fabric components?

portal/src/radix-typography.css
```diff
@@ -1,6 +1,7 @@
 /* See https://www.radix-ui.com/themes/docs/theme/typography#custom-font-weights */
 /* Reference https://www.figma.com/design/wAIzBnO9un4fYcq5ampR5k/Authgear---Radix-Themes--Storybook-?node-id=610-103 */
 .radix-themes {
+  --default-font-family: inherit;
```
louischan-oursky: Do we still need to override `--default-font-family`?

portal/src/index.html
```diff
@@ -26,6 +26,7 @@
     </script>
     <link rel="shortcut icon" href="./favicon.png" />
     <link rel="stylesheet" href="./index.css" />
+    <style data-merge-styles="true"></style>
```
louischan-oursky, line 29: Add the following comment above this line

```
<!-- See https://github.com/microsoft/fluentui/blob/%40fluentui/merge-styles_v8.6.13/packages/merge-styles/README.md#controlling-where-styles-are-injected -->
```

portal/src/radix-theme-overrides.css
```diff
@@ -0,0 +1,18 @@
+.radix-themes {
+  --default-font-family: inherit;
```
louischan-oursky, line 2: Do we still need to override `--default-font-family`?

louischan-oursky (changes requested):

portal/src/radix-theme-overrides.css
tung2744: Yes, because by default all radix component uses the default value which is not same as what we specified in index.css. Change it to inherit so radix component also uses the same font-family of the whole portal.

portal/src/index.html
tung2744: Added a commit for the comment

portal/src/radix-theme-overrides.css
louischan-oursky: I see. BTW, we forgot to update font-family to follow the latest value we observed on github.com 

`-apple-system,BlinkMacSystemFont,"Segoe UI","Noto Sans",Helvetica,Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji"`

tung2744: Added one more commit for this

louischan-oursky (approved):

PR #5181, opened 2025-05-15
Allow oauth provider configs to be inactive

tung2744:
ref DEV-2711

### DIFF
--- a/pkg/lib/config/identity.go
+++ b/pkg/lib/config/identity.go
@@ -282,7 +282,8 @@ func OAuthSSOProviderConfigSchemaBuilder(builder validation.SchemaBuilder) valid
 		Property("alias", validation.SchemaBuilder{}.Type(validation.TypeString).MinLength(1)).
 		Property("modify_disabled", validation.SchemaBuilder{}.Type(validation.TypeBoolean)).
 		Property("create_disabled", validation.SchemaBuilder{}.Type(validation.TypeBoolean)).
-		Property("delete_disabled", validation.SchemaBuilder{}.Type(validation.TypeBoolean))
+		Property("delete_disabled", validation.SchemaBuilder{}.Type(validation.TypeBoolean)).
+		Property("is_active", validation.SchemaBuilder{}.Type(validation.TypeBoolean))
 	builder.AddRequired("alias")
 	return builder
 }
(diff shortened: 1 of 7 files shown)


pkg/lib/config/identity.go
```diff
@@ -282,7 +282,8 @@ func OAuthSSOProviderConfigSchemaBuilder(builder validation.SchemaBuilder) valid
 		Property("alias", validation.SchemaBuilder{}.Type(validation.TypeString).MinLength(1)).
 		Property("modify_disabled", validation.SchemaBuilder{}.Type(validation.TypeBoolean)).
 		Property("create_disabled", validation.SchemaBuilder{}.Type(validation.TypeBoolean)).
-		Property("delete_disabled", validation.SchemaBuilder{}.Type(validation.TypeBoolean))
+		Property("delete_disabled", validation.SchemaBuilder{}.Type(validation.TypeBoolean)).
+		Property("is_active", validation.SchemaBuilder{}.Type(validation.TypeBoolean))
```
louischan-oursky, line 287: According to our naming convention, it should be `disabled` (Note the `d` at the end) instead of `is_active`

It is `disabled` because the zero value `false` means enabled.

pkg/lib/config/identity.go
```diff
@@ -277,14 +277,23 @@ var _ = Schema.Add("OAuthSSOConfig", `
...
 		Property("is_active", validation.SchemaBuilder{}.Type(validation.TypeBoolean))
 	builder.AddRequired("alias")
+
+	_if := validation.SchemaBuilder{}
+	_if.Properties().
+		Property("is_active", validation.SchemaBuilder{}.Const(false))
+
+	builder.AllOf(validation.SchemaBuilder{}.If(_if).
+		Else(providerSchemaBuilder))
+
```
louischan-oursky, line 296: Add tests for this change. Some cases to test

1. `disabled` is absent, and the original validation should be enforced.
2. `disabled` is `false`, and the original validation should be enforced.
3. `disabled` is `true`, and the original validation should be skipped.

louischan-oursky (changes requested):

pkg/lib/config/identity.go
tung2744: Test cases in config_tests already cover 1. I've already added `oauth-provider-inactive` for 3. So I will add one more test case for 2.

pkg/lib/config/identity.go
tung2744: My concern is disabled looks like the whole item is disabled, but in fact the other configs are still effective. But I don't have other idea so let me name it as disabled first.

tung2744:
@louischan-oursky Added 3 commits to
- Rename is_active to disabled, remove default value
- Add 1 test case

PR #5184, opened 2025-05-16
Handle inactive provider in auth flow

tung2744:
ref DEV-2715
ref DEV-2712
requires https://github.com/authgear/authgear-server/pull/5183
Screenshot 2025-05-16 at 6 21 16 PM


pkg/lib/authenticationflow/declarative/data_oauth.go
```diff
@@ -5,12 +5,21 @@ import (
...
 )
 
+type OAuthProviderStatus string
+
+const (
+	OAuthProviderStatusActive   OAuthProviderStatus = "active"
+	OAuthProviderStatusInactive OAuthProviderStatus = "inactive"
+	// TODO(tung): Add demo status
+)
+
```
louischan-oursky, line 15: Please also update the spec documentation.

> - status: The status about this option. It has the following valid values.
>   - active: The OAuth provider has credentials configured and it is usable.
>   - missing_credentials: The OAuth provider is missing credentials and it is NOT usable.
>   - using_demo_credentials: The OAuth provider is using a shared credentials for demo purpose. It is NOT for production use.

- Rename `inactive` to `missing_credentials`
- Use `using_demo_credentials` for the variant we has not yet introduce.

pkg/lib/authenticationflow/declarative/data_identification.go
```diff
@@ -32,6 +32,8 @@ type IdentificationOption struct {
 	Alias string `json:"alias,omitempty"`
 	// WechatAppType is specific to OAuth.
 	WechatAppType wechat.AppType `json:"wechat_app_type,omitempty"`
+	// ProviderStatus is specific to OAuth.
+	ProviderStatus OAuthProviderStatus `json:"provider_status,omitempty"`
```
louischan-oursky, line 36: If we call this `provider_status`, then we should also call it `provider_status` in the previous commit. So that the names are consistent.

pkg/api/errors.go
```diff
@@ -27,6 +27,7 @@ var ErrIdentityNotFound = apierrors.NotFound.WithReason("IdentityNotFound").New(
 
 var ErrInvalidCredentials = InvalidCredentials.New("invalid credentials")
 var ErrOAuthProviderNotFound = apierrors.NotFound.WithReason("OAuthProviderNotFound").New("oauth provider not found")
+var ErrOAuthProviderInactive = apierrors.Invalid.WithReason("OAuthProviderInactive").New("oauth provider is inactive because of missing credentials")
```
louischan-oursky: Rename the reason to `OAuthProviderMissingCredentials`

pkg/lib/authenticationflow/declarative/errors.go
```diff
@@ -0,0 +1,7 @@
+package declarative
+
+import (
+	"github.com/authgear/authgear-server/pkg/api/apierrors"
+)
+
+var InactiveOAuthProvider = apierrors.Invalid.WithReason("InactiveOAuthProvider")
```
louischan-oursky, line 7: This error seems unused. Remove it.

pkg/lib/authn/sso/oauth_provider.go
```diff
@@ -42,7 +43,12 @@ func (p *OAuthProviderFactory) getActiveOrDemoProvider(alias string) (provider o
 
 	if config.OAuthSSOProviderConfig(providerConfig).IsDisabled() {
 		// TODO(tung): handle demo status
-		err = api.ErrOAuthProviderInactive
+
+		details := apierrors.Details{
+			"OAuthProviderAlias": alias,
+			"OAuthProviderType":  providerConfig.Type(),
+		}
+		err = api.OAuthProviderInactive.NewWithInfo("oauth provider is inactive because of missing credentials", details)
```
louischan-oursky, line 51: I think this error construction should be a method of `config.OAuthSSOProviderConfig` Maybe we should even remove `IsDisabled()` and replace it with `GetUsableProviderConfig`

pkg/auth/handler/webapp/error_renderer.go
```diff
@@ -167,7 +167,12 @@ func (s *ErrorRenderer) MakeAuthflowErrorResult(ctx context.Context, w http.Resp
 		case config.UIImplementationAuthflowV2:
 			s.AuthflowV2Navigator.NavigateNonRecoverableError(r, &u, err)
 		}
-		return s.makeNonRecoverableResult(ctx, u, apierror)
+		fmt.Println("r.URL.String()", r.URL.String())
```
louischan-oursky, line 170: Remove log

pkg/auth/handler/webapp/error_renderer.go
```diff
@@ -167,7 +167,12 @@ func (s *ErrorRenderer) MakeAuthflowErrorResult(ctx context.Context, w http.Resp
...
+		apierr = apierr.CloneWithAdditionalInfo(apierrors.Details{
+			"FromURL": r.URL.String(),
+		})
+		fmt.Println("apiErr", apierr.Info_ReadOnly)
```
louischan-oursky, line 174: Remove log

louischan-oursky (changes requested):

pkg/lib/authenticationflow/declarative/data_oauth.go
tung2744: @louischan-oursky I am ok with the renaming. But I also suggest to rename `disabled` of the oauth config to `missing_credential_allowed`, so we can relate them. What do you think?

louischan-oursky: > But I also suggest to rename disabled of the oauth config to missing_credential_allowed, so we can relate them. What do you think?

OK.

pkg/lib/authn/sso/oauth_provider.go
tung2744: Do "usable" includes demo credential? If yes, the demo credential probably come from app dependencies, so we cannot construct it in config.OAuthSSOProviderConfig. And for the same reason, we can only determine if the error should be returned in OAuthProviderFactory.

tung2744:
@louischan-oursky 
Updated, thanks!

pkg/lib/authn/sso/oauth_provider.go
louischan-oursky: Got that. Let's keep what we have then.

louischan-oursky (approved):

PR #5191, opened 2025-05-20
Add a sso callback route for any project

tung2744:
ref DEV-2716

This pr is not testable at the moment, but you can test the endpoint with the following steps:

1. Update `.env`, set `SHARED_AUTHGEAR_ENDPOINT`
2. Have a valid item in `oauth.providers`. The provider should whitelist `http://localhost:3100/noproject/sso/oauth2/callback` as a redirect url.
3. Modify `GetSSOCallbackURL` in `AuthflowController`.

```diff
diff --git a/pkg/auth/handler/webapp/authflow_controller.go b/pkg/auth/handler/webapp/authflow_controller.go
index 8e4b27506..fd2030851 100644
--- a/pkg/auth/handler/webapp/authflow_controller.go
+++ b/pkg/auth/handler/webapp/authflow_controller.go
@@ -1180,10 +1180,11 @@ func (c *AuthflowController) finishSession(
 }
 
 func (c *AuthflowController) GetSSOCallbackURL(alias string) (string, error) {
-       oauthProviderConfig, ok := c.OAuthSSOConfig.GetProviderConfig(alias)
-       if !ok {
-               return "", fmt.Errorf("unknown alias %s", alias)
-       }
-       callbackURL := c.Endpoints.SSOCallbackURL(alias, config.OAuthSSOProviderConfig(oauthProviderConfig).IsMissingCredentialAllowed()).String()
+       // oauthProviderConfig, ok := c.OAuthSSOConfig.GetProviderConfig(alias)
+       // if !ok {
+       //      return "", fmt.Errorf("unknown alias %s", alias)
+       // }
+       // callbackURL := c.Endpoints.SSOCallbackURL(alias, config.OAuthSSOProviderConfig(oauthProviderConfig).IsMissingCredentialAllowed()).String()
+       callbackURL := c.Endpoints.SSOCallbackURL(alias, true).String()
        return callbackURL, nil
 }
```

PR #5194, opened 2025-05-21
Support oauth demo credentials in authflow

tung2744:
ref DEV-2719

requires https://github.com/authgear/authgear-server/pull/5191

## How to test

1. Add demo credentials to `authgear.secrets.yaml`.
```yaml
- data:
    items:
      - provider_config:
          claims:
            email:
              assume_verified: true
              required: true
          client_id: {CLIENT_ID}
          type: google
        client_secret:  {CLIENT_SECRET}
  key: sso.oauth.demo_credentials
```

2. Set `missing_credential_allowed` in the corresponding provider:
```yaml
    - alias: google
      create_disabled: false
      delete_disabled: false
      type: google
      missing_credential_allowed: true
```

3. Use authflow api to test oauth.

### DIFF
--- a/pkg/lib/authn/sso/oauth_provider.go
+++ b/pkg/lib/authn/sso/oauth_provider.go
@@ -21,6 +21,7 @@ type StandardAttributesNormalizer interface {
 type OAuthProviderFactory struct {
 	IdentityConfig               *config.IdentityConfig
 	Credentials                  *config.OAuthSSOProviderCredentials
+	SSOOAuthDemoCredentials      *config.SSOOAuthDemoCredentials
 	Clock                        clock.Clock
 	StandardAttributesNormalizer StandardAttributesNormalizer
 	HTTPClient                   OAuthHTTPClient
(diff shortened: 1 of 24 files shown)


pkg/lib/authn/sso/oauth_provider.go
```diff
@@ -42,13 +43,26 @@ func (p *OAuthProviderFactory) getActiveOrDemoProvider(alias string) (provider o
...
 
-		details := apierrors.Details{
-			"OAuthProviderAlias": alias,
-			"OAuthProviderType":  providerConfig.Type(),
+		deps = &oauthrelyingparty.Dependencies{
+			Clock:          p.Clock,
+			ProviderConfig: demoItem.ProviderConfig,
+			ClientSecret:   demoItem.ClientSecret,
+			HTTPClient:     p.HTTPClient.Client,
+			SimpleStore:    p.SimpleStoreRedisFactory.GetStoreByProvider(providerConfig.Type(), alias),
```
louischan-oursky, line 62: providerConfig should be `demoItem.ProviderConfig`?

pkg/lib/config/identity.go
```diff
@@ -352,6 +352,29 @@ func (c OAuthSSOProviderConfig) IsMissingCredentialAllowed() bool {
 	return v
 }
 
+type OAuthProviderStatus string
+
+const (
+	OAuthProviderStatusActive               OAuthProviderStatus = "active"
+	OAuthProviderStatusMissingCredentials   OAuthProviderStatus = "missing_credentials"
+	OAuthProviderStatusUsingDemoCredentials OAuthProviderStatus = "using_demo_credentials" // nolint: gosec
```
louischan-oursky, line 360: Rename to `using_shared_credentials`

louischan-oursky (changes requested):

pkg/lib/config/secret.go
```diff
@@ -331,6 +331,7 @@ const (
 	AdminAPIAuthKeyKey          SecretKey = "admin-api.auth"
 	// nolint: gosec
 	OAuthSSOProviderCredentialsKey SecretKey = "sso.oauth.client"
+	SSOOAuthDemoCredentialsKey     SecretKey = "sso.oauth.demo_credentials"
```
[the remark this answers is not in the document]
tung2744: Theoretically we can have demo credentials that are not shared, we make them shared by using deployment level secrets.
And I prefer name it as demo_credentials because it is always used as demo only.

pkg/lib/config/identity.go
tung2744: Similar to the secret key, I prefer name it demo credential.
In auth ui, we also call this as demo credential.

pkg/lib/authn/sso/oauth_provider.go
tung2744: The type should be same, so they are same. Let me still update to use demoItem so we don't have to think about this.

PR #5198, opened 2025-05-22
Fix calling WithTx() in DidCommitTx() will block

louischan-oursky:
ref DEV-2729


tung2744:
@louischan-oursky Some lint failed

tung2744 (approved):

PR #5199, opened 2025-05-22
Update portal ui for demo credentials

tung2744:
ref DEV-2718

https://github.com/user-attachments/assets/34a99e12-361f-4032-94d1-d3e01d85ba7f


portal/src/graphql/portal/SingleSignOnConfigurationScreen.tsx
```diff
@@ -258,7 +273,7 @@ const SingleSignOnConfigurationScreen1: React.VFC<{
     };
   }, [renderToString]);
 
-  if (form.isLoading || featureConfig.loading) {
+  if (form.isLoading || featureConfig.loading || isEffectiveSecretLoading) {
```
louischan-oursky, line 276: We should handle this like `if (form.isLoading || featureConfig.loading || effectiveSecretConfig.loading)` and handle the error too `if (form.loadError ?? featureConfig.error ?? effectiveSecretConfig.error)`

portal/src/graphql/portal/EditSingleSignOnConfigurationScreen.tsx
```diff
@@ -171,7 +188,12 @@ const EditSingleSignOnConfigurationScreen1: React.VFC<{
     }
   }, [isReadyToEdit, onRevealSecrets]);
 
-  if (!isReadyToEdit || form.isLoading || featureConfig.loading) {
+  if (
+    !isReadyToEdit ||
+    form.isLoading ||
+    featureConfig.loading ||
+    isEffectiveSecretLoading
+  ) {
```
louischan-oursky, line 196: Maybe we need a new abstraction to handle loading and error in case there are multiple queries being used in a component?

tung2744: Let me try to introduce a new hook or component for this.

tung2744:
Changes from [d27bfb3](https://github.com/authgear/authgear-server/pull/5199/commits/d27bfb3eccf080ce4000950f55c3d3996a321bee).

tung2744:
> Previously, if it is true, then the provider can have its client_id and client_secret missing. If the credentials are missing, AND a demo credentials is available, the demo one is used.

This is still true for authgear server.

> However, in this PR, the meaning of missing_credential_allowed has changed to: if it is true, then demo credentials is used, even the project has configured its credentials.

This is intended, as we allow choosing demo credential or custom credential in portal. And I am trying to keep this as a portal logic.

> I think that flag should be changed to an enum

When are we using these 3 cases?

1. If no demo credential, we always use `use_project_credentials_always`? The problem is that we want a flag to control whether the server will reject an invalid credential. If `use_project_credentials_always` will reject invalid credential, we need another case to tell server not to validate the credential.

2. If we have demo credential, and user choose demo credential in portal: Use `use_demo_credentials_always`, this make sense.

3. If we have demo credential, and user choose custom credential in portal: Use `use_project_credentials_always`? Then `use_demo_credentials_if_available` is never used.



My thought was `missing_credential_allowed` should be simpler: It is a flag to control whether the server turn on the validation of the credential or not. When true, validation can be bypassed, and therefore no guarantee the credential is valid. Otherwise, it is guaranteed that the stored credential is valid.

tung2744:
ok, let me add a commit to change missing_credential_allowed to a enum according to the above way.

tung2744:

PR #5202, opened 2025-05-26
Remove reserved word from generated app id

tung2744:
ref DEV-2730


portal/src/util/projectname.ts
```diff
@@ -1,7 +1,8 @@
 // ref https://github.com/bitcoinjs/bip39/blob/master/src/wordlists/english.json
 import wordlist from "./wordlist.json";
 
-function determineWord(index: number): string {
+function determineWord(seed: number): string {
+  const index = seed % wordlist.length;
```
louischan-oursky, line 5: That would make the function non-bijective.

Can we use this word list instead? https://github.com/satoshilabs/slips/blob/master/slip-0039/wordlist.txt Let's check if slip-0039 has fewer (or even better, no) reversed words.

Since the wordlist is shorter, we need to amend the algorithm to take different number of bits (from 11bits to 10bits)

Or we simply drop the fix to the issue. It is not very important.

tung2744: There are still reserved words in the new list.

tung2744: I think the way to randomize the word is not very important.
Let me create another pr for this fix so we can merge the rename commit first.

tung2744: @louischan-oursky Created https://github.com/authgear/authgear-server/pull/5203 for the renaming instead

louischan-oursky: I suggest we keep wordlist.json unchanged. Add a new file `reserved_words.json` to contain the words that is reserved in wordlist.json.

In the project name generation, if the random word matches a word in reserved_words.json, try again (That is, a infinite for loop).

tung2744: I feel like this is too complicated for a simple random project id generation, and this created one more list to maintain. What about simply modify `randomProjectID` so that it won't generate a number larger than the max index of the word list?

louischan-oursky: The problem is that we have this comment in the code

> // ref https://github.com/bitcoinjs/bip39/blob/master/src/wordlists/english.json
import wordlist from "./wordlist.json";

But after this change, the commend is no longer valid. I want to document that the list has been changed, and what words have been removed. So future reader will understand the list is no longer BIP39, and what words are excluded.

I think changing the comment to this is also acceptable.

> ./wordlist.json is originally https://github.com/bitcoinjs/bip39/blob/master/src/wordlists/english.json
> However, we found that some of the words are actually reserved words in reserved_app_id.txt
> Therefore, the following words are removed from the list.
> The implementation is also changed to accommodate the fact that the list is no longer 2048 items long.
> [a list of removed words here]

PR #5205, opened 2025-05-27
Specify new authentication blocking events

tung2744:
ref DEV-2725


docs/specs/amr.md
```diff
@@ -0,0 +1,24 @@
...
+The following AMR values are supported:
+
+| AMR Value     | Description                      | Authenticator Type                      |
+| ------------- | -------------------------------- | --------------------------------------- |
+| `pwd`         | Password-based authentication    | `password`                              |
+| `otp`         | One-time password authentication | `totp`, `oob_email`, `oob_sms`          |
+| `sms`         | SMS-based authentication         | `oob_sms`                               |
+| `mfa`         | Multi-factor authentication      | Added when multiple authenticators used |
+| `x_biometric` | Biometric authentication         | `biometric`                             |
+| `x_passkey`   | Passkey authentication           | `passkey`                               |
```
louischan-oursky: Add a note on Auth0 is using `phr` to represent passkey authentication. https://auth0.com/blog/all-you-need-to-know-about-passkeys-at-auth0/ The blogpost said `phr` is a standard value, but I googled a bit and found no standard document mentioning this.

docs/specs/event.md
```diff
@@ -172,6 +179,205 @@ Occurs right before the account anonymization is scheduled.
...
 
+#### authentication.pre_initialize
+
+Occurs right before any authentication, such as login.
+
+Example payload:
+
+```json5
+{
+  "authentication": {
```
louischan-oursky: This example is missing the root `payload` key.

docs/specs/event.md
```diff
@@ -172,6 +179,205 @@ Occurs right before the account anonymization is scheduled.
...
+            "https://authgear.com/claims/oauth/provider_type": "google",
+            "https://authgear.com/claims/oauth/subject_id": "1234567",
+            "name": "Test Authgear"
+          },
+          "type": "oauth",
+          "created_at": "2025-05-27T06:32:54.02264Z",
+          "updated_at": "2025-05-27T06:32:54.02264Z"
+        },
+      ],
+      "asserted_authenticators": [ /* Authenticator used during the authentication */
```
louischan-oursky: Instead of including `asserted_authenticators`, maybe we should just include `amr`? So the developer do not need to loop through our proprietary `asserted_authenticators` array to see what have been asserted.

Is it easy to add a new example to demo the use of `authentication.amr`?

docs/specs/event.md
```diff
@@ -530,206 +736,3 @@ Occurs after the user failed to input the OTP delivered to their phone.
   }
 }
 ```
-
```
louischan-oursky: Why were these removed?

tung2744: Should be removed by accident, let me restore them

docs/specs/event.md
tung2744: Let me include both `amr` and `asserted_authenticators`. And add an example of using `amr`.

tung2744:
Updated, thanks!

docs/specs/event.md
```diff
@@ -105,6 +108,10 @@ All fields are guaranteed that only backward-compatible changes would be made.
   - `portal`: The event originates from the management portal.
 - `preferred_languages`: User preferred languages which are inferred from the request. It is the value of `ui_locales` query parameter if it is provided, or the value of the `Accept-Language` header. It is an empty array when the event is not generated by the end user.
 - `language`: The locale which is derived based on user's preferred languages and app's languages config. The fallback value is the fallback language of the app.
+- `geo_location_code`: The (ISO 3166-1 alpha-2 code)[https://www.iban.com/country-codes] of the location derived from the ip address. `null` if the location cannot be determined by the ip address, for example, it is an internal ip address.
```
tung2744, line 114: Empty string if location cannot be determined

louischan-oursky: Empty string is no good. Omit the key if location cannot be determined.

tung2744: Then maybe I keep it null. The user may want to look at the json to know which key he should check.

tung2744:
post_identified should also be called when identified a user by id token. In such case, the payload will have `null` identity.
I also suggest to add a new field `identification`, which tells how this identification is done.
Possible values are values of `identification` in the authflow.
```
{
  "identity": null,
  "id_token": "abcde",
  "identification": "id_token"
}
```

- [x] Update the spec

tung2744:
- [x] Mention how recovery code participate in AMR

tung2744:
- [x] Mention passkey is not supported in AMR constraints at the moment

docs/specs/amr.md
```diff
@@ -0,0 +1,49 @@
...
+| `pwd`                       | Password-based authentication.                                                             |
+| `otp`                       | One-time password (OTP) authentication.                                                    |
+| `sms`                       | SMS-based authentication.                                                                  |
+| `mfa`                       | Multi-factor authentication; Added when multiple authenticators are used in a single flow, OR one authenticator with one recovery code. |
+| `x_biometric`               | Biometric authentication.                                                                  |
+| `x_primary_password`        | Indicates primary password authentication.                                                 |
+| `x_primary_oob_otp_email`   | Indicates primary one-time password (OTP) authentication via email.                        |
+| `x_primary_oob_otp_sms`     | Indicates primary one-time password (OTP) authentication via SMS.                          |
+| `x_primary_passkey`         | Indicates passkey authentication.                                                          |
```
tung2744: @louischan-oursky 
I've renamed `x_passkey` to `x_primary_passkey` to fit the convention `x_{authentication}`. I think existing users does not check the amr so I think we can rename it directly?

docs/specs/event.md
```diff
@@ -172,6 +179,187 @@ Occurs right before the account anonymization is scheduled.
...
+
+Fields in payload:
+- `authentication_context`: An [`AuthenticationContext`](./event_models.md#authenticationcontext) object.
+
+Example payload:
+
+```json5
+{
+  "payload": {
+    "authentication_context": {
```
tung2744, line 217: I've renamed authentication to `authentication_context` to avoid naming collision with `authentication`

tung2744:
@fungc-io @chpapa @louischan-oursky 

I've updated the spec per our latest discussion.

Major changes:
1. Authentication event payload updated:
    - asserted_identities -> asserted_identifications
    - asserted_authenticators -> asserted_authentications
    - You can see the updated payload here:
      - [authentication.pre_initialize](https://github.com/authgear/authgear-server/blob/5ce447fb55d23ccaae3c2a1232ed4f043ec45f89/docs/specs/event.md#authenticationpre_initialize)
      - [authentication.post_identified](https://github.com/authgear/authgear-server/blob/5ce447fb55d23ccaae3c2a1232ed4f043ec45f89/docs/specs/event.md#authenticationpost_identified)
      - [authentication.pre_authenticated](https://github.com/authgear/authgear-server/blob/5ce447fb55d23ccaae3c2a1232ed4f043ec45f89/docs/specs/event.md#authenticationpre_authenticated)
    - I've also added a page to document these common objects in event payloads. [Event Models](http://github.com/authgear/authgear-server/blob/5ce447fb55d23ccaae3c2a1232ed4f043ec45f89/docs/specs/event_models.md)
2. Updated [Glossary](https://github.com/authgear/authgear-server/blob/5ce447fb55d23ccaae3c2a1232ed4f043ec45f89/docs/specs/glossary.md) for common concepts appearing in APIs.
3. Updated [AMR](https://github.com/authgear/authgear-server/blob/5ce447fb55d23ccaae3c2a1232ed4f043ec45f89/docs/specs/amr.md)
    - Also added a unique amr value for all authentication methods. The convention is `x_{authentication}`.

- [x] Add back `x_passkey`

tung2744:
- [x] Document hook response in event

louischan-oursky:
@tung2744 If I remember correctly, we agreed that we are not going to add new types to authflow, right?

tung2744:
Yes. I've added [2ba0b2c](https://github.com/authgear/authgear-server/pull/5205/commits/2ba0b2ca985f86d65f7792f3803034906576eed8) to remove the new step types.

tung2744:
- [x] Document supported flows in pre_initialized. It should support login, signup, reauth, promote, signup_login

tung2744:
Changes:
1. Give a name to each rate limit, and also made some adjustments to the spec. See c387784.
2. Updated rate_limits in hook response according to our latest discussions. See 59065ca.

tung2744:
Added two commits to:
- Add example to check ip address against a list of cidr ranges
- Mention the fallback behavior of rate limits

docs/specs/rate-limit.md
```diff
@@ -29,217 +31,251 @@ Therefore, verifying a correct password would never exceed the rate limit.
...
+| `verification.email.trigger`            | Send Verification Email                 | Disabled                             |            |                                      |          | Mitigate mass/targeted message spam.                                                                                  |
+| `verification.email.validate`           | Verify Verification Email               | 60/minute                            |            |                                      |          | Mitigate credential brute-force.                                                                                      |
+| `verification.sms.trigger`              | Send Verification SMS                   | Disabled                             |            |                                      |          | Mitigate mass/targeted message spam.                                                                                  |
```
tung2744, line 62: verification.sms.trigger supports per user rate limit

docs/specs/rate-limit.md
```diff
@@ -29,217 +31,251 @@ Therefore, verifying a correct password would never exceed the rate limit.
...
+| -                                       | Login OTP failed verify attempts        |                                      | 5 attempts |                                      |          | Mitigate credential brute-force; revoke OTP when set limit exceeded.                                                  |
+| -                                       | Verification OTP failed verify attempts |                                      | 5 attempts |                                      |          |                                                                                                                       |
+| `verification.email.trigger`            | Send Verification Email                 | Disabled                             |            |                                      |          | Mitigate mass/targeted message spam.                                                                                  |
```
tung2744, line 60: verification.email.trigger supports per user rate limit

PR #5209, opened 2025-05-30
Support include_identity_attributes_in_id_token

louischan-oursky:
ref DEV-2754

### DIFF
--- a/pkg/lib/oauth/handler/handler_authz.go
+++ b/pkg/lib/oauth/handler/handler_authz.go
@@ -10,6 +10,7 @@ import (
 	"github.com/authgear/authgear-server/pkg/api/apierrors"
 	"github.com/authgear/authgear-server/pkg/api/model"
 	"github.com/authgear/authgear-server/pkg/lib/authn/authenticationinfo"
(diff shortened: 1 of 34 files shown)


pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -809,6 +814,7 @@ func (h *AuthorizationHandler) doHandleConsentRequest(
 			IDTokenHintSID:       idTokenHintSID,
 			Cookies:              []*http.Cookie{},
 			GrantAuthz:           opts.GrantAuthz,
+			IdentitySpecs:        opts.ConsentRequest.OAuthSessionEntry.T.IdentitySpecs,
```
tung2744, line 817: I think `AuthInfoEntry.T` is a better place to store `IdentitySpecs`, because it is something produced **after** the authentication. And `OAuthSessionEntry` seems to be storing something related to the request, to things that used during the authentication.

louischan-oursky: Added 2 commits to address this.

PR #5214, opened 2025-06-02
Allow calling blocking hooks between steps which affect the flow

tung2744:
ref DEV-2768

### DIFF
--- a/pkg/api/event/blocking/authentication_post_identified.go
+++ b/pkg/api/event/blocking/authentication_post_identified.go
@@ -34,7 +34,7 @@ func (e *AuthenticationPostIdentifiedBlockingEventPayload) ApplyHookResponse(ctx
 	if response.Constraints.AMR != nil {
 		e.Constraints = &response.Constraints
 	}
-	return event.ApplyHookResponseResult{UserMutationsEverApplied: false}
+	return event.ApplyHookResponseResult{MutationsEverApplied: false}
 }
 
 func (e *AuthenticationPostIdentifiedBlockingEventPayload) PerformEffects(ctx context.Context, effectCtx event.MutationsEffectContext) error {
--- a/pkg/api/event/blocking/authentication_post_identified.go
+++ b/pkg/api/event/blocking/authentication_post_identified.go
@@ -12,7 +12,8 @@ const (
 )
 
 type AuthenticationPostIdentifiedBlockingEventPayload struct {
-	Identity model.Identity `json:"identity"`
+	Identity    model.Identity    `json:"identity"`
+	Constraints event.Constraints `json:"constraints,omitempty"`
 }
 
 func (e *AuthenticationPostIdentifiedBlockingEventPayload) BlockingEventType() event.Type {
@@ -29,9 +30,11 @@ func (e *AuthenticationPostIdentifiedBlockingEventPayload) GetTriggeredBy() even
 
 func (e *AuthenticationPostIdentifiedBlockingEventPayload) FillContext(ctx *event.Context) {}
 
-func (e *AuthenticationPostIdentifiedBlockingEventPayload) ApplyHookResponse(ctx context.Context, response event.HookResponse) bool {
-	// no-op
-	return false
+func (e *AuthenticationPostIdentifiedBlockingEventPayload) ApplyHookResponse(ctx context.Context, response event.HookResponse) event.ApplyHookResponseResult {
+	if response.Constraints.AMR != nil {
+		e.Constraints = response.Constraints
+	}
+	return event.ApplyHookResponseResult{UserMutationsEverApplied: false}
 }
 
 func (e *AuthenticationPostIdentifiedBlockingEventPayload) PerformEffects(ctx context.Context, effectCtx event.MutationsEffectContext) error {
--- a/pkg/api/event/blocking/authentication_post_identified.go
+++ b/pkg/api/event/blocking/authentication_post_identified.go
@@ -12,8 +12,8 @@ const (
 )
 
 type AuthenticationPostIdentifiedBlockingEventPayload struct {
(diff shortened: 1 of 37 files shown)


pkg/api/event/blocking/authentication_post_identified.go
```diff
@@ -12,7 +12,8 @@ const (
 )
 
 type AuthenticationPostIdentifiedBlockingEventPayload struct {
-	Identity model.Identity `json:"identity"`
+	Identity    model.Identity    `json:"identity"`
+	Constraints event.Constraints `json:"constraints,omitempty"`
```
louischan-oursky: I think this behavior is unspecified in the document.

Previously, we only support mutations so the event payload is mutated directly to reflect the changes made by a previous hook.

I think we should specify how the event look like when there are multiple hooks

Here is just an example of how it could look like
```json
{
  "payload": {
    "identity": { ... },
    "authentication": {
      "authentication_flow": { ... },
      "user": { ... },
      "asserted_identities": { ... },
      "asserted_authenticators": { ... },
      "amr": { ... }
    },
    "previous_hook_response": {
      "constraints": { ... },
      "rate_limit": { ... },
      "bot_protection": { ... }
    }
  }
}

```

---

Another problem, we did not specify whether `mutations` is supported in these new blocking events. We need to specify in the spec. I think not supporting is fine, since the primary purpose of these new events is to control authentication.

pkg/lib/ldap/entry.go
```diff
@@ -24,5 +29,14 @@ func (e *Entry) ToJSON() map[string]interface{} {
...
 	}
+	// Ensure the types are correct by serialize once and deserialize again
+	bytes, err := json.Marshal(dict)
+	if err != nil {
+		panic(fmt.Errorf("failed to marshal ldap attribute"))
+	}
+	err = json.Unmarshal(bytes, &dict)
+	if err != nil {
+		panic(fmt.Errorf("failed to unmarshal ldap attribute"))
+	}
```
louischan-oursky, line 40: What bug does this change fix? Please add a test case.

pkg/lib/authenticationflow/declarative/node_do_use_identity_passkey.go
```diff
@@ -18,53 +14,26 @@ func init() {
...
 type NodeDoUseIdentityPasskey struct {
-	AssertionResponse       []byte                `json:"assertion_response,omitempty"`
-	Identity                *identity.Info        `json:"identity,omitempty"`
-	Authenticator           *authenticator.Info   `json:"authenticator,omitempty"`
-	RequireUpdate           bool                  `json:"require_update,omitempty"`
-	IsPostIdentifiedInvoked bool                  `json:"is_post_identified_invoked"`
-	Constraints             *eventapi.Constraints `json:"constraints,omitempty"`
+	*NodeDoUseIdentity
+	AssertionResponse []byte              `json:"assertion_response,omitempty"`
+	Identity          *identity.Info      `json:"identity,omitempty"`
```
louischan-oursky: This field `Identity` clashes with the field of the embedded `*NodeDoUseIdentity`. Is that an issue?

pkg/lib/authenticationflow/declarative/node_do_create_passkey.go
```diff
@@ -12,7 +12,39 @@ func init() {
...
+	return &authflow.NodeWithDelayedOneTimeFunction{
+		Node:                   authflow.NewNodeSimple(node),
+		DelayedOneTimeFunction: delayedFunction,
+	}, nil
+}
+
 type NodeDoCreatePasskey struct {
+	*NodeDoCreateIdentity
 	SkipCreate          bool                `json:"skip_create,omitempty"`
 	Identity            *identity.Info      `json:"identity,omitempty"`
```
louischan-oursky: Same here, now `NodeDoCreatePasskey` has two `Identity` fields, two `SkipCreate` fields.

pkg/api/event/blocking/authentication_post_identified.go
tung2744: 1. I think I will simply remove `constraints` in json serialization, I just need a place to store the hook response.
2. For multiple hook, I think we can simply read the last response. I think there is no need to chain hooks.

pkg/lib/ldap/entry.go
tung2744: The e2e tests failed with a panic without this fix in here:
https://github.com/authgear/authgear-server/blob/main/pkg/lib/authn/identity/ldap_identity.go#L42
Because the type of RawEntryJSON here is not []interface{}, but [][]uint8. As the field was named `RawEntryJSON`, I expect there should only be types specified in json.Unmarshal,. i.e. []interface{}.

I am not sure why it was not triggered before this pr.

tung2744:
Added 3 commits to address the comments, and also updated the spec. Thanks!

pkg/lib/authenticationflow/declarative/node_do_create_passkey.go
```diff
@@ -12,16 +12,47 @@ func init() {
...
+
+	return &authflow.NodeWithDelayedOneTimeFunction{
+		Node:                   authflow.NewNodeSimple(node),
+		DelayedOneTimeFunction: delayedFunction,
+	}, nil
+}
+
 type NodeDoCreatePasskey struct {
+	*NodeDoCreateIdentity
 	SkipCreate          bool                `json:"skip_create,omitempty"`
```
louischan-oursky: SkipCreate is still duplicated.

tung2744: Removed, thanks!

louischan-oursky (approved):

PR #5215, opened 2025-06-04
Add authentication details to event payload

tung2744:
ref DEV-2769
requires https://github.com/authgear/authgear-server/pull/5214


pkg/lib/authenticationflow/declarative/milestone.go
```diff
@@ -52,16 +52,38 @@ func getUserID(flows authflow.Flows) (userID string, err error) {
 }
 
 func collectAMR(ctx context.Context, deps *authflow.Dependencies, flows authflow.Flows) (amr []string, err error) {
```
tung2744, line 55: There are two changes to the current AMR behavior:
1. Originally, signup flow has no AMR. Now it will have amr depends on the authenticator created.
2. Added `mfa` if there are more than one authenticator used.

pkg/lib/authenticationflow/declarative/intent_account_linking.go
```diff
@@ -142,7 +142,7 @@ func (i *IntentAccountLinking) ReactTo(ctx context.Context, deps *authflow.Depen
 		if err != nil {
 			return nil, err
 		}
-		return NewNodeDoCreateIdentityReactToResult(ctx, deps, NodeDoCreateIdentityOptions{
+		return NewNodeDoCreateIdentityReactToResult(ctx, flows, deps, NodeDoCreateIdentityOptions{
```
louischan-oursky: The convention of the order of the argument should be `ctx, deps, flows` Please change them all.

pkg/lib/authenticationflow/declarative/utils_event_test.go
```diff
@@ -0,0 +1,338 @@
...
+	"github.com/authgear/authgear-server/pkg/lib/authenticationflow/declarative"
+	"github.com/authgear/authgear-server/pkg/lib/authn/authenticator"
+	"github.com/authgear/authgear-server/pkg/lib/authn/identity"
+	"github.com/authgear/authgear-server/pkg/lib/authn/user"
+	"github.com/authgear/authgear-server/pkg/util/accesscontrol"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+// MockUserService is a mock implementation of authenticationflow.UserService
+type MockUserService struct {
```
louischan-oursky: Why don't we introduce an interface and ask mockgen to generate the mock?

louischan-oursky (changes requested):

pkg/lib/authenticationflow/declarative/utils_event_test.go
tung2744: I generated the mock using LLM. Let me use mockgen to generate it.

pkg/lib/authenticationflow/dependencies.go
```diff
@@ -36,6 +36,8 @@ import (
 	"github.com/authgear/authgear-server/pkg/util/httputil"
 )
 
+//go:generate go tool mockgen -source=dependencies.go -destination=dependencies_mock_test.go -package authenticationflow
```
louischan-oursky: I wonder why the filename is not `dependencies_mock_test.go`. If it is ended with `_test.go`, it is not included in the final executable. Why did we change to `_gen.go`?

tung2744: This is because the mock was in authenticationflow package, but the test cases that need authenticationflow.Dependencies are usually in declarative package. And I found that files end with _test.go seems cannot be imported in other packages, so I change it go _gen.

louischan-oursky: You can write the minimal interface in the file that reference the interface. And then `//go:generate go tool mockgen` to generate it. In doing so, you can keep `_test.go`.

tung2744: But our functions in declarative package usually takes `*authenticationflow.Dependencies` as the argument, therefore I need to copy the interfaces used in this struct to the declarative package to generate a mock that can be used to construct `authenticationflow.Dependencies`, which seems a bit redundant.

louischan-oursky: You only need one method `deps.Users.XXX`. So the interface could just be that method. And then in the unit test, you construct a bare Dependencies object

```golang
deps := Dependencies{
  Users: mockUserService,
}
```

tung2744: I think `mockUserService` must implements all methods of `Dependencies.Users` so it can be assigned to it? So the interface which generates `mockUserService` must be the interface of `Dependencies.Users`.

tung2744: Added https://github.com/authgear/authgear-server/pull/5215/commits/94baf34f4e239b35a4d0e31c29c8b50e1c5c38ed so that we only generate user service in declarative package.

louischan-oursky:
Please rebase onto main to receive the AUDIT_LEVEL fix.

tung2744:
Rebased, thanks!

PR #5218, opened 2025-06-09
Modify authflow to support amr contraints & Add authentication.pre_authenticated

tung2744:
ref DEV-2770
ref DEV-2773

<del>requires https://github.com/authgear/authgear-server/pull/5215</del>


pkg/lib/authenticationflow/declarative/intent_signup_flow_ensure_contraints_fulfilled.go
```diff
@@ -0,0 +1,121 @@
...
+		case model.AuthenticatorTypePassword:
+			addOneOf(config.AuthenticationFlowAuthenticationSecondaryPassword, nil)
+		case model.AuthenticatorTypeOOBEmail:
+			addOneOf(config.AuthenticationFlowAuthenticationSecondaryOOBOTPEmail, getBotProtectionRequirementsOOBOTPEmail)
+		case model.AuthenticatorTypeOOBSMS:
+			addOneOf(config.AuthenticationFlowAuthenticationSecondaryOOBOTPSMS, getBotProtectionRequirementsOOBOTPSMS)
+		case model.AuthenticatorTypeTOTP:
+			addOneOf(config.AuthenticationFlowAuthenticationSecondaryTOTP, nil)
+		case model.AuthenticatorTypePasskey:
+			// FIXME(tung): We don't have a step to force user create passkey at the moment
```
tung2744: I found that we don't have a "Create Passkey" option in the create authenticator step at the moment, so if amr contraints includes `x_passkey`, we still cannot ensure user created passkey in signup flow.

tung2744:
- [x] Validate the constrains in hook response

pkg/lib/config/authentication_flow.go
```diff
@@ -1075,6 +1076,9 @@ type AuthenticationFlowLoginFlowStep struct {
 
 	// TargetStep is relevant when Type is change_password.
 	TargetStep string `json:"target_step,omitempty"`
+
+	// show_until_amr_constraints_fulfilled is relevant when Type is authenticate.
+	ShowUntilAMRConstraintsFulfilled *bool `json:"show_until_amr_constraints_fulfilled,omitempty"`
```
louischan-oursky, line 1091: I think we should not make this a public API. Reading the later commits, this new field works as a internal flag to make a step block until AMR constraints are fulfilled.

I forgot what the decision was about how to implement the blocking hooks in authflow.

pkg/lib/authenticationflow/declarative/utils_amr.go
```diff
@@ -0,0 +1,100 @@
...
+		amr = append(amr, model.AMRMFA)
+	}
+
+	amr = slice.Deduplicate(amr)
+	sort.Strings(amr)
+
+	return
+}
+
+// TODO(tung): Write a unit test for this
```
louischan-oursky: Remember to fix this.

pkg/api/event/blocking/authentication_post_identified.go
```diff
@@ -26,7 +26,13 @@ func (e *AuthenticationPostIdentifiedBlockingEventPayload) BlockingEventType() e
 }
 
 func (e *AuthenticationPostIdentifiedBlockingEventPayload) UserID() string {
-	return e.Identity.UserID
+	if e.Identity != nil {
+		return e.Identity.UserID
+	}
+	if e.Authentication.User != nil {
+		return e.Authentication.User.ID
```
louischan-oursky: In what case reading `e.Authentication.User.ID` is wrong? Shouldn't we always read it?

pkg/lib/authenticationflow/declarative/intent_signup_flow_step_create_authenticator.go
```diff
@@ -45,6 +46,7 @@ type IntentSignupFlowStepCreateAuthenticator struct {
 
 	Options                          []CreateAuthenticatorOptionInternal `json:"options,omitempty"`
 	ShowUntilAMRConstraintsFulfilled bool                                `json:"show_until_amr_constraints_fulfilled,omitempty"`
+	CannotBeSkipped                  bool                                `json:"cannot_be_skipped,omitempty"`
```
louischan-oursky, line 49: Originally, under what situation this step could be skipped?

pkg/api/event/hook_response.go
```diff
@@ -75,10 +78,21 @@ type HookResponse struct {
...
+var supportedAMRContraints = []string{model.AMRMFA, model.AMROTP, model.AMRPWD, model.AMRSMS}
+
 type Constraints struct {
 	AMR []string `json:"amr,omitempty"`
 }
 
+func (c *Constraints) Validate() error {
+	for _, amr := range c.AMR {
+		if !slice.ContainsString(supportedAMRContraints, amr) {
+			return fmt.Errorf("unsupported amr constraint %s", amr)
```
louischan-oursky: This error will become server error, as it is constructed with fmt.Errorf.

If you want to warn the developer, you should use our JSON schema validation. The error audience is the developer so in production end user should not see it.

pkg/lib/config/authentication_flow.go
tung2744: Exposing this flag allowing user to control the timing the "adaptive mfa" step appears in the custom authflow.

If the authflow does not contain such a step, the constraints are still checked at the end of authflow (IntentLoginFlowEnsureConstraintsFulfilled), but the ux may be weird because this step appears after steps like change_password.

You can also read https://github.com/authgear/authgear-server/pull/5222 to see how it can be used.

pkg/lib/authenticationflow/declarative/utils_amr.go
tung2744: Added the test in https://github.com/authgear/authgear-server/pull/5218/commits/114007b5bb9977eb73b87217dbfc2358d24a4f46

pkg/api/event/blocking/authentication_post_identified.go
tung2744: Right, if something was identified, then a user should always exist. Lets always use e.Authentication.User.ID

pkg/lib/authenticationflow/declarative/intent_signup_flow_step_create_authenticator.go
tung2744: In signup flow, create authenticator can always be skipped if `Options` is empty.

louischan-oursky: > In signup flow, create authenticator can always be skipped if Options is empty.

In non-customized authflow, how could that happen?

tung2744: Reading the code, there are two cases (No matter in customized or non-customized flow)
1. `oneOf` is empty.
2. The option points to a `target_step`, which was skipped (Probably happens in account linking). In this case, the option will be excluded. If all options were excluded, Options is empty.

tung2744:
@louischan-oursky Updated, thanks! Also added a commit to refactor some code so we don't need to mess with the post_identified payload in different node.

pkg/lib/config/authentication_flow.go
louischan-oursky: Opened this issue https://linear.app/authgear/issue/DEV-2820/discuss-how-do-we-implement-show-until-amr-constraints-fulfilled

louischan-oursky (approved):

PR #5221, opened 2025-06-09
Slightly change the meaning of `password` in createUser mutation

louischan-oursky:
ref DEV-2793


tung2744:
make check-tidy failed

pkg/admin/graphql/user_mutation.go
```diff
@@ -69,15 +70,23 @@ var _ = registerMutationField(
 				return nil, err
 			}
 
-			password, _ := input["password"].(string)
-			generatePassword := password == ""
```
tung2744, line 73: Just curious, why did it worked in latte before?

PR #5222, opened 2025-06-10
Update generated flow for amr constraints

tung2744:
ref DEV-2771
requires https://github.com/authgear/authgear-server/pull/5218

### DIFF
--- a/pkg/lib/authenticationflow/declarative/generate_config_login_flow_test.go
+++ b/pkg/lib/authenticationflow/declarative/generate_config_login_flow_test.go
@@ -14,29 +14,31 @@ import (
 
 func TestGenerateLoginFlowConfig(t *testing.T) {
 	Convey("GenerateLoginFlowConfig", t, func() {
-		test := func(cfgStr string, expected string) {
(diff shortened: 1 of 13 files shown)


pkg/lib/authenticationflow/declarative/generate_config_login_flow_test.go
```diff
@@ -68,12 +70,18 @@ steps:
           - authentication: device_token
         - type: change_password
           target_step: authenticate_primary_email
+- name: authenticate_amr_constraints
+  type: authenticate
```
tung2744, line 74: Change it to

```
- type: trigger_event_pre_authenticated
- type: enforce_constraints_amr
- type: change_password
- type: check_account_status
- type: terminate_other_sessions
-  
```

tung2744: https://linear.app/authgear/issue/DEV-2820/implicitly-trigger-pre-authenticated-after-all-authentication-steps

We will not add these steps.

Will handle the change in another pr.

PR #5225, opened 2025-06-11
Fix https://github.com/advisories/GHSA-v6h2-p8h4-qcjw

tung2744:
Fix https://github.com/advisories/GHSA-v6h2-p8h4-qcjw


tung2744:
brace-expansion is not patched yet, so npm audit fix does not fix it. Closing the pr first.

PR #5227, opened 2025-06-11
WeChat Login bug fixes

louischan-oursky:
ref DEV-2809


tung2744:
Is this still WIP?

louischan-oursky:
Not yet. I added two commits to handle OAuthError from Wechat. But I haven't tested whether other OAuth providers have the same problem.

louischan-oursky:
Should be ready for review

PR #5228, opened 2025-06-12
Replace doc links

louischan-oursky:
ref DEV-2796


tung2744:
There was a prettier error

tung2744 (approved):

PR #5234, opened 2025-06-16
Support bot_protection in hook response

tung2744:
ref DEV-2774 DEV-2838

PR #5237, opened 2025-06-19
Trigger pre_authenticated & enforce constraints after all authentication steps

tung2744:
ref DEV-2820
requires https://github.com/authgear/authgear-server/pull/5234


pkg/lib/authenticationflow/declarative/intent_login_flow_steps.go
```diff
@@ -50,66 +56,78 @@ func (i *IntentLoginFlowSteps) ReactTo(ctx context.Context, deps *authflow.Depen
...
-			JSONPointer:   authflow.JSONPointerForStep(i.JSONPointer, nextStepIndex),
+			JSONPointer:   authflow.JSONPointerForStep(i.JSONPointer, i.NextStepIndex),
 			UserID:        i.userID(flows),
-		}), nil
+		})
+		break
 	}
 
-	return nil, authflow.ErrIncompatibleInput
+	i.NextStepIndex = i.NextStepIndex + 1
```
louischan-oursky, line 128: Is ReactTo allowed to modify the input reactor? Can you point me to the relevant code in `accept.go`?

tung2744: You can read this:
https://github.com/authgear/authgear-server/blob/main/pkg/lib/authenticationflow/service.go#L466

The flow will be persisted after `accept`. So any changes made to the flow in accept will be persisted.

louischan-oursky: But we also have `ErrUpdateNode`, which is used to signify the intention of updating the input reactor.

tung2744: `ErrUpdateNode` was used to signify the an update of the latest node, without adding a new node. In our case, new node were also added so it is not used.

louischan-oursky: The behavior you mentioned is a implementation details of the authflow engine. It is not documented in the `InputReactor` interface

tung2744: Where is the documentation? Or you mean to add comments to it?

louischan-oursky: > Where is the documentation?

https://github.com/authgear/authgear-server/blob/main/pkg/lib/authenticationflow/input.go#L19

> Or you mean to add comments to it?

Actually not. What I am thinking is whether we should allow this. If we allow this, then how. In the very beginning, the nodes in authflow are supposed to be immutable. That's why we have introduced ErrUpdateNode as an escape hatch for the immutability rule.

tung2744:
```
// ErrUpdateNode means the input is reacted to, but instead of producing a new node to be appended,
// the returned node should replace the node.
// This error can only be returned by ReactTo.
var ErrUpdateNode = errors.New("update node")
```

The description of ErrUpdateNode actually mentions it is only for the case "no new node to be appended". And actually, once we returned this error, the flow returns and accept stops. We do not want to stop accept in our case.

louischan-oursky: Anyway, let me open another issue to discuss this problem. https://linear.app/authgear/issue/DEV-2846/how-do-we-allow-update-of-input-reactor

I will merge this PR first.

louischan-oursky (approved):

pkg/lib/authenticationflow/declarative/intent_login_flow_steps.go
tung2744: > Actually not. What I am thinking is whether we should allow this. If we allow this, then how. In the very beginning, the nodes in authflow are supposed to be immutable. That's why we have introduced ErrUpdateNode as an escape hatch for the immutability rule.

Actually it was not immutable. When you call NewXXNode or NewXXIntent inside ReactTo, some of them mutate the nodes. For example, intents in account linking sometimes mutate their ancestors (To skip creating authenticators, or update the owner user id.)

And in our recent pr, we also mutate nodes in DelayedFunction (for blocking hooks), so I did not assume the nodes are immutable.

PR #5247, opened 2025-06-24
Fix deno types cannot be loaded in monaco

tung2744:
Screenshot 2025-06-24 at 5 30 40 PM

### DIFF
--- a/portal/src/CodeEditor.tsx
+++ b/portal/src/CodeEditor.tsx
@@ -57,7 +57,13 @@ const CodeEditor: React.VFC<CodeEditorProps> = function CodeEditor(props) {
 
       // We only resolve first-level imports to avoid performance issue
       for (const pkg of parseDependencies(value)) {
-        const path = `inmemory://model/${pkg}`;
+        const encoder = new TextEncoder();
+        const uint8Array = encoder.encode(pkg);
+        const buffer = await window.crypto.subtle.digest("SHA-1", uint8Array);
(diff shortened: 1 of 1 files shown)


portal/src/CodeEditor.tsx
```diff
@@ -57,7 +57,13 @@ const CodeEditor: React.VFC<CodeEditorProps> = function CodeEditor(props) {
...
       // We only resolve first-level imports to avoid performance issue
       for (const pkg of parseDependencies(value)) {
-        const path = `inmemory://model/${pkg}`;
+        const encoder = new TextEncoder();
+        const uint8Array = encoder.encode(pkg);
+        const buffer = await window.crypto.subtle.digest("SHA-1", uint8Array);
+        const fileName = Array.from(new Uint8Array(buffer), (byte) =>
+          byte.toString(16).padStart(2, "0")
+        ).join("");
+        const path = `inmemory://model/${fileName}.d.ts`;
```
louischan-oursky, line 74: What is the cause of this problem?

tung2744: I don't know the actual cause, but I guess it is because of some special symbol in the path causing bugs in monaco.

louischan-oursky: I added a few commits for further improvements. The weird error message is now gone.

louischan-oursky: Amended the last commit to improve correctness

louischan-oursky (approved):

PR #5249, opened 2025-06-26
[WIP] Specify M2M authentication and authorization

louischan-oursky:
ref DEV-2848


docs/specs/m2m.md
```diff
@@ -0,0 +1,513 @@
...
+{
+  "iss": "https://dev-fnc259uk.auth0.com/",
+  "sub": "auth0|683d622f1a6ac540d22d1409",
+  "aud": [
+    "myapi",
+    "https://dev-fnc259uk.auth0.com/userinfo"
+  ],
+  "iat": 1750926072,
+  "exp": 1751012472,
+  "scope": "openid profile email",
```
tung2744: If we use scope as permission, another problem I can think of is in authorization code flow, do we need to write all scopes in the authorization request?

louischan-oursky: Actually, we will not.

I amended the document a lot, and it now includes `sub/scope`. In order for `sub/scope` to appear, `resource` is required. So without `resource`, the Authorization Code Grant behaves exactly the same. No `scope` nor `sub/scope` are modified or added.

If `resource` is additionally provided to the Authorization endpoint, the resulting access token will have 2 `aud`, one for the `resource`, another one for Authgear. `sub/scope` will contain all scope the `sub` (User) has, with respect to `aud`

In Client Credentials Grant, `resource` is required so `sub/scope` is always present, and its value is just equal to `scope`.

docs/specs/m2m.md
```diff
@@ -0,0 +1,884 @@
...
+
+> Access tokens are credentials used to access protected resources.
+> An access token is a string representing an authorization issued to the client.
+
+In `RFC6749 section-3.3`, `scope` is
+
+> The authorization and token endpoints allow the client to specify the scope of the access request using the "scope" request parameter.
+> In turn, the authorization server uses the "scope" response parameter to inform the client of the scope of the access token issued.
+
+In my own interpretation, **`scope` means the access of `client_id` to `aud` acting on behalf of `sub`**. The access of `sub` **SHOULD NOT** be included in `scope`.
```
tung2744: Just a comment for discussion, maybe not related to the m2m design:
If I understand it correctly, no matter scope of APIs (`orders:read`) or scopes of oidc (`profile email`), are not access of `sub`. They are [access of this access token](https://www.rfc-editor.org/rfc/rfc8693#name-scope-scopes-claim) to the resource (resource of `sub` in `aud`s). And client with `client_id` requested this token ([ref](https://www.rfc-editor.org/rfc/rfc8693#name-client_id-client-identifier)), therefore the meaning is "the client is allowed to access these resources". So they are access of the client, and there are no access of `sub` in the token.
For example, `orders:read` might means `client_id` can read orders of `sub` in `https://onlinestore.myapp.com`. (But not `sub` can read his own order in `https://onlinestore.myapp.com`)
 Similarly, `profile` means `client_id` can read profile of `sub` in `https://auth.myapp.com`. (But not `sub` can read his own profile in `https://auth.myapp.com`)

louischan-oursky: I added a commit https://github.com/authgear/authgear-server/pull/5249/commits/252e8291c6f307edc5e2952fdf5967bc5a57640e to remove the `subject_scope` thing, in flavor of Token exchange. See if it looks good to you.

docs/specs/m2m.md
```diff
@@ -320,97 +320,99 @@ In essence, Auth0 offers
...
+With this constraint, the `access_token` of `grant_type=authorization_code` always look like:
 
-The pain-points are
+```json
+{
+  "iss": "https://auth.myapp.com",
+  "sub": "johndoe",
+  "aud": ["https://auth.myapp.com"],
+  "client_id": "mobileapp",
+  "roles": ["onlinestore:admin"],
```
tung2744: The current key of roles is `https://authgear.com/claims/user/roles`.

louischan-oursky: Thanks. Fixed.

docs/specs/m2m.md
```diff
@@ -0,0 +1,1139 @@
...
+
+-- A sibling table of _auth_oauth_authorization, that takes resource_id into account.
+CREATE TABLE _auth_oauth_authorization_resource (
+  id text PRIMARY KEY,
+  app_id text NOT NULL,
+  client_id text NOT NULL,
+  user_id text NOT NULL REFERENCES _auth_user(id),
+  resource_id text NOT NULL REFERENCES _auth_resource(id),
+  scope_id text NOT NULL REFERENCES _auth_resource_scope(id)
+)
```
tung2744, line 924: @louischan-oursky 

I suggest to support a "{resource}/*" scope for each resource. For example, `https://onlinestore.myapp.com/*`.
It is because we probably don't want explicit assignment of scopes to each client. Say when you create an application, you want to allow it to access `https://onlinestore.myapp.com` with any permissions.
Then later, if you added new scopes to `https://onlinestore.myapp.com`, you probably won't remember to update the client you've created. Having a `*` scope in the client can automatically associate new scopes to the client.

tung2744:
I think the spec did not mention what will happen if the project did not turn on jwt access token.
The token returned in client_credential flow is still an `access_token`, so I think it should still respect the jwt token config?

docs/specs/m2m.md
louischan-oursky: Added a commit to document our offline discussion on this suggestion.

docs/specs/m2m.md
```diff
@@ -0,0 +1,1152 @@
...
+
+```sql
+CREATE TABLE _auth_resource (
+  id text PRIMARY KEY,
+  app_id text NOT NULL,
+  created_at timestamp without time zone NOT NULL,
+  updated_at timestamp without time zone NOT NULL,
+  uri text NOT NULL,
+  name text,
+  data jsonb
```
tung2744: @louischan-oursky What is this data for?

louischan-oursky: Maybe I should call this `metadata`. It is a column to store metadata about the entity, like the recent `_auth_user.metadata`

louischan-oursky:
Added a commit to clarify the format of the access token.

docs/specs/m2m.md
tung2744: I see, lets call it metadata so it looks the same as _auth_user

docs/specs/m2m.md
```diff
@@ -0,0 +1,1152 @@
...
+  deleteResource(input: DeleteResourceInput!): DeleteResourcePayload!
+
+  createScope(input: CreateScopeInput!): CreateScopePayload!
+  updateScope(input: UpdateScopeInput!): UpdateScopePayload!
+  deleteScope(input: DeleteScopeInput!): DeleteScopePayload!
+
+  addResourceToClientID(input: AddResourceToClientIDInput!): AddResourceToClientIDPayload!
+  removeResourceFromClientID(input: RemoveResourceFromClientIDInput!): RemoveResourceFromClientIDPayload!
+  addScopeToClientID(input: AddScopeToClientIDInput!): AddScopeToClientIDPayload!
+  removeScopeFromClientID(input: RemoveScopeFromClientIDInput!): RemoveScopeFromClientIDPayload!
```
tung2744: Will it be easier to use if there is a mutation to setScopesOfClientID?

If we only have add / remove apis, the user always have to compute the diff.

louischan-oursky: Someone has to do the diff anyway. In your proposal, the diff should be done in the server side. How did we deal with this in roles and groups?

docs/specs/m2m.md
```diff
@@ -0,0 +1,1152 @@
...
+  resourceID: ID!
+  clientID: String!
+}
+
+type RemoveResourceFromClientIDPayload {
+  resource: Resource!
+}
+
+input AddScopeToClientIDInput {
+  scopeID: ID!
```
tung2744: I think we should use `scope` to reference the scope here instead of ID, like how we use a key to reference a role. Therefore I suggest the input should be:

```
input AddScopeToClientIDInput {
  resourceUri: String!
  scope: String!
  clientID: String!
}
```

Actually, I think our api should always use `uri` to identify a resource, and `scope` string + `resourceUri` to identify a scope. This is easier for the user to understand the api.

louischan-oursky: Agree. But `resourceUri` should be `resourceURI`

louischan-oursky: Please update the spec to reflect this proposal.

docs/specs/m2m.md
tung2744: In my proposal, `setScopesOfClientID` directly set the scope to what the user provided. So no diff is needed. It overrides the current state by user input. (Or say, the user did the diff, and he decide the final state)

In roles and groups we do not have a similar `setRolesOfGroup` mutation. But I think there are a bit differences. Scopes are always directly under a resource. i.e. The user is always setting a subset of all scopes inside a resource to a client, but roles and groups exist independently and is many to many relation.

docs/specs/m2m.md
tung2744: Pushed a new commit for this change

docs/specs/m2m.md
louischan-oursky: addScopesToClientID
removeScopesFromClientID
replaceScopesOfClientID


All these operations should be idempotent. However, if the given scope does not exist, it should be an error.

tung2744: Updated the spec for renaming & new replace mutation

docs/specs/m2m.md
```diff
@@ -0,0 +1,1204 @@
...
+input DeleteResourceInput {
+  id: ID!
+}
+
+type DeleteResourcePayload {
+  ok: Boolean
+}
+
+input CreateScopeInput {
+  resourceID: ID!
```
tung2744: Maybe we should use `resourceURI` here too?

docs/specs/m2m.md
```diff
@@ -1066,22 +1067,32 @@ type RemoveResourceFromClientIDPayload {
 
 input AddScopeToClientIDInput {
```
louischan-oursky: The type name should also be updated. It should be `AddScopes...`

Apply this change to other input types as well.

docs/specs/m2m.md
```diff
@@ -0,0 +1,1204 @@
...
+  uri: String!
+  name: String
+}
+
+type CreateResourcePayload {
+  resource: Resource!
+}
+
+input UpdateResourceInput {
+  id: ID!
```
louischan-oursky: Change this to take resourceURI. This implies resourceURI cannot be updated.

docs/specs/m2m.md
```diff
@@ -0,0 +1,1204 @@
...
+  """The new name"""
+  name: String
+}
+
+type UpdateResourcePayload {
+  resource: Resource!
+}
+
+input DeleteResourceInput {
+  id: ID!
```
louischan-oursky: Change this to take resourceURI

docs/specs/m2m.md
louischan-oursky: Yes please.

docs/specs/m2m.md
```diff
@@ -0,0 +1,1204 @@
...
+  scope: String!
+  description: String
+}
+
+type CreateScopePayload {
+  scope: Scope!
+}
+
+input UpdateScopeInput {
+  id: ID!
```
louischan-oursky: Change this to take resourceURI and scope

docs/specs/m2m.md
```diff
@@ -0,0 +1,1204 @@
...
+  """The new description"""
+  description: String
+}
+
+type UpdateScopePayload {
+  scope: Scope!
+}
+
+input DeleteScopeInput {
+  id: ID!
```
louischan-oursky: Change this to take resourceURI and scope

tung2744:
Added a new commit to remove all `id` in update and delete mutations

louischan-oursky:
@tung2744 When I review the implementation, I noticed a detail I missed in the specification. In order to support Admin API authorization with M2M in the future, we are going to introduce a builtin resource representing the Admin API. Since the domain of Authgear can change at anytime, these resources are not resource in the `_auth_resource` table. Instead, the builtin **MUST** always take precedence over the `_auth_resource`. The `_auth_resource` with a conflicting URI is shadowed by the builtins.

I think we need to specify this behavior.

docs/specs/m2m.md
```diff
@@ -0,0 +1,1206 @@
...
+| `iss`       | Same value                       |                                                                |                                                                                      |
+| `sub`       | Different value                  | The user ID (`johndoe`)                                        | The string concatenation of `client_id_` and the `client_id` (`client_id_inventory`) |
+| `aud`       | The target audience of the token | The authentication server plus `resource`                      | The `resource` parameter (`https://onlinestore.myapp.com`)                           |
+| `client_id` | Different value                  | The client acting on behalf of an end-user (`mobileapp`)       | The client acting on behalf of itself (`inventory`)                                  |
+| `scope`     | Same meaning                     | The access of the `client_id` to `aud` on `sub`. See Remarks 1 | The access of the `client_id` to `aud` on behalf of `sub`. See Remarks 2             |
```
tung2744, line 275: I just found that our user access token does not have scope

louischan-oursky: What do you mean?

tung2744: https://github.com/authgear/authgear-server/blob/main/pkg/lib/oauth/token_encoding.go#L62-L149

We do not have scope in the access token now

louischan-oursky: https://datatracker.ietf.org/doc/html/rfc9068#section-2.2.3

It is not required to have it. In this specification, I changed the behavior to always include it.

docs/specs/m2m.md
```diff
@@ -0,0 +1,1206 @@
...
+- `scope`: (Optional) The `scope` parameter defined by `RFC6749 section-4.4`. It is optional. When it is present, validation is done to ensure it is valid with respect to `resource`.
+
+#### The successful response
+
+```json
+{
+  "access_token": "A_JWT",
+  "token_type": "bearer",
+  "expires_in": 3600,
+  "scope": "scopea+scopeb+scopec"
```
tung2744: I think this should be space instead of +

louischan-oursky: Yes indeed.

docs/specs/m2m.md
```diff
@@ -0,0 +1,1206 @@
...
+  scopes(clientID: String, searchKeyword: String, after: String, before: string, first: Int, last: Int): ScopeConnection
+  """The list of client IDs associated with this Resource."""
+  clientIDs: [String!]!
+}
+
+type Scope implements Entity & Node {
+  id: ID!
+  createdAt: DateTime!
+  updatedAt: DateTime!
+  resource: Resource!
```
tung2744: I think this should be resourceID instead of resource, or else we can construct an looped query. And actually, in the api you always reference a scope together with an resource (Usually under a resource), so it is not very useful to have resource here.

louischan-oursky: Let's change it to `resourceID: ID!` then.

tung2744:
Added 2 commits to:
- Change resource to resourceID in Scope
- Add a section to mention behavior of built-in resources

docs/specs/m2m.md
```diff
@@ -0,0 +1,1218 @@
...
+  addScopesToClientID(input: AddScopesToClientIDInput!): AddScopesToClientIDPayload!
+  removeScopesFromClientID(input: RemoveScopesFromClientIDInput!): RemoveScopesFromClientIDPayload!
+  replaceScopesOfClientID(input: ReplaceScopesOfClientIDInput!): ReplaceScopesOfClientIDPayload!
+}
+
+type Resource implements Entity & Node {
+  id: ID!
+  createdAt: DateTime!
+  updatedAt: DateTime!
+  uri: String!
```
tung2744: Change all `uri` to `resourceURI`.

docs/specs/m2m.md
```diff
@@ -0,0 +1,1218 @@
...
+  name: String
+  """If clientID is null, then all scopes of this Resource is returned."""
+  """If clientID is specified, then only scopes that are associated with clientID is returned."""
+  """If searchKeyword is non-null, a prefix search of scope is performed."""
+  """If both clientID and searchKeyword are specified, they are AND-ed."""
+  scopes(clientID: String, searchKeyword: String, after: String, before: string, first: Int, last: Int): ScopeConnection
```
tung2744, line 971: @louischan-oursky I am thinking should we remove the pagination parameters from this query, and always return all scopes under a resource, because looking at the portal UI design makes me thinks that actually we want to get all scopes in most of the cases.

docs/specs/m2m.md
tung2744: Added a commit to rename

docs/specs/m2m.md
louischan-oursky: In the Resource Details page, there is actually pagination. I think you are referring to the page where scopes are listed and searched in order to be added to the client.

Maybe we can do this.

1. Place a hard limit on the total number of scopes a resource can have. Say 100. Then we always fetch `first: 100` in that page.
2. Keep the pagination API unchanged. So that we can life the limit AND change the UI design to handle more than 1000 scopes in the future.

@fungc-io

louischan-oursky: @tung2744 @fungc-io Per our discussion in the product meeting, let's place a hard limit of the number of resources and scopes.

1. A project can only have at most 10000 resources. `apierrors.Forbidden.WithReason("ResourceQuotaExceeded")`
2. A resource can only have at most 10000 scopes. `apierrors.Forbidden.WithReason("ScopeQuotaExceeded")`

tung2744: Per discussion we will not set a limit on max resource / scope. Instead we will set a maximum number item in the query.

tung2744:
I think all comments have been addressed, let me merge it.

tung2744 (approved):

PR #5251, opened 2025-06-26
Group rate limits by name

tung2744:
ref DEV-2772


pkg/lib/authn/otp/kind_forgot_password.go
```diff
@@ -57,35 +46,35 @@ func (k kindForgotPassword) ValidPeriod() time.Duration {
...
-		ip,
-	)
-}
-
-func (k kindForgotPassword) RateLimitTriggerPerUser(userID string) ratelimit.BucketSpec {
-	return ratelimit.BucketSpecDisabled
+func (k kindForgotPassword) RateLimitTrigger(
+	featureConfig *config.FeatureConfig,
+	envConfig *config.RateLimitsEnvironmentConfig,
+	ip string, userID string,
```
louischan-oursky, line 52: The original code does not take userID.

pkg/lib/authn/otp/kind_forgot_password.go
```diff
@@ -97,26 +86,26 @@ func (k kindForgotPassword) RateLimitTriggerCooldown(target string) ratelimit.Bu
...
-		ip,
-	)
-}
-
-func (k kindForgotPassword) RateLimitValidatePerUserPerIP(userID string, ip string) ratelimit.BucketSpec {
-	return ratelimit.BucketSpecDisabled
+func (k kindForgotPassword) RateLimitValidate(
+	featureConfig *config.FeatureConfig,
+	envConfig *config.RateLimitsEnvironmentConfig,
+	ip string, userID string,
```
louischan-oursky, line 92: The original code does not take userID.

pkg/lib/ratelimit/ratelimits.go
```diff
@@ -75,91 +75,111 @@ const (
...
+	get := func(c *config.RateLimitConfig, fallback *config.RateLimitConfig, featureConfig *config.RateLimitConfig) *config.RateLimitConfig {
+		effectiveConfig := c
+
+		if !effectiveConfig.IsEnabled() && fallback != nil {
+			effectiveConfig = fallback
 		}
-		if fallback != nil && fallback.IsEnabled() {
-			return fallback
+
+		if featureConfig != nil && featureConfig.Rate() < c.Rate() {
```
louischan-oursky: Should we check featureConfig is enabled instead of checking `!= nil`?

pkg/lib/ratelimit/ratelimits.go
```diff
@@ -183,6 +203,26 @@ func (n RateLimit) resolvePerUser(cfg *config.AppConfig) *config.RateLimitConfig
 	return nil
 }
 
+func (n RateLimit) resolvePerTarget(cfg *config.AppConfig, featureCfg *config.FeatureConfig) *config.RateLimitConfig {
+	get := func(c *config.RateLimitConfig, featureConfig *config.RateLimitConfig) *config.RateLimitConfig {
+		effectiveConfig := c
+		if featureConfig != nil && featureConfig.Rate() < c.Rate() {
```
louischan-oursky: Should we check featureConfig is enabled instead of checking `!= nil`?

pkg/lib/ratelimit/ratelimits.go
```diff
@@ -75,91 +75,111 @@ const (
...
 
-func (n RateLimit) resolvePerIP(cfg *config.AppConfig) *config.RateLimitConfig {
-	get := func(c *config.RateLimitConfig, fallback *config.RateLimitConfig) *config.RateLimitConfig {
-		if c != nil && c.IsEnabled() {
-			return c
+func (n RateLimit) resolvePerIP(cfg *config.AppConfig, featureCfg *config.FeatureConfig) *config.RateLimitConfig {
+	get := func(c *config.RateLimitConfig, fallback *config.RateLimitConfig, featureConfig *config.RateLimitConfig) *config.RateLimitConfig {
+		effectiveConfig := c
+
+		if !effectiveConfig.IsEnabled() && fallback != nil {
```
louischan-oursky: The meaning seems changed. In the original code, it was checking `config.enabled == nil`, if it it true, then `fallback` is used. Here, if the rate limit was explicitly disabled, the fallback is used. This means we cannot explicitly disable a rate limit **AND** not falling back to the `fallback`.

louischan-oursky (changes requested):
This is a big refactoring. Please add tests to ensure we do not break things. Specifically we need to test

1. The resolution of each rate limit resolves the intended spec.
2. The fallback mechanism should be the same as before.
3. The fallback-to-feature-config mechanism.
4. Extensive tests to ensure we do not make any typos.

I suggest in your test, you initialize a AppConfig struct and a FeatureConfig struct with each rate limit having a distinct value of `burst`, and the enabled flag being `nil`.

And then you have a lot of `Convey` to test each scenario. At the beginning of the function body, you mutate the structs to set up the test scenario, and assert the resolved spec match the distinct value.

```golang
func TestRateLimitResolution(t *testing.T) {
  Convey("TestRateLimitResolution", t, func() {
    appConfig := &config.AppConfig{}
    featureConfig := &config.FeatureConfig{}
    
    Convey("resolve blahblahblah without fallback", func() {
      
    })
    Convey("resolve blahblahblah with fallback", func() {
    })
    Convey("resolve blahblahblah with fallback to feature", func() {
    })
  })
}
```

pkg/lib/authn/otp/kind_forgot_password.go
tung2744: Some otp kind take userID, some does not. As all otp kind implement the same interface, I can only make it always take user ID, like originally, forgot password kind also implements RateLimitTriggerPerUser.

I think the responsibility of deciding whether a userID should participate in rate limit should be in the type `RateLimit`. If that rate limit does not support `perUser` or `perUserPerIP` rate limits, it will just ignore the userID and do not return related buckets. The user of the RateLimit should always provide all available information.

pkg/lib/ratelimit/ratelimits.go
tung2744: Checking `Rate` implicitly checked IsEnabled, because Rate will be infinity if it is not enabled. See the original logic: https://github.com/authgear/authgear-server/pull/5251/files#diff-a0734821e46d3730ae898ffffb7b590bce65c74b8dbc267ff8a2cac059b2e5e5L82

I added a nil check here because we can pass nil to featureConfig, meaning that there is no corresponding feature config for a rate limit. Currently, only messaging rate limits has feature config.

pkg/lib/ratelimit/ratelimits.go
tung2744: Same as https://github.com/authgear/authgear-server/pull/5251#discussion_r2174200305

pkg/lib/ratelimit/ratelimits.go
tung2744: You are right, let me fix this.

tung2744:
@louischan-oursky Added a few commits to:
1. Fix some bugs in RateLimit, including the one you mentioned and some other bugs discovered in tests.
2. Added a test file for all rate limits.

louischan-oursky:
Nice. The added tests did catch some bugs!

louischan-oursky (approved):

PR #5252, opened 2025-06-27
Support rate_limits in hook

tung2744:
ref DEV-2772
requires https://github.com/authgear/authgear-server/pull/5251

TODO:
- [ ] Update deno portal types: https://github.com/authgear/authgear_deno_hook/pull/12

### DIFF
--- a/pkg/lib/ratelimit/gcra.go
+++ b/pkg/lib/ratelimit/gcra.go
@@ -34,7 +34,7 @@ else
 	tat = tonumber(tat)
 end
 
-local increment = emission_interval * n
+local increment = math.ceil(emission_interval * n)
 local new_tat = math.max(tat, now_timestamp) + increment
 local dvt = emission_interval * tolerance
 
--- a/pkg/lib/ratelimit/gcra.go
+++ b/pkg/lib/ratelimit/gcra.go
@@ -55,7 +55,7 @@ type gcraResult struct {
 	TimeToAct    time.Time
 }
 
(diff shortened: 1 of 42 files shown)


pkg/lib/ratelimit/gcra.go
```diff
@@ -55,7 +55,7 @@ type gcraResult struct {
 	TimeToAct    time.Time
 }
 
-func gcra(ctx context.Context, conn redis.Redis_6_0_Cmdable, key string, period time.Duration, burst int, n int) (*gcraResult, error) {
+func gcra(ctx context.Context, conn redis.Redis_6_0_Cmdable, key string, period time.Duration, burst int, n float64) (*gcraResult, error) {
```
louischan-oursky, line 58: We cannot simply change it to float64.

The analysis is as follows:

If you look at the Lua script, we can see that

- `n` is used to derive `increment`.
- `increment` is used to derive `new_tat`.
- `new_tat` is used as an argument to Redis command `EXPIREAT`. And `EXPIREAT` takes integer only.

I think we need to math.round non-integer `n` into an integer. Do not modify the GCRA algorithm.

tung2744: I think `n` itself can be float64, but the timestamp must be integer. So I should ceil `increment`.

tung2744: Added https://github.com/authgear/authgear-server/pull/5252/commits/32fa241314d42ad9832fa46be7a63c49916877f8 to fix, thanks!

louischan-oursky (approved):

PR #5260, opened 2025-07-02
Improve error handling in request too large in portal

louischan-oursky:
ref DEV-2845
ref DEV-2834

https://github.com/user-attachments/assets/abc1767f-2bc0-4d87-9fc0-d88fc06d5a6b

SCR-20250702-qvxi

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -32,6 +32,8 @@ http {
         '' close;
     }
 
+    client_max_body_size 10M;
+
     server {
         server_name _;
         listen 8010;
@@ -254,7 +256,6 @@ http {
         }
 
         location /_images {
-            client_max_body_size 10M;
             proxy_pass http://host.docker.internal:3004;
             proxy_http_version 1.1;
             proxy_set_header Host $http_host;
--- a/portal/src/components/common/BaseImagePicker.tsx
+++ b/portal/src/components/common/BaseImagePicker.tsx
@@ -1,6 +1,16 @@
-import React, { useCallback, useRef } from "react";
+import React, {
+  useCallback,
+  useRef,
+  useState,
+  useMemo,
+  useContext,
+} from "react";
 import cn from "classnames";
+import { Dialog, DialogFooter } from "@fluentui/react";
+import { Context } from "@oursky/react-messageformat";
 import { dataURIToBase64EncodedData } from "../../util/uri";
+import PrimaryButton from "../../PrimaryButton";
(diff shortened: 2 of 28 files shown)


nginx.conf
```diff
@@ -32,6 +32,8 @@ http {
         '' close;
     }
 
+    client_max_body_size 10M;
```
tung2744, line 35: I think this config does not affect saas production, do we have the same body size limit there?

portal/src/components/common/BaseImagePicker.tsx
```diff
@@ -91,6 +144,26 @@ const BaseImagePicker: React.VFC<BaseImagePickerProps> =
...
+              )}
+            />
+            <DefaultButton
+              onClick={onDialogDismiss}
+              text={renderToString(
+                "BaseImagePicker.size-dialog.button-cancel-label"
+              )}
+            />
+          </DialogFooter>
+        </Dialog>
```
tung2744, line 166: There is a `ImageInput` component in the v2 ui, do we need to update it so the ux is the same? (But it is not fluent style 🤔 )

nginx.conf
louischan-oursky: Yes, this setting is unrelated to SaaS. But we do have the same setting. See https://github.com/authgear/helm-charts/blob/authgear-10.23.0/authgear/values.yaml#L51

portal/src/components/common/BaseImagePicker.tsx
louischan-oursky: `ImageInput` has handled size limit already. But the size limit prop is not mandatory.

I am going to rename the prop to `sizeLimitInBytes` and make it required. So all these image picker components have the same prop dealing with size limit.

However, but the error handling is not trivial to port. `ImageInput` is v2 ui and it is backed by `FormField`, an abstraction that have native support for field-level error.

louischan-oursky: Actually I am thinking if size limit error should also be handled internally by `ImageInput`, like in `BaseImagePicker`. Now that both components require `sizeLimitInBytes`, if the end-user selects an image larger than the limit, the component should have handled it internally. Otherwise, we have to repeat the same error handling everywhere `ImageInput` is used, like what we did in `<Step3>`.

Another problem is, in the storybook, currently if I select a large file, there is no visual feedback, because `<ImageInput>` does not handle size limit itself.

PR #5261, opened 2025-07-03
Allow creating password in portal even the user has no password

tung2744:
ref DEV-2811

https://github.com/user-attachments/assets/4b5e5e8b-20ed-4e9d-9801-52067b5ace32

### DIFF
--- a/portal/src/locale-data/en.json
+++ b/portal/src/locale-data/en.json
@@ -450,12 +450,6 @@
   "Add2FAScreen.error.duplicated-email": "2FA email already exist",
 
   "ChangePasswordScreen.title": "Change Password",
-  "ChangePasswordScreen.email": "Email",
-  "ChangePasswordScreen.password-creation-type.manual": "Enter a password",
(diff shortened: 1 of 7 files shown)


portal/src/locale-data/en.json
```diff
@@ -395,6 +395,8 @@
   "UserDetails.account-security.primary": "Primary Authenticators",
   "UserDetails.account-security.secondary": "2-Factor Authenticators",
   "UserDetails.account-security.secondary.add": "Add 2FA",
+  "UserDetails.account-security.primary.password.empty": "No Primary Authenticator yet.",
```
louischan-oursky: Shouldn't this message be `No Primary Password yet.`?

louischan-oursky (changes requested):

portal/src/locale-data/en.json
tung2744: It looks like this so I think it should be primary authenticator. Let me update the message id.

Screenshot 2025-07-03 at 1 05 00 PM

PR #5266, opened 2025-07-03
Handle email blocked by plus sign error

tung2744:
ref DEV-2566

### DIFF
--- a/pkg/lib/authn/identity/loginid/type_checkers.go
+++ b/pkg/lib/authn/identity/loginid/type_checkers.go
@@ -184,7 +184,7 @@ func (c *EmailChecker) Validate(ctx context.Context, validationCtx *validation.C
 
 	if *c.Config.BlockPlusSign {
 		if strings.Contains(local, "+") {
-			validationCtx.EmitError("format", map[string]interface{}{"format": "email"})
+			validationCtx.EmitError("format", map[string]interface{}{"format": "email", "reason": "BlockPlusSign"})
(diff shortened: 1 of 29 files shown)


pkg/lib/authn/identity/loginid/type_checkers.go
```diff
@@ -184,7 +184,7 @@ func (c *EmailChecker) Validate(ctx context.Context, validationCtx *validation.C
 
 	if *c.Config.BlockPlusSign {
 		if strings.Contains(local, "+") {
-			validationCtx.EmitError("format", map[string]interface{}{"format": "email"})
+			validationCtx.EmitError("format", map[string]interface{}{"format": "email", "reason": "BlockPlusSign"})
```
louischan-oursky: If we follow the convention of this file, this emission is actually wrong. The correct one should be

```golang
validationCtx.EmitError("blocked", map[string]interface{}{"reason": "EmailPlusSign"})
```

We should use `blocked` to sigify the email address is syntactically correct, but otherwise rejected (blocked) due to a reason


---

On the other hand, I think the error message is not very helpful, it just says invalid but did not tell it is the plus sign that makes it blocked. Is that intention to confuse the end-user? @fungc-io

louischan-oursky (changes requested):

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #5268, opened 2025-07-04
Support skip passkey upselling

tung2744:
ref DEV-2480

https://github.com/user-attachments/assets/476ea805-cc49-477a-9214-ed43de3d68a9

Screenshot 2025-07-04 at 8 04 49 PM


cmd/authgear/cmd/cmddatabase/migrations/authgear/20250704145030-add_user_metadata.sql
```diff
@@ -0,0 +1,7 @@
+-- +migrate Up
+
+ALTER TABLE _auth_user ADD COLUMN metadata jsonb NOT NULL DEFAULT '{}'::jsonb;
```
louischan-oursky: Make it nullable. Handle null in application side.

pkg/lib/config/ui.go
```diff
@@ -34,12 +35,13 @@ var _ = Schema.Add("UIConfig", `
...
 
 type UIConfig struct {
-	SignupLoginFlowEnabled bool              `json:"signup_login_flow_enabled,omitempty"`
-	PhoneInput             *PhoneInputConfig `json:"phone_input,omitempty"`
-	DarkThemeDisabled      bool              `json:"dark_theme_disabled,omitempty"`
-	LightThemeDisabled     bool              `json:"light_theme_disabled,omitempty"`
-	WatermarkDisabled      bool              `json:"watermark_disabled,omitempty"`
-	DirectAccessDisabled   bool              `json:"direct_access_disabled,omitempty"`
+	SignupLoginFlowEnabled   bool              `json:"signup_login_flow_enabled,omitempty"`
+	AllowOptOutPasskeyUpsell bool              `json:"allow_opt_out_passkey_upsell,omitempty"`
```
louischan-oursky: Maybe name it `passkey_upselling_opt_out_enabled`

pkg/lib/authn/user/store.go
```diff
@@ -44,7 +44,7 @@ type Store struct {
 
 var _ store = &Store{}
 
-const keySkipPasskeyCreation = "skip_passkey_creation"
+const keyOptOutPasskeyUpsell = "opt_out_passkey_upsell"
```
louischan-oursky: `opt_out_passkey_upselling`

cmd/authgear/cmd/cmddatabase/migrations/authgear/20250704145030-add_user_metadata.sql
tung2744: Why? We can give it a default empty object in the migration so we don't need to handle null in application logic.

louischan-oursky: Generally, adding a nullable column is safer. See https://stackoverflow.com/a/49924935

tung2744: I see. Ok let me change it to no default.

pkg/lib/config/ui.go
louischan-oursky: Maybe also change all upsell (verb) to upselling (noun) wherever appropriate.

tung2744:
- Changed metadata to nullable
- Renamed configs and related words

louischan-oursky (approved):

PR #5272, opened 2025-07-08
Upgrade eslint and stylelint

louischan-oursky:
ref DEV-2864

### DIFF
--- a/authui/src/authflowv2/turboForm.ts
+++ b/authui/src/authflowv2/turboForm.ts
@@ -31,7 +31,7 @@ export class TurboFormController extends Controller {
     e.preventDefault();
     const form = e.currentTarget;
     if (!(form instanceof HTMLFormElement)) {
-      return;
+      throw new Error("expected event.currentTarget to be a HTMLFormElement");
     }
 
     if (form.querySelector('[data-turbo="false"]')) {
@@ -47,6 +47,8 @@ export class TurboFormController extends Controller {
     formData.forEach((value, name) => {
       if (typeof value === "string") {
         params.set(name, value);
+      } else {
+        console.error("ignoring non-string value: ", name);
       }
     });
     let loadingButton: HTMLButtonElement | null = null;


authui/src/authflowv2/customSelect.ts
```diff
@@ -212,11 +212,11 @@ export class CustomSelectController extends Controller {
...
-    if (!item) return;
-
-    const value = item.dataset.value;
-    this._selectValue(value);
+    const item = event.target;
+    if (item instanceof HTMLLIElement) {
+      const value = item.dataset.value;
+      this._selectValue(value);
+    }
   }
```
tung2744: This might be incorrect: It could be other type of element inside the li element if we use event.target. (Like the span inside it)
- Maybe we should use currentTarget, and checking instanceof HTMLLIElement will be redundant because it is always the element we attached the listener to.

louischan-oursky: You mean the original code is correct? I just rewrite it in a way that preserve the original meaning.

authui/src/authflowv2/date.ts
```diff
@@ -208,12 +208,14 @@ export class FormatInputDateController extends Controller {
...
     }
 
-    const dateSpan = this.element as HTMLSpanElement;
-    const rfc3339 = this.dateValue;
-    if (rfc3339 !== "") {
-      const jsDate = new Date(rfc3339);
-      if (!isNaN(jsDate.getTime())) {
-        dateSpan.textContent = new Intl.DateTimeFormat().format(jsDate);
+    const dateSpan = this.element;
+    if (dateSpan instanceof HTMLSpanElement) {
```
tung2744: I feel like we should not do `instanceof HTMLXXElement` checks on places which originally are just type assertions, because this actually changes the runtime logic.

Like here we assumed this.element must be a span, but I am not sure if this assumption is correct.
The original code works even it is not a span because it is not checked during runtime, but now the code will not work if the actual element is not span.

louischan-oursky: I checked this particular case before I made the changes. The only place this controller is used is really on a `<span>`. I agree that this controller can be used on other HTML element as well, as long as `.textContent` is available.

But the type assertion is really unsafe. Checking `instanceof` is actually preferred as it is a runtime checking, and tsc is smart enough to know its effect.

authui/src/authflowv2/customSelect.ts
tung2744: I've left a new comment about the type assertion error:
https://github.com/authgear/authgear-server/pull/5272#discussion_r2192113520

For this code, I feel like the original code should use currentTarget.

authui/src/authflowv2/date.ts
tung2744: > But the type assertion is really unsafe

I agree. However checking `instanceof` might break some existing code because it will just do nothing if the element type does not match, and it would be difficult for us to find out which code now just do nothing because of unmatched type.

I think for most of the places:
1. We should just check `instanceof Element `, unless you really know the exact element type (Like in change event of input element).
2. We should use currentTarget so the type of the element is known.
3. We should throw error instead of no-op if the type does not match. (Or at least console.error)

louischan-oursky: Added a number of commits to

1. Remove the instanceof checking if `Element` is sufficient.
2. Use event.currentTarget wherever appropriate.
3. Throw or log error.

authui/src/authflowv2/customSelect.ts
louischan-oursky: Updated to use currentTarget.

tung2744 (approved):

PR #5273, opened 2025-07-08
Fix portal audit logs are not using the correct audit db

tung2744:
ref DEV-2671

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -205,13 +205,6 @@ use flake
    go run ./cmd/portal internal configsource create ./var
    ```
 
-3. Add domain
-
-   ```sh
-   go run ./cmd/portal internal domain create-default --default-domain-suffix ".localhost"
-   go run ./cmd/portal internal domain create-custom accounts --apex-domain="accounts.portal.localhost" --domain="accounts.portal.localhost"
-   ```
-
 ## Set up MinIO
 
 ```sh
@@ -530,13 +523,15 @@ Note the `data` column should be obtained by running the following command:
 go run ./cmd/portal internal configsource pack -i ./var
 ```
 
-3. Create rows in `_portal_domain`:
+3. Create domains for `accounts`:
 
-```sql
-INSERT INTO "public"."_portal_domain"("id","app_id","created_at","domain","apex_domain","verification_nonce","is_custom")
-VALUES
-(E'accounts.portal.localhost',E'accounts',NOW(),E'accounts.portal.localhost',E'accounts.portal.localhost',E'-',TRUE),
-(E'localhost',E'accounts',NOW(),E'localhost',E'localhost',E'-',FALSE);
+```sh
+# This allows portal to access the admin api with accounts.localhost
+go run ./cmd/portal internal domain create-default --default-domain-suffix ".localhost"
+# This allows using accounts.portal.localhost:3100
+go run ./cmd/portal internal domain create-custom accounts --apex-domain="accounts.portal.localhost" --domain="accounts.portal.localhost"
+# This allows using localhost:3100
+go run ./cmd/portal internal domain create-custom accounts --apex-domain="localhost" --domain="localhost"
 ```
 
 4. Create a free plan
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,7 @@
     * [Configure Authgear](#configure-authgear)
     * [Start with the profile ldap](#start-with-the-profile-ldap)
   * [Switching between sessionType=refresh\_token and sessionType=cookie](#switching-between-sessiontyperefresh_token-and-sessiontypecookie)
+  * [Switch to Database config source](#switch-to-database-config-source)
 * [Storybooks](#storybooks)
 
 # Contributing guide
@@ -199,11 +200,6 @@ use flake
    go run ./cmd/portal database migrate up
    ```
 
-3. Create the project in the database
-
-   ```sh
-   go run ./cmd/portal internal configsource create ./var
-   ```
 
 ## Set up MinIO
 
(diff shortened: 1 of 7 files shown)


hack/custom-resources/authgear.secrets.yaml
```diff
@@ -3,10 +3,6 @@ secrets:
     data:
       database_schema: app
       database_url: 'postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable'
-  - data:
```
louischan-oursky, line 6: Actually I never use this file. How are you using it?

CONTRIBUTING.md
```diff
@@ -508,6 +508,47 @@ In case you need to switch to sessionType=cookie, you
 - Use `AUTHGEAR_WEB_SDK_SESSION_TYPE=cookie` in your .env
 - Access the portal at port 8001 or 8011
 
+## Switch to Database config source
+
+1. In your `.env` set these values:
+
+```
+CONFIG_SOURCE_TYPE=database
+CUSTOM_RESOURCE_DIRECTORY=./hack/custom-resources
```
louischan-oursky: I suggest we remove the directory `./hack`. It is a legacy directory that was used to host a local k8s API server.

CONTRIBUTING.md
```diff
@@ -508,6 +508,47 @@ In case you need to switch to sessionType=cookie, you
...
+```sql
+INSERT INTO "public"."_portal_config_source"("id","app_id","created_at","updated_at","data","plan_name")
+VALUES
+('00000000-0000-0000-0000-000000000000','accounts',NOW(),NOW(),'{{ YOUR_DATA }}','free');
+```
+
+Note the `data` column should be obtained by running the following command:
+```
+go run ./cmd/portal internal configsource pack -i ./var
+```
```
louischan-oursky, line 513: I think Step 2 is actually line 205: `go run ./cmd/portal internal configsource create ./var`

CONTRIBUTING.md
```diff
@@ -508,6 +508,47 @@ In case you need to switch to sessionType=cookie, you
...
+
+3. Create rows in `_portal_domain`:
+
+```sql
+INSERT INTO "public"."_portal_domain"("id","app_id","created_at","domain","apex_domain","verification_nonce","is_custom")
+VALUES
+(E'accounts.portal.localhost',E'accounts',NOW(),E'accounts.portal.localhost',E'accounts.portal.localhost',E'-',TRUE),
+(E'localhost',E'accounts',NOW(),E'localhost',E'localhost',E'-',FALSE);
+```
```
louischan-oursky, line 524: I think Step 3 is actually line 211-212

CONTRIBUTING.md
```diff
@@ -508,6 +508,47 @@ In case you need to switch to sessionType=cookie, you
...
+
+4. Create a free plan
```
louischan-oursky, line 526: This step should be moved to `## Set up the database`

CONTRIBUTING.md
```diff
@@ -508,6 +508,47 @@ In case you need to switch to sessionType=cookie, you
...
+
+```sql
+INSERT INTO "public"."_portal_plan"("id","name","feature_config","created_at","updated_at")
+VALUES
+(E'free',E'free',E'{}',NOW(),NOW());
+```
+
+Restart your server, then it should be running with database config source.
```
louischan-oursky, line 536: So, switching to database configsource should be as simple as changing the few environment variables. No additional steps are needed.

hack/custom-resources/authgear.secrets.yaml
tung2744: This file set secrets for all projects. Removing the audit db secret here so that new projects will not have audit db secret.

CONTRIBUTING.md
tung2744: I think these steps are not necessary when you use local config source, so maybe I move them to the new "Switch to Database config source" section?

CONTRIBUTING.md
tung2744: When using local configsource, free plan is actually not needed in the db. So I prefer keep this in the new section.

louischan-oursky: Let's keep the new section `Switch to Database config source`. This new section includes

1. Tell you that you need to change some environment variables.
2. Ask you to run `go run ./cmd/portal internal configsource create ./var`
3. Ask you create the plan if you haven't done it before.

tung2744:
I found that analytic.ChartService is actually using the global audit db credential, so I cannot simply remove it.
Let me try to create a new type for analytic audit database credential.

tung2744:
I think the problem is analytic data (Which always store in global audit database), is mixed with audit log data (Which should be stored in app specific db). So we need to change `CountCollector` so that it collect counts from app specific audit database, but write the result counts into the global audit db. Let me create a new issue to separate these two database.

CONTRIBUTING.md
tung2744: Looks like this file is still needed, or the new created projects will missing some required secrets (db, redis)

louischan-oursky: I just re-use `./var` for this purpose. Any reason why you need `./hack`?

tung2744: Oh I see. I think using `./var` will also work, the different will be the new projects always share all secrets with the `accounts` app.

tung2744:
Created issue https://linear.app/authgear/issue/DEV-2867/separate-analytic-database-and-audit-database to follow up the separation of analytic / audit db.

CONTRIBUTING.md
tung2744: Updated to use `./var` in the guide.

louischan-oursky (approved):

PR #5277, opened 2025-07-10
Add CRUD mutations and query for resource and scope

tung2744:
ref DEV-2868
ref DEV-2869


pkg/api/model/resource.go
```diff
@@ -0,0 +1,14 @@
+package model
+
+type Resource struct {
+	Meta Meta    `json:"meta"`
+	URI  string  `json:"uri"`
+	Name *string `json:"name,omitempty"`
```
louischan-oursky: We should use `omitzero` in new code.

pkg/api/model/resource.go
```diff
@@ -0,0 +1,14 @@
...
+}
+
+type Scope struct {
+	Meta        Meta    `json:"meta"`
+	ResourceID  string  `json:"resource_id"`
+	Scope       string  `json:"scope"`
+	Description *string `json:"description,omitempty"`
```
louischan-oursky: `omitzero`

pkg/lib/resourcescope/resource.go
```diff
@@ -0,0 +1,38 @@
+package resourcescope
+
+import (
+	"time"
+
+	"github.com/authgear/authgear-server/pkg/api/model"
+)
+
+type NewResourceOptions struct {
+	URI  string
```
louischan-oursky, line 12: Who is the intended user of this struct? If this struct appears in service method, then it is OK to use `string` here. The service is responsible to perform the validation and return error.

pkg/lib/resourcescope/store_resource.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/google/uuid"
+)
+
+func (s *Store) NewResource(options *NewResourceOptions) *Resource {
+	now := s.Clock.NowUTC()
+	return &Resource{
+		ID:        uuid.NewString(),
+		CreatedAt: now,
+		UpdatedAt: now,
+		URI:       options.URI,
```
louischan-oursky, line 22: Remember to perform the validation.

pkg/lib/resourcescope/store_resource.go
```diff
@@ -0,0 +1,162 @@
...
+	result, err := s.SQLExecutor.ExecWith(ctx, q)
+	if err != nil {
+		var pqError *pq.Error
+		if errors.As(err, &pqError) {
+			if pqError.Code == "23505" {
+				return ErrResourceDuplicateURI
+			}
+		}
+		return err
+	}
```
louischan-oursky, line 83: 23505 should not happen here.

pkg/lib/resourcescope/store_scope.go
```diff
@@ -0,0 +1,170 @@
...
+	result, err := s.SQLExecutor.ExecWith(ctx, q)
+	if err != nil {
+		var pqError *pq.Error
+		if errors.As(err, &pqError) {
+			if pqError.Code == "23505" {
+				return ErrScopeDuplicate
+			}
+		}
+		return err
+	}
```
louischan-oursky, line 90: 23505 should not happen here.

pkg/admin/graphql/resource.go
```diff
@@ -41,4 +42,34 @@ var nodeResource = node(
...
+			"clientID":      &graphql.ArgumentConfig{Type: graphql.String},
+			"searchKeyword": &graphql.ArgumentConfig{Type: graphql.String},
+		}),
+		Resolve: func(p graphql.ResolveParams) (interface{}, error) {
+			source := p.Source.(*model.Resource)
+			ctx := p.Context
+			gqlCtx := GQLContext(ctx)
+
+			// TODO(tung): Support client ID & searchKeyword filter
+			scopes, err := gqlCtx.ResourceScopeFacade.ListScopes(ctx, source.ID)
```
louischan-oursky, line 60: When I see this, I realized we did not define an order when we return a list of scopes. We should return a list of scopes ordered by the created at ASC. Please add the ORDER BY clause to the SQL query.

Add an index `(app_id, resource_id, created_at ASC)` on `_auth_resource_scope`

pkg/admin/facade/resoucescope.go
```diff
@@ -62,3 +65,25 @@ func (f *ResourceScopeFacade) GetScope(ctx context.Context, id string) (*model.S
 func (f *ResourceScopeFacade) ListScopes(ctx context.Context, resourceID string) ([]*model.Scope, error) {
 	return f.ResourceScopeQueries.ListScopes(ctx, resourceID)
 }
+
+func (f *ResourceScopeFacade) ListResources(ctx context.Context, options *resourcescope.ListResourcesOptions, pageArgs graphqlutil.PageArgs) ([]model.PageItemRef, *graphqlutil.PageResult, error) {
```
louischan-oursky, line 69: Again, I forgot to specify the order of the list. The list should be ordered by created_at ASC. Add an index.

louischan-oursky (changes requested):

pkg/lib/resourcescope/resource.go
tung2744: It appears in service method. But I indeed forgot the validation.

pkg/admin/graphql/resource.go
tung2744: I've made a change to this method in later pr to support pagination & search & filter. The scopes will be ordered by `scope ASC`
https://github.com/authgear/authgear-server/pull/5278/files#diff-0788f8780a7bbd02a9db89539c8e45bdba8eaa1912950600a5069ad9e8bd56b5R160-R185

pkg/admin/facade/resoucescope.go
tung2744: I've added ordering in https://github.com/authgear/authgear-server/pull/5278/files#diff-ef4ddd48e9637013f387a224ca40b2a7f31c0ef6cedda79eeb351c23739d7b51R179 by `uri`.

pkg/lib/resourcescope/store_scope.go
tung2744: I've done a bit refactoring for this logic in the next pr, so I will remove it there.

pkg/admin/graphql/resource.go
tung2744: I did not change the order in this pr, but I still added the index so we can change it any time.

tung2744:
Updated to add validations and add index

pkg/lib/resourcescope/resource_test.go
```diff
@@ -0,0 +1,38 @@
...
+			So(ValidateResourceURI(ctx, "invalid"), ShouldBeError, `invalid value:
+<root>: format
+  map[error:invalid scheme:  format:x_resource_uri]`)
+		})
+		Convey("https scheme should be valid", func() {
+			So(ValidateResourceURI(ctx, "https://host"), ShouldBeNil)
+		})
+		Convey("https scheme with path should be valid", func() {
+			So(ValidateResourceURI(ctx, "https://host/path"), ShouldBeNil)
+		})
```
louischan-oursky, line 36: Add test for these cases

1. It should disallow URI with query
2. It should disallow URI with fragment

pkg/lib/resourcescope/scope.go
```diff
@@ -39,3 +43,23 @@ func (s *Scope) ToModel() *model.Scope {
...
+	}
+
+	for _, blacklisted := range blacklist {
+		if blacklisted == scope {
+			validationCtx.EmitError("blocked", map[string]interface{}{"reason": "ReservedScope", "scope": scope})
+		}
+	}
+
+	return validationCtx.Error("invalid scope")
+}
```
louischan-oursky, line 65: Actually it should be possible to define this validation as a JSON schema (at least the regex part). See pkg/lib/rolesgroups/key.go for an example.

For the blocklist, I think we can use `"not"` and `"enum"`.

The reason to use JSON schema to that we can produce informative error message out-of-the-box.

tung2744: I did not use jsonschema because we cannot tell why the scope is rejected if we use one json schema. (blacklist or invalid format). We can do it with two json schema but I think it is unnecessary. We also need dynamic schema here if we want to update the schema automatically by oauth.AllowedScopes. So I decided not to use it.

pkg/lib/resourcescope/resource_test.go
tung2744: Added a commit for this https://github.com/authgear/authgear-server/pull/5277/commits/bc183fd0b78253c930f565a02fd22f2a33fdfacf, thanks!

pkg/lib/resourcescope/scope.go
tung2744: Thinking again, validations should not happens in command level because we cannot tell the user which field in the input was incorrect. Let me try to do the validation in graphql handler, and create a new type for scope and resource uri to ensure values are already correct when calling the create method in command.

Let me do this refactoring in later prs to avoid conflicts.

louischan-oursky: You can write a single jsonschema, for example,

```golang
{
  "type": "string",
  "minLength": 1,
  "maxLength": 100,
  "pattern": "^  the regex here  $",
  "not": {
     "enum": ["not", "allowed", "these"]
  }
}
```

With the help of SchemeBuilder and `func init`, you can build a static json schema instance that does the validation.

tung2744: Let me do it in https://github.com/authgear/authgear-server/pull/5278#issuecomment-3082886201

pkg/lib/resourcescope/resource_test.go
louischan-oursky: I forgot to mention that we should also forbid userinfo in the URL. But let's do that in a later PR.

tung2744: But the userinfo endpoint can be changing because the public endpoint can be changing.
Maybe we should forbid it in token endpoint instead of creation.

louischan-oursky: No, I meant the userinfo (`user:password`) in `https://user:password@host/path`

tung2744: Oh I see.
However, I am thinking about maybe we should block userinfo endpoint (Or any uri prefixed by the current public endpoint) as resource uri?

louischan-oursky: Actually, we should block all domains (both verified and pending) as resource URI

pkg/lib/resourcescope/scope.go
tung2744: The error message of using `not` is like
```
{
  "location": "/scope",
  "kind": "not"
}
```
Which seems unclear what was blocked.

PR #5278, opened 2025-07-11
Add resource-client and scope-client mutations

tung2744:
ref DEV-2870
requires https://github.com/authgear/authgear-server/pull/5277

### DIFF
--- a/portal/src/graphql/adminapi/globalTypes.generated.ts
+++ b/portal/src/graphql/adminapi/globalTypes.generated.ts
@@ -456,7 +456,7 @@ export type CreateResourceInput = {
   /** The optional name of the resource. */
   name?: InputMaybe<Scalars['String']['input']>;
   /** The URI of the resource. */
-  uri: Scalars['String']['input'];
+  resourceURI: Scalars['String']['input'];
 };
 
 export type CreateResourcePayload = {
@@ -1400,12 +1400,12 @@ export type Resource = Entity & Node & {
   id: Scalars['ID']['output'];
   /** The optional name of the resource. */
   name?: Maybe<Scalars['String']['output']>;
+  /** The URI of the resource. */
+  resourceURI: Scalars['String']['output'];
   /** The list of scopes for this resource. */
   scopes?: Maybe<ScopeConnection>;
   /** The update time of entity */
   updatedAt: Scalars['DateTime']['output'];
-  /** The URI of the resource. */
-  uri: Scalars['String']['output'];
 };
 
 
--- a/portal/src/graphql/adminapi/schema.graphql
+++ b/portal/src/graphql/adminapi/schema.graphql
@@ -773,7 +773,7 @@ input CreateResourceInput {
   name: String
 
   """The URI of the resource."""
-  uri: String!
+  resourceURI: String!
 }
 
 """"""
@@ -1549,14 +1549,14 @@ type Resource implements Entity & Node {
   """The optional name of the resource."""
   name: String
 
+  """The URI of the resource."""
+  resourceURI: String!
+
   """The list of scopes for this resource."""
   scopes(after: String, before: String, clientID: String, first: Int, last: Int, searchKeyword: String): ScopeConnection
 
   """The update time of entity"""
   updatedAt: DateTime!
-
-  """The URI of the resource."""
-  uri: String!
 }
 
 """A connection to a list of items."""


tung2744:
TODO
- [x] Move validations to graphql handlers https://github.com/authgear/authgear-server/pull/5277#discussion_r2212482246

tung2744:
Added commits to:
- Remove duplicate key checks on update
- Use created_at to order resource. Note that I did not change the order of scopes because I think order by `scope` is easier to use.
- Added jsonschema to validate user input in admin api mutations level
- Introduced new types for uri and scopes to ensure caller has validated them

pkg/lib/resourcescope/commands.go
```diff
@@ -105,3 +105,118 @@ func (c *Commands) UpdateScope(ctx context.Context, options *UpdateScopeOptions)
...
+	for _, scopeStr := range scopes {
+		scope, err := c.Store.GetScopeByResourceIDAndScope(ctx, resource.ID, scopeStr)
+		if err != nil {
+			return nil, err
+		}
+		err = c.Store.AddScopeToClientID(ctx, resource.ID, scope.ID, clientID)
+		if err != nil {
+			return nil, err
+		}
+	}
```
louischan-oursky, line 163: There are N queries here. We should use 1 query to fetch all given scopes, and 1 query to insert them.

pkg/lib/resourcescope/commands.go
```diff
@@ -105,3 +105,118 @@ func (c *Commands) UpdateScope(ctx context.Context, options *UpdateScopeOptions)
...
+	for _, scopeStr := range scopes {
+		scope, err := c.Store.GetScopeByResourceIDAndScope(ctx, resource.ID, scopeStr)
+		if err != nil {
+			return nil, err
+		}
+		err = c.Store.RemoveScopeFromClientID(ctx, scope.ID, clientID)
+		if err != nil {
+			return nil, err
+		}
+	}
```
louischan-oursky, line 200: Same again, 1 query to fetch scopes, 1 query to remove.

pkg/lib/resourcescope/commands.go
```diff
@@ -105,3 +105,118 @@ func (c *Commands) UpdateScope(ctx context.Context, options *UpdateScopeOptions)
...
+	// Remove extra scopes
+	for scopeStr, s := range currentSet {
+		if _, ok := desiredSet[scopeStr]; !ok {
+			err := c.Store.RemoveScopeFromClientID(ctx, s.ID, clientID)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
```
louischan-oursky, line 269: Compute the change set and use 1 query to add, 1 query to remove.

pkg/lib/resourcescope/store_scope.go
```diff
@@ -291,3 +291,41 @@ func (s *Store) scanScope(scanner db.Scanner) (*Scope, error) {
...
+	return err
+}
+
+func (s *Store) RemoveScopeFromClientID(ctx context.Context, scopeID, clientID string) error {
+	q := s.SQLBuilder.
+		Delete(s.SQLBuilder.TableName("_auth_client_resource_scope")).
+		Where("client_id = ? AND scope_id = ?", clientID, scopeID)
+	_, err := s.SQLExecutor.ExecWith(ctx, q)
+	return err
+}
```
louischan-oursky, line 330: Replace these 3 functions as they promote N-query pattern.

pkg/lib/resourcescope/store_resource.go
```diff
@@ -175,6 +175,20 @@ func (s *Store) GetManyResources(ctx context.Context, ids []string) ([]*Resource
...
+func (s *Store) GetClientResource(ctx context.Context, clientID, resourceID string) (*Resource, error) {
+	q := s.selectResourceQuery("r").
+		Join(s.SQLBuilder.TableName("_auth_client_resource"), "acr", "acr.resource_id = r.id").
+		Where("r.id = ? AND acr.client_id = ?", resourceID, clientID)
+	resources, err := s.queryResources(ctx, q)
+	if err != nil {
+		return nil, err
+	}
+	if len(resources) == 0 {
+		return nil, ErrResourceNotFound
```
louischan-oursky: We should introduce a new error for this, `ErrClientResourceNotFound`. It means the association is not found, not the resource itself.

pkg/lib/resourcescope/store_resource.go
```diff
@@ -173,7 +173,7 @@ type storeListResourceResult struct {
 
 func (s *Store) ListResources(ctx context.Context, options *ListResourcesOptions, pageArgs graphqlutil.PageArgs) (*storeListResourceResult, error) {
 	q := s.selectResourceQuery("r").
-		OrderBy("r.uri ASC")
+		OrderBy("r.created_at DESC")
```
louischan-oursky, line 176: ASC. The index you created is for ASC.

pkg/lib/resourcescope/formats.go
```diff
@@ -0,0 +1,74 @@
...
+		return fmt.Errorf("invalid scope-token: forbidden character")
+	}
+
+	return nil
+}
+
+func init() {
+	jsonschemaformat.DefaultChecker["x_resource_uri"] = FormatResourceURI{}
+	jsonschemaformat.DefaultChecker["x_scope_token"] = FormatScopeToken{}
+}
```
louischan-oursky: Put `func init` at the beginning of the file.

louischan-oursky (changes requested):

pkg/lib/resourcescope/store_resource.go
tung2744: https://www.postgresql.org/docs/current/indexes-ordering.html#:~:text=By%20default%2C%20B%2Dtree%20indexes,BY%20x%20ASC%20NULLS%20LAST%20).

b-tree indexes can be scanned backward, therefore it should work in both direction. Or we need two indexes in queries which allow sorting in both directions.

pkg/lib/resourcescope/store_resource.go
tung2744: There was an error `ErrResourceNotAssociatedWithClient` for this, let me use it instead.

pkg/lib/resourcescope/store_scope.go
tung2744: <del> Which 3 functions? I think only AddScopeToClientID and RemoveScopeFromClientID should be changed to batch query. </del>

Let me also rewrite `GetScopeByResourceIDAndScope` to take multiple scopes and return multiple scopes at once.

tung2744:

louischan-oursky (changes requested):
# Passing empty scopes will crash

SCR-20250721-lebg

```
unexpected error when executing GraphQL query  app=accounts context_cause=<context-err-is-nil> error=insert statements must have at least one set of values or select clause
```

# Inconsistent naming of uri

In most types, Resource URI is called `resourceURI` in the schema. But in CreateResourceInput, it is named `uri`. Please change it to `resourceURI`.
```

tung2744:
Added a commit to fix empty scope causing crash.
For the naming problem, the field inside `Resource` is named `uri`, therefore it is named `uri` inside the create mutation. The current difference between `uri` and `resourceURI` is `resourceURI` always reference an existing resource's uri, but `uri` is the `uri` of the new created resource. So I am not sure if we should change them.

louischan-oursky:
Actually I also want to rename the `uri` field inside `Resource` to `resourceURI`. So there is no ambiguity whenever we see `resourceURI`. It always means Resource URI. And there is no more `uri` that people have no idea what it is.

tung2744:
Ok, let me rename both to resourceURI then.

louischan-oursky:
Are you going to do the rename in this PR or in a later PR?

tung2744:
Let me do it in this PR

louischan-oursky:
I fixed up the 3 commits to exclude the file `pkg/util/rot13/rot13.go`

PR #5280, opened 2025-07-14
Migrate to slog

louischan-oursky:
ref DEV-2879


louischan-oursky:
I notice there is a problem in panic middleware. Let me have a look.

cmd/authgear/images/server/server.go
```diff
@@ -2,45 +2,49 @@ package server
...
 
 func (c *Controller) Start(ctx context.Context) {
+	logger := logger.GetLogger(ctx)
+
 	vipsutil.LibvipsInit()
 
 	cfg, err := LoadConfigFromEnv()
 	if err != nil {
-		golog.Fatalf("failed to load server config: %v", err)
+		logger.WithError(err).Error(ctx, "failed to load server config")
```
tung2744: If we log the error and panic, isn't the error logged twice?

louischan-oursky: Actually yes.

Or should we just `panic(fmt.Errorf("failed to load server config: %w"))`?

The panic is not `recover()` by anybody else. The program is going to exit very soon, so the SentryHandler has no time to kick in and do its job.

tung2744: Yes, I think panic should be enough?

louischan-oursky: Used ` rg --multiline  'WithError.*\n\s*panic\('` to find such patterns and corrected them.

pkg/util/slogutil/context_cause_handler.go
```diff
@@ -0,0 +1,73 @@
...
+			attrValue = "<context-err-is-nil>"
+		} else {
+			cause := context.Cause(ctx)
+			if cause == nil {
+				attrValue = "<context-cause-is-nil>"
+			} else {
+				attrValue = cause.Error()
+			}
+		}
+	}
```
tung2744: It seems ctxErr here was not logged? We only logged cause.Error().

louischan-oursky: Good question. I just copied the legacy implementation, and did not think about this.

When slog, we now always have access tot the context. We should use `ctx.Err()` to determine whether the ctx has ended or not. The key `context_cause` should only appear when the ctx has ended.

Let me update the test to reflect the intended behavior.

louischan-oursky: Added a commit to address this.

tung2744 (approved):

PR #5281, opened 2025-07-15
Housekeep refresh tokens in OfflineGrants

tung2744:
ref DEV-2882


pkg/lib/oauth/redis/store.go
```diff
@@ -303,6 +303,21 @@ func (s *Store) UpdateOfflineGrantLastAccess(ctx context.Context, grantID string
 	}
 
 	grant.AccessInfo.LastAccess = accessEvent
+	// If the refresh token actually used is known, update the individual access info
+	if refreshTokenHash != "" {
+		for i := range grant.RefreshTokens {
+			token := grant.RefreshTokens[i]
+			if token.MatchHash(refreshTokenHash) && token.AccessInfo != nil {
+				if token.AccessInfo == nil {
```
louischan-oursky: The condition here seems incorrect. The outer if has `&& token.AccessInfo != nil`, while the inner if has `token.AccessInfo == nil`.


BTW, did we test this function? The test should have caught this.

pkg/lib/oauth/grant_offline_service.go
```diff
@@ -201,3 +216,52 @@ func (s *OfflineGrantService) AddSAMLServiceProviderParticipant(
...
+	tokenHashesToRemove := []string{}
+	for idx, token := range grant.RefreshTokens {
+		if idx == 0 {
+			// Never remove the root token
+			continue
+		}
+		var lastAccess time.Time
+		// For backward compatibility
+		if token.AccessInfo == nil {
+			lastAccess = grant.AccessInfo.InitialAccess.Timestamp
```
louischan-oursky: This should be the creation time of the refresh token, instead of the initial access time of the grant. The initial access time of the grant should be very close to the creation time of the grant. In real world situation, this will invalidate all refresh tokens immediately.

pkg/lib/oauth/grant_offline.go
```diff
@@ -63,7 +63,8 @@ type OfflineGrant struct {
 	// ExpireAtForResolvedSession is a transient field that tells when the session will exire at, computed now.
 	// Note that ExpireAtForResolvedSession will keep changing if idle timeout is enabled.
 	// This is NOT supposed to be stored, hence it is json-ignored.
-	// TODO(tung): Revisit the logic of setting this field
+	// This field is always the expiry of the OfflineGrant, but not the expiry of any of the children refresh tokens
+	// Exipry of children refresh tokens could be shorter due to diffrent last access time
```
louischan-oursky: Typo in `Exipry`

louischan-oursky (changes requested):

pkg/lib/oauth/redis/store.go
tung2744: The store was mocked in the test because we do not run a redis, so it was not tested. Maybe I should try to keep logic in the store as simple as possible.

tung2744:
- Fixed the bug inside LastAccess update and added tests
- Fixed the bug inside housekeepOfflineGrant and fixed the test so that the bug can be catched

louischan-oursky (approved):

PR #5279, opened 2025-07-14
Add e2e tests to CRUD mutations of resource / scopes

tung2744:
ref DEV-2878

requires https://github.com/authgear/authgear-server/pull/5278

PR #5284, opened 2025-07-17
Implement client_credentials flow

tung2744:
ref DEV-2871
requires https://github.com/authgear/authgear-server/pull/5279

### DIFF
--- a/pkg/lib/oauth/grant_type.go
+++ b/pkg/lib/oauth/grant_type.go
@@ -7,6 +7,7 @@ import (
 const (
 	AuthorizationCodeGrantType = "authorization_code"
 	RefreshTokenGrantType      = "refresh_token"
+	ClientCredentialsGrantType = "client_credentials"
 	// nolint:gosec
 	TokenExchangeGrantType = "urn:ietf:params:oauth:grant-type:token-exchange"
 
@@ -51,6 +52,10 @@ func GetAllowedGrantTypes(c *config.OAuthClientConfig) []string {
 			allowed = append(allowed, g)
 			seen[g] = struct{}{}
 		}
+	}
+
+	if c.ApplicationType.IsClientCredentialsFlowAllowed() {
+		allowed = append(allowed, ClientCredentialsGrantType)
 	}
 
 	return allowed
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -1951,7 +1951,7 @@ func (h *TokenHandler) handleClientCredentials(
 	resource, err := h.ClientResourceScopeService.GetClientResourceByURI(ctx, client.ClientID, r.Resource())
 	if err != nil {
 		if errors.Is(err, resourcescope.ErrResourceNotFound) {
-			return nil, protocol.NewError("invalid_request", "resource not found")
+			return nil, protocol.NewError("invalid_request", "resource not found: "+r.Resource())
 		}
 		if errors.Is(err, resourcescope.ErrResourceNotAssociatedWithClient) {
 			return nil, protocol.NewError("invalid_request", "resource is not associated with the client")
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -35,6 +35,7 @@ import (
 	"github.com/authgear/authgear-server/pkg/lib/oauth/oidc"
 	"github.com/authgear/authgear-server/pkg/lib/oauth/protocol"
 	"github.com/authgear/authgear-server/pkg/lib/otelauthgear"
+	"github.com/authgear/authgear-server/pkg/lib/resourcescope"
 	"github.com/authgear/authgear-server/pkg/lib/session"
 	"github.com/authgear/authgear-server/pkg/lib/session/access"
 	"github.com/authgear/authgear-server/pkg/lib/session/idpsession"
@@ -183,6 +184,11 @@ func (s SimpleSessionLike) SessionType() session.Type {
 	return s.GrantSessionKind.SessionType()
 }
 
+type TokenHandlerClientResourceScopeService interface {
+	GetClientResourceByURI(ctx context.Context, clientID string, uri string) (*resourcescope.Resource, error)
+	GetClientResourceScopes(ctx context.Context, clientID string, resourceID string) ([]*resourcescope.Scope, error)
+}
+
 type TokenHandler struct {
 	AppID                  config.AppID
 	AppDomains             config.AppDomains
@@ -200,6 +206,7 @@ type TokenHandler struct {
 	AppSessionTokens                TokenHandlerAppSessionTokenStore
 	OfflineGrantService             TokenHandlerOfflineGrantService
 	PreAuthenticatedURLTokenService PreAuthenticatedURLTokenService
+	ClientResourceScopeService      TokenHandlerClientResourceScopeService
 	Graphs                          GraphService
 	IDTokenIssuer                   IDTokenIssuer
 	Clock                           clock.Clock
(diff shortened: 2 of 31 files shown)


pkg/lib/oauth/grant_type.go
```diff
@@ -53,5 +54,9 @@ func GetAllowedGrantTypes(c *config.OAuthClientConfig) []string {
 		}
 	}
 
+	if c.ApplicationType.IsClientCredentialsFlowAllowed() {
+		allowed = append(allowed, ClientCredentialsGrantType)
+	}
```
louischan-oursky, line 59: This function has test. Update the test.

pkg/lib/oauth/protocol/token.go
```diff
@@ -29,6 +29,7 @@ func (r TokenRequest) SubjectToken() string        { return r["subject_token"] }
 func (r TokenRequest) ActorTokenType() string      { return r["actor_token_type"] }
 func (r TokenRequest) ActorToken() string          { return r["actor_token"] }
 func (r TokenRequest) DeviceSecret() string        { return r["device_secret"] }
+func (r TokenRequest) Resource() string            { return r["resource"] }
```
louischan-oursky: I remember `resource` can appear more than once. That implies we need to change `type TokenRequest url.Values`

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -1934,8 +1934,14 @@ func (h *TokenHandler) handleClientCredentials(
 	if r.Resource() == "" {
 		return nil, protocol.NewError("invalid_request", "resource is required")
 	}
+	if strings.HasPrefix(r.Resource(), h.IDTokenIssuer.Iss()) {
+		return nil, protocol.NewError("invalid_request", "invalid resource uri")
```
louischan-oursky: In https://datatracker.ietf.org/doc/html/rfc8707#name-resource-parameter it introduces `invalid_target`. We should use that.

- When resource is empty, `invalid_target`, `resource is required`
- When resource matches our prefix, `invalid_target`, `resource URI must not be a prefix of iss`
- When resource is not found, `invalid_target`, `resource not found`
- When resource is not associated with the client, `invalid_target`, `client is not associated with the resource`

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -364,6 +364,13 @@ func (h *TokenHandler) validateRequest(r protocol.TokenRequest, client *config.O
 		// The validation logics can be different depends on requested_token_type
 		// Do the validation in methods for each requested_token_type
 		break
+	case oauth.ClientCredentialsGrantType:
+		if r.Resource() == "" {
+			return protocol.NewError("invalid_request", "resource is required")
+		}
+		if r.ClientSecret() == "" {
+			return protocol.NewError("invalid_request", "client secret is required")
```
louischan-oursky: According to https://datatracker.ietf.org/doc/html/rfc6749#section-5.2 when client secret is required but not provided, `invalid_client` `client secret is required` should be returned.

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -364,6 +364,13 @@ func (h *TokenHandler) validateRequest(r protocol.TokenRequest, client *config.O
 		// The validation logics can be different depends on requested_token_type
 		// Do the validation in methods for each requested_token_type
 		break
+	case oauth.ClientCredentialsGrantType:
+		if r.Resource() == "" {
+			return protocol.NewError("invalid_request", "resource is required")
```
louischan-oursky: `invalid_target` `resource is required`

pkg/lib/oauth/handler/handler_token_test.go
```diff
@@ -221,5 +226,175 @@ func TestTokenHandler(t *testing.T) {
...
+					"scope":         "admin",
+				}
+				ctx := context.Background()
+				resp := handle(ctx, req, r)
+
+				So(resp.Result().StatusCode, ShouldEqual, 400)
+				var body map[string]interface{}
+				err := json.Unmarshal(resp.Body.Bytes(), &body)
+				So(err, ShouldBeNil)
+				So(body["error"], ShouldEqual, "invalid_scope")
```
louischan-oursky, line 353: Please also assert the error description.

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -1949,6 +1949,9 @@ func (h *TokenHandler) handleClientCredentials(
 		if errors.Is(err, resourcescope.ErrResourceNotFound) {
 			return nil, protocol.NewError("invalid_request", "resource not found")
 		}
+		if errors.Is(err, resourcescope.ErrResourceNotAssociatedWithClient) {
+			return nil, protocol.NewError("invalid_request", "resource is not associated with the client")
```
louischan-oursky: Again, use `invalid_target`

pkg/lib/oauth/token_encoding.go
```diff
@@ -166,6 +166,8 @@ func (e *AccessTokenEncoding) EncodeClientAccessToken(ctx context.Context, optio
 
 	claims := jwt.New()
 
+	// sub
```
louischan-oursky: The comment is incorrect. It should be `jti`??

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -1947,7 +1947,7 @@ func (h *TokenHandler) handleClientCredentials(
 	resource, err := h.ClientResourceScopeService.GetClientResourceByURI(ctx, client.ClientID, r.Resource())
 	if err != nil {
 		if errors.Is(err, resourcescope.ErrResourceNotFound) {
-			return nil, protocol.NewError("invalid_request", "resource not found")
+			return nil, protocol.NewError("invalid_request", "resource not found: "+r.Resource())
```
louischan-oursky: No need to echo the resource back in the error description.

louischan-oursky (changes requested):
1. Please review the relevant specs to ensure we are returning the most appropriate OAuth error code. For example, I asked Claude `In OAuth, when client_secret is required but not provided, what OAuth error have to be returned according to the RFCs?` and he pointed to RFC 6749 Section 5.2

pkg/lib/oauth/handler/handler_token.go
tung2744: For the second case, lets use `resource URI must not be a prefixed by authgear endpoint` which is clearer. `iss` in this context could be difficult to understand.

pkg/lib/oauth/handler/handler_token.go
tung2744: @louischan-oursky However, our original implementation returns invalid_request in this case in other grant types. Not sure if we want to change the behavior now.

pkg/lib/oauth/handler/handler_token_test.go
tung2744: I can add an assertion, but I actually thinks the description is just a text for human to read so it is not expected to be used to compare in any code.

tung2744:
Updated, changes start from [bf46828](https://github.com/authgear/authgear-server/pull/5284/commits/bf468286247a448c2ddb3a65be30731155891f41). Thanks!

pkg/lib/oauth/handler/handler_token.go
louischan-oursky: Maybe just say `resource URI is reserved by Authgear`

pkg/lib/oauth/handler/handler_token.go
louischan-oursky: It should be OK, since this error indicates a programming error anyway, and it is not a recoverable error in a running deployment.

pkg/lib/oauth/handler/handler_token_test.go
louischan-oursky: Adding assertion on the error description makes sure the code return the expected error. It could catch bugs.

pkg/lib/oauth/handler/handler_token.go
tung2744: Added a commit to change the error to `invalid_client`

louischan-oursky (approved):

PR #5287, opened 2025-07-18
Add m2m client type

tung2744:
ref DEV-2888


e2e/var/authgear.yaml
```diff
@@ -153,6 +153,14 @@ oauth:
     - none
     x_application_type: confidential
     issue_jwt_access_token: true
+  - access_token_lifetime_seconds: 1800
+    client_id: e2em2mclient
+    client_name: M2M app
+    name: M2M app
+    redirect_uris:
+    - http://localhost
```
louischan-oursky: M2M client does not need to have any redirect URIs. Or our current JSON schema always require redirect URIs to be non-empty? If that is the case, we need to fix that.

tung2744: Added a new commit for this

PR #5302, opened 2025-07-28
Apply a more shorter lifetime for pre-authenticated URL refresh tokens

louischan-oursky:
ref DEV-2909

### DIFF
--- a/pkg/lib/oauth/grant_offline_service.go
+++ b/pkg/lib/oauth/grant_offline_service.go
@@ -162,11 +162,12 @@ func (s *OfflineGrantService) computeRefreshTokenExpiryWithClient(token expirabl
 }
 
 type CreateNewRefreshTokenOptions struct {
-	OfflineGrant    *OfflineGrant
-	ClientID        string
-	Scopes          []string
-	AuthorizationID string
-	DPoPJKT         string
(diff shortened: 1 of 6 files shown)


pkg/lib/oauth/grant_offline_service.go
```diff
@@ -266,10 +259,11 @@ func (s *OfflineGrantService) housekeepOfflineGrant(ctx context.Context, grant *
...
 
-		expiry := s.computeRefreshTokenExpiryWithClient(expirableRefreshToken{
-			CreatedAt:    token.CreatedAt,
-			LastAccessAt: lastAccess,
-		}, clientConfig)
+		// Assume other refresh tokens are pre-authentication URL refresh token.
+		// These refresh tokens are never returned.
+		// So their expiry should match the expiry of the access token.
+
+		expiry := token.CreatedAt.Add(clientConfig.AccessTokenLifetime.Duration())
```
tung2744: Not all refresh tokens can be cleaned up. There are two refresh tokens generated, only the one generated with `PreAuthenticatedURLTokenService.ExchangeForAccessToken` can be cleaned up using the access token lifetime. The others are generated using normal token exchange flow and the webapp will use that refresh token.

tung2744 (approved):

PR #5303, opened 2025-07-28
[Portal] Resource Applications

tung2744:
ref DEV-2897

https://github.com/user-attachments/assets/1a6deee1-fe81-4d08-9826-53baea1297e2


portal/src/screens/api-resources/APIResourceDetailsScreenApplicationsTab.tsx
```diff
@@ -1,5 +1,57 @@
...
+  resource: Resource;
+}): JSX.Element {
+  const { appID } = useParams() as { appID: string };
+  const appConfigQuery = useAppAndSecretConfigQuery(appID);
+
+  const isLoading = appConfigQuery.isLoading;
+
+  const applications = useMemo((): ApplicationListItem[] => {
+    return (
+      appConfigQuery.effectiveAppConfig?.oauth?.clients?.map(
```
louischan-oursky: We should only show applications whose `x_application_type=m2m`

.claude/commands/add_portal_admin_api_mutation.md
```diff
@@ -0,0 +1,6 @@
+Provided the graphql schema: @/portal/src/graphql/adminapi/schema.graphql
```
louischan-oursky: The path should be a relative path `@./portal/src/graphql/adminapi/schema.graphql`

.claude/commands/add_portal_admin_api_mutation.md
```diff
@@ -0,0 +1,6 @@
...
+
+Add a file under portal/src/graphql/adminapi/mutations for $ARGUMENTS.
```
louischan-oursky: `@./portal/src/graphql/adminapi/mutations`

.claude/commands/add_portal_admin_api_query.md
```diff
@@ -0,0 +1,7 @@
+Provided the graphql schema: @/portal/src/graphql/adminapi/schema.graphql
```
louischan-oursky: Use relative path (relative to project root)

.claude/commands/add_portal_admin_api_query.md
```diff
@@ -0,0 +1,7 @@
...
+
+Add a file under portal/src/graphql/adminapi/query for $ARGUMENTS.
```
louischan-oursky: Use relative path (relative to project root)

pkg/admin/graphql/resource.go
```diff
@@ -36,6 +36,21 @@ var nodeResource = node(
...
 			},
+			"clientIDs": &graphql.Field{
+				Type:        graphql.NewNonNull(graphql.NewList(graphql.NewNonNull(graphql.String))),
+				Description: "The list of client IDs associated with this Resource.",
+				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
+					resource := p.Source.(*model.Resource)
+					ctx := p.Context
+					gqlCtx := GQLContext(ctx)
+
+					clientIDs, err := gqlCtx.ResourceScopeFacade.ListClientIDsByResourceID(ctx, resource.ID)
```
louischan-oursky: When the query fetches a list of resources, and the query fetches the field `clientIDs`, then it is N+1 query pattern.

I notice we also have this pattern in User.identities, user.authenticators, etc.

For this particular instance of this pattern, maybe we introduce a loader to load the client IDs of resource ID.

louischan-oursky (commented):
When we toggle, the toggle should

1. Immediately switch to the intended state (That is, optimistic update)
2. Become disabled when the request is loading
3. Become enabled after the request resolves or rejects (That is, finally)
4. If the request rejects, roll back the optimistic update.

.claude/commands/add_portal_admin_api_mutation.md
tung2744: In cursor or cline, looks like starting with `@/` is correct (I think @ means project root). But both should be understandable to the ai.

tung2744:
> When we toggle, the toggle should

> Immediately switch to the intended state (That is, optimistic update)
Become disabled when the request is loading
Become enabled after the request resolves or rejects (That is, finally)
If the request rejects, roll back the optimistic update.

I think optimistic updates usually do not disable the input (Because that is a kind of loading state)? For the remaining it is the current behavior already.

portal/src/screens/api-resources/APIResourceDetailsScreenApplicationsTab.tsx
tung2744: Should be handled in later commits. Only m2m and confidential clients are shown.

.claude/commands/add_portal_admin_api_mutation.md
louischan-oursky: But Claude Code interprets `@~/` and `@/` like shell. See https://docs.anthropic.com/en/docs/claude-code/memory#claude-md-imports So this will confuse Claude Code.

tung2744: Do claude code understand it if we remove the `@`?
I am trying to write the commands in a way that any ai can understand and do not rely on a specific syntax.

louischan-oursky: Actually yes, if the path is explicitly relative (`./`)

louischan-oursky:
> I think optimistic updates usually do not disable the input (Because that is a kind of loading state)? For the remaining it is the current behavior already.

I asked Claude and Gemini, they both gave implementation of disabling the input when there is a pending request.

tung2744:
Gemini's answer:

> Generally, you do NOT disable the input in an optimistic update.
> ...
> When you might consider disabling/showing a "pending" state:
> A "Like" button: The button might immediately show "Liked" and the count increases (optimistic update). During the server request, you might disable further clicks on that specific like button to prevent multiple rapid submissions of the same action.

> Form Submission: After a user submits a form with an optimistic update (e.g., adding an item to a list), you might disable the submit button itself until the server response, but the newly added item would still appear in the list.

> Actions with dependencies: If a subsequent action heavily depends on the successful completion of the current action (e.g., editing a newly created item that needs a server-generated ID), you might delay enabling that specific edit functionality until the server confirms the creation.

In our case, it is similar the "Like" button, but it is not an action with dependencies. Actually I think it depends on whether the next possible action (Toggling off after toggling on immediately) will have problem if the previous action isn't completed.

tung2744:
Added 3 commits
- Update claude commands to use relative paths
- Use loader to load client ids
- Disable the toggle to prevent multiple request triggered at same time

louischan-oursky:
For like button, it makes sense to not disable the like button, as the only possible operations are like and unlike.

In this particular case, when toggle is on (that is, authorized), the developer can further click the Manage button and assign scopes. So it falls into the category of `Actions with dependencies`

louischan-oursky (approved):

PR #5306, opened 2025-07-29
Add user.reauthenticated and do not update last login in reauthentication

louischan-oursky:
ref DEV-2912
ref DEV-2913

SCR-20250729-qabb

### DIFF
--- a/docs/specs/event.md
+++ b/docs/specs/event.md
@@ -551,7 +551,7 @@ Occurs after user signed up or signed in.
 }
 ```
 
-#### user.authenticated
+#### user.reauthenticated
 
 Occurs after user reauthenticated.
 
--- a/docs/specs/event.md
+++ b/docs/specs/event.md
@@ -17,6 +17,7 @@
       - [user.created](#usercreated)
       - [user.profile.updated](#userprofileupdated)
       - [user.authenticated](#userauthenticated)
+      - [user.reauthenticated](#userreauthenticated)
       - [user.signed_out](#usersigned_out)
       - [user.session.terminated](#usersessionterminated)
       - [user.anonymous.promoted](#useranonymouspromoted)
@@ -482,6 +483,7 @@ Use this event to add custom fields to the JWT access token.
 - [user.created](#usercreated)
 - [user.profile.updated](#userprofileupdated)
 - [user.authenticated](#userauthenticated)
+- [user.reauthenticated](#userreauthenticated)
 - [user.signed_out](#usersigned-out)
 - [user.session.terminated](#usersessionterminated)
 - [user.anonymous.promoted](#useranonymouspromoted)
@@ -538,7 +540,20 @@ Occurs when the user profile is updated.
 
 #### user.authenticated
 
-Occurs after user logged in.
+Occurs after user signed up or signed in.
+
+```json5
+{
+  "payload": {
+    "user": { /* ... */ },
+    "session": { /* ... */ }
+  }
+}
+```
+
+#### user.authenticated
+
+Occurs after user reauthenticated.
 
 ```json5
 {
--- a/pkg/lib/authenticationflow/declarative/intent_signup_flow.go
+++ b/pkg/lib/authenticationflow/declarative/intent_signup_flow.go
@@ -129,6 +129,15 @@ func (i *IntentSignupFlow) GetEffects(ctx context.Context, deps *authflow.Depend
 		authflow.OnCommitEffect(func(ctx context.Context, deps *authflow.Dependencies) error {
 			userID, createUser := i.userID(flows)
 			if !createUser {
+				return nil
+			}
+
+			now := deps.Clock.NowUTC()
+			return deps.Users.UpdateLoginTime(ctx, userID, now)
+		}),
+		authflow.OnCommitEffect(func(ctx context.Context, deps *authflow.Dependencies) error {
+			userID, createUser := i.userID(flows)
+			if !createUser {
 				// The creation is skipped for some reason, such as entered account linking flow
 				return nil
 			}
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -1556,6 +1556,20 @@ func (h *TokenHandler) doIssueTokensForAuthorizationCode(
 						return nil, err
 					}
 				}
+
+				// Dispatch user.reauthenticated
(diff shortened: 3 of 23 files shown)


docs/specs/event.md
```diff
@@ -538,7 +540,20 @@ Occurs when the user profile is updated.
...
+```json5
+{
+  "payload": {
+    "user": { /* ... */ },
+    "session": { /* ... */ }
+  }
+}
+```
+
+#### user.authenticated
```
tung2744: Looks like the title is wrong

pkg/lib/authenticationflow/declarative/intent_signup_flow.go
```diff
@@ -126,6 +126,15 @@ func (i *IntentSignupFlow) GetEffects(ctx context.Context, deps *authflow.Depend
 			}
 			return nil
 		}),
+		authflow.OnCommitEffect(func(ctx context.Context, deps *authflow.Dependencies) error {
+			userID, createUser := i.userID(flows)
+			if !createUser {
+				return nil
+			}
```
tung2744, line 133: Even if the user is not created (I think it occurs in account linking), the login time should still be updated

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -1556,6 +1556,20 @@ func (h *TokenHandler) doIssueTokensForAuthorizationCode(
...
+						Meta: model.Meta{
+							ID: info.UserID,
+						},
+					},
+					Session:  *offlineGrant.ToAPIModel(),
+					AdminAPI: false,
+				})
+				if err != nil {
+					return nil, err
+				}
```
tung2744, line 1572: Why it was not done in reauth flow? (user.authenticated and user.created both are in the corresponding flow)

pkg/lib/oauth/redis/store.go
```diff
@@ -568,6 +569,15 @@ func (s *Store) DeleteOfflineGrant(ctx context.Context, grant *oauth.OfflineGran
...
 			return err
 		}
+
+		// TODO(slog): Before we have fine-grained logging, use WithSkipLogging().Error() to force logging to stderr.
+		logger.WithSkipLogging().Error(ctx,
+			"delete offline grant",
+			slog.String("offline_grant_id", grant.ID),
+			slog.String("offline_grant_initial_client_id", grant.InitialClientID),
+			slog.String("user_id", grant.Attrs.UserID),
+		)
```
tung2744: Maybe also log the create_at of offline grant so we can map the grant to a log.

pkg/lib/oauth/handler/handler_token.go
louischan-oursky: When x_sso_enabled is false, ShouldFireAuthenticatedEventWhenIssueOfflineGrant is set to true, and user.authenticated is actually done in handler_token.go (actually just a few lines below counting from here.)

I do it here because reauth is primarily triggered via oauth authorization endpoint, and it ends here.

louischan-oursky: Another reason is that we need to include a Session in the Event, and this code block is the block that does the reauthentication logic (update auth time of the offline grant, etc)

pkg/lib/authenticationflow/declarative/intent_signup_flow.go
louischan-oursky: Actually I followed the logic of the below effect. When it is account linking, the below effect (the one that calls deps.Users.AfterCreate, which should in turn dispatch user.created) is skipped.

If we update login time, then we will introduce the same problem again: Login time was updated but no audit log events are dispatched.

tung2744: I see, in account linking we have a login flow inside a signup flow so that flow will update the login time

pkg/lib/oauth/handler/handler_token.go
tung2744: https://github.com/authgear/authgear-server/blob/main/pkg/lib/authenticationflow/declarative/intent_login_flow.go#L126-L134

The event was also triggered in the login flow. It seems related to `x_sso_enabled` option.
Looking at the current code, it seems actually `user.authenticated` means "session created". If session was not created, even user completed the login flow, it is not counted as `user.authenticated`. If this is the case, why we don't have `user.authenticated` event in signup flow which also creates a session? (And user.created event does not log the session, so we do not know how the first session was created)

---

And looks like token endpoint only update the offline grant authenticated_at, but do not call `UpdateLoginTime`. So updating user's login time and `user.authenticated` / `user.reauthenticated` event is now still separated, therefore it is still possible that the login time updated but no corresponding authenticated event exist.

louischan-oursky: `user.created` does not log the session because creating a User via Admin API also fires this event.

tung2744: Yes. So it seems we fire `user.authenticated` in signup flow make more sense. We need a way to track the first session.

Or actually, we should separate session.created and user.authenticated. `session.created` called whenever session (idpsession or offline grant) created, and `user.authenticated` should be defined as "completing the login flow" so it is always called at the end of flow. But this might be breaking change. (Actually there is `user.session.terminated` but no `user.session.created`)

louischan-oursky: We begin discussing another problem.

The original comment is about why user.reauthenticated is not dispatched in authflow code, but here.

The answer is reauthentication, in its current form, is triggered via OIDC exclusively. So I dispatched the event is the most relevant code block.

tung2744: I think they are related, because it is about the definition of `user.authenticated` (And therefore, the similar event `user.reauthenticated` should share the same logic). i.e. whether it should be triggered just after the flow, or in code exchange.

And in the current logic (Trigger in code exchange), there are following problems:
1. login time of user and the event is still not a pair, so the problem of update login time / authenticated event do not occur together still exist, even after this pr.
2. signup flow session is not tracked in audit log.

So what I am suggesting is:
- `user.created` / `user.authenticated` / `user.reauthenticated` should always be called in the flow to solve 1.
- Add `user.session.created` to solve 2.

tung2744: Created https://linear.app/authgear/issue/DEV-2915/user-last-login-can-be-updated-without-userauthenticated for further discussion, let me merge this pr for the new event first.

tung2744 (approved):

PR #5307, opened 2025-07-29
[Portal] Application scopes

tung2744:
ref DEV-2898

https://github.com/user-attachments/assets/53db2f77-f7ac-4fd7-bc2e-fd421b347eea

### DIFF
--- a/.claude/commands/add_portal_admin_api_query.md
+++ b/.claude/commands/add_portal_admin_api_query.md
@@ -5,4 +5,4 @@ Add a .graphql file under ./portal/src/graphql/adminapi/query for $ARGUMENTS.
 - You should first read the graphql schema to understand how to write the query before you start.
 - Include cursor, pageInfo, totalCount if available unless the user mentioned they are not needed.
 - Query all fields inside the entity, except nested entities, unless the user explicitly mentioned they are required.
-- After adding the graphql file, run `npm run gentype` inside `/portal` to generate the ncessary code.
+- After adding the graphql file, run `npm run gentype` inside ./portal to generate the ncessary code.
--- a/.claude/commands/add_portal_admin_api_query.md
+++ b/.claude/commands/add_portal_admin_api_query.md
@@ -1,7 +1,8 @@
 Provided the graphql schema: ./portal/src/graphql/adminapi/schema.graphql
 
-Add a file under ./portal/src/graphql/adminapi/query for $ARGUMENTS.
+Add a .graphql file under ./portal/src/graphql/adminapi/query for $ARGUMENTS.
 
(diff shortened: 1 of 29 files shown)


.claude/commands/add_portal_admin_api_query.md
```diff
@@ -1,7 +1,8 @@
 Provided the graphql schema: ./portal/src/graphql/adminapi/schema.graphql
 
-Add a file under ./portal/src/graphql/adminapi/query for $ARGUMENTS.
+Add a .graphql file under ./portal/src/graphql/adminapi/query for $ARGUMENTS.
 
 - You should first read the graphql schema to understand how to write the query before you start.
 - Include cursor, pageInfo, totalCount if available unless the user mentioned they are not needed.
 - Query all fields inside the entity, except nested entities, unless the user explicitly mentioned they are required.
+- After adding the graphql file, run `npm run gentype` inside `/portal` to generate the ncessary code.
```
louischan-oursky: inside ./portal

That is, do not quote it in backticks and use explicitly relative path.

portal/src/screens/api-resources/APIResourceDetailsScreenScopesTab.tsx
```diff
@@ -87,10 +82,11 @@ export function APIResourceDetailsScreenScopesTab({
     variables: {
       resourceID: resource.id,
       first: pageSize,
-      after: offset === 0 ? undefined : encodeOffsetToCursor(offset),
+      after: offset === 0 ? undefined : encodeOffsetToCursor(offset - 1),
```
louischan-oursky: This logic appears in GroupsScreen.tsx:33 as well. I think we should extract this as a function: `makeAfterFromOffset`

e2e/tests/m2m/token.test.yaml
```diff
@@ -114,3 +114,28 @@ steps:
           "expires_in": 1800,
           "scope": "resource2_scope1 resource2_scope2"
         }
+
+---
+name: M2M - client_credentials flow with unassigned resource
+focus: true
```
louischan-oursky: I think the CI should have caught this. Remove `focus: true`

louischan-oursky (changes requested):

portal/src/screens/api-resources/APIResourceDetailsScreenScopesTab.tsx
tung2744: It seems that this logic should exist in encodeOffsetToCursor, because by nature offset to cursor is always - 1

tung2744:
Updated, thanks!

louischan-oursky (approved):

PR #5308, opened 2025-07-30
[Portal] Test m2m token exchange

tung2744:
ref DEV-2899

https://github.com/user-attachments/assets/ee5f71ac-bf15-4475-9427-d99923c54d6f


portal/src/graphql/portal/Authenticated.tsx
```diff
@@ -14,7 +14,7 @@ import authgear, {
   AuthenticateResult,
   ConfigureOptions,
 } from "@authgear/web";
-import { useNavigate } from "react-router-dom";
+import { useNavigate, createPath } from "react-router-dom";
```
louischan-oursky, line 17: It seems that createPath is not a documented API

https://github.com/remix-run/react-router/blob/v6.3.0/docs/api.md

It also does not appear in v6.30

What does `createPath` do?

tung2744: It is for creating a string, taking `pathname`, `hash`, `search`.
I saw it is documented in the v7 API. Looks like it was just v6 api doc missing it.
https://reactrouter.com/api/utils/createPath

louischan-oursky (approved):

PR #5313, opened 2025-08-04
Count m2m token generated

tung2744:
ref DEV-2889

### DIFF
--- a/pkg/lib/analytic/posthog.go
+++ b/pkg/lib/analytic/posthog.go
@@ -56,13 +56,13 @@ type PosthogIntegration struct {
 }
 
 type PosthogGroup struct {
-	ProjectID              string
-	MAU                    int
-	UserCount              int
-	CollaboratorCount      int
-	ApplicationCount       int
-	MonthlyM2MTokenCreated int
-	ProjectPlan            string
+	ProjectID                 string
+	MAU                       int
+	UserCount                 int
+	CollaboratorCount         int
+	ApplicationCount          int
+	M2MTokenCreatedLast30Days int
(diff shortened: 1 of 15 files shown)


pkg/lib/analytic/posthog.go
```diff
@@ -149,6 +154,21 @@ func (p *PosthogIntegration) preparePosthogGroup(ctx context.Context, appID stri
 		return nil, err
 	}
 
+	var m2mTokenCreatedCount int
+	err = p.AuditDBReadHandle.WithTx(ctx, func(ctx context.Context) error {
+		rangeTo := now
+		rangeFrom := now.AddDate(0, 0, -30)
```
louischan-oursky: For consistency, we should add a new function in timeutil `PreviousMonthRange`

```
// PreviousMonthRange is like PreviousMonth, except that it returns a end-exclusive range representing the result.
func PreviousMonthRange(t time.Time) (rangeFrom time.Time, rangeTo time.Time)
```

louischan-oursky (changes requested):

pkg/lib/analytic/posthog.go
tung2744: Added a commit for this, I named it Last30Days because it seems to be more accurate.

louischan-oursky: No, I really meant PreviousMonthRange. So the collected metrics share the same time span.

tung2744: Added a commit to rename the key, as discussed in standup.

louischan-oursky (approved):

PR #5317, opened 2025-08-07
Add rate limits to token endpoint

tung2744:
ref DEV-2917
ref DEV-2919
ref DEV-2907

### DIFF
--- a/pkg/lib/config/oauth.go
+++ b/pkg/lib/config/oauth.go
@@ -224,6 +224,23 @@ var _ = Schema.Add("OAuthClientConfig", `
 			"then": {
 				"required": ["redirect_uris"]
 			}
+		},
+		{
+			"if": {
+				"properties": {
+					"x_application_type": {
+						"enum": ["m2m"]
+					}
+				},
+				"required": ["x_application_type"]
+			},
+			"then": {
+				"properties": {
+					"access_token_lifetime_seconds": {
+						"maximum": 2419200
+					}
+				}
+			}
 		}
 	]
 }
--- a/pkg/lib/config/oauth.go
+++ b/pkg/lib/config/oauth.go
@@ -5,7 +5,11 @@ var _ = Schema.Add("OAuthConfig", `
 	"type": "object",
 	"additionalProperties": false,
 	"properties": {
-		"clients": { "type": "array", "items": { "$ref": "#/$defs/OAuthClientConfig" } }
+		"clients": {
+			"type": "array",
+			"items": { "$ref": "#/$defs/OAuthClientConfig" },
+			"maxItems": 50
+		}
 	}
 }
 `)
--- a/pkg/lib/config/oauth.go
+++ b/pkg/lib/config/oauth.go
@@ -7,8 +7,7 @@ var _ = Schema.Add("OAuthConfig", `
 	"properties": {
 		"clients": {
 			"type": "array",
-			"items": { "$ref": "#/$defs/OAuthClientConfig" },
-			"maxItems": 50
+			"items": { "$ref": "#/$defs/OAuthClientConfig" }
 		}
 	}
 }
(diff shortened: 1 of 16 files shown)


pkg/lib/oauth/handler/handler_token.go
```diff
@@ -130,6 +131,10 @@ type TokenHandlerOfflineGrantService interface {
 	GetOfflineGrant(ctx context.Context, id string) (*oauth.OfflineGrant, error)
 }
 
+type TokenHandlerRateLimiter interface {
```
louischan-oursky, line 134: Where did we apply the Per-IP rate limit?

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -1885,6 +1939,12 @@ func (h *TokenHandler) IssueTokensForSettingsActionCode(
 		return nil, err
 	}
 
+	if err := h.checkRateLimits(ctx, ratelimit.RateLimitOAuthTokenGeneral, ratelimit.ResolveBucketSpecOptions{
+		UserID: "", // We don't know the user id
```
louischan-oursky: But we should know it, right? Store the User ID the settings action grant.

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -2099,7 +2099,9 @@ func (h *TokenHandler) validateClientSecret(client *config.OAuthClientConfig, cl
 }
 
 func (h *TokenHandler) checkRateLimits(ctx context.Context, rl ratelimit.RateLimit, opts ratelimit.ResolveBucketSpecOptions) error {
-	specs := rl.ResolveBucketSpecs(nil, nil, nil, &opts)
+	finalOpts := opts
+	finalOpts.IPAddress = string(h.RemoteIP)
```
louischan-oursky: The IP rate limit should be applied very early in the process pipeline.

If the IP rate limit is applied very late, compute resources are still wasted.

pkg/lib/config/oauth.go
```diff
@@ -5,7 +5,11 @@ var _ = Schema.Add("OAuthConfig", `
 	"type": "object",
 	"additionalProperties": false,
 	"properties": {
-		"clients": { "type": "array", "items": { "$ref": "#/$defs/OAuthClientConfig" } }
+		"clients": {
+			"type": "array",
+			"items": { "$ref": "#/$defs/OAuthClientConfig" },
+			"maxItems": 50
+		}
```
louischan-oursky, line 11: Changing this may break existing projects. I suggest we apply the limit when authgear.yaml is updated. So it only breaks when someone tries to add more than 50 clients.

pkg/lib/oauth/handler/handler_token.go
tung2744: I think this comment is related to the below comment? https://github.com/authgear/authgear-server/pull/5317#discussion_r2261856640

Will check it before opening db connection.

pkg/lib/ratelimit/ratelimits.go
```diff
@@ -175,13 +175,6 @@ func (n RateLimit) resolvePerIP(cfg *config.AppConfig, featureCfg *config.Featur
 	case RateLimitMessagingEmail:
 		return resolveConfig(cfg.Messaging.RateLimits.EmailPerIP, nil, featureCfg.Messaging.RateLimits.EmailPerIP)
 
-	// OAuth
-	case RateLimitOAuthTokenGeneral:
-		return resolveConfig(&config.RateLimitConfig{
-			Enabled: func() *bool { var t = true; return &t }(),
-			Period:  "1m",
-			Burst:   120,
-		}, nil, nil)
```
louischan-oursky: Why did we move the defintions? I remember we decided to move all definitions to here recently.

tung2744: This is because in token endpoint, the per ip rate limit and per user rate limit is not checked at the same point. Therefore, unlike other ratelimits, `ResolveBucketSpecs` cannot be used. Instead, I have to define separate buckets and use them separately. So I decided to move this specific bucket into the place which uses it.

And actually the rate limit is not from any config but hardcoded, like `AccountManagementAPIPerIP`, I also did not move it to `ResolveBucketSpecs`.

louischan-oursky (approved):

PR #5318, opened 2025-08-07
Prevent double tap on all form submit buttons

louischan-oursky:
ref DEV-2935


authui/src/authflowv2/loading.ts
```diff
@@ -16,7 +48,27 @@ export class LoadingController extends Controller {
     this.revert = null;
   };
 
-  startLoading(e: HTMLElement | null): LoadingHandle {
+  onSubmit = (e: SubmitEvent) => {
+    // We do not call preventDefault() nor stopPropgation() here.
+    // We want the browser to submit the form.
+    // We just want to make the buttons no longer react to pointer events.
+
+    const form = e.target;
```
tung2744, line 56: Use currentTarget which is always the element we attached the event listener?

tung2744: I see, we are now listening to the event in document.

PR #5321, opened 2025-08-11
Implement client secret rotation

tung2744:
ref DEV-2905

https://github.com/user-attachments/assets/02447843-2817-4013-b30d-f845aa0b2c9b


louischan-oursky (commented):
The delete secret UI should show a dialog for confirmation. Right now it is too easy to accidentally delete a secret.

tung2744:
Added a confirmation dialog, thanks!

louischan-oursky:
The type checking still failed.

tung2744:
Added one more commit to fix the type error, thanks!

PR #5329, opened 2025-08-12
Use custom error message for access token lifetime

tung2744:
ref DEV-2907

Screenshot 2025-08-12 at 7 39 52 PM

### DIFF
--- a/portal/src/graphql/portal/EditOAuthClientForm.tsx
+++ b/portal/src/graphql/portal/EditOAuthClientForm.tsx
@@ -35,6 +35,7 @@ import {
 } from "../../components/applications/DeleteClientSecretConfirmationDialog";
 import Tooltip from "../../Tooltip";
 import { LocationState } from "./EditOAuthClientScreen";
+import { makeValidationErrorMatchUnknownKindParseRule } from "../../error/parse";
 
 const MASKED_SECRET = "***************";
 
@@ -835,6 +836,13 @@ const EditOAuthClientForm: React.VFC<EditOAuthClientFormProps> =
               value={
                 clientConfig.access_token_lifetime_seconds?.toFixed(0) ?? ""
               }
+              errorRules={[
+                makeValidationErrorMatchUnknownKindParseRule(
+                  "maximum",
+                  /\/access_token_lifetime_seconds$/,
+                  "EditOAuthClientForm.access-token.error.maximum"
+                ),
+              ]}
               onChange={onAccessTokenLifetimeChange}
             />
             <div>
--- a/portal/src/graphql/portal/EditOAuthClientForm.tsx
+++ b/portal/src/graphql/portal/EditOAuthClientForm.tsx
@@ -35,7 +35,7 @@ import {
 } from "../../components/applications/DeleteClientSecretConfirmationDialog";
 import Tooltip from "../../Tooltip";
 import { LocationState } from "./EditOAuthClientScreen";
-import { makeValidationErrorMatchUnknownKindParseRule } from "../../error/parse";
+import { makeValidationErrorCustomMessageIDRule } from "../../error/parse";
 
 const MASKED_SECRET = "***************";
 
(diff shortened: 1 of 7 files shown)


portal/src/graphql/portal/EditOAuthClientForm.tsx
```diff
@@ -826,6 +827,13 @@ const EditOAuthClientForm: React.VFC<EditOAuthClientFormProps> =
               value={
                 clientConfig.access_token_lifetime_seconds?.toFixed(0) ?? ""
               }
+              errorRules={[
+                makeValidationErrorMatchUnknownKindParseRule(
```
louischan-oursky: I am surprised that `maximum` is unknown kind.

Then this custom message will be broken when we handle `maximum` by default.

So my suggestion is

1. Handle `maximum` now. I think we should also handle `minimum`, `exclusiveMaximum`, and `exclusiveMinimum`
2. Switch to use `makeValidationErrorCustomMessageIDRule`, given that `maximum` is now handled and we want to override the message.

louischan-oursky (changes requested):

portal/src/graphql/portal/EditOAuthClientForm.tsx
tung2744: Actually, I am not sure what is meant by "unknown kind". I use this function just because I don't know there is a makeValidationErrorCustomMessageIDRule.

tung2744: https://github.com/authgear/authgear-server/blob/main/portal/src/error/parse.ts#L77-L91

I think known kind means messages listed here?
So only exclusiveMaximum and exclusiveMinimum is missing, which we did not use in the whole codebase.

And I am not sure why we need makeValidationErrorCustomMessageIDRule, it seems `makeValidationErrorMatchUnknownKindParseRule` can always be used to provide custom message.

louischan-oursky: I compared the implementation of `makeValidationErrorMatchUnknownKindParseRule` and `makeValidationErrorCustomMessageIDRule`. They look very similar, the only difference is the context being passed to the MessageFormat.

In `makeValidationErrorCustomMessageIDRule`, the validation error details is passed to the MessageFormat.

In `makeValidationErrorMatchUnknownKindParseRule`, the optional `values` is passed to the MessageFormat.

We can merge the functionality of `makeValidationErrorMatchUnknownKindParseRule` into `makeValidationErrorCustomMessageIDRule`, and remove `makeValidationErrorMatchUnknownKindParseRule`.

The pseudo code

```
export function makeValidationErrorCustomMessageIDRule(
  kind: string,
  locationRegExp: RegExp,
  errorMessageID: string,
  values?: Values,
): ErrorParseRule {
  // ...
          parsedAPIErrors.push({
            messageID: errorMessageID,
            arguments: {
              ...cause.details,
              ...values, // Allow given values to override cause.details.
            },
          });

}
```

tung2744:
Added a commit to remove makeValidationErrorMatchUnknownKindParseRule.

And also a commit trying to fix the problem of portal development server always having stale requests. It seems solved the problem for me. Maybe you can see if it also solved the problem for you.

nginx.conf
```diff
@@ -91,7 +91,7 @@ http {
             proxy_http_version 1.1;
             proxy_set_header Host $http_host;
             proxy_set_header Upgrade $http_upgrade;
-            proxy_set_header Connection $connection_upgrade;
+            proxy_set_header Connection "upgrade";
```
tung2744: I'm not sure why this seems working. $connection_upgrade was defined in http block and seems to be correct.

louischan-oursky: The effect of this change is that we no longer send `Connection: close` to the dev server.

I tend not to include this change in this PR. I am going to drop it.

louischan-oursky (approved):

PR #5335, opened 2025-08-14
Enable vite dev server debug log

louischan-oursky:
Continue from https://github.com/authgear/authgear-server/pull/5329#discussion_r2275244867

If the change worked for you, then the vite dev server may have problem when the request has `Connection: close`.

I enabled the debug log for the dev server. Let's see if we have any insights from the log when the problem occurs again.

### DIFF
--- a/portal/package.json
+++ b/portal/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "postinstall": "./scripts/npm-postinstall.sh",
     "test": "jest",
-    "start": "vite",
+    "start": "vite --debug",
     "build": "vite build",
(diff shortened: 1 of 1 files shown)


tung2744:
I've actually done it already and saw those pending requests do not reach vite server.

louischan-oursky:
I added one more commit. See if that works for you? I cannot reproduce the stale problem very often.

tung2744:
This fix seems working.
It is quite easy to trigger the problem if you navigate in different feature pages, and try to reauth in different places.
My guess is there are some unclosed connections, and when your loaded modules add up the number of connection saturate at some point.

PR #5342, opened 2025-08-21
Open database connection only on valid request received

tung2744:
ref DEV-2930

### DIFF
--- a/pkg/lib/oauth/handler/handler_authz.go
+++ b/pkg/lib/oauth/handler/handler_authz.go
@@ -734,7 +734,7 @@ func (h *AuthorizationHandler) doHandleConsentRequest(
 	ctx context.Context,
 	opts doHandleConsentRequestOptions,
 ) (httputil.Result, error) {
-	if err := h.validateRequestInternal(
+	if err := h.doValidateRequestWithoutTx(
 		opts.ConsentRequest.Client,
 		opts.ConsentRequest.OAuthSessionEntry.T.AuthorizationRequest,
 	); err != nil {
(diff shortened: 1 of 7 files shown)


pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -400,17 +345,43 @@ func (h *AuthorizationHandler) prepareConsentRequest(ctx context.Context, req *h
 	}, nil
 }
 
+func (h *AuthorizationHandler) HandleRequestWithTx(
```
louischan-oursky, line 348: Context should be the first argument. Never embed it in an options struct.

pkg/lib/oauth/handler/handler_authz.go
```diff
@@ -843,7 +814,66 @@ func (h *AuthorizationHandler) validatePreAuthenticatedURLTokenRequest(
 }
 
 // nolint:gocognit
-func (h *AuthorizationHandler) validateRequest(
+func (h *AuthorizationHandler) ValidateRequestWithoutTx(
+	ctx context.Context,
+	r protocol.AuthorizationRequest,
+) (*AuthorizationParams, *AuthorizationResultError) {
```
louischan-oursky: Change the return type to `(context.Context, *AuthorizationParams, *AuthorizationResultError)`.

The semantics is: we always return `ctx` even in case of error. So the caller can always write something like

```go
ctx := ...
ctx, params, err := ValidateRequestWithoutTx(ctx, ...)
if err != nil {
  // Handle error. ctx should be unmodified
}
// ctx should be modified here.
```

pkg/lib/oauth/handler/handler_authz.go
tung2744: I wanted to enforce the Context used here must be the context returned by ValidateRequestWithoutTx, therefore I put it inside the struct, but this pattern is indeed a bit weird.

tung2744:
Added a commit to remove context from the struct, thanks!

louischan-oursky (approved):

PR #5351, opened 2025-08-26
Support Alibaba Cloud OSS

louischan-oursky:
Depend on https://github.com/authgear/authgear-server/pull/5349

ref DEV-2992


tung2744 (approved):
I am thinking if it will be better to support s3 rest api integration instead of adding sdk for each provider

louischan-oursky:
Actually it is better to use SDK, to support Application Default Credentials. For now, we only support that for GCS, but literally all Cloud Provider support that.

PR #5353, opened 2025-08-27
Support admin api forgot password in authflow

tung2744:
ref DEV-2993

### DIFF
--- a/pkg/auth/handler/webapp/authflowv2/reset_password.go
+++ b/pkg/auth/handler/webapp/authflowv2/reset_password.go
@@ -10,7 +10,6 @@ import (
 	"github.com/authgear/authgear-server/pkg/auth/webapp"
 	"github.com/authgear/authgear-server/pkg/lib/authenticationflow/declarative"
 	"github.com/authgear/authgear-server/pkg/lib/authn/authenticator/password"
-	"github.com/authgear/authgear-server/pkg/lib/authn/otp"
 	"github.com/authgear/authgear-server/pkg/lib/feature/forgotpassword"
 	"github.com/authgear/authgear-server/pkg/util/httproute"
 	pwd "github.com/authgear/authgear-server/pkg/util/password"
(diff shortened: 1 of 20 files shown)


pkg/lib/authenticationflow/flow.go
```diff
@@ -121,3 +121,45 @@ func FindCurrentFlowReference(flow *Flow) *FlowReference {
 	}, flow)
 	return ref
 }
+
+// FindStepByType recursively finds the first step in the flow that matches the given stepType
+// and returns its JSON pointer.
+func FindStepByType(flowRoot config.AuthenticationFlowObjectFlowRoot, stepType config.AuthenticationFlowStepType) *jsonpointer.T {
```
louischan-oursky: It is better to return `(jsonpointer.T, bool)`, since a zero value of jsonpointer.T is a valid value. It points to the root of the JSON document.

pkg/auth/handler/webapp/authflowv2/reset_password.go
```diff
@@ -231,7 +173,3 @@ func (h *AuthflowV2ResetPasswordHandler) serveHTTPAuthflow(w http.ResponseWriter
 		h.Controller.HandleStep(r.Context(), w, r, &handlers)
 	}
 }
-
-func isURLFromAdminAPI(r *http.Request) bool {
-	return r.URL.Query().Get(otp.FromAdminAPIQueryKey) == "true"
```
louischan-oursky, line 236: I think we can keep the constant commented, but no longer writes it. So we can grep it later in the codebase and git blame the history if needed.

louischan-oursky (changes requested):

pkg/lib/authenticationflow/flow.go
louischan-oursky: This was addressed in a later commit.

pkg/auth/handler/webapp/authflowv2/reset_password.go
tung2744: Amended the commit so that we only commented this code, and added link to this pr.

louischan-oursky (approved):

PR #5356, opened 2025-09-02
Do not show recovery code as option if no authenticator can be used

tung2744:
ref DEV-2960

Screenshot 2025-09-02 at 3 42 01 PM


pkg/lib/authenticationflow/declarative/utils_common.go
```diff
@@ -436,6 +441,10 @@ func getAuthenticationOptionsForLogin(ctx context.Context, deps *authflow.Depend
 		}
 	}
 
+	if recoveryCodeBranch != nil && userHasRecoveryCode && (len(options) > 0) {
```
louischan-oursky, line 444: The last condition `(len(options) > 0)` should be `len(secondaryAuthenticators) > 0`

resources/authgear/templates/en/translation.json
```diff
@@ -1276,6 +1276,7 @@
   "v2.page.settings-mfa-password.default.remove-button-label": "Remove additional password",
   "v2.page.settings-mfa-password.default.title": "Additional password",
   "v2.page.settings-mfa-view-recovery-code.default.regenerate-button-label": "Generate new codes",
+  "v2.page.settings-mfa-view-recovery-code.default.empty": "You do not have any recovery codes yet. Generate new codes by clicking the button below.",
```
louischan-oursky: Rename this key to `v2.page.settings-mfa-view-recovery-code.default.no-existing-codes-description`

louischan-oursky (changes requested):

pkg/lib/authenticationflow/declarative/utils_common.go
tung2744: I think no, because `secondaryAuthenticators` includes authenticators which are disabled in config. `len(options)` is correct because at this point it is options excluding recovery code.

louischan-oursky: I see. You're correct.

resources/authgear/templates/en/translation.json
tung2744: Updated, thanks!

louischan-oursky (approved):

PR #5363, opened 2025-09-10
Use whatsapp cloudapi webhook to update message status

tung2744:
ref DEV-2608


pkg/lib/config/secret_whatsapp.go
```diff
@@ -111,10 +111,15 @@ type WhatsappCloudAPIAuthenticationTemplateCopyCodeButtonConfig struct {
 	Languages []string `json:"languages"`
 }
 
+type WhatsappCloudAPIWebhook struct {
+	VerifyToken string `json:"verify_token"`
```
louischan-oursky, line 115: ~~Any links or screenshots showing the official name of this token and how is it supposed to work?~~

I saw that you added a commit to document how to set things up. Nice.

pkg/auth/routes.go
```diff
@@ -505,6 +505,7 @@ func NewRouter(p *deps.RootProvider, configSource *configsource.ConfigSource) ht
 
 	router.Add(webapphandler.ConfigureWechatAuthRoute(webappPageRoute), p.Handler(newWechatAuthHandler))
 
+	router.Add(webapphandler.ConfigureWhatsappCloudAPIWebhookRoute(webappAPIRoute), p.Handler(newWhatsappCloudAPIWebhookHandler))
```
louischan-oursky, line 508: This route should be mounted to the no-project route, as this route is supposed to be shared by all projects.

We need to be able to tell which project by either

1. If whatsapp API allows us to include our own reference code that will be included in the webhook body, then use this.
2. Otherwise, we use the whatsapp message reference to look up the project.

pkg/lib/infra/whatsapp/message_store_redis.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type MessageStore struct {
+	// We use global redis because the whatsapp credential might be shared between different apps
+	Redis       *globalredis.Handle
+	Credentials *config.WhatsappCloudAPICredentials
+}
+
+func redisMessageStatusKey(phoneNumberID string, messageID string) string {
+	return fmt.Sprintf("whatsapp:phone-number-id:%s:message-id:%s", phoneNumberID, messageID)
```
louischan-oursky: phone number id looks sensitive. We should hash it with sha256 hex before use. Rename the key to `whatsapp:hashed-phone-number-id:%s:message-id:%s`.

pkg/auth/handler/webapp/whatsapp_webhook.go
```diff
@@ -55,24 +59,67 @@ func ConfigureWhatsappCloudAPIWebhookRoute(route httproute.Route) httproute.Rout
...
+	hubVerifyToken := r.URL.Query().Get("hub.verify_token")
+
+	if hubChallenge != "" {
+		w.Header().Set("Content-Type", "text/plain")
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte(hubChallenge))
+		return
+	}
+
+	if hubVerifyToken != h.Credentials.Webhook.VerifyToken {
```
louischan-oursky: Use constant time compare.

pkg/auth/handler/webapp/whatsapp_webhook.go
```diff
@@ -55,24 +59,67 @@ func ConfigureWhatsappCloudAPIWebhookRoute(route httproute.Route) httproute.Rout
...
+			slog.String("app_id", string(h.AppID)),
+		).WithError(err).Error(ctx, "invalid request body")
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
 	for _, entry := range payload.Entry {
 		for _, change := range entry.Changes {
 			if change.Field == "messages" {
+				if change.Value.Metadata.PhoneNumberID != h.Credentials.PhoneNumberID {
```
louischan-oursky: Use constant time compare

pkg/lib/infra/whatsapp/service.go
```diff
@@ -89,16 +96,69 @@ func (s *Service) SendAuthenticationOTP(ctx context.Context, opts *SendAuthentic
...
 			ctx,
 			opts,
 			lang,
 		)
+		if err != nil {
+			return err
+		}
+		success := make(chan bool, 1)
+		// Wait for 5 seconds for the message status
+		go s.waitUntilSent(ctx, success, messageID, 5*time.Second)
```
louischan-oursky: It is better to make this an environment variable.

pkg/lib/infra/whatsapp/service.go
```diff
@@ -104,6 +105,9 @@ func (s *Service) SendAuthenticationOTP(ctx context.Context, opts *SendAuthentic
 		if err != nil {
 			return err
 		}
+		if !s.shouldWaitForMessageStatusUpdate() {
+			return nil
+		}
```
louischan-oursky: This should be handled by `waitUntilSent`. So the caller can always simply call `waitUntilSent` directly.

When the webhook credentials are not configured, `waitUntilSent` should not block and return almost immediately.

pkg/auth/routes.go
tung2744: Actually, this route is shared by projects which shares the same whatsapp credential (Same phone number id).

For 1, it is not possible.
For 2, we need to store the message id after we get a response from the message api, after that, when we receive the webhook we are able to look up the project. However in my testing, I cannot tell if the webhook is always called after the request returned a response. That is, there might be cases that the webhook called before we know the message id. Therefore I afraid this might not work properly.

The current implementation should work properly, even the webhook was called before we know the message id. For projects which shares the same credential with `accounts`, we simply use the `accounts` endpoint. For project using their own credentials, we use their own project endpoint.

tung2744: ~I found there is a param `biz_opaque_callback_data` in the api, which seems can be used to store the app id. So let me try to use approach 1.~

---

I still find it difficult to use a no project route, because we have verify token set in the credential. So I prefer to keep the current implementation. Different project will have different webhook endpoint, however, if they share the same phone number id, using a shared endpoint will work. In our production, we will use the `accounts` endpoint. If any project have their own whatsapp credential, then that project will use its own webhook endpoint.

pkg/lib/infra/whatsapp/service.go
tung2744: Maybe let me put it in the app config with a default 5s.

tung2744:
Updated, thanks!

pkg/auth/routes.go
louischan-oursky: > I still find it difficult to use a no project route, because we have verify token set in the credential.

Can you explain what is the difficulty?

If the incoming request is `POST https://accounts.portal.authgear.cloud/noproject/callback?verify_token=token`, we can first look at the webhook body before verifying the request. After we have identified the project, we know the credential we have to use, then we check the verify_token.

tung2744: There are two type of requests the webhook will receive. Verify request and notification requests. On registering the endpoint, we get a verify request. And when you send message, we get notification request.

In verify request, there is no `biz_opaque_callback_data` (No request body). Therefore we cannot identify the project.
We can set a different path for different project, like `/whatsapp/webhook/{appID}`, but then this seems to be the same as using project specific authgear endpoint.

louischan-oursky: I am a bit confused. For the second type of webhook (notification request), how do we check the authenticity of the request to ensure it was sent from Whatsapp Cloud API, not from a malicious attacker?

For the first type of webhook (verify request), how often does it happen? Does it only happen once when we register a new endpoint to receive webhook?

If the 1st type of webhook happens only once during registering a new endpoint, then we may

1. Use https://<project>.authgear.cloud/whatsapp/webhook/verify to serve 1st request. It uses the project specific credentials if available, otherwise fall back to the deployment credentials.
2. Use https://noproject.authgear.cloud/whatsapp/webhook/notification to serve 2nd request. It always use  biz_opaque_callback_data to resolve the project.

tung2744: > If the 1st type of webhook happens only once during registering a new endpoint, then we may
Use https://.authgear.cloud/whatsapp/webhook/verify to serve 1st request. It uses the project specific credentials if available, otherwise fall back to the deployment credentials.
Use https://noproject.authgear.cloud/whatsapp/webhook/notification to serve 2nd request. It always use biz_opaque_callback_data to resolve the project.

We need to server both type of requests in the same endpoint, so it is not possible. Facebook did not mention how many times / how often the verify request happens. In my testing, it always occur once at the time I register the webhook, but I think our endpoint should be always able to handle two type of requests.

> how do we check the authenticity of the request to ensure it was sent from Whatsapp Cloud API, not from a malicious attacker?

I indeed missed this, we should check the `X-Hub-Signature-256` header to verify the payload.

tung2744: Added a commit to check the signature, and also updated helm chart for this
https://github.com/authgear/helm-charts/pull/50

PR #5374, opened 2025-09-29
Move access token hook outside transaction

louischan-oursky:
ref DEV-3061

Manually tested the following cases:

- Sign in with mobile SDK
- Anonymous user sign in with mobile SDK
- Biometric login with mobile SDK
- Sign in with web SDK
- Anonymous user sign in with web SDK
- Pre-authenticated URL with mobile SDK
- Refresh access token
- The /tester endpoint

### DIFF
--- a/pkg/auth/handler/oauth/authorize.go
+++ b/pkg/auth/handler/oauth/authorize.go
@@ -22,7 +22,7 @@ var AuthorizeHandlerLogger = slogutil.NewLogger("handler-authz")
 
 type ProtocolAuthorizeHandler interface {
 	ValidateRequestWithoutTx(ctx context.Context, r protocol.AuthorizationRequest) (context.Context, *handler.AuthorizationParams, *handler.AuthorizationResultError)
-	HandleRequestWithTx(ctx context.Context, r protocol.AuthorizationRequest, params *handler.AuthorizationParams) httputil.Result
+	HandleRequest(ctx context.Context, r protocol.AuthorizationRequest, params *handler.AuthorizationParams) httputil.Result
 }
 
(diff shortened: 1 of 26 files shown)


pkg/lib/oauth/grant_access_service.go
```diff
@@ -34,7 +34,7 @@ type IssueAccessGrantResult struct {
 func (s *AccessGrantService) IssueAccessGrant(
```
tung2744: Should we rename the method too?

pkg/auth/handler/oauth/authorize.go
```diff
@@ -52,14 +50,10 @@ func (h *AuthorizeHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
 	var result httputil.Result
-	err = h.Database.WithTx(ctx, func(ctx context.Context) error {
-		result = h.AuthzHandler.HandleRequestWithTx(ctx, req, params)
-		if result.IsInternalError() {
-			return errAuthzInternalError
-		}
-		return nil
-	})
-
+	result = h.AuthzHandler.HandleRequestWithTx(ctx, req, params)
```
tung2744: Because there is no Tx here, maybe we should rename it to `HandleRequest`.

pkg/lib/oauth/grant_access_service.go
louischan-oursky: Addressed.

pkg/auth/handler/oauth/authorize.go
louischan-oursky: Addressed.

PR #5375, opened 2025-09-29
Send whatsapp asynchronously in authflow

tung2744:
ref DEV-3055


pkg/lib/authn/otp/code.go
```diff
@@ -34,4 +35,8 @@ type Code struct {
 	AuthenticationFlowType                 string        `json:"authentication_flow_type,omitempty"`
 	AuthenticationFlowName                 string        `json:"authentication_flow_name,omitempty"`
 	AuthenticationFlowJSONPointer          jsonpointer.T `json:"authentication_flow_json_pointer,omitempty"`
+
+	// The message which carries the otp
+	MessageID      string                  `json:"message_id,omitzero"`
```
louischan-oursky: It should be `WhatsappCloudAPIMessageID` `whatsapp_cloud_api_message_id`

pkg/lib/authn/otp/code.go
```diff
@@ -37,6 +37,7 @@ type Code struct {
 	AuthenticationFlowJSONPointer          jsonpointer.T `json:"authentication_flow_json_pointer,omitempty"`
 
 	// The message which carries the otp
-	MessageID      string                  `json:"message_id,omitzero"`
-	DeliveryStatus model.OTPDeliveryStatus `json:"delivery_status,omitzero"`
+	MessageID      string                        `json:"message_id,omitzero"`
+	OOBChannel     model.AuthenticatorOOBChannel `json:"channel,omitzero"`
+	DeliveryStatus model.OTPDeliveryStatus       `json:"delivery_status,omitzero"`
```
louischan-oursky: I think we should not store `delivery_status` in Code. Instead, when `getCode`, check `channel` and derive `delivery_status`.

If `channel` is `sms` or `email`, then it should be `sent`.

If it is `whatsapp`, then read the whatsapp message status redis key. If the key does not exist, then it is `sending`. Timeout should also be determined here, so we need to store the `created_at` of the code. With `created_at` stored in the code, we can determine timeout by checking the current time and `created_at`, thus we can also eliminate the `go func() {}` that updates the message status.

We should also remove `UpdateOTPMessageStatus` so that we no longer need to keep the states in sync.

With this change, the implementation should be simpler.

tung2744: I think timeout should be determined by the whatsapp service instead of the otp service, because it seems to be logic of the message sender instead of the otp. I have a later commit which moved the timeout logic to whatsapp service.

But I think I can try to remove `UpdateOTPMessageStatus` and compute the status on each inspect.

pkg/lib/authn/otp/code.go
tung2744: I name it WhatsappMessageID because I don't want the otp service to know about the whatsapp api implementation.

tung2744:

tung2744:
As per latest discussion:
- [x] Rename the error reason
- [x] Do not read the latest used channel of user
- [x] Do not wait for whatsapp cloud api call

tung2744:
Added some commits according to latest discussion

louischan-oursky (approved):

PR #5382, opened 2025-10-06
Rename hook related errors

tung2744:
ref DEV-3066


docs/specs/hook.md
```diff
@@ -92,7 +92,7 @@ If any hook fails the operation, the operation fails with error
 }
 ```
 
-> For backward compatibility, the reason is called "WebHookDisallowed".
+> For backward compatibility, the reason is called "HookDisallowed".
```
louischan-oursky: The changed statement is incorrect. Change it to

```
> For backward compatibility, the reason was called "WebHookDisallowed".
> In DEV-3066, however, we decided to break backward compatibility and rename it to `HookDisallowed`.
```

resources/authgear/templates/en/web/authflowv2/__error.html
```diff
@@ -239,7 +239,7 @@
       </span>
     {{ else if eq .Error.reason "WebHookInvalidResponse" }}
       <span>{{ include "v2.error.webhook-invalid-response" nil }}</span>
```
louischan-oursky: We should also rename the translation keys.

tung2744: There could be projects customized the text. This would be a breaking change.

docs/specs/hook.md
tung2744:
```
> In the past, the reason was `WebHookDisallowed`. However, we decided to break backward compatibility and rename it to `HookDisallowed`.
```

I write it in this way, and do not mention the issue number in a spec.

tung2744:
Updated, thanks!

resources/authgear/templates/de/translation.json
```diff
@@ -1029,9 +1031,7 @@
...
   "v2.error.web-ui-session-completed": "Einen Moment bitte, Sie werden zu Ihrer App weitergeleitet.",
   "v2.error.webhook-delivery-failed": "Der Vorgang ist nicht zulässig, da die Webhook-Zustellung zu einem unbekannten Fehler geführt hat.",
-  "v2.error.webhook-delivery-timeout": "Der Vorgang ist nicht zulässig, da die Webhook-Übermittlung zu einem Timeout geführt hat.",
   "v2.error.webhook-disallowed": "Der Vorgang ist nicht zulässig",
-  "v2.error.webhook-invalid-response": "[An den Entwickler] Ihr Webhook-Handler hat KEINE gültige Antwort zurückgegeben. Bitte besuchen Sie <a class=\"link\" target=\"_blank\" href=\"https://docs.authgear.com/customization/events-hooks/webhooks\">https://docs.authgear.com/customization/events-hooks/webhooks</a>",
   "v2.error.webhook-pre-signup-disallowed": "Die Anmeldung ist nicht zulässig",
```
louischan-oursky: We still have some keys not renamed, for example:

1. `v2.error.webhook-delivery-failed`
2. `v2.error.webhook-disallowed`
3. `v2.error.webhook-pre-signup-disallowed`

I also wonder why we have `v2.error.webhook-pre-signup-disallowed`. It is checking `if eq .Error.info.event_type "pre_signup"` But we never had a event called `pre_signup`.

Searching the git history brought me to e7117d92ded25ffc63a23e56aeee76ca75d9c1db, the first commit this line was introduced. I think it is no longer valid. So

1. Remove `v2.error.webhook-pre-signup-disallowed` in translation.json
2. Remove the code that reference `v2.error.webhook-pre-signup-disallowed`

tung2744:
@louischan-oursky Updated to rename more translation keys, and removed `v2.error.webhook-pre-signup-disallowed`.

I did not rename `v2.error.webhook-delivery-failed` because it is webhook specific.

tung2744:
Added a commit to rename WebHookDeliveryUnknownFailure to HookDeliveryUnknownFailure, also rebased to solve the errors.

PR #5384, opened 2025-10-08
Fix errors missing stacktrace in sentry

tung2744:
ref DEV-3073

### DIFF
--- a/pkg/util/slogutil/mask_handler.go
+++ b/pkg/util/slogutil/mask_handler.go
@@ -3,6 +3,7 @@ package slogutil
 import (
 	"context"
 	"log/slog"
+	"reflect"
 	"slices"
 
 	"github.com/jba/slog/withsupport"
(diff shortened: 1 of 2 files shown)


pkg/util/slogutil/mask_handler.go
```diff
@@ -45,23 +46,38 @@ func (o MaskHandlerOptions) maskString(s string) string {
...
 		// Otherwise coerce to string and mask.
 		str := o.maskString(a.Value.String())
 		return slog.Attr{
-			Key:   a.Key,
-			Value: slog.StringValue(str),
+			Key: a.Key,
+			Value: slog.AnyValue(&MaskedError{
+				Type:    reflect.TypeOf(a.Value.Any()).String(),
+				Message: str,
+			}),
```
louischan-oursky, line 95: We should keep using StringValue here as the `anything` here is not really an `error`. It is inappropriate to coerce everything to `error` if it is not an `error` originally.

tung2744: But it is not a string too. I think we can keep using AnyValue but the content will be a masked string.

tung2744:
Updated, thanks!

PR #5373, opened 2025-09-29
Re-specify account status

louischan-oursky:
ref DEV-3051

PR #5385, opened 2025-10-09
Account Status - Part 1

louischan-oursky:
ref DEV-3067

### DIFF
--- a/pkg/lib/authn/user/account_status.go
+++ b/pkg/lib/authn/user/account_status.go
@@ -349,14 +349,14 @@ func (s AccountStatusWithRefTime) Check() error {
 		return ErrAnonymizedUser
 	}
 
+	if s.isOutsideValidPeriod() {
+		return ErrUserOutsideValidPeriod
+	}
+
 	variant := s.variant()
 	err := variant.check()
 	if err != nil {
 		return err
-	}
-
-	if s.isOutsideValidPeriod() {
-		return ErrUserOutsideValidPeriod
 	}
 
 	return nil
--- a/pkg/lib/authn/user/account_status_test.go
+++ b/pkg/lib/authn/user/account_status_test.go
@@ -560,7 +560,7 @@ func TestAccountStatusAccountValidPeriod(t *testing.T) {
 			state2, err := state1.ScheduleDeletionByAdmin(deleteAt)
 			So(err, ShouldBeNil)
 			So(state2.Check(), ShouldNotBeNil)
-			So(state2.Check(), ShouldBeError, "user was scheduled for deletion by admin")
+			So(state2.Check(), ShouldBeError, "user is outside valid period")
 			So(state2.IsDisabled(), ShouldEqual, true)
 			So(state2.accountStatus.isDisabled, ShouldEqual, true)
 			So(state2.variant().getAccountStatusType(), ShouldEqual, accountStatusTypeScheduledDeletionDisabled)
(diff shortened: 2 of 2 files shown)


pkg/lib/authn/user/account_status.go
```diff
@@ -22,40 +22,47 @@ const (
...
+	TemporarilyDisabledUntil *time.Time
+	AccountValidFrom         *time.Time
+	AccountValidUntil        *time.Time
+	DeleteAt                 *time.Time
+	AnonymizeAt              *time.Time
+	AnonymizedAt             *time.Time
+	IsAnonymized             *bool
 }
 
 func (s AccountStatus) Type() AccountStatusType {
```
tung2744: Should we check `AccountStatusStaleFrom` here?

pkg/lib/authn/user/account_status.go
```diff
@@ -37,259 +154,402 @@ type AccountStatus struct {
...
+		return AccountStatusVariantAnonymized{}
+	}
+	if s.accountStatus.anonymizeAt != nil {
+		return AccountStatusVariantScheduledAnonymizationByAdmin{
+			anonymizeAt: *s.accountStatus.anonymizeAt,
+		}
+	}
+	if *s.accountStatus.isDeactivated {
+		return AccountStatusVariantDeactivated{}
+	}
```
tung2744, line 421: I remember according to our previous discussion, `Deactivated` is not a valid state of a user because it is always set together with deleteAt?

pkg/lib/authn/user/account_status.go
```diff
@@ -367,6 +367,39 @@ func (s AccountStatusWithRefTime) DisableIndefinitely(reason *string) (*AccountS
 	}
 }
 
+func (s AccountStatusWithRefTime) DisableTemporarily(from time.Time, until time.Time, reason *string) (*AccountStatusWithRefTime, error) {
```
tung2744: Should until be nullable, indicating disabled until forever?

pkg/lib/authn/user/account_status.go
```diff
@@ -232,13 +222,57 @@ func (s AccountStatusWithRefTime) IsAnonymized() bool {
...
 	variant := s.variant()
-	return variant.check()
+	err := variant.check()
+	if err != nil {
+		return err
+	}
+
+	if s.isOutsideValidPeriod() {
+		return ErrUserOutsideValidPeriod
+	}
```
tung2744: Looks like `isOutsideValidPeriod` should be checked before all other checks? (Have a higher priority?)

cmd/authgear/cmd/cmddatabase/migrations/authgear/20251003144527-add_new_account_status_columns.sql
```diff
@@ -0,0 +1,16 @@
+-- +migrate Up
+ALTER TABLE _auth_user ADD COLUMN is_indefinitely_disabled boolean;
```
tung2744, line 2: Do we set the initial value from the current `is_disabled`?

pkg/admin/graphql/user.go
```diff
@@ -333,6 +333,22 @@ var nodeUser = node(
...
+				Description: "The end timestamp of the temporarily disabled period",
+			},
+			"accountValidFrom": &graphql.Field{
+				Type:        graphql.DateTime,
+				Description: "The start timestamp of the account valid period",
+			},
+			"accountValidUntil": &graphql.Field{
+				Type:        graphql.DateTime,
+				Description: "The end timestamp of the account valid period",
+			},
```
tung2744, line 351: I remember we have said that we want to expose one new enum as the "current status" of the user? So that the user do not need to compute it from all these fields to know why the user is disabled.

pkg/lib/authn/user/account_status.go
tung2744: Should be handled in later commits

cmd/authgear/cmd/cmddatabase/migrations/authgear/20251003144527-add_new_account_status_columns.sql
louischan-oursky: I thought of that. But there is a delay between the time when the migration is run, and the time when the image is updated. There could be some rows with is_indefinitely_disabled being null. So I gave it up.

pkg/lib/authn/user/account_status.go
louischan-oursky: Indefinitely disabled is represented by `is_indefinitely_disabled=true`. `temporarily_disabled_xxx` is for use of temporarily disabled only.

tung2744: I am not sure, but isn't we said using the new disabled timestamps for future disable user? the `is_indefinitely_disabled` column should be only for storing existing `is_disabled` data?

pkg/lib/authn/user/account_status.go
louischan-oursky: The spec says this has lower precedence than others. https://github.com/authgear/authgear-server/pull/5373/files#diff-69efaa0592587589e7bc78fab44085c48a2713976267b003f287e88eec6f3ed9R553

Or you think we should amend the order in the spec?

tung2744: 🤔 Seems valid period should be checked first, because if the account is not in valid period, it probably means "the user haven't joined yet or already leaved", so any other status, such as temp ban / disabled is not very meaningful. Maybe also @fungc-io.

pkg/lib/authn/user/account_status.go
louischan-oursky: Currently it is not a valid state. But if we allow end-user to disable their own account, then it will be a valid state. I can keep the enum variant but panic here.

pkg/admin/graphql/user.go
louischan-oursky: I did not do this because account status is not always a single status. For example, an account can be anonymized and scheduled for deletion at the same time. Introducing a single status field cannot represent this. So I tend to keep what we have now.

pkg/lib/authn/user/account_status.go
louischan-oursky: If your suggestion is applied, then we have 2 representations

```
# Both means is_indefinitely_disabled
is_indefinitely_disabled=true
disabled_from=XXX,disabled_until=null
```

In the product meeting, we discussed that the database schema should be intuitive to understand. So I did not consider this design, as it is not intuitive.

The current design is more intuitive because indefinitely disabled and temporarily disabled has only one representation.

If we want to store when the user was indefinitely disabled, we can add a column `indefinitely_disabled_at`. I remember that we do not need this for now, so I did not add this column.

```
is_indefinitely_disabled=true,indefinitely_disabled_at=null # indefinitely disabled, but the time is unknown, probably because the user was disabled before we start tracking the time
is_indefinitely_disabled=true,indefinitely_disabled_at=XXX # indefinitely disabled at XXX
temporarily_disabled_from=XXX,temporarily_disabled_until=YYY # temporarily disabled from XXX to YYY
```

pkg/lib/authn/user/account_status.go
louischan-oursky: Let's discuss in the standup meeting.

pkg/lib/authn/user/account_status.go
louischan-oursky: Added a commit to comment out deactivated and panic instead.

pkg/admin/graphql/user.go
tung2744: However, we already have a defined order in the original `Check` method. In your example, the account status is considered `scheduled_deletion_disabled`.
As we always return a single api error reason on checking account status, there is always only one single status.

louischan-oursky: I realized returning a single error reason in `Check()` is normally fine because IsAnonymized does not overlap with other account status in authflow check_account_status. If the user is anonymized, it should have no identities, therefore, will not be able to identify itself.

But in Admin API, it is possible to query an anonymized user, in which the user can also be scheduled_deletion_disabled.

Maybe a bitfield is more suitable than a enum. But I tend not to add this now. Most of the time, the developer can just check `isDisabled` to see if the user is disabled.

pkg/lib/authn/user/account_status.go
louischan-oursky: Let me re-order and put OutsideValidPeriod after Anonymized.

louischan-oursky: Added a commit to address this.

PR #5386, opened 2025-10-10
Do not allow user to create primary authenticator in login flow

tung2744:
ref DEV-3082

### DIFF
--- a/pkg/lib/authenticationflow/declarative/intent_login_flow_step_authenticate.go
+++ b/pkg/lib/authenticationflow/declarative/intent_login_flow_step_authenticate.go
@@ -41,12 +41,13 @@ func init() {
 //     NodeDoConsumeRecoveryCode (MilestoneDidAuthenticate)
 
 type IntentLoginFlowStepAuthenticate struct {
-	FlowReference      authflow.FlowReference `json:"flow_reference,omitempty"`
-	JSONPointer        jsonpointer.T          `json:"json_pointer,omitempty"`
-	StepName           string                 `json:"step_name,omitempty"`
-	UserID             string                 `json:"user_id,omitempty"`
-	Options            []AuthenticateOption   `json:"options"`
-	DeviceTokenEnabled bool                   `json:"device_token_enabled"`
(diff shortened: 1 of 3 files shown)


pkg/lib/authenticationflow/declarative/intent_login_flow_step_authenticate.go
```diff
@@ -41,12 +41,13 @@ func init() {
...
-	UserID             string                 `json:"user_id,omitempty"`
-	Options            []AuthenticateOption   `json:"options"`
-	DeviceTokenEnabled bool                   `json:"device_token_enabled"`
+	FlowReference                  authflow.FlowReference `json:"flow_reference,omitempty"`
+	JSONPointer                    jsonpointer.T          `json:"json_pointer,omitempty"`
+	StepName                       string                 `json:"step_name,omitempty"`
+	UserID                         string                 `json:"user_id,omitempty"`
+	Options                        []AuthenticateOption   `json:"options"`
+	DeviceTokenEnabled             bool                   `json:"device_token_enabled"`
+	DidAuthenticatedBeforeThisStep bool                   `json:"did_authenticated_before_this_step"`
```
louischan-oursky, line 50: Why do we need to store this field? You can modify `canCreateAuthenticator` to also take `flows` and derive DidAuthenticatedBeforeThisStep just-in-time.

tung2744: This is because `MilestoneDidAuthenticate` can be created within `IntentLoginFlowStepAuthenticate`. Therefore, if we find the milestone in `canCreateAuthenticator`, the result of `canCreateAuthenticator` can change. And what we actually want to know is before entering this authenticate step, is there at least one `MilestoneDidAuthenticate` exist. Therefore, compute and store the flag in `NewIntentLoginFlowStepAuthenticate` seems to be the correct way to do it.

louischan-oursky: I see.

PR #5387, opened 2025-10-10
Return InvalidCode error when getting invalid reset password code

tung2744:
ref DEV-3083

### DIFF
--- a/pkg/auth/handler/api/authenticationflow_v1.go
+++ b/pkg/auth/handler/api/authenticationflow_v1.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 
 	"github.com/authgear/authgear-server/pkg/api"
+	"github.com/authgear/authgear-server/pkg/api/apierrors"
 	authflow "github.com/authgear/authgear-server/pkg/lib/authenticationflow"
 	"github.com/authgear/authgear-server/pkg/util/httputil"
 )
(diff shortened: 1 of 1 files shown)


pkg/auth/handler/api/authenticationflow_v1.go
```diff
@@ -58,7 +58,14 @@ func batchInput0(
...
 	if err != nil {
-		return nil, err
+		// If we failed to get the state, simply return the api error
+		if errors.Is(err, authflow.ErrFlowNotFound) {
+			return &api.Response{
+				Error: flowErr,
+			}, nil
+		} else {
+			return nil, err
+		}
```
louischan-oursky: I suggest we change the function body to

```golang
// In cases like reset password, state token is empty.
if stateToken == "" {
  return &api.Response{
    Error: flowErr,
  }
}

// The original function body
output, err := service.Get(ctx, stateToken)
if err != nil {
  return nil, err
}
result := output.ToFlowResponse()
return &api.Response{
  Error: flowErr,
  Result: result,
}
```

tung2744:
```
// The original function body
output, err := service.Get(ctx, stateToken)
if err != nil {
  return nil, err
}
```

This seems to be incorrect, because it swallowed the `flowErr` which is the real error we encountered and returned an irrelevant `err`.

tung2744: But I agree that we can add the `if stateToken == "" {}` block to handle the case if `stateToken` is empty string so we don't even need to get the flow.

tung2744: Tried to rewrite the function so that no unnecessary redis connection are needed, and no errors are dropped.

PR #5390, opened 2025-10-13
Add "No Password" option in portal

tung2744:
ref DEV-3084

Screenshot 2025-10-13 at 4 18 33 PM
Screenshot 2025-10-13 at 4 18 06 PM
Screenshot 2025-10-13 at 4 17 53 PM

### DIFF
--- a/portal/src/graphql/adminapi/AddUserScreen.tsx
+++ b/portal/src/graphql/adminapi/AddUserScreen.tsx
@@ -288,15 +288,17 @@ const AddUserContent: React.VFC<AddUserContentProps> = function AddUserContent(
             <Text block={true}>
               <FormattedMessage id="AddUserScreen.password-creation-type.manual.description" />
             </Text>
-            <ManualEntryPasswordField
-              disabled={!props?.checked}
-              passwordPolicy={passwordPolicy}
-              password={password}
-              onPasswordChange={onPasswordChange}
-              selectedLoginIDType={selectedLoginIDType}
-              sendPassword={state.manualEntrySendPassword}
-              onChangeSendPassword={onChangeSendPassword}
-            />
+            {props?.checked ? (
+              <ManualEntryPasswordField
+                disabled={!props.checked}
+                passwordPolicy={passwordPolicy}
+                password={password}
+                onPasswordChange={onPasswordChange}
+                selectedLoginIDType={selectedLoginIDType}
+                sendPassword={state.manualEntrySendPassword}
+                onChangeSendPassword={onChangeSendPassword}
+              />
+            ) : null}
           </div>
         </>
       );


fungc-io:
Thanks @tung2744 , I think the style is not consistent with other pages
image


for instance, the options are not bold in this page. Let me ask the designer to make some adjustment on the design.

portal/src/graphql/adminapi/AddUserScreen.tsx
```diff
@@ -183,77 +238,225 @@ const AddUserContent: React.VFC<AddUserContentProps> = function AddUserContent(
...
+            case PasswordCreationType.AutoGenerate:
+              newSendPassword = true;
+              break;
+            case PasswordCreationType.NoPassword:
+              newSendPassword = false;
+              newSetPasswordExpired = false;
+              break;
+            case PasswordCreationType.ManualEntry:
+              newSendPassword = prev.sendPassword;
+              newSetPasswordExpired = prev.setPasswordExpired;
```
louischan-oursky: I have these suggestions

1. Rename `sendPassword` to `sendPasswordManualEntry`. This means this state is only for PasswordCreationType.ManualEntry. For PasswordCreationType.AutoGenerate, we always send password so we can just hard-code true and ignore this state. For PasswordCreationType.NoPassword, we never send password so we can just hard-code false. Then we can keep this state unchanged when PasswordCreationType changes.
2. Rename `password` to `passwordManualEntry`. For a similar reason as in Point 1.
3. Keep `setPasswordExpired` unchanged when PasswordCreationType changes.
4. The whole form is reset to a deterministic state when the login ID type change.

pkg/lib/facade/coordinator.go
```diff
@@ -681,19 +681,6 @@ func (c *Coordinator) createPrimaryAuthenticatorsForAdminAPICreateUser(ctx conte
 		return nil, err
 	}
 
-	if len(authenticatorInfos) == 0 {
```
louischan-oursky, line 684: Instead of removing this code, it is better to comment out this and add a comment like

> In DEV-3084, we decided to relax this restriction and allow user to be created without a primary authenticator.

portal/src/graphql/adminapi/AddUserScreen.tsx
louischan-oursky: The upside is that now changing among the PasswordCreationType does not reset the form, but `passwordManualEntry` and `sendPasswordManualEntry` is kept.

tung2744:
@fungc-io 

> for instance, the options are not bold in this page

I made it bold because there is another line of description under the option in this page. In endpoint direct access there is only one line.

tung2744:
Updated, thanks!
Changes:
- Comment code in coordinator instead of delete
- Rename the form states and ensure they will not be reset on changing creation type

louischan-oursky:
The code looks good to me. What blocks the merge is stylistic issue reported by Fung.

tung2744:
Rebased and pushed new commits for new ui adjustments. Ready to review.

PR #5391, opened 2025-10-14
Add dpop logs

tung2744:
ref DEV-3087

### DIFF
--- a/pkg/util/slogutil/error_detail_handler.go
+++ b/pkg/util/slogutil/error_detail_handler.go
@@ -2,17 +2,18 @@ package slogutil
 
 import (
 	"context"
-	"fmt"
 	"log/slog"
 
+	"github.com/jba/slog/withsupport"
 	slogmulti "github.com/samber/slog-multi"
 
 	"github.com/authgear/authgear-server/pkg/util/errorutil"
 )
 
 type ErrorDetailHandler struct {
-	Next    slog.Handler
-	Details errorutil.Details
+	Next slog.Handler
+	// See https://github.com/golang/example/blob/master/slog-handler-guide/README.md#the-withgroup-method
+	groupOrAttrs *withsupport.GroupOrAttrs
 }
 
 var _ slog.Handler = (*ErrorDetailHandler)(nil)
@@ -30,44 +31,40 @@ func (s *ErrorDetailHandler) Enabled(context.Context, slog.Level) bool {
 }
 
 func (s *ErrorDetailHandler) Handle(ctx context.Context, record slog.Record) error {
-	details := s.Details
 
-	recordAttrs := []slog.Attr{}
+	allAttrs := []slog.Attr{}
 	record.Attrs(func(a slog.Attr) bool {
-		recordAttrs = append(recordAttrs, a)
+		allAttrs = append(allAttrs, a)
 		return true
 	})
+	s.groupOrAttrs.Apply(func(groups []string, a slog.Attr) {
+		allAttrs = append(allAttrs, a)
+	})
 
-	recordDetail := s.collectDetails(recordAttrs)
-	if recordDetail != nil {
-		details = recordDetail
-	}
+	details := s.collectDetails(allAttrs)
 
+	detailAttrs := []any{}
 	for k, v := range details {
-		record.AddAttrs(slog.Attr{
-			Key:   fmt.Sprintf("details.%s", k),
+		detailAttrs = append(detailAttrs, slog.Attr{
+			Key:   k,
 			Value: slog.AnyValue(v),
 		})
 	}
+	newRecord := record.Clone()
+	newRecord.AddAttrs(slog.Group("details", detailAttrs...))
 
-	if s.Next.Enabled(ctx, record.Level) {
-		return s.Next.Handle(ctx, record)
+	if s.Next.Enabled(ctx, newRecord.Level) {
+		return s.Next.Handle(ctx, newRecord)
 	}
 	return nil
 }
 
 func (s *ErrorDetailHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
-	details := s.collectDetails(attrs)
-	if details != nil {
-		return &ErrorDetailHandler{
-			Next:    s.Next.WithAttrs(attrs),
-			Details: details,
-		}
-	}
 	return &ErrorDetailHandler{
-		Next:    s.Next.WithAttrs(attrs),
-		Details: s.Details,
+		Next:         s.Next.WithAttrs(attrs),
+		groupOrAttrs: s.groupOrAttrs.WithAttrs(attrs),
 	}
+
 }
 
 func (s *ErrorDetailHandler) collectDetails(attrs []slog.Attr) errorutil.Details {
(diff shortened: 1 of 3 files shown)


pkg/lib/dpop/errors.go
```diff
@@ -20,3 +24,20 @@ var ErrInvalidHTU = newInvalidDPoPProofError("htu in the DPoP proof is not a val
 var ErrUnmatchedMethod = newInvalidDPoPProofError("htm in the DPoP proof does not match request method")
 var ErrUnmatchedURI = newInvalidDPoPProofError("htu in the DPoP proof does not match request uri")
 var ErrUnsupportedAlg = newInvalidDPoPProofError("unsupported alg in DPoP jwt")
+
+var UnmatchedJKT = apierrors.Invalid.WithReason("UnmatchedJKT")
```
louischan-oursky: Since we are not going to return this apierror to public, I suggest we do not use apierror to represent this.

The first reason is that this error contains the jkt in the code grant, which should not be returned to the public. So using apierror is prone to incorrect use.

We can just return plain error with formatted error message.

```golang
var ErrMissingJKT = errors.new("dpop: missing jkt")

type ErrorUnmatchedJKT struct {
  ExpectedJKT string
  ActualJKT string
}

func (e *ErrorUnmatchedJKT) Error() string {
  return fmt.Sprintf("dpop: unmatched jkt %s %s", e.ExpectedJKT, e.ActualJKT)
}

func (t *SomethingThatHasJKT) MatchDPoPJKT(proof *dpop.DPoPProof) error {
 // Use ErrMissingJKT and ErrorUnmatchedJKT
}
```

If we really want to support structural logging (slog), we can introduce a new interface in slogutil, and make any error to conform to it.

```golang
// In slogutil

type ToSlogAttrs interface {
  ToSlotAttrs() []slog.Attr
}

// in dpop
var _ slogutil.ToSlogAttrs = &ErrorUnmatchedJKT{}

func (e *ErrorUnmatchedJKT) ToSlogAttrs() []slog.Attr {
  return []slog.Attr{
    slog.String("expected_jkt", e.ExpectedJKT),
    slog.String("actual_jkt", e.ActualJKT),
  }
}
```

pkg/lib/oauth/handler/handler_token.go
```diff
@@ -585,7 +585,11 @@ func (h *TokenHandler) IssueTokensForAuthorizationCode(
 	}
 
 	dpopProof := dpop.GetDPoPProof(ctx)
-	if !codeGrant.MatchDPoPJKT(dpopProof) {
+	if err := codeGrant.MatchDPoPJKT(dpopProof); err != nil {
+		logger.WithSkipLogging().Error(ctx,
+			fmt.Sprintf("failed to match dpop jkt on issue tokens: %s", err.Message),
+			err.Info_ReadOnly.ToSlogAttrs()...,
+		)
```
louischan-oursky, line 592: With my suggestion implemented,

1. We no longer need to call `err.Info_ReadOnly.ToSlogAttrs()...`. The handling of ToSlogAttrs should be handled by slogutil behind the scene. I suggest we add a new handler to do this. This handler has to come before the mask handler. Note that the handler has to use errorutil.Unwrap to traverse the error chain and collect all ToSlogAttrs(). errorutil.Unwrap is smart enough to handle all kinds of error wrapping.
2. Use WithError to include the error.

So this code should look like

```golang
// Assume the slog setup supports ToSlogAttrs for errors, including wrapped ones.
logger.WithSkipLogging().WithError(err).Error(ctx, "failed to match dpop jkt on issue tokens")
```

pkg/util/slogutil/stack_trace_handler.go
```diff
@@ -12,8 +12,11 @@ import (
 
 type StackTraceHandler struct {
 	Next slog.Handler
+	Skip bool
 }
 
+const AttrKeySkipStackTrace = "__authgear_skip_stacktrace"
```
louischan-oursky, line 18: Add tests in stack_trace_handler_test.go to verify SkipStackTrace works as intended.

louischan-oursky (changes requested):
Please also consider Point 4 in this Slack message. https://oursky.slack.com/archives/C03AH1ACM/p1760408666196229?thread_ts=1760349221.931509&cid=C03AH1ACM I meant if the JWT failed to validate due to expiration, the client may have a large clock skew. If the error message is clear enough, we can tell this situation easily.

tung2744:

Changes:
- Added ErrorDetailHandler to add details to the log
- Changed `UnmatchedJKT` to non apierror
- Added unit tests
- Added unique error message to all cases dpop error

pkg/util/slogutil/error_detail_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	return errorutil.CollectDetails(err, nil)
+}
+
+func (s *ErrorDetailHandler) WithGroup(name string) slog.Handler {
+	return &ErrorDetailHandler{
+		Next: s.Next.WithGroup(name),
+	}
+}
```
louischan-oursky, line 93: Please follow the implementation of SkipLoggingHandler. In particular, we need to use withsupport. SkipLoggingHandler is similar to ErrorDetailHandler because they both add extra attributes. So the general logic should be the same.

Problem in the current implementation is, for example, `WithGroup` constructs a new instance of ErrorDetailHandler without the original `Details`.

pkg/lib/dpop/provider.go
```diff
@@ -83,21 +85,31 @@ func (p *Provider) validateProofJWT(header jws.Headers, payload jwt.Token) (jwk.
 		var jwkBytes []byte
 		jwkBytes, err = json.Marshal(jwkIface)
 		if err != nil {
-			return nil, nil, ErrInvalidJwk
+			// Anything in a jwt header should always be a valid json
+			panic(fmt.Errorf("unexpected: cannot marshal jwk in jwt %w", err))
```
louischan-oursky: `... jwt: %w`

louischan-oursky (changes requested):

pkg/util/slogutil/error_detail_handler.go
tung2744: @louischan-oursky Where did you find doc of github.com/jba/slog/withsupport? I cannot find any doc / example about it and not sure what is it for.

tung2744: It seems I just have to retain the `Details` in the new `ErrorDetailHandler` and no need to use withsupport?

louischan-oursky: You can read this https://golang.org/s/slog-handler-guide The doc of OtelMetricHandler mentions this as well. But I forgot to mention it in every handler that follows this pattern. Maybe you can add back the doc for me :)

louischan-oursky: This guide is also mentioned in https://pkg.go.dev/log/slog#hdr-Writing_a_handler

tung2744: @louischan-oursky Updated to use `GroupOrAttrs` in `ErrorDetailHandler`, thanks!

louischan-oursky (approved):

PR #5395, opened 2025-10-16
Track whatsapp metrics

louischan-oursky:
ref DEV-3090

---

Now Whatsapp metrics is tracked in whatsapp service. Do we also want to move the tracking of email metrics and SMS metrics away from lib/messaging?


tung2744:
I think we can refactor email / sms part later when needed

PR #5401, opened 2025-10-23
Fix otp status can be sending forever

tung2744:
ref DEV-3101

### DIFF
--- a/pkg/lib/authn/otp/sender.go
+++ b/pkg/lib/authn/otp/sender.go
@@ -299,8 +299,8 @@ func (s *MessageSender) send(ctx context.Context, opts SendOptions, preferAsync 
 		updateErr := s.updateCodeAfterSent(ctx, opts, afterSentResult{
 			SendError: err,
 		})
-		SenderLogger.GetLogger(ctx).WithError(updateErr).Error(ctx, "failed to update code after sent")
 		if updateErr != nil {
+			SenderLogger.GetLogger(ctx).WithError(updateErr).Error(ctx, "failed to update code after sent")
 			err = errors.Join(err, updateErr)
 		}
 	}()


pkg/lib/authn/otp/sender.go
```diff
@@ -290,29 +288,27 @@ func (s *MessageSender) SendAsync(ctx context.Context, opts SendOptions) error {
...
-	case model.AuthenticatorOOBChannelEmail:
-		err := s.sendEmail(ctx, opts)
-		if err != nil {
-			return err
+func (s *MessageSender) send(ctx context.Context, opts SendOptions, preferAsync bool) (err error) {
+	defer func() {
+		updateErr := s.updateCodeAfterSent(ctx, opts, afterSentResult{
+			SendError: err,
+		})
+		SenderLogger.GetLogger(ctx).WithError(updateErr).Error(ctx, "failed to update code after sent")
```
louischan-oursky: This log should be put in the below if statement.

tung2744: @louischan-oursky Fixed, thanks!

pkg/auth/webapp/session_authflow.go
```diff
@@ -840,15 +841,23 @@ func (s *AuthflowScreenWithFlowResponse) makeVerifyOOBOTPOutputTransformer(chann
...
-					return nil, fmt.Errorf("timed out waiting for delivery_status")
+					return nil, errorutil.WithDetails(fmt.Errorf("timed out waiting for delivery_status"),
+						errorutil.Details{
+							"otp_form":                           data.OTPForm,
+							"channel":                            data.Channel,
+							"failed_attempt_rate_limit_exceeded": data.FailedAttemptRateLimitExceeded,
+							"delivery_status":                    data.DeliveryStatus,
+							"delivery_error":                     data.DeliveryError,
+						},
+					)
```
louischan-oursky, line 858: It seems that details cannot be logged. https://oursky.slack.com/archives/CUUDP0XJQ/p1761230768391579?thread_ts=1761230129.453129&cid=CUUDP0XJQ

tung2744: The error cannot be logged in that event because the error was constructed by `NewWithInfo`. It converted the details to errorutil.DetailTag.

Here I use `WithDetails` so it should be ok.

tung2744: Created https://linear.app/authgear/issue/DEV-3122/handle-tagged-detail-in-logger for the issue you mentioned

PR #5402, opened 2025-10-23
Ignore some known errors and convert to metric

tung2744:
ref DEV-3102

### DIFF
--- a/pkg/util/slogutil/metric.go
+++ b/pkg/util/slogutil/metric.go
@@ -7,10 +7,12 @@ import (
 	"errors"
 	"net"
 	"os"
+	"syscall"
 
 	"github.com/lib/pq"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
+	"golang.org/x/sys/unix"
 
 	"github.com/authgear/authgear-server/pkg/util/otelutil"
 )
@@ -92,6 +94,12 @@ func MetricOptionsForError(err error) []otelutil.MetricOption {
 	if errors.As(err, &netOpError) {
 		opts = append(opts, MetricOptionAttributeKeyValue{attribute.Key("net_op_error.op").String(netOpError.Op)})
 		opts = append(opts, MetricOptionAttributeKeyValue{attribute.Key("net_op_error.net").String(netOpError.Net)})
+
+		var syscallErrno syscall.Errno
+		if errors.As(netOpError.Err, &syscallErrno) {
+			symbolicName := unix.ErrnoName(syscallErrno)
+			opts = append(opts, MetricOptionAttributeKeyValue{attribute.Key("net_op_error.syscall_errno").String(symbolicName)})
+		}
 	}
 
 	return opts
--- a/pkg/util/slogutil/otel_metric_handler_test.go
+++ b/pkg/util/slogutil/otel_metric_handler_test.go
@@ -62,8 +62,12 @@ func TestOtelMetricHandler_Integration(t *testing.T) {
 
 			So(len(trackedErrorOptions), ShouldEqual, 1)
 			opts := trackedErrorOptions[0]
+			So(len(opts), ShouldEqual, 4)
+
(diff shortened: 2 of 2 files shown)


pkg/lib/config/configsource/database.go
```diff
@@ -337,9 +337,16 @@ func (d *Database) invalidateAppByDomain(ctx context.Context, domain string) {
...
 	if err != nil {
 		logger := DatabaseLogger.GetLogger(ctx)
-		logger.WithError(err).Error(ctx, "failed to invalidate app cache by domain",
-			slog.String("domain", domain),
-		)
+		if errors.Is(err, ErrAppNotFound) {
+			// If the domain was deleted
+			logger.Info(ctx, "failed to invalidate app cache because domain not found",
+				slog.String("domain", domain),
+			)
```
louischan-oursky, line 344: The condition is checking AppNotFound, but the error message says domain not found. Maybe there is some misunderstanding here?

pkg/lib/oauthrelyingparty/oauthrelyingpartyutil/oidc.go
```diff
@@ -50,7 +51,7 @@ func FetchOIDCDiscoveryDocument(ctx context.Context, client *http.Client, endpoi
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("failed to fetch OIDC discovery document: unexpected status code: %d", resp.StatusCode)
+		return nil, slogutil.ErrOIDCDiscoveryInvalidStatusCode
```
louischan-oursky: Actually what we are missing is a general mechanism to collect metrics on the outgoing HTTP request we send to the external service. Ideally, we should track https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-client

So I suggest we just leave this alone first.

pkg/lib/config/configsource/database.go
tung2744: This is because the method to find the app is `GetAppIDByDomain`. And it is considered app not found if the domain does not exist with error `ErrAppNotFound`. We don't know the app id here.

pkg/lib/oauthrelyingparty/oauthrelyingpartyutil/oidc.go
tung2744: This error actually can be easily produced by setting a wrong url in a sso config, so we probably don't want to see it in sentry. I think we better ensure the error is not sending to sentry first. Tracking other metric of outgoing http request could be another task to do.

I actually thinks this error should be an error log for the project admin, so he can know he configured something incorrect in his project. However we don't have this feature yet.

louischan-oursky: > I think we better ensure the error is not sending to sentry first.

Then I think we should not introduce a APIError to represent this (yet). Let's use a custom error type and implement `SkipLogging()` to skip logging to sentry first.

> Tracking other metric of outgoing http request could be another task to do.

Agree. This PR should be kept small.

> I actually thinks this error should be an error log for the project admin, so he can know he configured something incorrect in his project. However we don't have this feature yet.

Actually we should have validate the URL before the configuration is saved. This should be done in https://linear.app/authgear/project/support-more-social-logins-5a62ad4313fd/overview

tung2744: I did not add APIError. It is still a simple error constructed with `errors.New`. I moved the package to avoid import cycle. And also added a metric for it so we can count the error numbers if we need.

tung2744: Added a commit to add a custom type of this error. And the metric was removed.

louischan-oursky (approved):
I added a commit to stop tracking syscall.ECONNRESET seperately.

louischan-oursky:
Added 3 more commits to recognize syscall.Errno

PR #5406, opened 2025-10-28
Improve sms error logging

tung2744:
ref DEV-3104

Need https://github.com/authgear/authgear-sms-gateway/pull/28

### DIFF
--- a/pkg/lib/messaging/sender.go
+++ b/pkg/lib/messaging/sender.go
@@ -236,22 +236,7 @@ func (s *Sender) sendSMS(ctx context.Context, msgType translation.MessageType, o
 				options := []otelutil.MetricOption{
 					otelauthgear.WithStatusError(),
 				}
-				if smsapiErr.APIErrorKind != nil {
-					options = append(options, otelauthgear.WithAPIErrorReason(smsapiErr.APIErrorKind.Reason))
-				}
-				if smsapiErr.CustomResponseCode != "" {
-					options = append(options, otelauthgear.WithCustomResponseCode(smsapiErr.CustomResponseCode))
-				}
-				if smsapiErr.ProviderName != "" {
-					options = append(options, otelauthgear.WithProviderName(smsapiErr.ProviderName))
-				}
-				if smsapiErr.ProviderType != "" {
-					options = append(options, otelauthgear.WithProviderType(smsapiErr.ProviderType))
-				}
-				if smsapiErr.ProviderErrorCode != "" {
-					options = append(options, otelauthgear.WithProviderErrorCode(smsapiErr.ProviderErrorCode))
-				}
-				options = append(options, otelauthgear.WithIsNonCritical(smsapiErr.IsNonCritical))
+				options = append(options, s.applySMSAPIErrorMetrics(smsapiErr)...)
 				otelutil.IntCounterAddOne(
 					ctx,
 					otelauthgear.CounterSMSRequestCount,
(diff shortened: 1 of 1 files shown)


pkg/lib/infra/sms/twilio/client.go
```diff
@@ -149,6 +149,7 @@ func (t *TwilioClient) makeError(
 	err := &smsapi.SendError{
 		DumpedResponse:    dumpedResponse,
 		ProviderName:      "twilio",
+		ProviderType:      "twilio",
```
louischan-oursky: Cross-check other locations as well

```
$ rg 'SendError\{'
pkg/lib/infra/sms/twilio/client.go
90:             return nil, nil, errors.Join(err, &smsapi.SendError{
110:            return errors.Join(err, &smsapi.SendError{
132:                    return &smsapi.SendError{
136:                    return errors.Join(err, &smsapi.SendError{
149:    err := &smsapi.SendError{

pkg/lib/infra/sms/custom/client.go
95:             return errors.Join(err, &smsapi.SendError{
107:            return &smsapi.SendError{
170:            return errors.Join(err, &smsapi.SendError{
225:    err := &smsapi.SendError{

pkg/lib/infra/sms/nexmo/client.go
68:             return nil, nil, errors.Join(err, &smsapi.SendError{
84:             return errors.Join(err, &smsapi.SendError{
93:             return &smsapi.SendError{
99:                     return &smsapi.SendError{
```

pkg/lib/infra/sms/smsapi/api.go
```diff
@@ -78,6 +78,9 @@ type SendError struct {
 	ProviderType      string          `json:"provider_type,omitempty"`
 	ProviderErrorCode string          `json:"provider_error_code,omitempty"`
 	IsNonCritical     bool            `json:"is_non_critical,omitempty"`
+
+	// custom provider specific
+	CustomResponseCode string `json:"custom_response_code,omitempty"`
```
louischan-oursky, line 83: I prefer we just name it `code`. And it does not need to be custom provider specific. Actually we can translate Twilio error code into this code.

pkg/lib/infra/sms/custom/model.go
```diff
@@ -22,6 +22,7 @@ type ResponseBody struct {
 	ProviderType      string `json:"provider_type,omitempty"`
 	ProviderErrorCode string `json:"provider_error_code,omitempty"`
 	GoError           string `json:"go_error,omitempty"`
+	IsNonCritical     bool   `json:"is_non_critical,omitempty"`
```
louischan-oursky, line 25: As mentioned in another PR, we should use `code` to determine whether to log it.

tung2744: I added this flag because it is not enough to use a single code to do it. To decide whether to log the error, we need to know which provider we do not want to log, and which provider we want to log. And we do not have this information in authgear server.

pkg/lib/infra/sms/smsapi/api.go
tung2744: In authgear server, sms errors are classified by APIError Reason. Therefore this `Code` is custom provider specific. I don't want to add new reason to APIError, but at the same time I want to have a field to store the raw code returned by custom client for logging purpose, therefore we have this field.

pkg/lib/infra/sms/custom/model.go
louischan-oursky: For timeout, we do not need to log it to Sentry. We can track it with metrics and set up alerts.

For the special case that accessyou sometimes return non-json, I think we can still log it.

tung2744: > For timeout, we do not need to log it to Sentry. 

There are different reasons for timeout, which is mentioned in the go error message. So maybe it is still useful to log to sentry, especially for providers which usually do not timeout. Accessyou seems to be the special case here. The json error should share the same rationale. Basically the logic here is only to ignore the specific error we want to ignore and leave other untouched.

I actually think we better do the logging in the sms gateway side so we don't have to break the custom gateway / authgear server abstraction. And in authgear server, we log any error between custom webhook with authgear server as metric.

pkg/lib/infra/sms/twilio/client.go
tung2744: I only set ProviderType if originally ProviderName is set. Let me add back these two fields to all construction of SendError

pkg/lib/infra/sms/custom/model.go
louischan-oursky: The reason I prefer doing the logging in Authgear side is because we do not deploy authgear-sms-gateway in other environments. Centralizing the logging in Authgear ensures we always log no matter whether the custom gateway is written by us or not.

tung2744:
I will create another pr for this

PR #5407, opened 2025-10-28
Handle the case that otp code is not generated therefore no delivery status

tung2744:
ref DEV-3074

How to reproduce:
1. Set the cooldown to 60s
2. Go to login with a phone number
3. Saw otp page, enter a correct otp to consume the code
4. Logout
5. Within the cooldown, login again by entering the same phone number
6. See panic because the code is not generated, and delivery_status is ""

### DIFF
--- a/pkg/auth/webapp/session_authflow.go
+++ b/pkg/auth/webapp/session_authflow.go
@@ -859,10 +859,6 @@ func (s *AuthflowScreenWithFlowResponse) makeVerifyOOBOTPOutputTransformer(chann
 				}
 
 				switch data.DeliveryStatus {
-				case "":
-					// Empty string indicates the code is not generated, maybe still in cooldown
-					// Simply proceed to next page
-					return output, err
 				case model.OTPDeliveryStatusFailed:
 					if apierrors.IsKind(data.DeliveryError, whatsapp.WhatsappMessageStatusCallbackTimeout) {
 						if !hasNonWhatsappChannels {
@@ -878,7 +874,7 @@ func (s *AuthflowScreenWithFlowResponse) makeVerifyOOBOTPOutputTransformer(chann
 					return output, data.DeliveryError
 				case model.OTPDeliveryStatusSent:
 					return output, err
-				case model.OTPDeliveryStatusSending, model.OTPDeliveryStatusPending:
+				case model.OTPDeliveryStatusSending:
 					{
 						// Wait until sent or failed
 						time.Sleep(500 * time.Millisecond)
--- a/pkg/api/model/otp.go
+++ b/pkg/api/model/otp.go
@@ -3,8 +3,7 @@ package model
 type OTPDeliveryStatus string
 
 const (
-	OTPDeliveryStatusPending = "pending"
-	OTPDeliveryStatusSending = "sending"
-	OTPDeliveryStatusFailed  = "failed"
-	OTPDeliveryStatusSent    = "sent"
+	OTPDeliveryStatusSending OTPDeliveryStatus = "sending"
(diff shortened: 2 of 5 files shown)


pkg/auth/webapp/session_authflow.go
```diff
@@ -859,6 +859,10 @@ func (s *AuthflowScreenWithFlowResponse) makeVerifyOOBOTPOutputTransformer(chann
 				}
 
 				switch data.DeliveryStatus {
+				case "":
+					// Empty string indicates the code is not generated, maybe still in cooldown
+					// Simply proceed to next page
+					return output, err
```
louischan-oursky: The problem is that VerifyOOBOTPData is part of the public API, and leaving it as empty `means the code is not generated` is not intuitive. Instead, we should make sure it always output something.

I checked the code and VerifyOOBOTPData is constructed by NewVerifyOOBOTPData, the delivery status is from InspectState. I think we should think InspectState instead of checking empty string here.

I think we should output `pending`

tung2744: @louischan-oursky Actually, the current api is not only outputting empty string for delivery_status. If you look at InspectState, you will find that all other field require a `Code` will be empty if the code does not exist, therefore I keep it as empty.

> I think we should output pending

We cannot use `pending`, because "not generated" is actually another state. If the ui received this state, the ui should not wait for status change. We need another string value for it if we do not want empty string here. And seems empty string is clear enough that it means the otp is not generated.

louischan-oursky: > And seems empty string is clear enough that it means the otp is not generated.

I think a random guy looking at the API response `"delivery_status": ""` probably will not have an idea what it means.

What does `pending` mean actually?

tung2744: In the current implementation:
`pending` means the server did not finish all the work required to send the code yet. (e.g. Still calling whatsapp api, the api did not return yet)
`sending` means the server did finish all work to send the code so the code should be already sending to your phone. (e.g. Called whatsapp api, waiting for webhook update)

Originally, these two status are both `sending`. We added `pending` in previous pr change to distinguish these two cases. I think this could be useful for distinguishing problems like "we did not receive a webhook from whatsapp" between "we failed to send the whatsapp at all".

> I think a random guy looking at the API response "delivery_status": "" probably will not have an idea what it means.

What do you think if we call it `not-sending`?

I actually think the problem is the current api did not disclose whether the code was generated or not, therefore we have such problem. If the user knows the code was not generated (Or does not exist), they should be able to understand `delivery_status` is not relevant.

louischan-oursky: Whether the code is generated or not should be kept secret from the user. Whether the code was not generated or was `pending` is not distinguishable from the user's point of view. The observable fact is the same: the user will not receive the code.

Do you mean if we output `pending`, then the client (whether or not it is AuthUI) will not be able to display a resend button to the end-user?

tung2744: `pending` only means the authgear server have not call the sms/whatsapp/smtp api yet, does not mean it will never call.

The expected action for the auth ui is:
pending: wait until it is sending or sent
sending: wait until it is sent
failed: show error
"" (Not generated): Our auth ui handled this case as entering the otp page pretending it was sent. But I think some ui might prefer showing "Please try again later".

tung2744:
@louischan-oursky Updated according to our latest discussion:
1. Defined all fields of otp State when otp does not exist
2. Removed `pending` state in api status
3. Return `sent` in api if the otp does not exist

louischan-oursky (approved):

PR #5410, opened 2025-10-30
Refine sms gateway api and logging

tung2744:
ref DEV-3104

### DIFF
--- a/pkg/lib/messaging/metric_test.go
+++ b/pkg/lib/messaging/metric_test.go
@@ -3,12 +3,13 @@ package messaging
 import (
 	"testing"
 
+	. "github.com/smartystreets/goconvey/convey"
+
 	"github.com/authgear/authgear-server/pkg/api/apierrors"
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	smsapi "github.com/authgear/authgear-server/pkg/lib/infra/sms/smsapi"
 	"github.com/authgear/authgear-server/pkg/lib/otelauthgear"
 	"github.com/authgear/authgear-server/pkg/util/otelutil"
-	. "github.com/smartystreets/goconvey/convey"
 )
 
 func TestApplySMSAPIErrorMetrics(t *testing.T) {


docs/specs/sms_gateway.md
```diff
@@ -135,33 +134,74 @@ And then use `template_variables` to interpolate the template.
...
+  // A message authgear server might display to the end-user when there is an error.
+  description?: string;
+
+  // An optional JSON object. Only for error logging purpose.
+  // If exist, authgear server will send all details in the object to the error logging service (Sentry).
+  info?: Record<string, any>;
 }
 ```
 
+#### Response Codes
```
tung2744, line 163: Something we might want to further discuss:
1. Finally I did not add `unknown_response`, because I think it is difficult to explain to reader (The implementer of SMS Custom gateway) the difference between `unknown_response` and `unknown_error`. In both cases, the authgear server react with same behavior. As a result, unknown response error will still be logged to sentry.
2. Looks like `timeout` can be renamed to `temporary_error` which seems to be closer to the actual meaning: Authgear server should report error to end-user but ask them to try again.

docs/specs/sms_gateway.md
```diff
@@ -135,33 +134,74 @@ And then use `template_variables` to interpolate the template.
...
-  dumped_response?: string;
+  // The code indicating the result. Read the Response Code section for details.
+  code: CustomSMSGatewayResponseCode;
+
+  // A message authgear server might display to the end-user when there is an error.
+  description?: string;
+
+  // An optional JSON object. Only for error logging purpose.
+  // If exist, authgear server will send all details in the object to the error logging service (Sentry).
+  info?: Record<string, any>;
```
tung2744, line 159: Maybe we can put info into `sms.error` event payload?

docs/specs/sms_gateway.md
louischan-oursky: 1. Agree we can just have `unknown_error`.
2. Let's keep `timeout`.

resources/authgear/templates/en/translation.json
```diff
@@ -1016,6 +1016,7 @@
   "v2.error.sms-gateway-delivery-rejected": "SMS delivery failed. The SMS provider rejected the message. If the problem persists, please contact support.",
   "v2.error.sms-gateway-invalid-phone-number": "Please enter a correct phone number.",
   "v2.error.sms-gateway-rate-limited": "High-traffic on SMS service. Please try again later.",
+  "v2.error.sms-gateway-unsupported-request": "SMS delivery failed. Developer: The SMS provider does not support the request. {Description}",
```
louischan-oursky, line 1019: Not sure if it will crash or render `nil` if Description is nil

pkg/util/slogutil/metric.go
```diff
@@ -90,6 +104,13 @@ func MetricOptionsForError(err error) []otelutil.MetricOption {
 		opts = append(opts, MetricOptionAttributeKeyValue{attribute.Key("error_name").String(string(errorName))})
 	}
 
+	var metricErr MetricError
+	if errors.As(err, &metricErr) {
+		if _, ok := metricErr.GetMetricErrorName(); ok {
+			opts = append(opts, metricErr.GetMetricOptions()...)
+		}
+	}
+
```
louischan-oursky, line 113: Add unit test for the changes in this commit.

pkg/lib/infra/sms/custom/client.go
```diff
@@ -223,30 +223,37 @@ func (c *CustomClient) Send(ctx context.Context, opts smsapi.SendOptions) error
...
-		err.ProviderName = responseBody.ProviderName
-	} else {
-		err.ProviderName = gatewayType
-	}
-	if responseBody.ProviderErrorCode != "" {
-		err.ProviderErrorCode = responseBody.ProviderErrorCode
+
+	// Handle info returned by authgear-sms-gateway
+	if responseBody.Info != nil {
+		if providerName, ok := responseBody.Info["provider_name"].(string); ok {
```
louischan-oursky, line 232: It does not handle `"provider_type"`. It should store that to CustomProviderType, and include it as attribute in metrics.

The following is an evidence.
SCR-20251031-kzbb

resources/authgear/templates/en/translation.json
tung2744: I've tested and it won't crash.
Description will be empty string if the gateway response does not have such key.

pkg/lib/infra/sms/custom/client.go
tung2744: @louischan-oursky Fixed, thanks!

louischan-oursky:
The build failed due to `make check-tidy`. Otherwise, it looks good to me.

tung2744:
Fixed the ci, thanks!

louischan-oursky (approved):

PR #5413, opened 2025-11-03
Specifiy SDK settings actions

tung2744:
ref DEV-3027

### DIFF
--- a/docs/specs/user-profile/design.md
+++ b/docs/specs/user-profile/design.md
@@ -487,6 +487,14 @@ A boolean. True if the user is a verified user. i.e. Have at least one verified 
 
 A boolean. True if the user can perform reauthentication.
 
+#### https://authgear.com/claims/user/authenticators
+
+A list of authenticators owned by the user. See [this doc](../sdk-settings-actions.md).
+
+#### https://authgear.com/claims/user/recovery_code_enabled
+
+A boolean. See [this doc](../sdk-settings-actions.md).
+
 
 ## Webhook
 


docs/specs/user-profile/discussions.md
```diff
@@ -0,0 +1,16 @@
+### https://authgear.com/claims/user/has_primary_password
```
tung2744: @chpapa We would like to add a new oidc claim to the userinfo endpoint, to tell the frontend if the user has a primary password or not.

We decide to add a simple boolean flag for it instead of exposing authenticator details in the hook event payload.

Maybe see if you have time to take a look on this change and see if you have any opinion on it?

chpapa: While we only have one `has_primary_password` now, I envision we will probably need to add all of the other authenticators type in future.

So, what is the different between:

- Have multiple “simple boolean flag” to indicate if an certain type  of authenticator is available.
- Have a single claim with a list of all authenticators that is available.

Both options seems identical, while the list seems simpler and reduce the number of claims?

What are the attack vectors considered if we expose authenticators a user have on `user_info` endpoint?

tung2744: > While we only have one has_primary_password now, I envision we will probably need to add all of the other authenticators type in future.

> So, what is the different between:
> - Have multiple “simple boolean flag” to indicate if an certain type of authenticator is available.
> - Have a single claim with a list of all authenticators that is available.
Both options seems identical, while the list seems simpler and reduce the number of claims?


1. We will not add has_xx_xx flag for all kind of authenticators. The reason the frontend need to know if the user has a primary password or not, is because in our SDK we have a method `changePassword()`. The frontend need to know whether the user has a password to decide whether to show this button or not. And I think we have no plan to add other methods like `changeTOTP()`, `changeOOBOTPPhone()`, therefore we might not need to add flags for all authenticators. However, we do considered adding a method `updateMFA()` in the SDK could be useful, therefore there could be another flag `hasMFA` in the future.
2. We think a simple boolean in the UserInfo is easier to use and understand for the app developer, compared to an array of authenticator objects. We have done some research in this thread comparing how to do the same thing in different auth providers: https://linear.app/authgear/issue/DEV-3027#comment-16467205. And our opinion is if we need to write a hook to compute this flag, it seems to be over complicated.

> What are the attack vectors considered if we expose authenticators a user have on userinfo endpoint?

Actually, if we only expose the authenticator type (primary_password, primary_oob_otp_email, etc) in the userinfo endpoint, the two options should be the same in terms of security. That is, the client side app now have information about what login methods the user has enabled. And we think this should be acceptable because we only support first party app for now. We can hide this flag in third party app in the future when it is supported.

I think mainly it is about easier for understanding and use. And we did not plan to add things like `has_totp` or `has_secondary_oob_otp_email`. `has_mfa` might be useful, tho.

chpapa: 1. Is that `changePassword()` will fire up the setting page and go to the Change Password screen directly? If yes I remember we had the same conversations before that we will add `changeTOTP()` and `changeOOBOTPPhone()` (although probably won’t call it this way? lol), so is that applying the same logic we will have all kind of authenticators flags eventually?

2. I can understand the arguments of a simple Boolean is easier than an array, but it seems very tedious? However if Boolean is really so much better than array than it is ok. Yet I think we better add all reasonable options at once, rather than add it one by one (so we know how it impact the API design eventually?)

> I think mainly it is about easier for understanding and use. And we did not plan to add things like has_totp or has_secondary_oob_otp_email. has_mfa might be useful, tho.

Um, I think you should think of why users want a function to jump to change password page directly? Than what else will make sense?

Than my question will be, why we did not plan to add things like `has_totp` or `has_secondary_oob_otp_email`? They seems equally make sense (maybe TOTP is not as it is not quite as popular, but `oob_top_email` seems quite likely will be needed?)

chpapa: Quick brainstorm, maybe we just shouldn’t add this option in user_info / claims at all? But to make it easy for users to customize the JWT claims / user_info and add these type of options in it?

tung2744: > Is that changePassword() will fire up the setting page and go to the Change Password screen directly? If yes I remember we had the same conversations before that we will add changeTOTP() and changeOOBOTPPhone() (although probably won’t call it this way? lol), so is that applying the same logic we will have all kind of authenticators flags eventually?

We have the following types of authenticators:

- Primary Password: We have `changePassword` for this.
- Primary OOB OTP Phone / Email: These are tied to the email / phone identity of the user, so they should use the existing changeEmail / changePhone instead. They can check the `email` or `phone_number` claim in userinfo to know if the user have email / phone number or not.
- Passkeys: Probably need to be designed together with native passkey support, so I think not to do it in this spec.

- Secondary TOTP
- Secondary OOB OTP Email / Phone
- Recovery Codes (Technically not an authenticator, but is related to user's MFA)

For the above three, I agree that there could be a need to add methods to manage these authenticators in SDK. We've discussed this in our standup, and concluded that usually they will be shown as a single button "Manage your 2FA methods". This is why I've mentioned we may want `hasMFA` and `manageMFA()`.
However, if the app want to self implement a ui for managing MFA, these methods / flags could be useful. Maybe let me discuss with teammates again for this.




> Quick brainstorm, maybe we just shouldn’t add this option in user_info / claims at all? But to make it easy for users to customize the JWT claims / user_info and add these type of options in it?

This is indeed another way we've considered. But it will be the first time a `Authenticator` object appears in the hook payload. So if we can solve if by boolean maybe we don't want to add one more type of object to hook payloads yet.

Another problem is that we don't have a hook to customize the userinfo claims yet. (We can do it on user.pre_create, but that is only called on user creation therefore cannot handle cases the user added password later).

louischan-oursky: > Quick brainstorm, maybe we just shouldn’t add this option in user_info / claims at all? But to make it easy for users to customize the JWT claims / user_info and add these type of options in it?

Actually we can just add the new information (either boolean or other forms, we are still discussing this) to the user model. I mean [this](https://github.com/authgear/authgear-server/blob/master/portal/src/graphql/adminapi/schema.graphql#L849) and [this](https://docs.authgear.com/customization/events-hooks/blocking-events#user.pre_create) 

The downside is that it requires the developer to always write a hook to tell whether a user has primary password or not. According to https://linear.app/authgear/issue/DEV-3027#comment-16467205 Auth0 does this.

Ben brought up a good point that whether we should add this information to userinfo. Tung said we can conditionally expose this information depending on 1st party client / 3rd party client. Though this behavior is not hard to document, but my gut feeling is that people will be surprised by this behavior.

If we take a step back and look at why we are adding this feature, it is we have a client whose project has enabled password historically, and now they plan to remove it. So the client wants to know whether a user has password or not. I think this is not a usual case. Maybe our design does not need to make unusual case so easy to do.

> Than my question will be, why we did not plan to add things like has_totp or has_secondary_oob_otp_email? They seems equally make sense (maybe TOTP is not as it is not quite as popular, but oob_top_email seems quite likely will be needed?)

Tung mentioned that the developer can check if the user has `email` or `phone_number` or `preferred_username` to determine whether it is possible to change/remove email/phone/username. This is not always true. For example, if the project has Email Login ID and Google Login enabled, a user with Google Login will have `email`, but that email is not modifiable.

If we follow the convention of OAuth2/OIDC to expose information, then we should introduce a new scope to represent the scope for this information (whether a user has a particular kind of identity/authenticator). For 1st party client, since they are using the superpower scope `https://authgear.com/scopes/full-access`, this information is exposed. For 3rd party client, the scope is required in order to have this information exposed in userinfo/id token/jwt access token without hook.

We introduced a list of proprietary AMR values to represent our authenticator/recovery code when we implemented Authentication hooks. We can introduce a list of proprietary values to represent what identities the user has, and then we can introduce two new claims like this

```json
{
  "https://authgear.com/claims/user/owned_authentication_methods": ["x_primary_password", "x_primary_oob_otp_email", "x_primary_passkey", "x_secondary_totp", "x_recovery_code"],
  "https://authgear.com/claims/user/owned_identification_methods": ["email", "phone", "username", "oauth:google", "oauth:azureadv2", "passkey"]
}
```

These 2 claims are exposed in userinfo if the access token being used bear the new scope. Since ID token and access token may be passed around, they are not included. The developer has to write their hooks for this.

tung2744: I think we should use the authentication value matching authflow api and the event payload, mentioned in [this spec](https://github.com/authgear/authgear-server/blob/main/docs/specs/event_models.md#authentication). So it should be:
```
["primary_password", "primary_oob_otp_email", "recovery_code"] // authentication
["email", "phone", "username", "oauth"] // identification
```

Note that compared to louis's suggested values, these values does not disclose the linked oauth provider.

louischan-oursky: Can you prepare a draft based on what we have discussed so far so we have a full picture of how this design looks like now? Thank you.

tung2744:
@chpapa @louischan-oursky @fungc-io 

I've made updated the PR to add two documents for the full design of setting actions:

1. https://github.com/authgear/authgear-server/blob/b229afbb75894b3325f15f10da95806907bf27f5/docs/specs/sdk-settings-actions.md
2. https://github.com/authgear/authgear-server/blob/b229afbb75894b3325f15f10da95806907bf27f5/docs/specs/sdk-settings-actions-alternate.md

The first one does not expose `authenticator` in UserInfo, and the second one do expose it.

As you may wonder, according to our latest discussion we should be talking about exposing `owned_authentication_methods` instead of authenticators, why in the latest spec I added a design for `authenticators`?
This is because I found that if we plan to support managing MFAs in the SDK, we will need a way to obtain details about the MFA authenticators like the email or phone number, so only a authentication name is not enough. I believe you should be able to get the idea looking at the latest API design.

Let me know if any further comments, thanks!

fungc-io:
The first design is easier to understand, it's more intuitive, and users don't need to understand the authenticator model. 
I do agree with the concern this could be dangerous if we expose this to third party app, perhaps it needs to be controlled with a scope?

docs/specs/sdk-settings-actions.md
```diff
@@ -0,0 +1,292 @@
+# SDK - Settings Actions
```
louischan-oursky, line 1: I prefer this design. It is much easier to understand. No need to understand what is an Authgear authenticator.

docs/specs/sdk-settings-actions.md
```diff
@@ -0,0 +1,292 @@
...
+### Intention
+
+App developers might want custom button to trigger the UI in native App which manage the user's Phone.
+
+### SDK Design
+
+- Add Phone
+
+```typescript
+await authgear.addPhone({ redirectURI: "com.example://complete" });
```
louischan-oursky, line 61: Actually since our claim in userinfo is `phone_number`, maybe we want to be consistent and rename these `addPhoneNumber`, `changePhoneNumber`, etc.

docs/specs/sdk-settings-actions.md
```diff
@@ -0,0 +1,292 @@
...
+### Intention
+
+App developers might want custom button to trigger the UI in native App which manage the user's Username.
+
+### SDK Design
+
+- Add Username
+
+```typescript
+await authgear.addUsername("example", {
```
louischan-oursky, line 90: Maybe we rename these to `addPreferredUsername`, ... for consistency?

docs/specs/sdk-settings-actions.md
```diff
@@ -0,0 +1,292 @@
...
+{
+  "sub": "00000000-0000-0000-0000-000000000000",
+  "email": "user@example.com",
+  "phone_number": "+85211111111",
+  "preferred_username": "example",
+  "https://authgear.com/claims/user/password_enabled": true,
+  "https://authgear.com/claims/user/mfa_enabled": true,
+  "https://authgear.com/claims/user/mfa_emails": ["mfa@example.com"],
+  "https://authgear.com/claims/user/mfa_phone_numbers": ["+85212222222"],
+  "https://authgear.com/claims/user/mfa_password_enabled": true
```
louischan-oursky: For reference: In userimport, we have a similar model

```json
{
  "mfa": {
    "email": "user@example.com",
    "phone_number": "+85298765432",
    "password": {
      "type": "bcrypt",
      "password_hash": "$2a$10$N9qo8uLOickgx2ZMRZoMyeIjZAgcfl7p92ldGxad68LJZdL17lhWy"
    },
    "totp": {
      "secret": "secret"
    }
  }
}
```

1. What is `mfa_enabled` true? Either emails / phone numbers / password is non-empty?
2. Do we take the project config into account when deriving `mfa_enabled`?
3. Do we take the project config into account when we populate `mfa_emails`, `mfa_phone_numbers`, `mfa_password_enabled`?

The claims are actually flat in userinfo endpoint, so my gut feeling is to keep the same flat structure in the SDK. so `mfaEnabled`, `mfaEmails`, ...

docs/specs/sdk-settings-actions.md
tung2744: I called it Phone because in our UI, such as settings and Portal, we called it Phone. But I do agree if looking at the claim, it should be named `addPhoneNumber`

docs/specs/sdk-settings-actions.md
tung2744: Same as the above `Phone` comment, I named it according to our terms in UI.

docs/specs/sdk-settings-actions.md
tung2744: > 1. What is mfa_enabled true? Either emails / phone numbers / password is non-empty?

True if the user have any active (In config) secondary authenticator. i.e. he can see something in his settings related to MFA tab.

> Do we take the project config into account when deriving mfa_enabled?

Yes. It should sync with what the user see in settings.

> Do we take the project config into account when we populate mfa_emails, mfa_phone_numbers, mfa_password_enabled?

Yes.

Basically the rationale is what the user see in claims should be always matching what he see in settings.

BTW, I find that I've missed TOTP in this design. Let me add it back.

louischan-oursky: Recovery code is missing too.

tung2744:
> The first design is easier to understand, it's more intuitive, and users don't need to understand the authenticator model. I do agree with the concern this could be dangerous if we expose this to third party app, perhaps it needs to be controlled with a scope?

Yes, we should check the `https://authgear.com/scopes/full-access` scope. Let me mention this in the spec as well.

tung2744:
- Follow authenticators design
- May need to review primary / secondary concept

tung2744:
Added a commit to mention we've chosen the authenticator design

louischan-oursky:
Please rebase.

tung2744:
Rebased, thanks!

docs/specs/user-model.md
```diff
@@ -726,3 +728,10 @@ account_anonymization:
   # scheduled_by_end_user_enabled: false
   grace_period_days: 30
 ```
+
+
+## Special Attributes
+
+### has_primary_password
+
+A computed boolean. True if the user has a primary password.
```
louischan-oursky: I think we do not have this implemented, right?

tung2744: You are right, and I think we won't add this flag in future. So let me remove it

tung2744:
Rebased to drop irrelevant commits, and added one commit to document the new claims in user profile spec. Thanks!

PR #5414, opened 2025-11-03
Add has_primary_password

tung2744:
ref DEF-3027


tung2744:
Closing because we will go for another design

PR #5416, opened 2025-11-04
Log wrapped details correctly in sentry

tung2744:
ref DEV-3122

### DIFF
--- a/pkg/util/slogutil/error_detail_handler.go
+++ b/pkg/util/slogutil/error_detail_handler.go
@@ -82,7 +82,8 @@ func (s *ErrorDetailHandler) collectDetails(attrs []slog.Attr) errorutil.Details
 		return nil
 	}
 
-	return errorutil.CollectDetails(err, nil)
+	d := errorutil.CollectDetails(err, nil)
+	return errorutil.UnsafeUnwrapDetails(d)
 }
 
 func (s *ErrorDetailHandler) WithGroup(name string) slog.Handler {
--- a/pkg/util/errorutil/details_kv.go
+++ b/pkg/util/errorutil/details_kv.go
@@ -31,6 +31,20 @@ func FilterDetails(d Details, tag DetailTag) Details {
 	return fd
 }
 
+// This method unwrap all details without looking at the Tag
+// Use it with caution
+func UnsafeUnwrapDetails(d Details) Details {
+	fd := Details{}
+	for key, value := range d {
+		if tv, ok := value.(DetailTaggedValue); ok {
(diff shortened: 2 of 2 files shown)


pkg/util/slogutil/error_detail_handler.go
```diff
@@ -82,7 +82,8 @@ func (s *ErrorDetailHandler) collectDetails(attrs []slog.Attr) errorutil.Details
 		return nil
 	}
 
-	return errorutil.CollectDetails(err, nil)
+	d := errorutil.CollectDetails(err, nil)
+	return errorutil.UnsafeUnwrapDetails(d)
```
louischan-oursky, line 86: Instead of using an unsafe function, I think we should fix the source position to explicitly declare what details should be logged.

The current behavior is actually correct, do not log unless it is absolutely safe to do so.

tung2744: But I think safe logs should be handled by masking?
If I understand it correctly, the detail tags are used to mark a data safe to output in API, but not safe to log.
Currently, any tagged values cannot be logged correctly, including values tagged as safe.

louischan-oursky: 1. Currently we primarily use `NewWithInfo` to construct an API error with info. This function wraps every key as APIErrorDetail so the information is exposed to API response
2. We have some use cases of injecting info that is not APIErrorDetail. I guess this is for callers in the call stack to get more details from the error and act differently.
3. For the use case of 2, I think this should be replaced by `errors.As(err, &a_strong_error_type)`. This makes the error easier to understand and trace.
4. If we have done 3, then the info is purely for returning to API. There is a distinction though. Safe to return to API != Safe to log. I think we should apply our masking pipeline to the info before we log it.
5. This implies we can eliminate `errorutil.DetailTag` entirely.

tung2744: I agree to refactor the code for 2, 3 and 5, but it is a big refactoring which we might not want to handle in this pr.
For 4, we already applied the masking before we log it, so the logs produced in this pr are still masked.

louischan-oursky: If we refer back to the original error, the details are the stderr and the stdout of a user-written hook. The content may contain PII like email addresses, URLs, and phone numbers that our current masking mechanism do not handle.

So I think we better not fix this issue now. What is your thoughts?

tung2744: Agree. Lets leave it first and keep the current behavior.

PR #5417, opened 2025-11-04
Allow using fallback languages for links

carmenlau:
ref DEV-3092

### DIFF
--- a/pkg/util/template/translation.go
+++ b/pkg/util/template/translation.go
@@ -34,6 +34,9 @@ var appSpecificKeysRegex = []*regexp.Regexp{
 	regexp.MustCompile(`^email\..+\.sender$`),
 	regexp.MustCompile(`^email\..+\.reply-to$`),
 	regexp.MustCompile(`^sms\..+\.sender$`),
+	regexp.MustCompile(`^terms-of-service-link$`),
+	regexp.MustCompile(`^privacy-policy-link$`),
+	regexp.MustCompile(`^customer-support-link$`),
 }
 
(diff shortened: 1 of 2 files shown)


pkg/util/template/translation.go
```diff
@@ -34,6 +34,9 @@ var appSpecificKeysRegex = []*regexp.Regexp{
 	regexp.MustCompile(`^email\..+\.sender$`),
 	regexp.MustCompile(`^email\..+\.reply-to$`),
 	regexp.MustCompile(`^sms\..+\.sender$`),
+	regexp.MustCompile(`^terms-of-service-link$`),
+	regexp.MustCompile(`^customer-support-link$`),
+	regexp.MustCompile(`^privacy-policy-link$`),
```
louischan-oursky, line 38: Add a test case in pkg/util/template/translation_test.go. It simply be the case mentioned in the issue. The test can also serve as a documentation of the intended behavior.

carmenlau: Updated, thanks!

louischan-oursky: Sorry I didn't make myself very clear. I actually mean adding a test as a sibling to https://github.com/authgear/authgear-server/blob/main/pkg/util/template/translation_test.go#L227-L250 like this

```
Convey("it should use the privacy-policy-link defined in the default language", func() {... })
Convey("it should use the customer-support-link defined in the default language", func() { ... })
Convey("it should use the terms-of-service-link defined in the default language", func() { ... })
```

carmenlau: Added, thank you!!

PR #5418, opened 2025-11-05
Update vite server config for portal design preview

carmenlau:


louischan-oursky (approved):
Thanks for taking a deeper look into why `sandbox="allow-scripts allow-same-origin"` is actually a bad idea.

PR #5421, opened 2025-11-10
Expose user account status related fields in user export

tung2744:
ref DEV-3156

### DIFF
--- a/pkg/lib/userexport/model.go
+++ b/pkg/lib/userexport/model.go
@@ -21,8 +21,8 @@ const PresignGetExpiresForUserExport time.Duration = 1 * duration.PerMinute
 const BatchSize = 1000
 
 var defaultCSVExportFields = []*FieldPointer{
+	// Standard claims
 	&FieldPointer{Pointer: "/sub"},
-	&FieldPointer{Pointer: "/created_at"},
 	&FieldPointer{Pointer: "/preferred_username"},
 	&FieldPointer{Pointer: "/email"},
 	&FieldPointer{Pointer: "/phone_number"},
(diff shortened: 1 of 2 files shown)


pkg/lib/userexport/model.go
```diff
@@ -48,6 +48,11 @@ var defaultCSVExportFields = []*FieldPointer{
 	&FieldPointer{Pointer: "/groups"},
 	&FieldPointer{Pointer: "/disabled"},
 	&FieldPointer{Pointer: "/delete_at"},
+	&FieldPointer{Pointer: "/last_login_at"},
```
louischan-oursky, line 55: I think we should also add `/is_anonymized`, `/anonymize_at`, `/anonymized_at`

pkg/lib/userexport/model.go
```diff
@@ -22,6 +22,7 @@ const BatchSize = 1000
 
 var defaultCSVExportFields = []*FieldPointer{
 	&FieldPointer{Pointer: "/sub"},
+	&FieldPointer{Pointer: "/created_at"},
```
louischan-oursky: For this list, it is better to group them into standard claims, and our proprietary claims, separated by a newline.

```
&FieldPointer{Pointer: "/sub"},
&FieldPointer{Pointer: "/preferred_username"},
...

&FieldPointer{Pointer: "/roles"},
...

```

pkg/lib/userexport/model.go
```diff
@@ -142,7 +143,8 @@ type MFA struct {
 }
 
 type Record struct {
-	Sub string `json:"sub,omitempty"`
+	Sub       string    `json:"sub,omitempty"`
+	CreatedAt time.Time `json:"created_at,omitempty"`
```
louischan-oursky: Apply the group I mentioned above to here as well.

pkg/lib/userexport/model.go
tung2744: For `anonymized_at`, we did not expose it in any api at the moment. Do you think we should expose it?

louischan-oursky: OK. Let's just expose `/is_anonymized` and `/anonymize_at` then.

tung2744:

louischan-oursky (approved):

PR #5423, opened 2025-11-11
Block disposable email domains

carmenlau:
ref DEV-3126

### DIFF
--- a/portal/src/graphql/portal/LoginMethodConfigurationScreen.tsx
+++ b/portal/src/graphql/portal/LoginMethodConfigurationScreen.tsx
@@ -2184,9 +2184,6 @@ function EmailSettings(props: EmailSettingsProps) {
           prev.loginIDEmailConfig.domain_blocklist_enabled = checked;
           if (prev.loginIDEmailConfig.domain_blocklist_enabled) {
             prev.loginIDEmailConfig.domain_allowlist_enabled = false;
-          } else {
-            prev.loginIDEmailConfig.block_free_email_provider_domains = false;
-            prev.loginIDEmailConfig.block_disposable_email_domains = false;
           }
         })
       );
@@ -2241,7 +2238,6 @@ function EmailSettings(props: EmailSettingsProps) {
           prev.loginIDEmailConfig.block_free_email_provider_domains = checked;
           if (prev.loginIDEmailConfig.block_free_email_provider_domains) {
             prev.loginIDEmailConfig.domain_allowlist_enabled = false;
-            prev.loginIDEmailConfig.domain_blocklist_enabled = true;
           }
         })
       );
@@ -2259,7 +2255,6 @@ function EmailSettings(props: EmailSettingsProps) {
           prev.loginIDEmailConfig.block_disposable_email_domains = checked;
           if (prev.loginIDEmailConfig.block_disposable_email_domains) {
             prev.loginIDEmailConfig.domain_allowlist_enabled = false;
-            prev.loginIDEmailConfig.domain_blocklist_enabled = true;
           }
         })
       );


carmenlau:
Updated

~~TODO for me:~~
~~Update checkbox label to~~
```
[ ] Block emails from known [disposable email domains](link)
```

pkg/lib/config/identity.go
```diff
@@ -126,6 +127,20 @@ var _ = Schema.Add("LoginIDEmailConfig", `
...
+					"block_disposable_email_domains": { "enum": [true] }
+				},
+				"required": ["block_disposable_email_domains"]
+			},
+			"then": {
+				"properties": {
+					"domain_blocklist_enabled": { "enum": [true] }
+				},
+				"required": ["domain_blocklist_enabled"]
+			}
```
louischan-oursky, line 141: When I see this, I wonder why block_free_email_provider_domains requires domain_blocklist_enabled.

I think we should relax this restriction.

Block free email provider domain and block disposable email domain should work independent of `domain_blocklist_enabled`.

1. If domain_allowlist_enabled is true AND a domain is in the allowlist, then it is always allowed.
2. Otherwise, if domain_blocklist_enabled is true AND a domain is in the blocklist, then it is blocked.
3. Otherwise, if block_free_email_provider_domains is enabled, then block the domain if it is in the list.
4. Otherwise, if block_disposable_email_domains is enabled, then block the domain if it is in the list.

makefiles/check-files-outdated.mk
```diff
@@ -0,0 +1,52 @@
+DISPOSABLE_EMAIL_DOMAINS_URL ::= https://raw.githubusercontent.com/disposable-email-domains/disposable-email-domains/refs/heads/main/disposable_email_blocklist.conf
+DISPOSABLE_EMAIL_DOMAINS_FILE ::= ./resources/authgear/disposable_email_domain_list.txt
+
+# Usage:
+#	$(call check-file,<remote_url>,<local_file>)
+define check-file
```
louischan-oursky: I prefer we write this as a shell script. It is easier to maintain. We of course keep this Makefile, but it is just a driver to invoke the shell scripts.

pkg/lib/config/identity.go
carmenlau: I agree that `domain_blocklist_enabled`, `block_free_email_provider_domains`, and `block_disposable_email_domains` can work independently. However, we may still want to enforce that if `domain_allowlist_enabled` is true, none of the blocklists should be enabled.

From my understanding, the use case of `domain_allowlist_enabled` is for organizations that want to restrict signups to their own domains. Domains not on the allowlist will be rejected, and no blocklists will be checked in this case. To avoid confusion, it’s better to keep this rule.

So it would be:
1. If `domain_allowlist_enabled`, none of the blocklists should be enabled.  
2. `domain_blocklist_enabled`, `block_free_email_provider_domains`, and `block_disposable_email_domains` can be enabled independently.

Thoughts?

makefiles/check-files-outdated.mk
louischan-oursky: BTW, we put shell scripts in `./scripts/sh`, and they are `/bin/sh` scripts, not bash script.

carmenlau: Updated, thanks!

pkg/lib/config/identity.go
louischan-oursky: Let me list out the common use cases to facilitate our discussion.

## Use case 1: No specific restriction on email domains

No options have to be set.
By default, no restriction on email domains.

## Use case 2: Block disposable email domains

Turn on `block_disposable_email_domains`.

## Use case 3: Block disposable email domains, but found a disposable domain not blocked as expected.

Turn on `block_disposable_email_domains`, and turn on `domain_blocklist_enabled`, add the missed domain to the blocklist.

## Use case 4: Block free email domains

Turn on `block_free_email_provider_domains`.

## Use case 5: Block free email domains, but found a free domain not blocked as expected.

Turn on `block_free_email_provider_domains`, and turn on `domain_blocklist_enabled`, add the missed domain to the blocklist.

## Use case 6: Block disposable email domains, but want to allow a specific blocked domain for testing purpose

I think we should support this.
If I were the developer, I think this should be

1. Turn on `block_disposable_email_domains`.
2. Turn on `domain_allowlist_enabled`.
3. Add the specific domain to the allowlist to explicitly allow it.

## Use case 7: Block free email domains, but want to allow a specific blocked domain

Similar to Use case 6.

## Use case 8: Only explicitly allowed domain

This is the use case Carmen mentioned.
We definitely must keep this.

But unfortunately this is conflicting with Use case 6 and Use case 7.

Previously, when `domain_allowlist_enabled` is true, it means

- All domains are blocked.
- Only the domains in the allowlist is allowed.

In Use case 6 and Use case 7, we need

- ~~All domains are blocked.~~
- Only the domains in the allowlist is allowed.

---

## Design 1: Add a new option to model exception

```
# Use case 8
# When domain_allowlist_enabled is true, then domain_blocklist_enabled, block_free_email_provider_domains, block_disposable_email_domains must all be false.
domain_allowlist_enabled: true
```

```
# Use case 2, 3, 4, 5, 6, 7
block_free_email_provider_domains: true
block_disposable_email_domains: true
# As an escape hatch to block domains that are not captured in the bundled lists.
domain_blocklist_enabled: true
# New option. As an escape hatch to allow domains that are considered mistakenly blocked in the bundled lists.
domain_blocklist_exception_enabled: true
```

## Design 2: Relax the restriction of `domain_allowlist_enabled`

When `domain_allowlist_enabled` is enabled alone, it means

- All domains are blocked.
- Only the domains in the allowlist is allowed.

Otherwise, it means

- ~~All domains are blocked.~~
- Only the domains in the allowlist is allowed.

carmenlau: I prefer design 1. Although it makes the configuration more complicated, it is more intuitive.

For design 2, enabling another configuration that alter the allowlist behavior could be confusing or unexpected for developers.

louischan-oursky: ## How other competitors do this

- Auth0: No builtin support. The developer has to write custom Actions to block manually. See https://support.auth0.com/center/s/article/How-to-combat-fradulent-signups-from-disposable-email-services
- Stytch: No builtin support. See https://forum.stytch.com/t/block-users-from-logging-in-based-on-email-address/107/2
- Keycloak: No builtin support. Has to resort to custom extension, like https://github.com/micedre/keycloak-mail-whitelisting
- FushionAuth: Support blocklist only. https://fusionauth.io/docs/operate/secure/advanced-threat-detection#registration-email-domain-blocking
- SuperTokens: No builtin support. Need to override https://community.supertokens.com/t/22943350/block-specific-email-domains-we-re-getting-a-lot-of-sign-ups
- Zitadel: No builtin support.
- Clerk: Yes see https://clerk.com/docs/guides/secure/restricting-access#allowlist and https://clerk.com/docs/guides/secure/restricting-access#blocklist
- Kinde: Support allowlist See https://docs.kinde.com/build/set-up-options/access-policies/#set-policies-for-all-organizations

carmenlau: - Updated the PR to allow enabling blocklists individually. Test case `login-id-email-domain-allowlist-enabled` and `login-id-email-domain-multiple-blocklist-enabled` are testing the existing config.
- Created issue for supporting blocklists exception.

makefiles/check-files-outdated.mk
carmenlau: I further updated the implementation for DEV-3158, see PR #5425.

PR #5433, opened 2025-11-18
Support "reason" in scheduled user deletion

tung2744:
ref DEV-3164


carmenlau:
I’m wondering if the user might want to retrieve the delete reason when querying the user node for scheduled deletion? The rest of the PR looks good to me!

tung2744:
Because we have disableReason, let me also expose deleteReason in the user query.

PR #5436, opened 2025-11-19
Expose authenticators and recovery code information in userinfo

tung2744:
ref DEV-3027

### DIFF
--- a/pkg/lib/oauth/oidc/id_token_test.go
+++ b/pkg/lib/oauth/oidc/id_token_test.go
@@ -2,6 +2,7 @@ package oidc
 
 import (
 	"context"
+	"encoding/json"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/api/model/userinfo.go
```diff
@@ -0,0 +1,19 @@
+package model
+
+import "time"
+
+type UserInfoAuthenticator struct {
```
carmenlau, line 5: I think it’s better to expose only the minimum information unless there’s a clear use case. It’s easy to extend the payload later, but removing fields would be difficult and could introduce breaking changes. So for now, should the authenticators payload include only the type and kind?

I may have missed parts of the discussion, so I might be wrong 🙏🏻😅

pkg/lib/oauth/oidc/id_token_test.go
```diff
@@ -206,3 +206,174 @@ func TestIDTokenIssuer(t *testing.T) {
...
+		}
+
+		clientConfig := &config.OAuthClientConfig{
+			ClientID:        "client-id",
+			ApplicationType: config.OAuthClientApplicationTypeSPA,
+		}
+		client := oauth.ClientClientLike(clientConfig, []string{"openid", "email", oauth.FullUserInfoScope, string(model.ClaimAuthenticators), string(model.ClaimPhoneNumber), string(model.ClaimEmail)})
+		userInfo, err := issuer.GetUserInfo(context.Background(), "user-id", client)
+		So(err, ShouldBeNil)
+		So(userInfo, ShouldResemble, map[string]interface{}{
```
louischan-oursky: We should serialize userinfo and assert on the actual JSON. This can also test we have put correct JSON struct tags.

pkg/api/model/userinfo.go
tung2744: The email, phone are for mfa management methods according to the spec, but they will not be implemented at the moment so we can remove them first.
The created_at and updated_at timestamps and the display_name exist such that there will not be two identical objects, for example if the user has two totp authenticators.

tung2744: Removed email, phone and display_name, and keep the timestamps to avoid identical objects.

tung2744:
Updated with all SDK prs, thanks!

louischan-oursky (approved):

pkg/lib/userinfo/userinfo.go
```diff
@@ -94,10 +103,42 @@ func (s *UserInfoService) getUserInfoFromDatabase(ctx context.Context, userID st
 		roleKeys[i] = roles[i].Key
 	}
 
+	authns, err := s.AuthenticatorService.List(ctx, userID, authenticator.FilterFunc(func(info *authenticator.Info) bool {
```
louischan-oursky, line 115: I think this code is problematic because it does not filter out `passkey`. In the SDKs, if the type is `passkey`, then it will crash.

PR #5437, opened 2025-11-19
Support application-client specific logo by logo_uri

carmenlau:
ref DEV-3170

Portal update is not included

### DIFF
--- a/pkg/lib/config/oauth.go
+++ b/pkg/lib/config/oauth.go
@@ -174,7 +174,7 @@ var _ = Schema.Add("OAuthClientConfig", `
 		"x_authentication_flow_allowlist": { "$ref": "#/$defs/AuthenticationFlowAllowlist" },
 		"x_pre_authenticated_url_enabled": { "type": "boolean" },
 		"x_pre_authenticated_url_allowed_origins": { "type": "array", "items": { "type": "string", "format": "http_origin" } },
-		"logo_uri": { "type": "string", "format": "x_https_strict_uri" }
+		"logo_uri": { "type": "string", "format": "x_public_https_url" }
 	},
 	"required": ["name", "client_id"],
 	"allOf": [
--- a/pkg/lib/config/testdata/config_tests.yaml
+++ b/pkg/lib/config/testdata/config_tests.yaml
@@ -1514,7 +1514,7 @@ name: oauth-client-logo-uri-non-https
 error: |-
   invalid configuration:
   /oauth/clients/0/logo_uri: format
-    map[error:url scheme must be https format:x_https_strict_uri]
+    map[error:url scheme must be https format:x_public_https_url]
 config:
   id: test
   http:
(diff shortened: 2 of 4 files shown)


tung2744:
We have a linter to prevent misuse of request.Context(). You need to update `.vettedpositions` to ignore the lines if you think the usage is correct.
You can also ask AI to read `.claude/commands/update_vettedpositions.md` then it will update the file for you.

pkg/util/urlutil/validate.go
```diff
@@ -0,0 +1,191 @@
...
+	if raw == "" {
+		return ErrEmptyURL
+	}
+
+	// 2. Reject control characters
+	for _, r := range raw {
+		if r == '\r' || r == '\n' || unicode.IsControl(r) {
+			return ErrControlChars
+		}
+	}
```
louischan-oursky: I think 1 and 2 are not needed. They should have been handled by `url.Parse`. So, try remove 1 and 2, and keep the test case to see if `url.Parse` does the right job.

pkg/util/urlutil/validate.go
```diff
@@ -0,0 +1,191 @@
...
+	// 5. Host must exist
+	if u.Host == "" {
+		return ErrBadHost
+	}
+
+	hostOnly := u.Host
+	if h, _, err2 := net.SplitHostPort(u.Host); err2 == nil {
+		hostOnly = h
+	}
+	hostOnly = strings.Trim(hostOnly, "[]")
```
louischan-oursky: Use https://pkg.go.dev/net/url#URL.Hostname It handles IP address for you. So you do not need `net.SplitHostPort` or manual trimming of `[]`, accessing `u.Host` directly


```golang
hostname := u.Hostname()
ipv4Oripv6 = net.ParseIP(hostname)
```

pkg/util/urlutil/validate.go
```diff
@@ -0,0 +1,191 @@
...
+	// 9. Validate hostname characters
+	if !isHostnameSafe(hostOnly) {
+		return ErrBadHost
+	}
+
+	return nil
+}
+
+// ref https://www.ietf.org/rfc/rfc1035.txt
+func isHostnameSafe(h string) bool {
```
louischan-oursky, line 144: Actually I think we should allow hostname whose characters are in ASCII range only. `unicode.IsLetter` and `unicode.IsDigit` accepts many other characters. FYI https://github.com/skygeario/go-confusable-homoglyphs

That is, we only accept common hostname (ASCII range), or punycode-encoded hostname (Also ASCII range after the encoding)

pkg/lib/config/oauth.go
```diff
@@ -174,7 +174,7 @@ var _ = Schema.Add("OAuthClientConfig", `
 		"x_authentication_flow_allowlist": { "$ref": "#/$defs/AuthenticationFlowAllowlist" },
 		"x_pre_authenticated_url_enabled": { "type": "boolean" },
 		"x_pre_authenticated_url_allowed_origins": { "type": "array", "items": { "type": "string", "format": "http_origin" } },
-		"logo_uri": { "type": "string", "format": "x_public_image_uri" }
+		"logo_uri": { "type": "string", "format": "x_https_strict_uri" }
```
louischan-oursky: I think the name `x_public_https_url` is more suitable, as it is intended to be a publicly accessible HTTPS URL. Note that it is `_url`, not `_uri`.

pkg/util/urlutil/validate.go
carmenlau: Agree, we can use the stricter rule first.

carmenlau:
Updated, thank you!:)

louischan-oursky (approved):

PR #5441, opened 2025-11-21
Fix graphiql

louischan-oursky:
1. Update graphiql to v4
2. Since graphiql v4 dropped support for React 17, we updated React to 19
3. Fix the bug that the built bundle fail to load graphql-ws

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -274,10 +274,6 @@ mc mb local/userexport
 
 cert-manager@v1.7.3 has transitive dependency problem.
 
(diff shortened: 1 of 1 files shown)


tung2744:
@louischan-oursky Do you know what actually is the cause the original build does not work after updating vite?

louischan-oursky:
In the console, it prints something like `fail to resolve module graphql-ws`.

tung2744:
Yes but it worked before and i was just curious why because we did not install graphql-ws before

PR #5444, opened 2025-11-24
Configure client logo uri on the portal

carmenlau:
ref DEV-3170

after #5437

### DIFF
--- a/portal/src/TextFieldWithImagePreview.tsx
+++ b/portal/src/TextFieldWithImagePreview.tsx
@@ -54,7 +54,7 @@ const TextFieldWithImagePreview: React.VFC<TextFieldWithImagePreviewProps> =
       return (
         <img
           src={imgURL}
-          className={cn("max-h-12.5", "max-w-[95%]")}
+          className={cn("max-h-12.5", "max-w-[95%]", "object-contain")}
           onError={onError}
         />
       );


portal/src/TextFieldWithImagePreview.tsx
```diff
@@ -1,16 +1,75 @@
...
+      setHasError(true);
+    };
+
+    const renderImg = useMemo(() => {
+      if (hasError || !rest.value) {
+        return (
+          <div
+            className={cn(
+              "h-[40px]",
+              "w-[40px]",
```
louischan-oursky: `h-10` and `w-10`

portal/src/TextFieldWithImagePreview.tsx
```diff
@@ -1,16 +1,75 @@
...
+              "flex",
```
louischan-oursky, line 38: Be explicit and write `flex flex-row`

portal/src/graphql/portal/EditOAuthClientForm.tsx
```diff
@@ -590,6 +601,32 @@ const EditOAuthClientForm: React.VFC<EditOAuthClientFormProps> =
...
+            }
+            value={clientConfig.logo_uri}
+            onChange={onLogoURIChange}
+            errorRules={[
+              makeValidationErrorCustomMessageIDRule(
+                "format",
+                /\/logo_uri$/,
+                "EditOAuthClientForm.logo-uri.error.format"
+              ),
+            ]}
```
louischan-oursky, line 631: We should specify a placeholder to give a hint to the developer what is expected to be filled in. In the Figma, there is `https://example.com/app/logo.png`. Please specify it.

portal/src/TextFieldWithImagePreview.tsx
```diff
@@ -1,16 +1,75 @@
...
+            "border-neutral-light",
+            "flex",
+            "justify-center",
+            "items-center",
+            "h-[116px]",
+            "mb-1"
+          )}
+        >
+          {renderImg}
+        </div>
```
louischan-oursky, line 80: The text field is quite buggy.

Case 1
1. Suppose the value is `https://avatars.githubusercontent.com/u/15082564?v=4`
2. I delete the characters until it is `h`.
3. Delete the last character `h`. The value was reverted to `https://avatars.githubusercontent.com/u/15082564?v=4`

Case 2
1. Suppose the value is `https://avatars.githubusercontent.com/u/15082564?v=4`
2. I delete one character per 0.5 second
3. See the preview keeps changing
4. I think the preview should update only when the text field is blurred.

carmenlau: 1. Added a fix, please check if it’s resolved.
2. I think users may expect the image to update even if they stay focused on the text field. I added a debounce instead. See if this feels better.

Thanks!

portal/src/TextFieldWithImagePreview.tsx
```diff
@@ -1,29 +1,35 @@
...
 
 import cn from "classnames";
+import { useDebouncedEffect } from "./hook/useDebouncedEffect";
 
 export interface TextFieldWithImagePreviewProps extends FormTextFieldProps {}
 
 const TextFieldWithImagePreview: React.VFC<TextFieldWithImagePreviewProps> =
   function TextFieldWithImagePreview(props: TextFieldWithImagePreviewProps) {
     const { label, ...rest } = props;
+    const [imgURL, setImgURL] = useState(rest.value ?? "");
```
louischan-oursky, line 16: I think you should use `useDebounced`.

portal/src/TextFieldWithImagePreview.tsx
```diff
@@ -0,0 +1,86 @@
...
+            {hasError ? (
+              <img src={iconPreviewUnavailable} />
+            ) : (
+              <img src={iconPreview} />
+            )}
+          </div>
+        );
+      }
+      return (
+        <img
```
tung2744, line 55: I think we might want to set `object-fit` to `contain`, which is same as the `brand-logo` class in auth ui.

portal/src/TextFieldWithImagePreview.tsx
carmenlau: I think we should use either `useDebouncedEffect` or `useDebounced`. Because we need to update both `imageURL` and `hasError` simultaneously, I used `useDebouncedEffect`. If we use `useDebounced` for `imageURL` and `hasError` separately, the effect looks buggy. Does that make sense? :)

This is how I used the `useDebounced`
```diff
diff --git a/portal/src/TextFieldWithImagePreview.tsx b/portal/src/TextFieldWithImagePreview.tsx
index 88b340ceff..eb49245082 100644
--- a/portal/src/TextFieldWithImagePreview.tsx
+++ b/portal/src/TextFieldWithImagePreview.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { Label } from "@fluentui/react";
 import FormTextField, { FormTextFieldProps } from "./FormTextField";

@@ -6,7 +6,7 @@ import iconPreview from "./images/preview.svg";
 import iconPreviewUnavailable from "./images/preview-unavailable.svg";

 import cn from "classnames";
-import { useDebouncedEffect } from "./hook/useDebouncedEffect";
+import { useDebounced } from "./hook/useDebounced";

 export interface TextFieldWithImagePreviewProps extends FormTextFieldProps {}

@@ -16,20 +16,20 @@ const TextFieldWithImagePreview: React.VFC<TextFieldWithImagePreviewProps> =
     const [imgURL, setImgURL] = useState(rest.value ?? "");
     const [hasError, setHasError] = useState(false);

-    useDebouncedEffect(
-      useCallback(() => {
-        setImgURL(rest.value ?? "");
-        setHasError(false);
-      }, [rest.value]),
-      1000
-    );
+    const [debouncedHasError] = useDebounced(hasError, 1000);
+    const [debouncedImgURL] = useDebounced(imgURL, 1000);
+
+    useEffect(() => {
+      setImgURL(rest.value ?? "");
+      setHasError(false);
+    }, [rest.value]);

     const onError = () => {
       setHasError(true);
     };

     const renderImg = useMemo(() => {
-      if (hasError || !imgURL) {
+      if (debouncedHasError || !debouncedImgURL) {
         return (
           <div
             className={cn(
@@ -43,7 +43,7 @@ const TextFieldWithImagePreview: React.VFC<TextFieldWithImagePreviewProps> =
               "rounded"
             )}
           >
-            {hasError ? (
+            {debouncedHasError ? (
               <img src={iconPreviewUnavailable} />
             ) : (
               <img src={iconPreview} />
@@ -53,12 +53,12 @@ const TextFieldWithImagePreview: React.VFC<TextFieldWithImagePreviewProps> =
       }
       return (
         <img
-          src={imgURL}
+          src={debouncedImgURL}
           className={cn("max-h-12.5", "max-w-[95%]", "object-contain")}
           onError={onError}
         />
       );
-    }, [hasError, imgURL]);
+    }, [debouncedHasError, debouncedImgURL]);
```

portal/src/TextFieldWithImagePreview.tsx
carmenlau: Updated, thank you!

portal/src/TextFieldWithImagePreview.tsx
louischan-oursky: I see. Let's keep the current implementation then.

tung2744 (approved):

PR #5446, opened 2025-11-25
Fix portal unable to create new projects in Helm deployments

CoreyJ87:
Creating a new project through the portal UI fails in Helm deployments with this error:

```
invalid secret config: invalid secrets:
<root>: database credentials (secret 'db') is required
<root>: redis credentials (secret 'redis') is required
```

This makes it impossible to create multiple projects through the UI - you have to work around it by manually inserting into the database.

I tracked this down to `pkg/portal/service/app.go` line 598 where `generateResources()` creates secrets for new projects. It was passing an empty `GenerateSecretConfigOptions{}` struct, which meant no database or Redis credentials were included.

The validation logic in `pkg/lib/config/secret.go` always expects these secrets to exist, but they were never being populated. Turns out the `AppService` struct already had `DatabaseCredentials` and `RedisCredentials` fields, but they were sitting there as nil because Wire wasn't injecting them.

## The Fix

Changed two files:

**pkg/portal/service/app.go** - Actually use the credentials when generating secrets:

```go
secretOpts := &config.GenerateSecretConfigOptions{}

if s.DatabaseCredentials != nil {
    secretOpts.DatabaseURL = s.DatabaseCredentials.DatabaseURL
    secretOpts.DatabaseSchema = s.DatabaseCredentials.DatabaseSchema
}
if s.RedisCredentials != nil {
    secretOpts.RedisURL = s.RedisCredentials.RedisURL
}

secretConfig := config.GenerateSecretConfigFromOptions(secretOpts, createdAt, corerand.SecureRand)
```

**pkg/portal/deps/deps.go** - Wire up the dependency injection properly:

```go
func ProvideDatabaseCredentials(cfg *config.EnvironmentConfig) *config.GlobalDatabaseCredentialsEnvironmentConfig {
	return &cfg.GlobalDatabase
}

func ProvideRedisCredentials(cfg *config.EnvironmentConfig) *config.GlobalRedisCredentialsEnvironmentConfig {
	return &cfg.GlobalRedis
}
```

Had to remove `GlobalDatabase` and `GlobalRedis` from the `wire.FieldsOf` call since that was providing values but `AppService` needs pointers. The provider functions handle the conversion.

## Notes
Ran this in a live k8s cluster with Helm-deployed Authgear. Portal can now create new projects, they get the right credentials, and existing stuff keeps working fine. I'm new to this project and I could totally be missing something glaringly obvious so let me know if this change is not the way you guys want to go about it. 

I considered a few alternatives but this seemed cleanest:
- It follows the same pattern as `ProvideAuditDatabaseCredentials`
- The fields already existed in `AppService`, just needed wiring
- Only touches 2 files
- No breaking changes


louischan-oursky:
The code is actually working as intended. The `authgear.secrets.yaml` of the project does not include the global credentials of the database.

Imagine you have a global database credentials `postgresql://user:password@host-of-the-managed-postgresql:5432/authgear`. You will not want to put this credentials into the `authgear.secrets.yaml` of the project, in which the owner of the project has access to. Doing this will accidentally leak the credentials of the database to anyone who can sign up the portal.

Instead, in your helm chart release, you should do these

1. Create a k8s secret like this

```
apiVersion: v1
kind: Secret
metadata:
  name: authgear-vendor-resources
type: Opaque
data:
  authgear.secrets.yaml: |-
    {{ include "authgear.authgearSecretsYAML" .Values.authgear | b64enc }}
```

2. Specify `authgear.appCustomResources` in your values.yaml

```
authgear:
  appCustomResources:
  path: "/var/resources/authgear"
  volume:
    secret:
      # This must match the name of the secret.
      secretName: authgear-vendor-resources
      items:
      - key: authgear.secrets.yaml
        path: authgear.secrets.yaml
```

---

`authgear.secrets.yaml` has a mechanism to merge multiple `authgear.secrets.yaml` files into an effective `authgear.secrets.yaml`. What was described above is to define a deployment-wise `authgear.secrets.yaml`, which contains the secret `db` and `redis`. At runtime, the `authgear.secrets.yaml` of the project overlays on top of the deployment-wise `authgear.secrets.yaml`

CoreyJ87:
Ahhhhh I knew I missed something. Thanks

fungc-io:
@CoreyJ87 , Thanks for pointing out some unclear parts in our docs! 
We've updated the documentation for Helm deployments: https://docs.authgear.com/deployment/helm
Let us know if you have any further comments or suggestions. 😄

PR #5447, opened 2025-11-26
Add feature flag for logo_uri

carmenlau:
ref DEV-3170

after #5444

### DIFF
--- a/portal/src/graphql/portal/EditOAuthClientForm.tsx
+++ b/portal/src/graphql/portal/EditOAuthClientForm.tsx
@@ -46,6 +46,7 @@ interface EditOAuthClientFormProps {
   className?: string;
   clientConfig: OAuthClientConfig;
   customUIEnabled: boolean;
+  showAppLogoSection: boolean;
   app2appEnabled: boolean;
   onClientConfigChange: (newClientConfig: OAuthClientConfig) => void;
   clientSecretHook: ClientSecretsHook;
@@ -101,6 +102,7 @@ const EditOAuthClientForm: React.VFC<EditOAuthClientFormProps> =
       publicOrigin,
       customUIEnabled,
       app2appEnabled,
+      showAppLogoSection,
       onClientConfigChange,
       clientSecretHook,
     } = props;
@@ -601,35 +603,37 @@ const EditOAuthClientForm: React.VFC<EditOAuthClientFormProps> =
             value={applicationTypeLabel}
             readOnly={true}
           />
-          <TextFieldWithImagePreview
-            parentJSONPointer={parentJSONPointer}
-            fieldName="logo_uri"
-            label={renderToString("EditOAuthClientForm.logo-uri.label")}
-            description={
-              (
-                <FormattedMessage
-                  id="EditOAuthClientForm.logo-uri.description"
-                  values={{
-                    href: `/project/${encodeURIComponent(
-                      appID
-                    )}/branding/design`,
-                  }}
-                />
-              ) as any as string
-            }
-            placeholder={renderToString(
-              "EditOAuthClientForm.logo-uri.placeholder"
-            )}
-            value={clientConfig.logo_uri ?? ""}
-            onChange={onLogoURIChange}
-            errorRules={[
-              makeValidationErrorCustomMessageIDRule(
-                "format",
-                /\/logo_uri$/,
-                "EditOAuthClientForm.logo-uri.error.format"
-              ),
-            ]}
-          />
+          {showAppLogoSection ? (
+            <TextFieldWithImagePreview
+              parentJSONPointer={parentJSONPointer}
+              fieldName="logo_uri"
+              label={renderToString("EditOAuthClientForm.logo-uri.label")}
+              description={
+                (
+                  <FormattedMessage
+                    id="EditOAuthClientForm.logo-uri.description"
+                    values={{
+                      href: `/project/${encodeURIComponent(
+                        appID
+                      )}/branding/design`,
+                    }}
+                  />
+                ) as any as string
+              }
+              placeholder={renderToString(
+                "EditOAuthClientForm.logo-uri.placeholder"
+              )}
+              value={clientConfig.logo_uri ?? ""}
+              onChange={onLogoURIChange}
+              errorRules={[
+                makeValidationErrorCustomMessageIDRule(
+                  "format",
+                  /\/logo_uri$/,
+                  "EditOAuthClientForm.logo-uri.error.format"
+                ),
+              ]}
+            />
+          ) : null}
         </Widget>
 
         {showClientSecret && clientSecrets && clientSecrets.length > 0 ? (
--- a/portal/src/graphql/portal/EditOAuthClientScreen.tsx
+++ b/portal/src/graphql/portal/EditOAuthClientScreen.tsx
@@ -339,6 +339,7 @@ interface EditOAuthClientContentProps {
   samlIdpEntityID: string;
   samlIdpSigningCertificates: SAMLIdpSigningCertificate[];
   customUIEnabled: boolean;
+  showAppLogoSection: boolean;
   app2appEnabled: boolean;
   onGeneratedNewIdpSigningCertificate: () => void;
   clientSecretHook: ClientSecretsHook;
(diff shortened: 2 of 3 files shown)


pkg/lib/config/feature_oauth.go
```diff
@@ -39,16 +39,18 @@ var _ = FeatureConfigSchema.Add("OAuthClientFeatureConfig", `
...
-	Maximum         *int  `json:"maximum,omitempty"`
-	SoftMaximum     *int  `json:"soft_maximum,omitempty"`
-	CustomUIEnabled *bool `json:"custom_ui_enabled,omitempty"`
-	App2AppEnabled  *bool `json:"app2app_enabled,omitempty"`
+	Maximum                   *int  `json:"maximum,omitempty"`
+	SoftMaximum               *int  `json:"soft_maximum,omitempty"`
+	CustomUIEnabled           *bool `json:"custom_ui_enabled,omitempty"`
+	App2AppEnabled            *bool `json:"app2app_enabled,omitempty"`
+	LogoURIInAuthnPageEnabled *bool `json:"logo_uri_in_authn_page_enabled,omitempty"`
 }
```
tung2744, line 52: I thought this should be a portal ui feature flag but not a flag in core feature config?

tung2744: Oh but portal feature config cannot be different in different projects. Just ignore this comment.

pkg/lib/config/feature_oauth.go
```diff
@@ -39,16 +39,18 @@ var _ = FeatureConfigSchema.Add("OAuthClientFeatureConfig", `
...
 type OAuthClientFeatureConfig struct {
-	Maximum         *int  `json:"maximum,omitempty"`
-	SoftMaximum     *int  `json:"soft_maximum,omitempty"`
-	CustomUIEnabled *bool `json:"custom_ui_enabled,omitempty"`
-	App2AppEnabled  *bool `json:"app2app_enabled,omitempty"`
+	Maximum                   *int  `json:"maximum,omitempty"`
+	SoftMaximum               *int  `json:"soft_maximum,omitempty"`
+	CustomUIEnabled           *bool `json:"custom_ui_enabled,omitempty"`
+	App2AppEnabled            *bool `json:"app2app_enabled,omitempty"`
+	LogoURIInAuthnPageEnabled *bool `json:"logo_uri_in_authn_page_enabled,omitempty"`
```
tung2744: Suggest to simply call it `logo_uri_enabled`.

carmenlau: I also thought about that, but this feature displays the logo URI on authentication pages. Since we may add a feature in the future to show the logo URI only on the consent page, so I chose a more specific name for this key. Maybe we can come up with an even better name? 🤔

pkg/lib/config/feature_oauth.go
carmenlau: Didn't know we have a portal feature config, good to know!

pkg/lib/config/feature_oauth.go
louischan-oursky: My original thought is that this is a feature of ui, not feature of oauth. For consent page, we should always show the logo_uri of the client, no need feature flag to control that.

carmenlau: Moving the feature flag to UI makes sense, let me update it!
Agree that we should always show logo_uri in the consent page which I haven't check it, I will do it. If that's the case, name `logo_uri_enabled` becomes ambiguous? It sounds like logo uri shouldn't be used anywhere.

louischan-oursky: Here is my idea

1. In feature config, add `ui.allow_showing_logo_uri_as_project_logo`.
2. In oauth client config, add `x_show_logo_uri_as_project_logo`.
3. If 1 is true, and 2 is true, then we show `logo_uri` instead of the project logo.

carmenlau: If this is the case, we may:  

1. In the feature config, add `ui.allow_showing_logo_uri_as_project_logo`.  
2. In the OAuth client config, add `x_show_logo_uri_as_project_logo`.  
3. If both 1 and 2 are true, show `logo_uri` instead of the project logo.  
4. The consent screen should always show `logo_uri`, regardless of whether 1 or 2 is true.  
5. If 1 is true, show a checkbox in the portal to configure `x_show_logo_uri_as_project_logo`. I will add a screenshot in the PR later.  
6. If 1 is false, hide the entire application logo section in the portal until we think the `image_uri` feature is ready for all users.  

Thoughts? cc @fungc-io

tung2744: I think what we want to hide is only the portal ui which allow user to configure logo_uri of the application, therefore:
1. I suggest we simply have one feature flag in ui config which is for hiding the image widget in portal only. It have no effect on the actual logic of authgear core. And no additional flags in oauth config.
2. For consent screen, I think now the consent screen is not really in a usable state. So we need to redefine the bahavior when we implement it. I suggest we leave it for now.

carmenlau: To keep it simple, I think using one feature flag is fine, but we still need to add checks in authgear core. Using the logo URI as the project logo might not be the ‘common’ behavior, and we may switch back to using the project logo when we implement it properly. If we don’t add the check in core now, we might break existing client apps when we implement it. So another proposal will be

1. In the feature config, add `ui.allow_showing_logo_uri_as_project_logo`.
2. In the Portal, show application logo section only if 1 is true.
3. Use `logo_uri` as project logo only if 1 is true. 

I call this proposal 2.
This is similar to what I did in this PR. What I need to do now is move and rename the feature config.

carmenlau: @tung2744 Does this mean we don’t have a consent screen now?

fungc-io: I think we can call it something like "allow_app_logo_in_authui" to be more precise. 
The user can still set logo_uri technically, and it's always a metadata of the app client, and this logo_uri may be useful in other scenario such as consent screen in the future.

It only controls if the app logo is shown in the auth ui. And for now, our portal hide the application logo section only if this is true.

tung2744: > Does this mean we don’t have a consent screen now?

We have a screen called consent, but it only do a redirect for 1st party app, and we don't have 3rd party app at the moment. So actually we never see it.

louischan-oursky: If this feature is uncommon, maybe we can just

1. Do not introduce the feature flag.
2. Introduce `x_show_logo_uri_as_project_logo` in OAuth client config
3. Keep the UI for setting `logo_uri`. Anyone can set it.
4. To turn on showing `logo_uri` as Auth UI logo, go to `/edit-config` and set `x_show_logo_uri_as_project_logo` manually.

I propose this because whether or not to show `logo_uri` as Auth UI logo should be client-specific. Thus, a feature flag cannot express this behavior.

tung2744: I guess the major reason we want a feature flag is to avoid people setting it in portal, because this feature is not mature yet.
And if people can set it in portal but they need to edit config manually to set `x_show_logo_uri_as_project_logo` to make it effective, then most of the user will not understand what is the widget in portal is for.

carmenlau: Thanks for thinking it through so thoroughly! Now we have 3 proposals:

1. Both key in feature config and core config
2. Single key in feature config
    a. check in portal only
    b. check in both portal and core
3. Single key in core config

For the latest proposal, we’re not sure whether this feature (show as the project logo) will be common or not, but we know that showing it as the project logo should probably not be the default behavior..

Since we don’t support third-party apps at the moment, setting only `logo_uri` will not affect the app’s behavior unless the user manually sets the core config flag. This may surprise users.

So I think it’s better to hide the portal logo URI section for now, until we support third-party apps. So for now, proposal 1 might be the better option.

How about this?

1. When the feature config `ui.logo_uri` is enabled, the portal will display the `logo_uri` setting UI.
2. The portal will include a checkbox to enable the core config flag `x_show_logo_uri_as_project_logo`. The logo will only change when this flag is enabled.

BTW, we finally have the `/edit-config` page, so convenient!!

louischan-oursky:
Please remember to rebase.

pkg/lib/config/feature_oauth.go
tung2744: i try to write down the conclusion according to our latest discussion:
- We want to hide the portal ui for logo_uri at the moment.
- We do not need feature flag
- We want to have a new flag in the oauth client config, which controls whether logo_uri will be shown as the branding logo in authui / settings.
- We expect to use this feature in /edit-config only for now.

carmenlau:
Updated in https://github.com/authgear/authgear-server/pull/5460

PR #5450, opened 2025-11-27
Use AUTHGEAR_ENDPOINT as Host, AUTHGEAR_ENDPOINT_INTERNAL for connection

louischan-oursky:
This change is necessary for DEV-3225

ref DEV-3225


tung2744:
Some lint failed

PR #5452, opened 2025-11-27
Add flag to disable dpop

tung2744:
ref DEV-3088

Screenshot 2025-11-27 at 8 37 29 PM

### DIFF
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -852,8 +852,7 @@ func (h *TokenHandler) verifyIDTokenDeviceSecretHash(ctx context.Context,
 			slog.Bool("dpop_logs", true),
 		)
 	}); dpopErr != nil {
-		err = ErrInvalidDPoPKeyBinding
-		return err
+		return dpopErr
 	}
 
 	return err
--- a/pkg/lib/oauth/dpop.go
+++ b/pkg/lib/oauth/dpop.go
@@ -2,10 +2,14 @@ package oauth
 
 import (
 	"context"
+	"errors"
 
 	"github.com/authgear/authgear-server/pkg/lib/config"
 	"github.com/authgear/authgear-server/pkg/lib/dpop"
+	"github.com/authgear/authgear-server/pkg/lib/oauth/protocol"
 )
+
+var ErrInvalidDPoPKeyBinding = protocol.NewError(dpop.InvalidDPoPProof, "Invalid DPoP key binding")
 
 type oauthDPoPChecker func(*dpop.DPoPProof) error
 
@@ -19,7 +23,7 @@ func checkDPoPWithClient(
 	dpopProof, err := maybeDpopProof.Get()
 	if err != nil {
 		if !client.DPoPDisabled {
-			return err
+			return errors.Join(ErrInvalidDPoPKeyBinding, err)
 		}
 	}
 
@@ -27,7 +31,7 @@ func checkDPoPWithClient(
 	if err != nil {
 		errorLogger(err)
 		if !client.DPoPDisabled {
-			return err
+			return errors.Join(ErrInvalidDPoPKeyBinding, err)
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/lib/oauth/handler/handler_token.go
```diff
@@ -590,7 +590,9 @@ func (h *TokenHandler) IssueTokensForAuthorizationCode(
 			fmt.Sprintf("failed to match dpop jkt on issue tokens: %s", dpopErr.Message),
 			slog.Bool("dpop_logs", true),
 		)
-		return nil, ErrInvalidDPoPKeyBinding
+		if !client.DPoPDisabled {
+			return nil, ErrInvalidDPoPKeyBinding
+		}
```
louischan-oursky: It is better to make `codeGrant.MatchDPoPJKT` takes `client` (`codeGrant.MatchDPoPJKT(client, dpopProof)`) so that MatchDPoPJKT never return error if client.DPoPDisabled is true.

In general, make all places that enforce DPoP takes `client`, instead of conditionally returning error.

- `codeGrant.MatchDPoPJKT`
- `offlineGrant.MatchDeviceSecretDPoPJKT`
- `offlineGrantSession.MatchDPoPJKT`

pkg/lib/oauth/handler/handler_anonymous_user.go
```diff
@@ -228,7 +228,12 @@ func (h *AnonymousUserHandler) signupAnonymousUserWithRefreshTokenSessionType(
 		return nil, err
 	}
 
-	dpopJKT, _ := dpop.GetDPoPProofJKT(ctx)
+	dpopJKT, _, err := dpop.GetDPoPProofJKT(ctx)
```
louischan-oursky: Make `dpop.GetDPoPProofJKT` take client, such that we are confident that `client.DPoPDisabled` is always respected.

pkg/lib/oauth/handler/handler_token.go
tung2744: We want to preserve the matching result in logs, only not to return error in response. So I guess doing the check outside is better.
But I think we can have a helper method to avoid code duplication.

louischan-oursky: What about you return struct containing two `error`s?

```golang
type DPoPJKTResult struct {
  ErrorForClient error
  ErrorForLogging error
}

func MatchDPoPJKT(client, proof) DPoPJKTResult
```

tung2744: What about having a helper function?
```

type OAuthDPoPChecker func(*dpop.DPoPProof) *dpop.UnmatchedJKTError

func CheckDPoPWithClient(
	ctx context.Context,
	client *config.OAuthClientConfig,
	checker OAuthDPoPChecker,
	errorLogger func(err error),
) error {
	maybeDpopProof := dpop.GetDPoPProof(ctx)
	dpopProof, err := maybeDpopProof.Get()
	if err != nil {
		if !client.DPoPDisabled {
			return err
		}
	}

	dpopErr := checker(dpopProof)
	if dpopErr != nil {
		errorLogger(dpopErr)
		if !client.DPoPDisabled {
			return err
		}
	}

	return nil
}
```

louischan-oursky: The problem is, we still cannot ensure the helper function is used everywhere. Maybe we can change the signature of the check function to take `(ctx, client, logger)`

tung2744: I see, ok let me change the check function

tung2744:
Updated, thanks!

carmenlau:
I’m wondering if there’s any way we can test this, perhaps by using an iphone with an incorrect clock or by making HTTP requests to the server directly?

tung2744:
Yes. Using incorrect clock should be able to test this pr.
I actually test it by modifying the SDK to send a incorrect key to server.

Comment this line:
https://github.com/authgear/authgear-sdk-ios/blob/main/Sources/dpop/DefaultDPoPProvider.swift#L88

tung2744:
Fixed in https://github.com/authgear/authgear-server/pull/5452/commits/5fb5937c922d043792611480d422b28808ced8b5, thanks!

carmenlau:
@tung2744 I cannot merge this PR as this is blocked and needs Louis approval. See if you can remove him from the reviewers, or open an new PR for this? Thanks!

tung2744:
I dismissed louis's review, let me merge it

PR #5453, opened 2025-11-28
Add user agent to all logs

tung2744:
ref DEV-3207

### DIFF
--- a/.vettedpositions
+++ b/.vettedpositions
@@ -339,4 +339,5 @@
 /pkg/util/pubsub/http_handler.go:51:40: requestcontext
 /pkg/util/sentry/request.go:25:9: requestcontext
 /pkg/util/slogutil/sentry_handler.go:92:11: contextbackground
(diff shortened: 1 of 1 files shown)


pkg/util/slogutil/useragent_middleware.go
```diff
@@ -0,0 +1,18 @@
+package slogutil
+
+import (
+	"log/slog"
+	"net/http"
+)
+
+type UserAgentMiddleware struct{}
+
+func (m *UserAgentMiddleware) Handle(next http.Handler) http.Handler {
```
louischan-oursky: This middleware can be simplified to a function, so we do not need the wire things.

tung2744: oh right, let me simplify it

tung2744:
Rebased, thanks!

tung2744:
CI should be fixed after https://github.com/authgear/authgear-server/pull/5462 merged

PR #5458, opened 2025-12-02
Specify IP blocklist

tung2744:
ref DEV-2398

### DIFF
--- a/docs/specs/network-protection.md
+++ b/docs/specs/network-protection.md
@@ -2,10 +2,11 @@
 
 This document specifies project-level network protection configurations.
 
-- [IP Blocklist](#ip-blocklist)
+ - [IP Filter](#ip-filter)
   - [authgear.yaml](#authgearyaml)
   - [Error response](#error-response)
   - [Use cases](#use-cases)
+  - [Portal behavior](#portal-behavior)
 
 ## IP Filter
 
@@ -156,6 +157,6 @@ The web portal may present a simplified UI for managing network protection, but 
 
 1. The portal will only modify a single rule named `__portal`. The portal never creates or edits other rules.
 
-2. The `default_action` field is not editable through the portal and cannot be modified by portal.
+2. The `default_action` field is not editable through the portal and cannot be modified by the portal.
 
-3. If the portal detects that there are multiple rules present, or the single existing rule is not named `__portal`, the portal UI will be disabled and no changes will be allowed. This is to prevent overwritting existing config accidentally.
+3. If the portal detects that there are multiple rules present, or the single existing rule is not named `__portal`, the portal UI will be disabled and no portal-managed changes will be allowed to prevent accidentally overwriting existing configuration.


docs/specs/protection.md
```diff
@@ -0,0 +1,34 @@
...
+- [IP Blocklist](#ip-blocklist)
+- [authgear.yaml](#authgearyaml)
+- [Error response](#error-response)
+
+## IP Blocklist
+
+### authgear.yaml
+
+```yaml
+protection:
```
tung2744: I found the keyword `protection` could be confusing because we already have `bot_protection` in config.

Naming of similar features in other auth services:

- Auth0: Access Control https://auth0.com/docs/secure/tenant-access-control-list
    - Auth0's access control is not only about blocking or allowing traffic, it have fine-grained access control according to different rules.
- Okta: Security > Networks https://help.okta.com/en-us/content/topics/security/healthinsight/blocklist-network-zone.htm
- AWS Cognito: Firewall Rule with WAF

I think I will just name it `network_protection`.

carmenlau:
- Design the config with fail2ban and ip allowlist

tung2744:
Updated the spec with a new `network_protection.exceptions` section.
Also added a use case section for example configuration for cases I can think of.

I did not add `fail2ban` design because in the latest meeting, it seems we think it should be placed together with rate limit configs, which is not in network_protection section.

docs/specs/network-protection.md
```diff
@@ -14,21 +15,91 @@ This document specifies project-level network protection configurations.
...
+
+```yaml
+network_protection:
+  ip_blocklist:
+    cidrs:
+      - 0.0.0.0/0
+      - ::/0
+    exceptions:
+      cidrs:
+        - 198.51.100.0/24
```
carmenlau: Under this config design, to set up authorized networks we have to block all IPs except for some. I think it’s clever, but not very obvious. I’m wondering if a rule-based configuration could make things clearer and more extensible in the future. Open to discussion.

Example:
```yaml
ip_filter:

  # Default action if no rules match
  # default is `allow` when it is not set
  default_action: deny  

  # Rules evaluated in priority order (lowest first)
  rules:
    # Allow office subnet in US
    - priority: 100
      name: allow-office-us
      action: allow
      source:
        cidr: "192.168.1.0/24"
        countries: ["US"]

    # Block known abusive IP range
    - priority: 200
      name: block-abuse-range
      action: deny
      source:
        cidrs:
          - "203.0.113.0/24"
          - "198.51.100.0/24"
      reason: "Detected abuse"

    # Block traffic from specific countries
    - priority: 300
      name: block-countries
      action: deny
      source:
        countries: ["KP", "IR", "SY", "VE", "CU"]

    # Optional catch-all deny rule (redundant with default_action)
    - priority: 1000
      name: deny-all
      action: deny
      source:
        cidr: "0.0.0.0/0"

```

To keep things simple in the first version, we can start with some required fields. The `priority` and `name` fields are just examples for future extension. We also need to define the details, for example, if both deny and allow rules match, perhaps deny should take precedence.

This design is similar to firewall models. We could also take GCP VPC or AWS WAF as references.

tung2744: I think this design is more extensible then the original design. However I have the following comments:
1. I think this design could be difficult to present in portal.
2. I understand the idea of `priority` is from firewall rules, however I think it is a bit over complicated. I think we can simply define priority by the index of the item. i.e. the first item matches the source take effect.
3. `cidrs` should always be a list, so we don't have to handle `cidr` & `cidrs`.

So I modified the example a bit:

```yaml
ip_filter:
  default_action: deny  
  rules: # Match from top to bottom, the first match take effect
    # Allow office subnet in US
    - name: allow-office-us
      action: allow
      source:
        cidrs: ["192.168.1.0/24"]
        geo_location_codes: ["US"] # I renamed this key to match the name in event context

    # Block known abusive IP range
    - name: block-abuse-range
      action: deny
      source:
        cidrs:
          - "203.0.113.0/24"
          - "198.51.100.0/24"
      reason: "Detected abuse"

    # Block traffic from specific countries
    - name: block-countries
      action: deny
      source:
        geo_location_codes: ["KP", "IR", "SY", "VE", "CU"]

    # Optional catch-all deny rule (redundant with default_action)
    - name: deny-all
      action: deny
      source:
        cidrs: ["0.0.0.0/0"]
```

For the portal UI, maybe we need to make it a list of rules if we use this design.

tung2744: Rename reason to description.

Per latest discussion, we will not have description, because we do not plan to support it in the portal at the moment.
We can extend the config to add such field when needed.

tung2744:
Updated:
1. Switched to the new ip_filter config structure.
2. Also updated use cases example with the new config structure.
3. Defined portal behavior.

@carmenlau @fungc-io

@chpapa Please also take a look if you have time. You can read the use case section directly: https://github.com/authgear/authgear-server/blob/37816b3d451f4ff191656d28b4cddd055d8360fb/docs/specs/network-protection.md#use-cases

fungc-io:
thanks, @tung2744 
lgtm

PR #5460, opened 2025-12-03
Add x_replace_project_logo_with_logo_uri for replacing project logo

carmenlau:
ref DEV-3170

### DIFF
--- a/.vettedpositions
+++ b/.vettedpositions
@@ -252,8 +252,8 @@
 /pkg/auth/handler/webapp/verify_identity.go:161:27: requestcontext
 /pkg/auth/handler/webapp/verify_identity_success.go:42:27: requestcontext
 /pkg/auth/handler/webapp/verify_login_link.go:92:27: requestcontext
-/pkg/auth/handler/webapp/viewmodels/base.go:209:9: requestcontext
-/pkg/auth/handler/webapp/viewmodels/base.go:422:9: requestcontext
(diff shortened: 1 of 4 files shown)


portal/src/graphql/portal/EditOAuthClientForm.tsx
```diff
@@ -601,35 +590,6 @@ const EditOAuthClientForm: React.VFC<EditOAuthClientFormProps> =
             value={applicationTypeLabel}
             readOnly={true}
           />
-          <TextFieldWithImagePreview
```
tung2744, line 604: I think we can simply hide it so we can restore it easily in the future.
```
// NOTE: We decided to hide this field at the moment
// See https://github.com/authgear/authgear-server/pull/5460
className="hidden"
```

carmenlau: I prefer not to include any unused code in the codebase unless we plan to use it in the very near future. Unused code can break and makes the codebase less clean. We can always retrieve it from the commit history if needed. How about I create a backlog issue for fully supporting logo_uri once we have the consent screen and add the relevant commit there for reference?

tung2744: I see, ok then lets keep it removed. I've created issue here https://linear.app/authgear/issue/DEV-3247/support-logo-uri-properly-in-portal

PR #5461, opened 2025-12-03
Implement network protection

tung2744:
ref DEV-2398

### DIFF
--- a/pkg/lib/config/network_protection.go
+++ b/pkg/lib/config/network_protection.go
@@ -21,7 +21,7 @@ var _ = Schema.Add("NetworkIPBlocklistConfig", `
 		},
 		"country_codes": {
 			"type": "array",
-			"items": { "type": "string" }
+			"items": { "type": "string", "minLength": 2, "maxLength": 2 }
 		}
 	}
 }


pkg/lib/config/network_protection.go
```diff
@@ -0,0 +1,37 @@
...
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"cidrs": {
+			"type": "array",
+			"items": { "type": "string", "format": "x_cidr" }
+		},
+		"country_codes": {
+			"type": "array",
+			"items": { "type": "string" }
```
carmenlau: I think we should validate the country code input instead of allowing free text.

pkg/lib/networkprotection/ip_blocklist_middleware.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+	}
+	return false
+}
+
+func (m *IPBlocklistMiddleware) isCountryBlocked(remoteIP string) bool {
+	if len(m.Config.IPBlocklist.CountryCodes) == 0 {
+		return false
+	}
+	if info, ok := geoip.IPString(remoteIP); ok {
```
carmenlau, line 55: Do we currently have a regular maintenance task to update the `GeoLite2-Country.mmdb` in our codebase? I think we might need one if we have a feature that relies on it.

pkg/lib/config/network_protection.go
tung2744: Right, let me restrict the input to a 2 character string.

pkg/lib/networkprotection/ip_blocklist_middleware.go
tung2744: We do not have a regular maintenance task to update this db. Let me create an issue for this.

pkg/lib/config/network_protection.go
carmenlau: Would it be better to check against a list, like `./pkg/util/territoryutil/alpha2.go`? Using a stricter rule first is usually better, we can relax it later if needed. From the comment at the top of `alpha2.go`, some items are excluded due to dependencies. Maybe we need a separate list for our validation here?

What do you think?

tung2744: The possible values of the country codes depends on the geo ip database, which we might update it later.
I do not want to maintain a list of valid country codes to validate against if we cannot derive the list from the database automatically, so I prefer a looser validation which only validate there are two characters. I believe loose validation here should not cause any problem practically.

carmenlau: OK, then restrict the input to a 2 character string only.

Thank you!

tung2744:
Rebased and added validation for country codes in https://github.com/authgear/authgear-server/pull/5461/commits/3d74e1db9a0232f734d960abb93a2b12d99fc9de, thanks!

carmenlau (approved):

PR #5463, opened 2025-12-04
Add metric for deno run counts

tung2744:
ref DEV-3231

### DIFF
--- a/pkg/lib/hook/deno_client.go
+++ b/pkg/lib/hook/deno_client.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"os"
 	"time"
 
 	"github.com/authgear/authgear-server/pkg/api/apierrors"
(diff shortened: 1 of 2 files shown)


pkg/lib/hook/deno_client.go
```diff
@@ -93,16 +95,27 @@ func (c *DenoClientImpl) Run(ctx context.Context, snippet string, input interfac
...
 		slog.Any("stderr", runResponse.Stderr),
 	).Info(ctx, "run deno script")
 
 	if runResponse.Error != "" {
+		otelutil.IntCounterAddOne(ctx,
+			otelauthgear.CounterDenoRunCount,
+			otelauthgear.WithStatusError(),
+			otelauthgear.WithDenoErrorCode(string(runResponse.ErrorCode)),
+		)
+
```
carmenlau, line 131: As mentioned, the core limits requests to 5s and 60s for blocking and non-blocking hooks. I think we should also track these as timeout cases. And this is even more useful because these timeouts are shorter or equal to Deno’s timeout, which means we may not have many opportunities to trigger Deno’s own timeout in practice.

tung2744: Added https://github.com/authgear/authgear-server/pull/5463/commits/48f69749799b1757e1edaa350cdd9d9a2eadf171 to ensure all cases are tracked by metric, thanks!

You can test it with this script with `authentication.pre_initialize` blocking event:
```typescript
import { EventAuthenticationPreInitialize, EventAuthenticationPreInitializeHookResponse } from "https://deno.land/x/authgear_deno_hook@v2.0.0/mod.ts";

export default async function(e: EventAuthenticationPreInitialize): Promise<EventAuthenticationPreInitializeHookResponse> {
  await new Promise(resolve => setTimeout(resolve, 10000));
  return { is_allowed: true };
}
```

PR #5464, opened 2025-12-05
Update error message for disposable/Free email error

carmenlau:
ref DEV-3248

### DIFF
--- a/pkg/lib/authn/identity/loginid/type_checkers.go
+++ b/pkg/lib/authn/identity/loginid/type_checkers.go
@@ -208,7 +208,10 @@ func (c *EmailChecker) Validate(ctx context.Context, validationCtx *validation.C
 			return
 		}
 		if matched {
-			validationCtx.EmitError("blocked", map[string]interface{}{"reason": "EmailDomainBlocklist"})
+			validationCtx.EmitError("blocked", map[string]interface{}{
+				"reason":    "EmailDomainBlocklist",
+				"blocklist": "custom",
+			})
 			return
 		}
 	}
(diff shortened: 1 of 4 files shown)


pkg/lib/authn/identity/loginid/type_checkers.go
```diff
@@ -233,7 +233,7 @@ func (c *EmailChecker) Validate(ctx context.Context, validationCtx *validation.C
 			return
 		}
 		if matched {
-			validationCtx.EmitError("blocked", map[string]interface{}{"reason": "EmailDomainBlocklist"})
+			validationCtx.EmitError("blocked", map[string]interface{}{"reason": "EmailDisposableDomainBlocklist"})
```
tung2744: I suggest we keep the api reason so there will be not breaking change.
Add one more field to tell which blocklist the email violate.
```
{
  "reason": "EmailDomainBlocklist",
  "blocklist": "free" | "disposable"
}
```

carmenlau: Sure, let me add `{ "blocklist": "free" | "disposable" | "custom" }`.

For the breaking change you mentioned, is it related to the importer? I just want to make sure I understand it correctly.

tung2744: For breaking change, there could be existing user enabled the free email blocklist.
And the error reason will be returned in apis, such as admin api and authflow api. If the reason changed, existing integration code which check the old reason might break.

carmenlau: Got it and updated. Thank you!!

PR #5466, opened 2025-12-05
Specify rate_limit.blocked event

tung2744:
ref DEV-3253

@fungc-io @carmenlau Please see the design of the new rate limit blocked event, and let me know if you think any other additional information is needed in the payload.

### DIFF
--- a/docs/specs/hook.md
+++ b/docs/specs/hook.md
@@ -221,7 +221,7 @@ The constraints are enforced based on the authentication flow type:
 
 ## Using Blocking Event with Rate Limits
 
-The `rate_limits` property in a blocking event response allows you to dynamically override rate limit weights. The value of this property is an object where keys are rate limit group names. For a list of available rate limit groups, see [the rate limit spec](./rate-limit.md).
+The `rate_limits` property in a blocking event response allows you to dynamically override rate limit weights. The value of this property is an object where keys are rate limit group. For a list of available rate limit groups, see [the rate limit spec](./rate-limit.md).
 
 Each rate limit key maps to an object with the following property:
 


docs/specs/event.md
```diff
@@ -107,6 +107,7 @@ All fields are guaranteed that only backward-compatible changes would be made.
 - `user_id`: The ID of the user associated with the event. It may be absent. For example, the user has not authenticated yet.
 - `ip_address`: The IP address of the HTTP request, if present.
 - `user_agent`: The User-Agent HTTP request header, if present.
+- `url`: The url of the HTTP request, if present.
```
carmenlau: I’m wondering whether adding the URL to the public API could be a concern. It might limit our ability to change the AuthUI route path freely. It’s fine if we already accept this restriction.

Also, does this URL include query parameters?

docs/specs/event.md
```diff
@@ -537,6 +538,7 @@ Use this event to add custom fields to the ID token.
 - [identity.oauth.disconnected](#identityoauthdisconnected)
 - [identity.biometric.enabled](#identitybiometricenabled)
 - [identity.biometric.disabled](#identitybiometricdisabled)
+- [rate_limit.blocked](#rate_limitblocked)
```
carmenlau, line 543: Is this event for audit logs only, or does it also apply to non-blocking hooks?

tung2744: Should be for audit logs only. However I think the limitation can be relaxed easily if there are needs.

docs/specs/event.md
tung2744: > I’m wondering whether adding the URL to the public API could be a concern. It might limit our ability to change the AuthUI route path freely. It’s fine if we already accept this restriction.

I think we will not guarantee the url will not be changing. So it exist only for a record purpose.

> does this URL include query parameters?

In my current plan, yes, because I think it could be useful for cases like authorization request.
One thing we might need to consider is the current query parameter might contain user PII (e.g. in id token), but I think in other audit logs we already have user phone number / email, and audit logs will be cleared after the retention period so I guess it should be ok.

docs/specs/event.md
carmenlau: That sounds good. But I think it would be better to have a separate section for audit logs in the specification, it’s a bit confusing as it is now.

docs/specs/event.md
carmenlau: > I think we will not guarantee the url will not be changing. So it exist only for a record purpose.

Putting the URL in the context means it will be included in all events. I think it will be difficult for us to control how users make use of this data, and we may get questions whenever something changes. In my opinion, I would only add it if it’s truly necessary, and I would try to avoid it otherwise.

So I’d like to clarify, are we adding this for debugging purposes, or does the bucket parameter in the rate_limit.blocked event already serve that purpose?

> One thing we might need to consider is the current query parameter might contain user PII

I was thinking the same thing :)

tung2744: > So I’d like to clarify, are we adding this for debugging purposes, or does the bucket parameter in the rate_limit.blocked event already serve that purpose?

The bucket name / rate limit name tells you which rate limit you hit, however, you do not know where do you hit this rate limit.
For example, the rate limit `verification.sms.trigger` could be triggered in authui, settings. If the client uses custom ui, it could also be triggered by authflow api. So we want to know the url to understand where the end-user hit the rate limit.

And I think it is mainly for the portal admin / project owner to answer end-user queries about what rate limits they are hitting. I imagine the portal admin can use it as follows:
1. They got a question about hitting rate limit in authui enter password screen.
2. The portal admin search for the `rate_limit.blocked` event in portal.
3. For each entry, they look at the url to check if it is enter password screen. Then, find an entry that most likely related. And look at the rate_limit name to understand why the end-user is hitting rate limit.

docs/specs/event.md
tung2744: I saw many events like `authentication.identity.login_id.failed` `authentication.primary.password.failed` are also audit only and listed in this page.
I think this page is for all events, no matter hook enabled / audit enabled.
I guess we are missing a document for audit events only, not sure if I should create such a document?

docs/specs/event.md
carmenlau: After checking the current events, I found we have something called `audit_context`. Should we put the URL there so it’s only for audit purposes?

However, I think `audit_context` is currently used only for admin-related events, right?

docs/specs/event.md
carmenlau: To keep things simple, maybe we could update the table of contents on this page to have three sections: `blocking events`, `non-blocking events`, and `audit logs`.

An event could appear in both the non-blocking events and audit logs sections if applicable.

What do you think?

docs/specs/event.md
tung2744: Right, it is used for providing additional context by admin-api triggered event, and can be set by a key in the admin api jwt.

I think we can simply mention the url of authui / settings ui is subject to change and we have no guarantee on the value:
https://docs.authgear.com/customization/events-hooks

docs/specs/event.md
tung2744: Sure, let me update the doc to add a audit logs section.

tung2744: Added a list of events with audit log support in https://github.com/authgear/authgear-server/pull/5466/commits/4985da340eeac37ebee46c7d3e03d710701ac72f

docs/specs/event.md
tung2744: Per latest discussion, lets put this into audit_context

tung2744:
Moved url into audit_context in [885e7b3](https://github.com/authgear/authgear-server/pull/5466/commits/885e7b348687a2ad043c5a21bf35af0ee57aa0f9)

For other fields of `audit_context`, and the feature of adding additional fields to it in admin api, maybe I will keep them as unspecified at the moment because maybe we need to design it carefully before exposing as a public api. The implementation will alway guarantee `audit_context` is an object with a field `http_url`.

carmenlau:
- Update event payload to remove bucket
- Check where have we used rate limit names
- [In authgear core] Add comments to the existing API to clarify that we are using the bucket field and that is for internal use only

tung2744:
Updated:
1. Rewritten the rate limit spec, now rate limit names include `per_ip` or `per_user` suffix.
2. Updated the rate_limit.blocked event payload to use the new rate limit names, and do not expose the bucket name.

tung2744:
Added one more commit to update the hook spec, which reference the rate limit names.

docs/specs/event.md
```diff
@@ -1123,6 +1127,68 @@ Occurs when biometric login is disabled. It will be triggered only when the user
...
+#### rate_limit.blocked
+
+Occurs after a request is blocked by rate limiting.
+
+Payload:
+```json5
+{
+  "payload": {
+    "rate_limit_group": "authentication.password",
+    "rate_limit_name": "authentication.password.per_ip",
```
tung2744: A second thought:

Maybe we can change this to an object:

```
{
  "rate_limit": {
    "name": "authentication.password.per_ip",
    "group": "authentication.password"
  }
}
```

carmenlau: Sorry for late reply. 🙇🏻‍♀️ This payload looks good to me. 

At the same time, I think we should also find out where we expose `rate_limit` in the public API. I believe it should be here: [error.go](https://github.com/authgear/authgear-server/blob/main/pkg/lib/ratelimit/error.go). We currently expose `bucket` there, and we should consider proposing some adjustments. One idea is to add `rate_limit.name` and `rate_limit.group` to the payload, and mark `rate_limit_name` and `bucket_name` as legacy while keeping them to avoid a breaking change. However, this would make the payload look a bit confusing.🫤

```json
{
  "name": "TooManyRequest",
  "reason": "RateLimited",
  "message": "request rate limited",
  "code": 429,
  "info": {
    "FlowType": "login",
    "bucket_name": "AccountEnumerationPerIP",
    "rate_limit_name": "authentication.account_enumeration",
    "rate_limit": {
      "name": "authentication.account_enumeration",
      "group": "authentication.account_enumeration.per_ip"
    }
  }
}

```

Also check if anywhere else have similar problems.

tung2744: I have checked we only use `bucket_name` at the moment. `rate_limit_name` should be safe to remove.

The place that actually uses the rate limit group name is hook. You can see this commit: https://github.com/authgear/authgear-server/pull/5466/commits/c1eaa57a882b5255975bfb48ab9db1f6b753f4c9

carmenlau: > I have checked we only use bucket_name at the moment. rate_limit_name should be safe to remove.

But is there any chance we return this error in our APIs that users might reference? Is it the Custom UI API?

> The place that actually uses the rate limit group name is hook.

We’re using the group name as the key there, not something like `rate_limit_name`, so it looks fine, right?

tung2744: > But is there any chance we return this error in our APIs that users might reference? Is it the Custom UI API?

I would not say there is no chance, but this thing is introduced recently, and I've also checked custom UI codes which I have access to and do not see any usage of the key. I think it should be safe to remove. And actually the original key is `rate_limit_name` but it is a group, which is more confusing if we do not remove it.

> We’re using the group name as the key there, not something like rate_limit_name, so it looks fine, right?

It is named `rate_limits`. An example:

```json
{
  "is_allowed": true,
  "rate_limits": {
    "authentication.account_enumeration": {
      "weight": 2
    }
  }
}
```

It is designed in this way because we do not want the user to list all related rate limits when he need to modify the weight.

docs/specs/hook.md
```diff
@@ -221,7 +221,7 @@ The constraints are enforced based on the authentication flow type:
 
 ## Using Blocking Event with Rate Limits
 
-The `rate_limits` property in a blocking event response allows you to dynamically override rate limit weights. The value of this property is an object where keys are rate limit names. For a list of available rate limit names, see [the rate limit spec](./rate-limit.md).
+The `rate_limits` property in a blocking event response allows you to dynamically override rate limit weights. The value of this property is an object where keys are rate limit group names. For a list of available rate limit groups, see [the rate limit spec](./rate-limit.md).
```
carmenlau: `'s/rate limit group names/rate limit groups/g'`

carmenlau (changes requested):
The rest looks fine

tung2744:
Updated, thanks!

carmenlau (approved):

PR #5468, opened 2025-12-09
Add rate limit blocked event

tung2744:
ref DEV-3253

### DIFF
--- a/e2e/tests/hook/authentication.post_identified/rate_limits.test.yaml
+++ b/e2e/tests/hook/authentication.post_identified/rate_limits.test.yaml
@@ -76,7 +76,10 @@ steps:
           "info": {
             "FlowType": "login",
             "bucket_name": "VerifyPasswordPerIP",
-            "rate_limit_name": "authentication.password.per_ip",
-            "rate_limit_group": "authentication.password"
+            "rate_limit_name": "authentication.password",
+            "rate_limit": {
+              "name": "authentication.password.per_ip",
+              "group": "authentication.password"
+            }
           }
         }
--- a/e2e/tests/hook/authentication.pre_initialize/rate_limits.test.yaml
+++ b/e2e/tests/hook/authentication.pre_initialize/rate_limits.test.yaml
@@ -60,7 +60,10 @@ steps:
           "info": {
             "FlowType": "login",
             "bucket_name": "AccountEnumerationPerIP",
-            "rate_limit_name": "authentication.account_enumeration.per_ip",
-            "rate_limit_group": "authentication.account_enumeration"
+            "rate_limit_name": "authentication.account_enumeration",
+            "rate_limit": {
+              "name": "authentication.account_enumeration.per_ip",
+              "group": "authentication.account_enumeration"
+            }
           }
         }


tung2744:
Found a bug, mark as WIP first

pkg/lib/ratelimit/error.go
```diff
@@ -5,17 +5,21 @@ import (
...
+func ErrRateLimited(rl RateLimitName, rlgroup RateLimitGroup, bucketName BucketName) error {
 	details := apierrors.Details{
+		// Deprecated field. Do not use.
+		// Use rate_limit_name instead.
 		bucketNameKey: bucketName,
 	}
 	// Some buckets do not have a rate limit name, so do not add the key if it is empty
 	if rl != "" {
 		details[rateLimitNameKey] = rl
+		details[rateLimitGroupKey] = rlgroup
```
carmenlau: Please update the payload per our discussion

pkg/lib/authn/authenticator/service/service_readonly.go
```diff
@@ -0,0 +1,134 @@
...
+// This service was created for read only methods and do not depends on RateLimiter to break this circular dependency
+type ReadOnlyService struct {
+	Store    *Store
+	Password PasswordAuthenticatorProvider
+	Passkey  PasskeyAuthenticatorProvider
+	TOTP     TOTPAuthenticatorProvider
+	OOBOTP   OOBOTPAuthenticatorProvider
+}
+
+func (s *ReadOnlyService) List(ctx context.Context, userID string, filters ...authenticator.Filter) ([]*authenticator.Info, error) {
```
carmenlau, line 22: After moving the functions here, maybe we can remove the implementation in `pkg/lib/authn/authenticator/service/service.go`?

pkg/api/event/hook_response.go
```diff
@@ -200,7 +199,9 @@ type BotProtectionRequirements struct {
 	Mode config.BotProtectionRiskMode `json:"mode,omitempty"`
 }
 
-type RateLimits map[ratelimit.RateLimit]RateLimitRequirements
+// This should be map[ratelimit.RateLimit]RateLimitRequirements,
+// but this will cause import cycle. So use string.
+type RateLimits map[string]RateLimitRequirements
```
carmenlau, line 204: This isn’t newly added. I’d like to understand why we didn’t have an import cycle issue before.

tung2744: https://github.com/authgear/authgear-server/pull/5468/commits/ca8161790957c3578ee701fb68ad0d9cfabedb12

In this commit, we import the event package in ratelimit package.
Then, then import cycle occurs because now ratelimit and event is importing each other.

pkg/lib/authn/authenticator/service/service_readonly.go
tung2744: Right, let me remove the duplicated method.

tung2744:
Updated, thanks!

pkg/lib/ratelimit/error.go
```diff
@@ -2,20 +2,26 @@ package ratelimit
 
 import (
 	"github.com/authgear/authgear-server/pkg/api/apierrors"
+	"github.com/authgear/authgear-server/pkg/api/model"
 )
 
-const rateLimitNameKey = "rate_limit_name"
```
carmenlau, line 7: Per our discussion, we should keep both `rate_limit_name` and `bucket_name`, and their values should remain unchanged.

This should be temporary until we confirm that our clients aren’t using it in their customizations, right?

tung2744: Yes you are correct, let me add back `rate_limit_name` which the value is the rate limit group.

PR #5472, opened 2025-12-16
Follow up of setting sms quota of free plan to 0

tung2744:
ref DEV-3264

Screenshot 2025-12-15 at 5 13 29 PM


carmenlau:
Show dialog only when phone is enabled

tung2744:
Updated, thanks!

carmenlau (approved):

PR #5474, opened 2025-12-16
Log an error when someone see direct access blocking

tung2744:
ref DEV-3262

You can test by setting
```yaml
ui:
  direct_access_disabled: true
```

### DIFF
--- a/.vettedpositions
+++ b/.vettedpositions
@@ -56,8 +56,7 @@
 /pkg/auth/handler/saml/logout.go:622:31: requestcontext
 /pkg/auth/handler/webapp/account_status.go:48:27: requestcontext
 /pkg/auth/handler/webapp/app_static_assets.go:39:33: requestcontext
-/pkg/auth/handler/webapp/auth_entry_point_middleware.go:31:31: requestcontext
-/pkg/auth/handler/webapp/auth_entry_point_middleware.go:32:35: requestcontext
+/pkg/auth/handler/webapp/auth_entry_point_middleware.go:35:10: requestcontext
 /pkg/auth/handler/webapp/authflow_change_password.go:96:26: requestcontext
 /pkg/auth/handler/webapp/authflow_controller.go:1034:30: requestcontext
 /pkg/auth/handler/webapp/authflow_controller.go:1039:24: requestcontext


pkg/auth/handler/webapp/auth_entry_point_middleware.go
```diff
@@ -47,6 +52,29 @@ func (m *AuthEntryPointMiddleware) Handle(next http.Handler) http.Handler {
...
+				samlSessionID = webSession.SAMLSessionID
+			}
+			allCookies := r.Cookies()
+			for _, cookie := range allCookies {
+				if cookie.Value != "" {
+					// Log all existing cookies
+					cookies = append(cookies, cookie.Name)
+				}
+			}
+			logger.WithSkipStackTrace().WithSkipLogging().Error(ctx, "auth direct access blocked",
```
carmenlau, line 72: - `WithSkipLogging` Does this mean we’ll only be able to see the logs in stderr and not anywhere else (e.g. sentry)?
- Since this endpoint is still accessible and I think it shouldn’t be treated as an error, maybe we could use a warning instead? Also, what is the log level in our deployments? I just want to make sure we’ll still be able to see it if we change it to a warning.

tung2744: > WithSkipLogging Does this mean we’ll only be able to see the logs in stderr and not anywhere else (e.g. sentry)?

Correct.

> Since this endpoint is still accessible and I think it shouldn’t be treated as an error, maybe we could use a warning instead? Also, what is the log level in our deployments? I just want to make sure we’ll still be able to see it if we change it to a warning.

Our log level is error, this is why I used error here. I think currently we do not have a way to log some non-error information.

carmenlau: https://github.com/authgear/authgear-server/blob/adb0fe418603d9da1da895b3febca8e939a8603c/pkg/lib/config/environment.go#L78

But the default log level is `warn`. Maybe we should check whether it’s explicitly set to `error`, and also check whether warning logs are visible on the servers.

tung2744: Yes, it was explicitly set to error in current deployements.

carmenlau: Okay, then let’s keep this log as an error for now. But if we find more logs like this, maybe we should review them and consider adjusting the server log level.”

tung2744: https://linear.app/authgear/issue/DEV-2438/support-fine-grained-leveled-logging

We have this issue opened but no time to do it yet 😿

carmenlau: It’s great that we’re aware of it, we can address it when we start working on improvements. :)

PR #5479, opened 2025-12-18
Fix crash when promoting anonymous user with cookie session

tung2744:
ref DEV-3286

You can test using the reactweb demo in js sdk with this fix:
https://github.com/authgear/authgear-sdk-js/pull/405

Notes:
- You need to use localhost:3100 as public_origin (Such that it share cookies with localhost)


tung2744:
@carmenlau This is an issue in production, please see if you can review first, thanks!

PR #5484, opened 2025-12-22
Add a temporary token for admin api page for testing

carmenlau:
ref DEV-3252

The implementation is a bit different from the design
1. I kept the copy icon in blue for consistency.
1. I didn't bold the word "Generate" since it is a text field placeholder

Screenshot 2025-12-22 at 06 42 57
Screenshot 2025-12-22 at 06 43 11

### DIFF
--- a/pkg/portal/graphql/token_mutation.go
+++ b/pkg/portal/graphql/token_mutation.go
@@ -88,7 +88,7 @@ var _ = registerMutationField(
 			keyID := secretConfig.AdminAPISecrets[0].KeyID
 			privateKeyPEM := secretConfig.AdminAPISecrets[0].PrivateKeyPEM
 			if privateKeyPEM == nil {
-				return nil, apierrors.NewInvalid("invalid secret token")
+				return nil, apierrors.NewForbidden("invalid secret token")
 			}
 
 			logger := tokenMutationLogger.GetLogger(ctx)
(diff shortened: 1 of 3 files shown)


pkg/portal/graphql/token_mutation.go
```diff
@@ -0,0 +1,105 @@
...
+				return nil, err
+			}
+
+			secretConfig, _, err := gqlCtx.AppService.LoadAppSecretConfig(ctx, app, sessionInfo, appSecretVisitToken)
+			if err != nil {
+				return nil, err
+			}
+
+			if len(secretConfig.AdminAPISecrets) == 0 {
+				return nil, apierrors.NewInvalid("no admin API auth key found")
```
tung2744: I think this can be a internal error, because project should have at least one key.

portal/src/graphql/portal/AdminAPIConfigurationScreen.tsx
```diff
@@ -167,8 +189,51 @@ const AdminAPIConfigurationScreenContent: React.VFC<AdminAPIConfigurationScreenC
...
+        });
+      };
+      // There are two possible cases:
+      // 1. secretToken is null
+      // 2. secretToken is not null but failed to generate short-lived admin API token
+      // in both cases, reauthentication is needed
+      if (secretToken != null) {
+        generateShortLivedAdminAPITokenHandle().catch(() => {
+          reauthentication();
+        });
```
tung2744: Can we check the actual error thrown, and only call `reauthentication` if the error is related?
I think there might be other errors which we want to log an error and probably display error ui  to user. e.g. Network error or server error.

carmenlau:
Updated, thank you!

tung2744 (approved):

PR #5485, opened 2025-12-23
Implement new ip filter config & portal UI

tung2744:
ref DEV-3293
ref DEV-3297

https://github.com/user-attachments/assets/d006cd8a-9348-4f18-80de-0e0b17669d1e


pkg/lib/config/network_protection.go
```diff
@@ -19,19 +26,70 @@ var _ = Schema.Add("NetworkIPBlocklistConfig", `
 			"type": "array",
 			"items": { "type": "string", "format": "x_cidr" }
 		},
-		"country_codes": {
+		"geo_location_codes": {
+			"type": "array",
+			"items": { "type": "string" }
```
carmenlau: We should keep the restriction that only allows two-character country codes

pkg/lib/networkprotection/ip_blocklist_middleware.go
```diff
@@ -11,31 +11,80 @@ import (
...
+	for _, rule := range m.Config.IPFilter.Rules {
+		if m.isRuleMatched(remoteIP, remoteIPStr, rule) {
+			return rule.Action
+		}
+	}
+
+	return defaultAction
+}
+
+func (m *IPFilterMiddleware) isRuleMatched(remoteIP net.IP, remoteIPStr string, rule *config.IPFilterRule) bool {
```
carmenlau: I think it would be better for the middleware and the portal `checkIP` mutation to use the same implementation, so we can ensure they produce the same results even if we update the logic in the future. Otherwise, it’s easy to miss updating the portal when changes are made here.

Thought?

portal/src/hook/useAppConfigForm.ts
```diff
@@ -134,6 +134,7 @@ export function useAppConfigForm<State>(
           appConfigChecksum: rawAppConfigChecksum,
           ignoreConflict: ignoreConflict,
         });
+        await reload();
```
carmenlau, line 137: > Fix app config form will show incorrect state for a while after save

I'd like to know how can I trigger this?

portal/src/locale-data/en.json
```diff
@@ -1922,6 +1924,19 @@
   "AnalyticsSignupMethodsWidget.chart.anonymous.label": "Anonymous",
   "AnalyticsSignupMethodsWidget.no-data-available.label": "No Data Available",
 
+  "IPBlocklistScreen.title": "IP Blocklist",
+  "IPBlocklistScreen.description": "Block traffic by IP address or geographic region to prevent unwanted access and reduce automated threats.",
+  "IPBlocklistForm.enable.label": "Enable IP Blocklist",
+  "IPBlocklistForm.ip-address.label": "Blocked IP Addresses",
+  "IPBlocklistForm.ip-address.description": "Enter a comma-separated list of IP addresses to block. IPv4, IPv6, and CIDR notation for ranges are supported.",
```
carmenlau: Should update the copywriting to use newline instead

carmenlau (changes requested):
Screenshot 2025-12-24 at 11 42 20

When clicking `Check` with an invalid IP address, the error message isn’t handled.

portal/src/hook/useAppConfigForm.ts
tung2744: You can try to revert the fix, when you save the form, you will see an instant the form reverted to the original state. This is because `setCurrentState(null)` will cause the form state reverted to the initial state. And before the config query finish, it is still the old config.

tung2744:
Updated, thanks!
Changes start from [3685257](https://github.com/authgear/authgear-server/pull/5485/commits/36852576e89b4293af5b4f6533d257b6f492607e).

PR #5500, opened 2026-01-14
Upgrade react router to v7, react to 18

tung2744:
ref DEV-2382
ref DEV-2345

### DIFF
--- a/portal/src/locale-data/en.json
+++ b/portal/src/locale-data/en.json
@@ -367,7 +367,7 @@
   "UserDetails.account-security.mark-as-expired-confirm-dialog.message": "The user will be asked to change password at the next login.",
   "UserDetails.account-security.mark-as-expired-confirm-dialog.message.revoke": "The user will NOT be asked to change password at the next login.",
   "UserDetails.account-security.secondary.grant-grace-period": "Grant grace period to set up 2FA",
-  "UserDetails.account-security.secondary.update-existing-grace-period": "<extend/> or <cancel/>",
+  "UserDetails.account-security.secondary.update-existing-grace-period": "<extend></extend> or <cancel></cancel>",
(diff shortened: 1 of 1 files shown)


tung2744:
Updated to migrate to react-intl

portal/src/locale-data/en.json
```diff
@@ -115,14 +106,14 @@
   "errors.password-policy.containing-excluded-keywords": "Password contains excluded keywords, please check password policies below",
   "errors.password-policy.unknown": "Password is invalid; please check password policies",
   "errors.resource-too-large": "The request is too big, try upload the images one at a time",
-  "errors.webhook.disallowed": "Operation is disallowed by one of your hook. Raw info: {code, react, children{{info}}}",
+  "errors.webhook.disallowed": "Operation is disallowed by one of your hook. Raw info: <code>{info}</code>",
   "errors.webhook.timeout": "Operation is disallowed because the hook delivery resulted in a timeout. Make sure your hooks return a valid response in reasonable time.",
-  "errors.webhook.invalid-response": "Operation is disallowed because one of your hook returned an invalid response. See {DocLink, react, href{https://docs.authgear.com/customization/events-hooks} children{the documentation}} for details.",
+  "errors.webhook.invalid-response": "Operation is disallowed because one of your hook returned an invalid response. See <docLink>the documentation</docLink> for details.",
```
carmenlau, line 111: I prefer the original approach where we can include links in the translation strings. It would be easier if we ever need different links for different languages, even though we don’t need that right now. However, if this is a limitation of `react-intl`, we can leave it as it is for now.

carmenlau: I did some checks, and it seems that having attributes in rich text tags is not supported. 😕

tung2744: Yes, we cannot define props in the messages, therefore I rewrite all messages to pass react components as values.

PR #5501, opened 2026-01-14
Fix unable to verify phone or email in settings when verification is not required 

carmenlau:
ref DEV-3333

**Verify email in settings**

https://github.com/user-attachments/assets/f543e6f0-e42a-47dd-8c0e-90227d70dfb3


tung2744:
I am not sure if it is a correct behavior to only show verified label if verification is required. Maybe @fungc-io
If this is the correct behavior, "Not Verified" will never be shown?

carmenlau:
The label and button will be hidden only if verification is not enabled (config: `verification.claims.email.enabled`). If verification is not required but still enabled, they will still be shown.

However, I just noticed that `verification.claims.email.enabled` actually means **“Allow users to verify their email address from the setting page.”** in the Portal. If that’s the case, I think maybe we should hide only the button, but not the label, when `enabled` is false. Thoughts?

tung2744:
I see. Agree to hide the button if verification is disabled.

carmenlau:
Updated, thanks!

PR #5510, opened 2026-01-26
Setup otel log handler and context

tung2744:
ref DEV-3341
ref DEV-3345

### DIFF
--- a/pkg/lib/config/environment.go
+++ b/pkg/lib/config/environment.go
@@ -74,8 +74,6 @@ type EnvironmentConfig struct {
 	TrustProxy TrustProxy `envconfig:"TRUST_PROXY" default:"false"`
 	// DevMode sets whether the server would be run under development mode
 	DevMode DevMode `envconfig:"DEV_MODE" default:"false"`
-	// Log configures the logging
-	Log LogEnvironmentConfig `envconfig:"LOG"`
 	// SentryDSN sets the sentry DSN.
 	SentryDSN SentryDSN `envconfig:"SENTRY_DSN"`
 	// AuthUISentryDSN sets the sentry DSN for auth ui.
--- a/pkg/util/otelutil/setup.go
+++ b/pkg/util/otelutil/setup.go
@@ -19,6 +19,7 @@ import (
 	"go.opentelemetry.io/otel/sdk/trace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.34.0"
 
+	"github.com/authgear/authgear-server/pkg/api/logging"
 	"github.com/authgear/authgear-server/pkg/version"
 )
 
@@ -97,7 +98,7 @@ func SetupOTelSDKGlobally(ctx context.Context) (outCtx context.Context, shutdown
 
 	shutdownFuncs = append(shutdownFuncs, traceProvider.Shutdown)
 
-	otlpEndpoint := os.Getenv("LOG_HANDLER_OTLP_ENDPOINT")
+	otlpEndpoint := logging.LoadConfig().OTLPEndpoint
 	var logOptions []log.LoggerProviderOption
 	logOptions = append(logOptions, log.WithResource(res))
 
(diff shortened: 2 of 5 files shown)


cmd/authgear/background/main.go
```diff
@@ -19,15 +17,6 @@ func (c *Controller) Start(ctx context.Context) {
...
 	}
-	ctx, shutdown, err := otelutil.SetupOTelSDKGlobally(ctx)
-	if err != nil {
-		err = fmt.Errorf("failed to setup otel: %w", err)
-		panic(err)
-	}
-	defer func() {
-		_ = shutdown(ctx)
-	}()
-	ctx = slogutil.Setup(ctx)
```
carmenlau, line 22: This line was here originally, or do we actually not need it?

pkg/lib/config/environment.go
```diff
@@ -74,8 +74,8 @@ type EnvironmentConfig struct {
 	TrustProxy TrustProxy `envconfig:"TRUST_PROXY" default:"false"`
 	// DevMode sets whether the server would be run under development mode
 	DevMode DevMode `envconfig:"DEV_MODE" default:"false"`
-	// LogLevel sets the global log level
-	LogLevel string `envconfig:"LOG_LEVEL" default:"warn"`
+	// Log configures the logging
+	Log LogEnvironmentConfig `envconfig:"LOG"`
```
carmenlau: The logger read the environment variable separately. It seems nowhere is referencing it, should we remove it to avoid confusion.


https://github.com/authgear/authgear-server/pull/5510/changes#diff-d2d4e93d25fdb9af8e6156be6f568ca5e978bfc2214602658dc790eee0134f00R75

pkg/util/otelutil/setup.go
```diff
@@ -80,6 +88,34 @@ func SetupOTelSDKGlobally(ctx context.Context) (outCtx context.Context, shutdown
...
+	// Set up trace provider.
+	traceProvider, err := newTracerProvider()
+	if err != nil {
+		return
+	}
+	otel.SetTracerProvider(traceProvider)
+
+	shutdownFuncs = append(shutdownFuncs, traceProvider.Shutdown)
+
+	otlpEndpoint := os.Getenv("LOG_HANDLER_OTLP_ENDPOINT")
```
carmenlau: The environment variable is defined in `LogEnvironmentConfig`, so it’s better to use it there to process it. Otherwise, it may cause problems if we need to update the name in the future.

CONTRIBUTING.md
```diff
@@ -654,3 +654,18 @@ Bad:
...
+
+In local development, logs are sent to a loki instance on `http://localhost:3102`.
+
+You can query the logs using the following command (logcli need to be installed):
+
+```sh
+logcli query --addr="http://localhost:3102" \
+  '{service_name="authgear"} | app="YOUR_APP"' \
+  --since=1h \
+  -o jsonl -q
```
carmenlau, line 668: It works great :)

cmd/authgear/background/main.go
tung2744: Yes, it is actually not needed. It was already called in the root cmd.

pkg/lib/config/environment.go
tung2744: I want to have a centralized place for all environment variables, therefore I still define it here, so we can know what environment variables exist by looking at just one package.

pkg/util/otelutil/setup.go
tung2744: The reason I put it here is because importing slogutil here will cause import cycle.
Maybe I should move the environment config struct to another independent package.

pkg/lib/config/environment.go
tung2744: Finally I removed this field, because I found it difficult to ensure the `LOG` prefix is always used to load the config.
I also moved the struct to api package to solve import cycles.

PR #5515, opened 2026-01-26
Fix high severity errors in sonarcloud

tung2744:
ref DEV-3350

### DIFF
--- a/once/docker-entrypoint.sh
+++ b/once/docker-entrypoint.sh
@@ -3,7 +3,7 @@
 # See https://www.shellcheck.net/wiki/SC3001
 
 check_user_is_correct() {
-	if [ "$(id -u -n)" != "authgear" ]; then
+	if [[ "$(id -u -n)" != "authgear" ]]; then
 		printf 1>&2 "docker-entrypoint.sh is supposed to be run with the user authgear.\n"
 		exit 1
 	fi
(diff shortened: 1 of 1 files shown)


pkg/lib/infra/db/hook_handle.go
```diff
@@ -263,18 +259,19 @@ func (*HookHandle) IsInTx(ctx context.Context) bool {
...
 	logger := HookHandleLogger.GetLogger(ctx)
 
 	// Pass a nil TxOptions to use default isolation level.
 	var txOptions *sql.TxOptions
 	tx, err := conn.BeginTx(ctx, txOptions)
 	if err != nil {
-		return nil, fmt.Errorf("hook-handle: failed to begin transaction: %w", err)
+		return fmt.Errorf("hook-handle: failed to begin transaction: %w", err)
 	}
+	defer func() { _ = rollbackTx(ctx, tx) }()
```
carmenlau, line 273: When I first saw this, I wasn’t sure whether calling rollback multiple times would be a problem (we also call it in do). However, it turns out it’s fine, as confirmed by the Go source code (see: https://go.dev/src/database/sql/sql.go#L2318). Could we add a comment to explain this?

pkg/lib/saml/service.go
```diff
@@ -699,7 +700,7 @@ func (s *Service) ConstructSignedQueryParameters(
 		return nil, err
 	}
 
-	rawSignature, err := rsa.SignPKCS1v15(nil, key, signingCtx.Hash, digest)
+	rawSignature, err := key.Sign(rand.Reader, digest, signingCtx.Hash)
```
carmenlau, line 703: That’s great that we have a test case to verify the signature, but have we tested it in a SAML setup?

tung2744: Yes. You can also try it in https://authgear-demo-saml-sp.pandawork.com/

There is another tool easier to use, but it seems to be bugged https://samlsp.com/en/

pkg/lib/infra/db/hook_handle.go
tung2744: This is actually documented in this example: https://go.dev/doc/database/execute-transactions

This seems to be the best practice. Let me add a comment to the link.

PR #5527, opened 2026-02-09
Specify fraud protection

tung2744:
ref DEV-3374

### DIFF
--- a/docs/specs/fraud-protection.md
+++ b/docs/specs/fraud-protection.md
@@ -64,6 +64,17 @@ fraud_protection:
     action: record_only
 ```
 
+**authgear.features.yaml**
+
+Fraud protection configurability is controlled by a feature flag. By default, fraud protection settings are **not modifiable** by project admins. To allow customization, set:
+
+```yaml
+fraud_protection:
+  is_modifiable: true
+```
+
+When `is_modifiable` is `false` (default), projects must use the default configuration above and cannot customize warnings or decisions.
+
 ### Warnings
 
 We have identified two primary patterns of SMS pumping attacks:


docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,373 @@
...
+### Config
+
+**authgear.yaml**
+
+```yaml
+fraud_protection:
+  enabled: true
+  geo_location_risks:
+    high:
+      - EG
```
chpapa: What's the standard used here? Or is it better to just use the phone number country code?

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,373 @@
...
+      enabled: true
+    - type: SMS_MANY_ATTEMPTS_PER_PHONE_NUMBER_COUNTRY
+      enabled: true
+    - type: SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY
+      enabled: true
+    - type: SMS_MANY_UNVERIFIED_OTPS_PER_IP
+      enabled: true
+    - type: SMS_UNMATCHED_PHONE_NUMBER_COUNTRIES_IP_GEO_LOCATION
+      enabled: true
+  decisions:
```
chpapa: Maybe we don't need it for now, as we do not know what would be good rules (unless we have done analysis of a few customers and see different patterns we want to cater?)

Or what's the use case of it?

chpapa: Instead, maybe:

1. A simple whitelist for country code (OR)
2. A simple whitelist for IP address block

And trigger of any enabled warning would block.

Would works?

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,373 @@
...
+Check if the number of distinct countries of requested phone numbers from a single IP exceeds the threshold in 24 hours.
+
+The threshold is 5.
+
+`enabled`: boolean. Whether this warning is enabled.
+
+#### SMS_MANY_FAILURES_PER_PHONE_NUMBER_COUNTRY
+Check if the number of SMS delivery failures for a specific phone number country exceeds the threshold in 24 hours.
+
+The threshold is 50.
```
chpapa: I think it's better to be a percentage instead of absolute number?

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,373 @@
...
+
+For Mid risk countries:
+
+```
+threshold = max(100, 7 day rolling mean of sms successfully sent to the country per day * 2)
+```
+
+`enabled`: boolean. Whether this warning is enabled.
+
+#### SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY
```
chpapa: I think this could be better as a percentage and a formula with some min absolute number as well

e.g. Verification Rate < 40% over last hours and sent > 3

This is a signal to block / challenge / delay SMS immediately.

docs/specs/fraud-protection.md
chpapa: I think this is better to be a percentage + min number. E.g. Delivery rate < 40% and sent > 10

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,373 @@
...
+`enabled`: boolean. Whether this warning is enabled.
+
+#### SMS_MANY_FAILURES_PER_PHONE_NUMBER_COUNTRY
+Check if the number of SMS delivery failures for a specific phone number country exceeds the threshold in 24 hours.
+
+The threshold is 50.
+
+`enabled`: boolean. Whether this warning is enabled.
+
+#### SMS_MANY_ATTEMPTS_PER_PHONE_NUMBER_COUNTRY
```
chpapa: Maybe better to...

1. use median, instead of average for daily baseline
2. use 14 days instead of 7 days (although week is good but sometimes it is not, think of X'mas / Black Friday holiday)
3. Use verified OTPs instead of sent

For 1,2,3:
daily_baseline = median(verified OTPs per day over 7 days)

4. Our blocking shouldn't be 24 hours, it is responding too slow. (actually consider this for other warnings as well.) We should add an hourly threadshold.
5. . Use a smaller max but with larger allowance from baseline, 

For 4 + 5:

Daily threshold: max(floor, 14d_median * multiplier)
Hourly threshold: max(ceil(floor/6), daily_threshold / 6)

Where:
High Risk: floor = 5, multiplier = 1.5
Mid Risk: floor = 10, multiplier = 2.5

6. Optional: even more sophisicated, we can calculate the floor based on the total volume of SMS verified per day

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,373 @@
...
+```
+threshold = max(30, 7 day rolling max of sms successfully verified to the country per day * 0.5)
+```
+
+`enabled`: boolean. Whether this warning is enabled.
+
+#### SMS_MANY_UNVERIFIED_OTPS_PER_IP
+Check if the number of unverified OTPs from a single IP exceeds the threshold.
+
+The threshold is 10.
```
chpapa: % + floor.

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,373 @@
...
+
+`enabled`: boolean. Whether this warning is enabled.
+
+#### SMS_UNMATCHED_PHONE_NUMBER_COUNTRIES_IP_GEO_LOCATION
+Check if the country of the requested phone number matches the geo-location of the IP address.
+
+`enabled`: boolean. Whether this warning is enabled.
```
chpapa: This could be a max threshold I guess.

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,373 @@
...
+    - NG # Nigeria
+    - OM # Oman
+    - PK # Pakistan
+    - PS # Palestinian Territory
+    - LK # Sri Lanka
+    - SY # Syria
+    - TJ # Tajikistan
+    - TN # Tunisia
+
+Low Risk:
```
chpapa: Maybe we can let users to whitelist country into low risk as they needed.

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,373 @@
...
+fraud_protection:
+  geo_location_risks:
+    high:
+      - EG
+      - UA
+    low:
+      - HK
+```
+ 
+### Environment Variables
```
chpapa: Per Project setting?

tung2744: Yes we have per project config in `fraud_protection.geo_location_risks`. I added the environment variable for a modifiable default. I think we probably want different defaults in different deployments. For example, HK in HK deployment as low risk.

docs/specs/fraud-protection.md
tung2744: There are two ways to do whitelisting in this config.

1. Use fraud_protection.geo_location_risks to add countries as low risk countries.
2. Add an item in `decisions` which always allow requests from a country. (The first item in the example config)

docs/specs/fraud-protection.md
tung2744: This warning is about whether the IP geo location matches the phone number country code.
I think this warning only have match / unmatch two states in a single record.

Or maybe you are suggesting another warning like counting the unmatches for a single IP?

docs/specs/fraud-protection.md
tung2744: This is the number of unverified OTPs of a single IP. Therefore I designed it as a low constant threshold.
However maybe percentage is better for cases like shared IP address of a shopping mall / office.

docs/specs/fraud-protection.md
tung2744: 1. What is the standard.

The default list is the countries marked as high risk in twilio.

2. Or is it better to just use the phone number country code

I uses alpha2 because phone number country code are sometimes shared between different regions, such as we cannot distinguish between US and other Caribbean Countries, and US we want to treat it as low risk and other Caribbean Countries should be high or mid risk.

chpapa: > we cannot distinguish between US and other Caribbean Countries, and US we want to treat it as low risk and other Caribbean Countries should be high or mid risk.

But if we cannot distinguish them in phone number, does it matters if we can distinglish them here?

docs/specs/fraud-protection.md
chpapa: Yea I was exactly thinking about the case of shared IP address / gateway.

docs/specs/fraud-protection.md
chpapa: Per my comments above, I'm thinking of do it without `decisions` and make it a simple block if any warning is triggered.

docs/specs/fraud-protection.md
tung2744: > 1. Using median.

I have considered using median, but I gave up because I am worried about it cannot handles patterns like mon-fri very low usage. sat - sun high usage.

> 3. Use verified OTPs instead of sent

There is another warning (SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY) based on the verified otps. This warning is designed to be a higher limit based on sms sent just in case if tracking verified otps does not fit the project.


Agree with other points, let me update the spec.

docs/specs/fraud-protection.md
tung2744: In the current design, I assume we allow 20% of unverified sms in high risk countries. 50% in mid risk countries. And 100% in low risk countries (i.e. everyone resend twice).

Like the comment above, I think I can add hourly limit.

I started with higher thresholds because I think we can lower them later. If we start with low thresholds I'm afraid we might unexpectedly block normal users in existing projects. But the actual number I think we can discuss.

docs/specs/fraud-protection.md
tung2744: Yes it matters. Caribbean islands (e.g. +1 876) are high risk (Or mid) in SMS pumping, while US (e.g. +1 212) is low risk. We can actually know exactly which country the phone number belongs to after parsing the number, even they share the same country code prefix.

docs/specs/fraud-protection.md
tung2744: The `decisions` section here is more for a flexible design for our future work, like block by risk score, or decide by custom hooks.
And I agree that we usually just want a whitelist + one block item. (In the example I have two block items, because I want to show we can have block_mode=silent & block_mode=error)

docs/specs/fraud-protection.md
chpapa: ok.

docs/specs/fraud-protection.md
chpapa: But I feel like it might be too flexible yet we don't know enough about the use cases (or I don't...?) so that's why I'm asking if we have a list of use cases and believe the current design is good, or if not maybe we just go with something simple and easy to migrate in future. (E.g. if we have a simple whitelist + one block items, we can easily migrate that to a more flexible schema programatically)

docs/specs/fraud-protection.md
chpapa: > I have considered using median, but I gave up because I am worried about it cannot handles patterns like mon-fri very low usage. sat - sun high usage.

Um, make sense, but it seems both mean and median would have it's problem, or a more complex formula...?

> There is another warning (SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY) based on the verified otps. This warning is designed to be a higher limit based on sms sent just in case if tracking verified otps does not fit the project.

I think the other warning is irrelevant here. If this warning only use historical SMS ATTEMPTS, than an attacker can slowly increase it's failed attempts to increase it's rate limit (Especially we're open source which meant the algo is simple and open)

Instead, using verified OTPS to limits attempts (we can have higher multiplier to cater the normal failure rate) as base line number have the advantage that the ATTEMPTS are limited by legit SMS OTP numbers

docs/specs/fraud-protection.md
chpapa: ok

docs/specs/fraud-protection.md
chpapa: ok, resolving it. But might consider remove it to avoid complexity.

docs/specs/fraud-protection.md
tung2744: > but it seems both mean and median would have it's problem, or a more complex formula...?

Looks like mean actually fit the purpose better. If we are going to use verified sms counts to calculate the threshold, then we don't have to worry about mean being increased too much by an attack?

tung2744: > High Risk: floor = 5, multiplier = 1.5
Mid Risk: floor = 10, multiplier = 2.5

I am not sure will this be too strict? The floor here seems to be daily threshold not hourly.
Considering there must be a period the project has no data, so initially there can only be 10 sms to mid risk country per day?

chpapa: Yap, I guess better to lean towards the strict side for us?

Alternatively, we can calculate the floor based on the total SMS OTP sent of the tenant (or MAU)?

tung2744: > I guess better to lean towards the strict side for us?

I think we should define it looser, and stricten the rules when we are confident that they do not block normal users, maybe after we reviewed the actual data running on some projects.
But alternatively, we can also configure the decision such that it never blocks for now.

> Alternatively, we can calculate the floor based on the total SMS OTP sent of the tenant

I guess we better use a fixed number for floor, because this can also provide a value when the project do not have data initially.

docs/specs/fraud-protection.md
tung2744: I've updated the config to a simpler decision section. Please see if it looks better. I also added 3 example configs which I think we can apply to most of the projects.

docs/specs/fraud-protection.md
tung2744: Separated hourly and daily limits to `SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY_PER_DAY` and `SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY_PER_HOUR`.

docs/specs/fraud-protection.md
tung2744: Updated the equation to

```
threshold = max(0.6 * total number of sms delivered in the past 24 hours, 10)
```

Which is 0.6 fail rate with 10 minimum.

docs/specs/fraud-protection.md
tung2744: - I still kept using mean.
- Updated to use 14 days rolling average.
- Use verified otps to increase the threshold instead of sms sent count.
- Added hourly limits, which is 1/6 of the daily limits.

docs/specs/fraud-protection.md
tung2744: Updated to use percentage of unverified otps.

Daily:

```
threshold = max(20, 0.5 * verified OTPs in the past 24 hours)
```

Hourly:

```
threshold = max(5, 0.5 * verified OTPs in the past 24 hours / 6)
```

docs/specs/fraud-protection.md
tung2744: Maybe I don't get your comment here, but I have updated decisions section to a simpler structure as mentioned in the above comment.

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,539 @@
...
+      weight: 1
+      enabled: true
+    - type: SMS_MANY_UNVERIFIED_OTPS_PER_IP_PER_DAY
+      weight: 1
+      enabled: true
+    - type: SMS_MANY_UNVERIFIED_OTPS_PER_IP_PER_HOUR
+      weight: 1
+      enabled: true
+    - type: SMS_UNMATCHED_PHONE_NUMBER_COUNTRIES_IP_GEO_LOCATION
+      weight: 1
```
fungc-io: I think the default for UNMATCHED_PHONE_NUMBER_COUNTRIES_IP_GEO_LOCATION  should be weight:0 because there may be too many false positives (VPNs, travelers, expats). User can turn it on if there business requirement is strict

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,539 @@
...
+
+#### SMS_MANY_PHONE_NUMBER_COUNTRIES_PER_IP_PER_DAY
+Check if the number of distinct countries of requested phone numbers from a single IP exceeds the threshold in 24 hours.
+
+The threshold is 5.
+
+`enabled`: boolean. Whether this warning is enabled.
+
+#### SMS_MANY_FAILURES_PER_PHONE_NUMBER_COUNTRY_PER_DAY
+Check if the number of SMS delivery failures for a specific phone number country exceeds the threshold in 24 hours.
```
fungc-io: This will block the whole country? Concern: If a carrier in the country is down, this may block other users in the same country.

tung2744: 🤔 You are right. Maybe we need a way to distinguish gateway down from delivery failure before we have this warning.

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,539 @@
...
+  geo_location_risks:
+    high:
+      - EG
+      - UA
+    low:
+      - HK
+  warnings:
+    - type: SMS_MANY_PHONE_NUMBER_COUNTRIES_PER_IP_PER_DAY
+      weight: 1 # (Optional) Supported values: 0 or 1. If 0, the warning does not contribute to the risk_score.
+      enabled: true
```
carmenlau: We previously discussed this and agreed that, as a general rule of thumb, we should avoid using an enabled field. The configuration should only store whether something is effective.

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,539 @@
...
+    "recipient": "+12341234",
+    "type": "verification",
+  },
+  "triggered_warnings": [
+    "SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY_PER_DAY",
+    "SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY_PER_HOUR",
+    "SMS_MANY_UNVERIFIED_OTPS_PER_IP_PER_DAY",
+    "SMS_MANY_UNVERIFIED_OTPS_PER_IP_PER_HOUR",
+    "SMS_MANY_ATTEMPTS_PER_PHONE_NUMBER_COUNTRY_PER_DAY",
+    "SMS_MANY_ATTEMPTS_PER_PHONE_NUMBER_COUNTRY_PER_HOUR",
```
carmenlau: At the beginning of this feature, we should have a period of time to fine-tune the warning rules, so that we can evaluate the following:

1. Do we have enough data to fine-tune the rules?
Currently, we only know which warnings are triggered. However, the actual values may be close to or far from the threshold. We may also need access to some numbers, otherwise, it will be difficult to make proper adjustments.
1. Are the thresholds hardcoded or configurable?
If they are hardcoded, the implementation is simpler. If they are configurable, we can adjust them more quickly. We should decide which approach we prefer.

docs/specs/fraud-protection.md
carmenlau: I can think of one use case for having a list of rules. If the system’s users are primarily in one country and they want to enforce certain mandatory rules there, they could handle it this way:

```yaml
decisions:
- decision: block
  name: block if number of unverified otp is high
  block_mode: error
  block_with_warnings:
    - SMS_MANY_UNVERIFIED_OTPS_PER_IP_PER_HOUR
  block_with_thresholds:
    risk_score: 1
- decision: allow
  name: always allow major business location
  allow_when_matches:
    ip_address:
      geo_location_codes: ["US"]
    phone_number:
      geo_location_codes: ["US"]
- decision: block
  name: block if triggered 3 warnings
  block_mode: error
  block_with_warnings:
    - "*"
  block_with_thresholds:
    risk_score: 3
```

If this is too complicated at this stage, we can just keep the map-based version for now.

docs/specs/fraud-protection.md
```diff
@@ -0,0 +1,539 @@
...
+
+```yaml
+fraud_protection:
+  decisions:
+    block:
+      risk_score: 10
+      block_mode: error
+```
+
+### Decision Record
```
carmenlau, line 192: I would also like to know how the decision records are stored and what the retention period is.

docs/specs/fraud-protection.md
tung2744: In the latest version, I've removed `block_with_warnings` because I found it confusing when used with risk_score:
When I have multiple `block_with_warnings` rules, how the score is calculated? I think because one record should only have one score, the score is still calculated by all rules. Then the problem is what happens when using `block_with_warnings` together with `block_with_thresholds.risk_score`. So finally I removed this config and use `weight` to control it.

docs/specs/fraud-protection.md
tung2744: They are stored as audit logs (The payload of the audit log), so sharing the same retention period with audit logs.

docs/specs/fraud-protection.md
tung2744: For 1, I have not planned technically how to store the data yet. But I think we can add otel metrics as a way to record the data.

For 2, originally I designed it as configurable, however, I changed it to non configurable to simplify the config because I think it could be difficult for each project to fine tune the parameters, and I think what we are trying to make should be some generic rules that can be applied to most of the projects. So I removed the configs. In current design it is hard coded.

tung2744:
Updated to:
1. Removed enabled flag in warnings
2. Added a section for default config

tung2744:
1. We should only keep the following rules for now:
  - SMS_MANY_PHONE_NUMBER_COUNTRIES_PER_IP_PER_DAY
  - SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY_PER_DAY
  - SMS_MANY_UNVERIFIED_OTPS_PER_PHONE_NUMBER_COUNTRY_PER_HOUR
  - SMS_MANY_UNVERIFIED_OTPS_PER_IP_PER_DAY
  - SMS_MANY_UNVERIFIED_OTPS_PER_IP_PER_HOUR
2. We will not introduce weight or score at the moment. Only block / observe mode (diagnose / deny mode)
3. Formulas need to be refined for an effective 20% unverified otp budget, handling cases like initial launch of project, or sudden spikes.
4. Unverified otp count should exclude otps sent in a completed login / signup / forgot password flow.
5. Document cases we have studied, and the reason we decided to keep these rules.
6. Make the warning names easier to read.

tung2744:
- naming change to SMS__[]__EXCEED_THRESHOLD
- Forbid config modifications on server side, use feature config to control it. For now, projects cannot change it by default.
- Further lower the lower bound of thresholds

tung2744:
@chpapa @carmenlau @fungc-io Updated again according to our latest discussion, thanks!

docs/specs/fraud-protection.md
tung2744: The rule was removed, resolving.

docs/specs/fraud-protection.md
tung2744: The rule was removed, resolving.

docs/specs/fraud-protection.md
tung2744: The rule was removed, resolving.

docs/specs/fraud-protection.md
tung2744: Now, we only have the root `enabled` field for convention. Items inside the config should not have `enabled`.

docs/specs/fraud-protection.md
tung2744: 1. We will add otel metrics to record the count of trigger of each rule in all projects. We can later relax or stricten the rules according to the data.
2. The threshold will remain non-configurable, such that we have full control on whether to stricten or relax later.

docs/specs/fraud-protection.md
tung2744: Finally we decided to use 20% unverified rate for all countries.

carmenlau (approved):

carmenlau:
@chpapa I think all the requested changes have been addressed. Could you approve the PR? Otherwise, I can dismiss your review and merge it. :)

chpapa:
> @chpapa I think all the requested changes have been addressed. Could you approve the PR? Otherwise, I can dismiss your review and merge it. :)

@carmenlau oh ya sorry I guess if you two have reviewed it is ok.

PR #5541, opened 2026-02-23
Support phone input allowlist in features config

carmenlau:
ref DEV-3298

### DIFF
--- a/pkg/lib/config/configsource/loader.go
+++ b/pkg/lib/config/configsource/loader.go
@@ -43,22 +43,8 @@ func LoadConfig(ctx context.Context, res *resource.Manager) (*config.Config, err
 		FeatureConfig: featureConfig,
 	}
 
-	// Filter phone countries based on feature config
-	if cfg.AppConfig.UI != nil && cfg.AppConfig.UI.PhoneInput != nil &&
-		cfg.FeatureConfig.UI != nil && cfg.FeatureConfig.UI.PhoneInput != nil {
-		if cfg.AppConfig.UI.PhoneInput.AllowList != nil {
-			cfg.AppConfig.UI.PhoneInput.AllowList = config.IntersectAllowlist(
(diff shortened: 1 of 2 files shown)


tung2744 (approved):

pkg/lib/config/configsource/loader.go
```diff
@@ -42,6 +42,24 @@ func LoadConfig(ctx context.Context, res *resource.Manager) (*config.Config, err
...
+				cfg.FeatureConfig.UI.PhoneInput.AllowList,
+			)
+		}
+		if cfg.AppConfig.UI.PhoneInput.PinnedList != nil {
+			cfg.AppConfig.UI.PhoneInput.PinnedList = config.IntersectAllowlist(
+				cfg.AppConfig.UI.PhoneInput.PinnedList,
+				cfg.FeatureConfig.UI.PhoneInput.AllowList,
+			)
+		}
+	}
```
tung2744: Instead of doing it in loader, maybe we can create another method similar to PopulateDefaultValues to do this? I think we should also ensure all places called PopulateDefaultValues calls that method.

carmenlau: I agree that moving the logic into a separate method makes the code clearer. I’ve added a new function, `ApplyFeatureConfigConstraints`.

However, I’m still calling this function in the loader rather than in the places where we call `PopulateDefaultValues`. Based on my review of how `PopulateDefaultValues` is used, not all call sites are suitable for or require `ApplyFeatureConfigConstraints`.

For example, `config.Parse` is a low-level function that constructs the config from YAML and should not depend on feature config. `AuthgearYAMLDescriptor` should not need to read feature config.

See if it looks good for you?

tung2744: Yes, rereading the code, they are tests and empty configs. So calling only in LoadConfig make sense. Thanks!

PR #5535, opened 2026-02-13
Update golang and fix the build

tung2744:
Also fixed a test build error

PR #5543, opened 2026-02-25
Cleanup most of the v1 code

tung2744:
ref DEV-3215
ref DEV-3216

requires:
- https://github.com/authgear/authgear-server/pull/5535

### DIFF
--- a/.vettedpositions
+++ b/.vettedpositions
@@ -91,7 +91,7 @@
 /pkg/auth/handler/webapp/authflowv2/change_password.go:115:26: requestcontext
 /pkg/auth/handler/webapp/authflowv2/change_password_success.go:65:35: requestcontext
 /pkg/auth/handler/webapp/authflowv2/create_password.go:179:26: requestcontext
-/pkg/auth/handler/webapp/authflowv2/enter_oob_otp.go:222:26: requestcontext
+/pkg/auth/handler/webapp/authflowv2/enter_oob_otp.go:226:26: requestcontext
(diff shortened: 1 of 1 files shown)


carmenlau (commented):
Could you share more background on this update? Are there any existing projects or clients currently using the interaction / auth v1 UI implementation? If so, what does the migration plan look like, or is a direct switch to v2 UI feasible for them?

tung2744:
@carmenlau We have already migrated all existing clients to v2, and set the default to v2. What remaining is just to remove the v1 code, which is what this PR for. Note that, interaction is not removed yet. I only removed v1 UI and the authflow v1 code (The intermediate UI between v1 and v2). Note that there a still a few v1 component remain because they seems to be still in use.
I created two issues for the remaining cleanup:
https://linear.app/authgear/issue/DEV-3406/cleanup-remaining-v1-code
https://linear.app/authgear/issue/DEV-3414/cleanup-interaction-engine

carmenlau:
Does this mean there are still projects using the interaction UI? This commit (https://github.com/authgear/authgear-server/pull/5543/changes/abc9bb2a38c31ff0c430bf3e5345c2115eadacf4) looks like it could affect those projects. I think having more context on how the UI has evolved would really help here, would we have a quick call when you have time?

tung2744:
There should not be projects using interaction UI. Even if there is, we will drop the support.
In this PR, all code path should treat ui implementation to be always v2. So in case there are such projects, they will just see the v2 ui.

PR #5618, opened 2026-04-02
[Portal] Add Fraud Protection Screen

tung2744:
ref DEV-3482
ref DEV-3505
ref DEV-3506

Screenshot 2026-04-02 at 8 12 49 PM
Screenshot 2026-04-02 at 8 12 03 PM
Screenshot 2026-04-02 at 8 11 55 PM
Screenshot 2026-04-02 at 8 11 52 PM

How to test:

1. Modify your `authgear.features.yaml`:
```yaml
fraud_protection:
  is_modifiable: true
```

2. Trigger sms and resend a lot of times.
3. See some records in the portal


tung2744:
Rebased and updated. New screenshots:

Screenshot 2026-04-14 at 3 57 30 PM

Screenshot 2026-04-14 at 3 57 33 PM

pkg/lib/audit/fraud_protection_overview.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (s *ReadStore) fraudProtectionDecisionRecordsQuery() db.SelectBuilder {
+	return s.SQLBuilder.
+		Select(
+			"created_at",
+			"user_id",
+			"activity_type",
+			"data",
+			"COALESCE(data#>>'{payload,record,ip_address}', '') AS ip_address",
```
carmenlau: Is this `ip_address` the same as the `ip_address` column on the `_audit_log` table? If so, we could use that column directly instead of extracting from JSONB, since we're grouping, ordering, and counting it for the top source IPs overview, the performance difference could be noticeable at scale.

cmd/authgear/cmd/cmdaudit/migrations/audit/20260402190000-add_fraud_protection_decision_indexes.sql
```diff
@@ -0,0 +1,51 @@
...
+  )
+  WHERE activity_type = 'fraud_protection.decision_recorded';
+
+CREATE INDEX _audit_log_idx_fraud_decision_app_id_ip_created_at
+  ON _audit_log (
+    app_id,
+    (data #>> '{payload,record,ip_address}') text_pattern_ops,
+    created_at DESC
+  )
+  WHERE activity_type = 'fraud_protection.decision_recorded';
```
carmenlau, line 48: If we switch to the ip_address column, we should update this index accordingly. A plain column index would also be simpler and more reusable, for example, filtering audit logs by IP across other activity types in the future.

pkg/lib/audit/fraud_protection_overview.go
```diff
@@ -93,19 +97,21 @@ func (s *ReadStore) GetFraudProtectionOverview(ctx context.Context, opts QueryPa
...
-		TotalActions:   int(totalActions),
-		BlockedActions: int(blockedActions),
-		WarnedActions:  int(warnedActions),
-		TopSourceIPs:   topSourceIPs,
+		SendSMS: FraudProtectionOverviewSendSMS{
+			TotalActions:   int(totalActions),
+			BlockedActions: int(blockedActions),
+			WarnedActions:  int(warnedActions),
+			TopSourceIPs:   topSourceIPs,
+		},
```
carmenlau, line 145: Since this count is specific to `send_sms` now, it might be worth adding an explicit filter to the query even though it's currently the only action. This keeps the intent clear and avoids confusion if new activity types are added to fraud protection in the future.

.claude/skills/add-portal-screen/SKILL.md
```diff
@@ -0,0 +1,385 @@
+---
+name: add-portal-screen
```
carmenlau, line 2: That's great to have the skill for adding new screen to the portal!

portal/src/screens/fraud-protection/FraudProtectionConfigurationScreen.tsx
```diff
@@ -0,0 +1,387 @@
...
+    }
+    return item;
+  });
+}
+
+function escapeRegExp(input: string): string {
+  return input.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
+
+function toPhoneRegex(raw: string): string[] {
```
carmenlau, line 90: Currently when users input something invalid in the phone number allowlist field, it will be stored straight away to the config. Maybe we should add some validation for this.

portal/src/graphql/portal/FraudProtectionConfigurationScreen.tsx
```diff
@@ -0,0 +1,387 @@
...
+                        description={renderToString(
+                          "FraudProtectionConfigurationScreen.allowlist.ip.description"
+                        )}
+                        placeholder="127.0.0.1/32"
+                        multiline={true}
+                        resizable={false}
+                        disabled={!isModifiable}
+                        value={state.ipAllowlist}
+                        onChange={onIPAllowlistChange}
+                      />
```
carmenlau: When users input an invalid value to the IP allowlist, the error message should be shown under the text field. Missing `errorRules`?

pkg/lib/audit/fraud_protection_overview.go
tung2744: Yes.
About the performance, I expect this query only run by portal admins in fraud protection page. So I expect the frequency of execution is very low.
If we are worried about the performance, maybe we should add cache. What do you think?

carmenlau: I agree that the execution frequency should be low, so we can keep it simple without caching for now. We could start by switching to the `ip_address` column first.

portal/src/screens/fraud-protection/FraudProtectionConfigurationScreen.tsx
tung2744: I want to keep the flexibility to input regex here to support e.g. prefix match.
So I guess we can validate:
- Is a possible phone number
- OR, a valid regex

tung2744:
Updated, thanks!

carmenlau (approved):

PR #5640, opened 2026-04-09
[Site Admin] Collaborators API

carmenlau:
ref DEV-3488

after #5617


docs/plans/siteadmin-api/04-collaborators-api.md
```diff
@@ -0,0 +1,831 @@
...
+    ├── CollaboratorServiceAdminAPI → *portalservice.AdminAPIService    (email resolution)
+    └── CollaboratorServiceHTTPClient                                   (HTTP for GraphQL)
+```
+
+### `ListCollaborators` flow
+
+```
+1. portalservice.CollaboratorService.ListCollaborators(ctx, appID) → []*model.Collaborator
+2. Collect all userIDs from result
+3. After the DB work is finished, Admin API batch call: getUserNodes(userIDs) → map[userID]email
```
tung2744, line 71: I saw the current portal api is using a Loader pattern to load users from admin api.
Is it possible to reuse the same pattern?

pkg/siteadmin/transport/handler_collaborator_remove.go
```diff
@@ -37,14 +36,12 @@ func parseCollaboratorRemoveParams(r *http.Request) CollaboratorRemoveParams {
...
-		}
+	if err := h.Service.RemoveCollaborator(r.Context(), params.AppID, params.CollaboratorID); err != nil {
+		writeError(w, r, err)
+		return
 	}
 
-	writeError(w, r, apierrors.NewNotFound("collaborator not found"))
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	_ = json.NewEncoder(w).Encode(struct{}{})
```
tung2744: I suggest we add a struct `SiteAdminAPISuccessResponse` which automatically do the following:
- Set Content-Type
- Set the http status
- Encode the response to json

And all site admin handler should only use this struct to return the response in the handler.

docs/plans/siteadmin-api/04-collaborators-api.md
carmenlau: The DataLoader is designed for GraphQL resolvers where IDs arrive asynchronously across multiple resolvers. Since REST handlers have all IDs available upfront, I think a direct batch call is simpler and achieves the same result.

pkg/siteadmin/transport/handler_collaborator_remove.go
carmenlau: Updated, thank you!

tung2744 (approved):

PR #5647, opened 2026-04-14
[Site Admin] Usage API 

carmenlau:
ref DEV-3489


pkg/siteadmin/transport/handler_messaging_usage.go
```diff
@@ -35,6 +48,19 @@ func parseMessagingUsageParams(r *http.Request) (MessagingUsageParams, error) {
 		return MessagingUsageParams{}, err
 	}
 
+	start, _ := time.Parse("2006-01-02", startDate)
+	end, _ := time.Parse("2006-01-02", endDate)
```
tung2744, line 52: Should we handle format errors?

tung2744: oh it was handled by getDateParam already

tung2744 (approved):

PR #5654, opened 2026-04-15
[Portal] Show Endpoint field for all application types including OIDC/SAML

fungc-io:
## Summary

- The **Endpoint** field was hidden for `confidential`, `third_party_app`, and `m2m` application types. OIDC and SAML clients both use the `confidential` type, so they were affected.
- Removed the `showEndpoint` conditional so the Endpoint field is always rendered, regardless of application type.
- No new components, translation keys, or GraphQL changes needed — everything was already wired up.

ref DEV-2381

image

## Test plan

- [ ] Open a **confidential** (OIDC/SAML) application in the portal — Endpoint field should now appear with the `publicOrigin` value and a working copy button
- [ ] Confirm `spa`, `traditional_webapp`, and `native` types still show the Endpoint field (no regression)
- [ ] `npm run typecheck` passes in `portal/`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/portal/src/graphql/portal/EditOAuthClientForm.tsx
+++ b/portal/src/graphql/portal/EditOAuthClientForm.tsx
@@ -471,15 +471,6 @@ const EditOAuthClientForm: React.VFC<EditOAuthClientFormProps> =
       [clientConfig.x_application_type]
     );
 
-    const showEndpoint = useMemo(
-      () =>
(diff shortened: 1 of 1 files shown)


fungc-io:
Reopening from personal fork instead.

PR #5660, opened 2026-04-15
Improve CONTRIBUTING.md local setup instructions

carmenlau:


fungc-io (approved):

CONTRIBUTING.md
```diff
@@ -48,7 +52,45 @@ This project supports asdf, and there is a .tool-versions file at the project ro
...
    asdf install
    ```
-3. Install pkg-config
+
+   Ensure the asdf shims are in your `PATH`, otherwise `go` may still resolve to the system version after installation.
+   Preferably add the following to your shell startup script.
+
+   ```sh
+   export PATH="${ASDF_DATA_DIR:-$HOME/.asdf}/shims:$PATH"
+   ```
```
tung2744, line 61: I believe this was done in asdf installation:
https://asdf-vm.com/guide/getting-started.html

Maybe we can just reference the doc here?

CONTRIBUTING.md
```diff
@@ -191,6 +233,13 @@ use flake
 
    You can either do this by editing `/etc/hosts` or using `dnsmasq`.
 
+   Example `/etc/hosts` entries:
+
+   ```text
+   127.0.0.1 portal.localhost
+   127.0.0.1 accounts.portal.localhost
+   ```
```
tung2744, line 241: I actually think we do not need these complicated hosts now because portal is not using cookie now. But we can change that later

CONTRIBUTING.md
```diff
@@ -210,6 +259,14 @@ use flake
...
    ```
 
+3. Create the initial portal app config source in the database
+
+   Run this after `go run ./cmd/authgear init ...` has generated `./var`.
+
+   ```sh
+   go run ./cmd/portal internal configsource create ./var
+   ```
+
```
tung2744, line 269: I am not sure why this is needed. In local setup isn't we are using local config source so database should not require such row? (CONFIG_SOURCE_TYPE=local_fs)

tung2744: I see, the problem was actually `go run ./cmd/authgear internal admin-api invoke` assumed database configsoucre. I think this is a bug in the command, so we should fix that instead of created unused rows in setup guide, what do you think?

fungc-io: Are there reasons why in local dev the config from yaml file instead of from the database is used?

tung2744: Modifying config in database is trouble so we prefer a plain text yaml which you can edit easily during development.

carmenlau: I created the issue DEV-3541 for the bug fix. Before we fix it, should we keep the change here first to avoid a broken flow?

CONTRIBUTING.md
carmenlau: The installation guide was mentioned in step 1, I adjusted the content here to remove the redundant PATH note and clarify Go verification output. See if it looks better.

CONTRIBUTING.md
tung2744: ok. Lets update the instruction after the bug fix.

tung2744 (approved):

PR #5661, opened 2026-04-16
Fix OTP form double-submission on fast typing

tung2744:
ref DEV-3486


carmenlau:
@tung2744 Could you share how you reproduce it? Are you pasting the OTP? That should be the quickest way.

tung2744:
@carmenlau I've tested the followings:
- Paste the code two times
- Use fixed otp, type 00000000 very fast

Note that I can only reproduce in safari

carmenlau (approved):

PR #5662, opened 2026-04-16
Improve robustness of auth ui finish handlers

tung2744:
ref DEV-3472
ref DEV-3487

- Display correct error message if the session is invalid
- Allow finish endpoint to be visited multiple times
  - However, in /consent, the authentication info will be consumed (See AuthorizationHandler.doHandleConsent), so you still cannot go to the app redirect uri again by visiting /finish. I am not sure if we should keep the auth info such that this become possible. Maybe we can discuss on this.


carmenlau (commented):
Discussion summary:
- The finish path should be safe to replay.
- We should update the method name / add comments to make `HandleWithoutScreenAllowCompleted` more understandable.

tung2744:
@carmenlau Added a commit, see if it is easier to understand?

carmenlau:
It looks better, thank you!

carmenlau (approved):

PR #5663, opened 2026-04-16
[Portal] Add Authentication Mode selector for Traditional Web Application

fungc-io:
## Summary

- Add \`x_traditional_webapp_session_type\` field (\`cookie\` | \`access_token\`) to the OAuth client config schema and Go struct
- Add an **Authentication Mode** radio group on the Traditional Web Application detail page (between Basic Info and URIs); Server-side SDK is shown first and is the default for new apps; Cookie Session is preserved as the default for existing apps that have no value stored
- Add a **Select Authentication Mode** step in the \`/configuration/apps/add\` creation wizard, inserted between type selection and save for \`traditional_webapp\` clients only
- Update the quickstart guide (both the sidebar widget and the post-creation screen) to always show **Next.js** and **Other Framework** links for Traditional Web Application clients
- Cookie Session mode shows the Cookie-based authentication section; Server-side SDK mode shows the Refresh Token and Access Token fields
- Inline note shown when switching from Server-side SDK → Cookie Session within an unsaved editing session

ref DEV-3469

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


tung2744:
Some checks are failing

fungc-io:
@tung2744 , let's discuss on this change before merging it

tung2744:
@fungc-io As per latest discussion I think this PR will be closed?

fungc-io:
Yes, I will redesign and re submit

PR #5679, opened 2026-04-23
Update usage alert emails

tung2744:
ref DEV-3503

Screenshot 2026-04-23 at 7 38 14 PM
Screenshot 2026-04-23 at 7 37 34 PM


carmenlau (approved):
In the email content, we don’t specify whether the limit reached is daily or monthly. Users might infer it from the number, but I think it would be more intuitive to include it explicitly in the email.
I’ll merge this PR first, and we can open another PR if we want to update it.

PR #5682, opened 2026-04-24
fraud protection: Only count unverified otp in countries per ip bucket

tung2744:
ref DEV-3540


pkg/lib/fraudprotection/leaky_bucket_store.go
```diff
@@ -225,10 +243,29 @@ func (s *LeakyBucketStore) RecordSMSOTPVerified(ctx context.Context, ip, phoneCo
 	})
 }
 
+func (s *LeakyBucketStore) RecordSMSOTPVerifiedCountry(ctx context.Context, ip, phoneCountry string) error {
+	now := float64(s.Clock.NowUTC().Unix())
+
+	return s.Redis.WithConnContext(ctx, func(ctx context.Context, conn redis.Redis_6_0_Cmdable) error {
+		return conn.Eval(ctx, `
+redis.call('ZADD', KEYS[1], ARGV[1], ARGV[2])
+redis.call('EXPIRE', KEYS[1], ARGV[3])
```
carmenlau, line 257: When I first read this, I wondered why there is no `ZREMRANGEBYSCORE` cleanup when adding a key to the verified-country set. It turns out
  the cleanup is intentionally deferred to read time in `ipCountriesScript`, which is also more accurate in timing since now reflects when
  the data is actually evaluated.

I think it worths to add a comment here so that if we want to use the list somewhere else, we won't accidentally miss the cleanup.

carmenlau (approved):

pkg/lib/fraudprotection/leaky_bucket_store.go
tung2744: I actually did not think too much about the cleanup when I modify this part.
`ZREMRANGEBYSCORE` actually must be called in `ipCountriesScript` because what we want is verified countries in the past 24 hour relative to the time when `ipCountriesScript` is called (At the time the sms is sent), but not relative to the time of the sms being verified.
Maybe I add some comments in `ipCountriesScript` to mention this.

carmenlau (approved):

PR #5684, opened 2026-04-24
fraud protection: Add metric to track drained bucket count

tung2744:
ref DEV-3540

requires
- https://github.com/authgear/authgear-server/pull/5682


pkg/lib/fraudprotection/metrics_store.go
```diff
@@ -38,18 +39,35 @@ type MetricsStore struct {
...
 		Insert(tableName).
-		Columns("id", "name", "key", "created_at").
-		Values(id1, metricsNameSMSOTPVerified, ipKey, now).
-		Values(id2, metricsNameSMSOTPVerified, countryKey, now)
+		Columns("id", "name", "key", "created_at")
+
+	for i := 0; i < count; i++ {
+		builder = builder.
+			Values(uuid.New(), metricName, ipKey, now).
+			Values(uuid.New(), metricName, countryKey, now)
```
carmenlau, line 69: It looks like we could add a `count` column to the table?

tung2744: I would like to keep the design simple so that 1 count = 1 row, so that the SQLs can simply count the rows and do not need to use SUM. So I prefer not to add this column until there is use cases we need a non-trivial number of counts.

carmenlau: No problem!

carmenlau (approved):

PR #5685, opened 2026-04-24
[Portal] Add Storybook setup and component stories

fungc-io:
## Summary
- Add Storybook v9 setup under `portal/.storybook/` with component stories for buttons, inputs, dialogs, dropdowns, toggles, etc.
- Add `portal/docs/components-guidelines/` docs bundle and Portal ARCHITECTURE.md / FRONTEND.md.
- Restructure AGENTS.md documentation map to reference `portal/docs`.

## Test plan
- [x] `cd portal && npm install && npm run storybook` — verify dev server runs at http://localhost:6006
- [x] Browse each story category (buttons, inputs, selection, feedback, overlay, navigation)
- [x] `npm run build-storybook` builds without errors


tung2744:
I believe docs in portal/docs is not for human so lets put it in .agents?

fungc-io:
I think it's easier to manage if the harness rules/references live in their respective area folders. So component references would go in `portal/docs/...`, with a map added to `AGENTS.md` / `CLAUDE.md` to point the agent to the right files.

If that creates confusion with other spec documents, how about putting them into `portal/references` instead?

That said, for other docs/specs, I think they'll be read by both humans and AI agents (they should follow the same spec after all). So I'd lean towards keeping those documents in `docs` and having the rules point to them.

tung2744:
I see your point, but my concern is this added many docs to the repo which will likely become outdated after half year. It will be ok if they are expected to be managed and read only by agent. But if they are for human, I think it create confusion.
Also, the docs doesn't seems to be easy for human to read. For example, I do not know when should I read docs inside ‎portal/docs/components-guidelines. So I think eventually only agents are reading them.

So I think:
Human -> Look at the storybook directly. Use common sense to determine which component to use.
Agent -> Read these docs.

tung2744:
@fungc-io 

I suggest the following changes to this PR:
1. Move the long docs including the ARCHITECTURE.md, FRONTEND.md, and component-guidelines into .agents.
2. Have a portal/README.md which only document the local development setup and useful npm commands, like npm start and npm gentype.
3. storybook.md can be keep but I might simplify it later.

tung2744:
Maybe I can just merge this and relocate the docs later, as they are not affecting the portal code.

tung2744 (approved):

PR #5688, opened 2026-04-27
fraud protection: Support configurable unverified otp budget

tung2744:
ref DEV-3540

requires:
- https://github.com/authgear/authgear-server/pull/5684


pkg/lib/config/fraud_protection.go
```diff
@@ -15,6 +18,69 @@ var _ = Schema.Add("FraudProtectionConfig", `
...
+var _ = Schema.Add("FraudProtectionSMSUnverifiedOTPBudgetByPhoneCountryConfig", `
+{
+	"type": "object",
+	"additionalProperties": false,
+	"required": ["country_codes"],
+	"properties": {
+		"country_codes": {
+			"type": "array",
+			"items": {
+				"type": "string",
```
carmenlau, line 65: Should it be `ISO31661Alpha2`?

pkg/lib/fraudprotection/service_test.go
```diff
@@ -283,33 +286,65 @@ func TestComputeThresholds(t *testing.T) {
...
+				},
+			}
+			svc := newSvc(600, 1000, 10, 600)
+			svc.Config = cfg
+
+			thresholds, err := svc.ComputeThresholds(ctx, "1.2.3.4", "SG")
+			So(err, ShouldBeNil)
+			So(thresholds.CountryDaily, ShouldEqual, 150)
+			So(thresholds.CountryHourly, ShouldEqual, 20)
+			So(thresholds.IPDaily, ShouldEqual, 180)
```
carmenlau, line 348: It would be great if we could add a comment showing the formula, similar to the test `scales with large historical counts`. 🙏🏻

pkg/lib/config/fraud_protection.go
tung2744: I think I will follow the validation of geo_location_codes, which only validate it is a string with length = 2

tung2744: oh actually the validation is correct already, we have minLength and maxLength

pkg/lib/fraudprotection/service_test.go
tung2744: Added a commit to show the formula

tung2744:
Updated, thanks!

pkg/lib/config/fraud_protection.go
carmenlau: I am still thinking about whether we should validate it against the ISO31661Alpha2 list, because:

1. The phone input allow list is bounded by the alpha2 list, so allowing users to input a country that is not in the list here should not have any effect.
1. Alpha2 codes are sometimes easily input incorrectly, e.g. people usually use "UK" for United Kingdom, but the actual code is "GB".

Due to the reason in https://github.com/authgear/authgear-server/pull/5461#discussion_r2598203052, maybe we should use ISO31661Alpha2 for phone country and string for geo IP country. We can discuss this.

The rest of the PR looks good to me. I am going to merge it first. We can open another PR for changes if needed.

carmenlau (approved):

pkg/lib/config/fraud_protection.go
tung2744: > The phone input allow list is bounded by the alpha2 list, so allowing users to input a country that is not in the list here should not have any effect.

The current alpha 2 list is from pkg/util/territoryutil/alpha2.go
Which might not be in sync with alpha2 supported by nyaruka/phonenumbers.

Reading the comments there they seems to be intentionally different.

But you are correct that actually if that country code is not in `ISO31661Alpha2`, the user cannot type that phone number in our input.

PR #5690, opened 2026-04-28
Reduce cardinality of high volume metric

tung2744:
High volume metrics such as http duration histogram and database operation metrics occupy high memory if cardinality is high, so remove project id and client id from those metrics.
ref DEV-3529


pkg/lib/deps/middleware_request.go
```diff
@@ -33,6 +34,7 @@ func (m *RequestMiddleware) Handle(next http.Handler) http.Handler {
 		err := m.ConfigSource.ProvideContext(ctx, r, func(ctx context.Context, appCtx *config.AppContext) error {
 			ctx, ap := m.RootProvider.NewAppProvider(ctx, appCtx)
 			ctx = WithAppProvider(ctx, ap)
+			ctx = otelutil.ContextWithClonedLabeler(ctx)
```
carmenlau, line 41: I think it would be better to add a comment explaining why a new labeler is needed.

tung2744: Added a commit for the comment, thanks!

carmenlau (approved):

PR #5694, opened 2026-04-29
[Site Admin] Add e2e tests for siteadmin apis

carmenlau:
ref DEV-3518


e2e/tests/siteadmin/apps.test.yaml
```diff
@@ -0,0 +1,353 @@
...
+    path: fixture/seed_siteadmin_apps.sql
+steps:
+# ── Basic list ────────────────────────────────────────────────────────────────
+- name: list all apps
+  action: http_request
+  http_request_method: GET
+  http_request_url: http://127.0.0.1:4003/api/v1/apps
+  http_request_headers:
+    X-Authgear-Session-Valid: "true"
+    X-Authgear-User-Id: "00000000-0000-0000-0000-000000000001"
```
tung2744, line 15: I thought we need to provide a access token here?

carmenlau: The test setup is simplified. It uses `WEB_SDK_SESSION_TYPE=cookie` and assumes the server is behind the resolver proxy, so an access token is not needed.

tung2744: I was actually thinking, did we exposed a security risk in our helm chart if `webSDKSessionType` is set to cookies? It seems in portal when `webSDKSessionType` is set to `cookies`, we automatically setup a nginx proxy and expose the nginx instead. However, in siteadmin, looks like we do not have such setup. Therefore, if the user set `webSDKSessionType` to `cookies`, the api caller can claim himself as anyone? Did I missed anything?

carmenlau: Yes, good catch. Users should never set `webSDKSessionType` to cookie, so maybe we should hardcode the env and remove `webSDKSessionType` from the helm chart for site admin.

tung2744: What if we remove the `SITEADMIN_AUTHGEAR_WEB_SDK_SESSION_TYPE` environment variable and always use access token instead?
I think practically we do not need cookies auth in site admin. This prevent anyone to set it up wrongly.

tung2744: Will handle later in https://linear.app/authgear/issue/DEV-3553/always-use-web-sdk-session-typerefresh-token-for-site-admin

tung2744 (approved):

PR #5696, opened 2026-04-30
[Site Admin] Isolate app plan e2e mutation

carmenlau:
ref DEV-3518

### DIFF
--- a/e2e/tests/siteadmin/plans.test.yaml
+++ b/e2e/tests/siteadmin/plans.test.yaml
@@ -39,10 +39,12 @@ steps:
       ]
 
 # ── Change App Plan ───────────────────────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


tung2744:
It seems this test passed in your original PR, do you know the reason?

carmenlau:
Yes, the tests share the same seeded app data, and they run in parallel.

`plans.test.yaml` changed `alpha`'s plan, but `apps.test.yaml` assumed `alpha` would stay on startups. Since the tests run in parallel, `apps.test.yaml` would fail if `plans.test.yaml` updated `alpha` first, and pass if it checked the data before that update happened. 

The original PR likely passed because the tests happened to run in a safe order that time.

tung2744 (approved):

PR #5709, opened 2026-05-07
specs: Allow admin to unlock user and query account lockout status

tung2744:
ref DEV-3508

### DIFF
--- a/docs/specs/account-lockout.md
+++ b/docs/specs/account-lockout.md
@@ -1,8 +1,29 @@
 # Account Lockout
 
-## Abstract
+## Table of Contents
 
-In order to increase resistance to brute forcing, we will lock the account for a period of time after a several times of authentication failure of a single account.
+- [Motivation](#motivation)
+- [Configuration](#configuration)
+- [Algorithm](#algorithm)
+  - [Failed attempts](#failed-attempts)
+  - [Lockout duration](#lockout-duration)
+- [Admin API](#admin-api)
+  - [User.accountLockout Query](#useraccountlockout-query)
+  - [unlockUser Mutation](#unlockuser-mutation)
+- [Usecases](#usecases)
+  - [1. Customer Support – Emergency Account Access](#1-customer-support--emergency-account-access-per_user-lockout)
+  - [2. Diagnosing Lockout](#2-diagnosing-lockout-per_user_per_ip-lockout)
+  - [3. Failed Brute Force Attempt Recovery](#3-failed-brute-force-attempt-recovery-both-lockout-types)
+  - [4. Brute Force Attack Investigation](#4-brute-force-attack-investigation-per_user_per_ip-lockout)
+  - [5. Account Recovery After Password Reset](#5-account-recovery-after-password-reset-both-lockout-types)
+  - [6. Testing & Development](#6-testing--development)
+- [References](#references)
+
+## Motivation
+
+Account lockout is a defensive mechanism that:
+- **Stops brute force attacks** by temporarily disabling login after repeated failed attempts
+- **Allows legitimate users to regain access** after waiting or with admin help
 
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


docs/specs/account-lockout.md
```diff
@@ -80,6 +101,121 @@ min(minimum_duration * backoff_factor^(failedAttemps - max_attempts), maximum_du
...
+- `lockedUntil`: DateTime in UTC when the global lock expires (null if not locked)
+- `lockedIPs`: Empty array (not applicable for this type)
+
+**For `per_user_per_ip` lockout type:**
+- `lockoutType`: "per_user_per_ip"
+- `isLocked`: true if user is locked from ANY IP, false if no IPs have locks
+- `lockedUntil`: null (not applicable; different IPs have different expiration times)
+- `lockedIPs`: Array of IP addresses currently locked (e.g., ["192.168.1.1", "203.0.113.45"]).
+
+### unlockUser Mutation
```
fungc-io: I lean toward `resetAccountLockout` for clarity. Since the feature is officially called "**Account Lockout**", using the same terminology in the API makes the intent unambiguous.

Authgear has several distinct mechanisms that can prevent a user from logging in: Account Lockout, Block User, and Account Valid Period... If we name this mutation "unlockUser", it could be misread as reversing any of these states, which risks confusing admins about what the action actually does.

`resetAccountLockout` ties directly to the specific feature it operates on, leaving less ambiguity.

docs/specs/account-lockout.md
```diff
@@ -80,6 +101,121 @@ min(minimum_duration * backoff_factor^(failedAttemps - max_attempts), maximum_du
...
+- Calls `unlockUser` to clear the old lockout state
+- User can now immediately log in with their new password (no waiting required)
+- Improves user experience by allowing immediate access after password reset
+
+### 6. Testing & Development
+
+**Scenario**: During development or testing, engineers need to simulate lockout scenarios and then reset account state without waiting for natural timeout.
+
+**Solution**: Test automation calls `unlockUser` to reset account state between test runs, enabling faster test cycles (works for both lockout types).
+
```
fungc-io, line 225: We also want to show this status in the portal's user details page. Is this covered in the spec?

docs/specs/account-lockout.md
```diff
@@ -80,6 +101,121 @@ min(minimum_duration * backoff_factor^(failedAttemps - max_attempts), maximum_du
...
+```graphql
+query {
+  node(id: "<USER_NODE_ID>") {
+    ... on User {
+      id
+      accountLockout {
+        lockoutType: "per_user" | "per_user_per_ip"
+        isLocked: Boolean!
+        lockedUntil: DateTime
+        lockedIPs: [String!]
```
carmenlau: Users may also be interested in lockedUntil for per_user_per_ip cases. Use case 4 (Brute Force Attack Investigation) benefits from knowing which IPs expire soon vs. just got locked.

Maybe something like this?

```
lockedIPs: [LockedIP!]!

type LockedIP {
  ipAddress: String!
  lockedUntil: DateTime!
}
```

Worth also specifying ordering (e.g. most recent lockedUntil first?).

docs/specs/account-lockout.md
tung2744: This spec do not cover portal UI update. However, it covers the required changes in Admin API.

docs/specs/account-lockout.md
tung2744: I agree that using the same terminology is better for user to find this mutation.
However, I think there is only one thing in authgear called lock, which is account lockout.
And "reset" seems to be not intuitive compared to "unlock", I agree it is still a understandable name tho.
So I think both names are acceptable.

tung2744:
Updated, thanks!

carmenlau (approved):

carmenlau:
The spec is included in 
- #5718

PR #5710, opened 2026-05-08
Specify pii masking & admin api scoped access token

tung2744:
ref DEV-3509


tung2744:
- We might need separated scope for pii search / read
- We might need additional design for custom attribute search

tung2744:
The priority of this feature is lowered. But I think this spec can be keep open here for review.

PR #5717, opened 2026-05-12
Support username in account recovery flow

carmenlau:
ref DEV-3577

### DIFF
--- a/pkg/lib/authenticationflow/declarative/intent_account_recovery_flow_step_select_destination.go
+++ b/pkg/lib/authenticationflow/declarative/intent_account_recovery_flow_step_select_destination.go
@@ -114,12 +114,88 @@ func (i *IntentAccountRecoveryFlowStepSelectDestination) ReactTo(ctx context.Con
 		if authflow.AsInput(input, &inputTakeAccountRecoveryDestinationOptionIndex) {
 			optionIdx := inputTakeAccountRecoveryDestinationOptionIndex.GetAccountRecoveryDestinationOptionIndex()
 			option := i.Options[optionIdx]
+			resolved, err := i.resolveUsernameTarget(ctx, deps, flows, option)
+			if err != nil {
+				return nil, err
+			}
 			return authflow.NewNodeSimple(&NodeUseAccountRecoveryDestination{
-				Destination: option,
+				Destination: resolved,
 			}), nil
 		}
 	}
 	return nil, authflow.ErrIncompatibleInput
+}
+
+// resolveUsernameTarget overrides TargetLoginID with the user's first identity
+// matching the picked option's Channel, but only for the
+// "username + enumerate_destinations=false + user found" sub-case.
+// All other cases return the option unchanged.
+func (i *IntentAccountRecoveryFlowStepSelectDestination) resolveUsernameTarget(
+	ctx context.Context,
+	deps *authflow.Dependencies,
+	flows authflow.Flows,
+	option *AccountRecoveryDestinationOptionInternal,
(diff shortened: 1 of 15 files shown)


pkg/lib/authenticationflow/declarative/intent_account_recovery_flow_step_select_destination.go
```diff
@@ -114,14 +114,100 @@ func (i *IntentAccountRecoveryFlowStepSelectDestination) ReactTo(ctx context.Con
...
+		return ""
+	}
+	for _, ui := range userIdens {
+		if ui.Type != model.IdentityTypeLoginID {
+			continue
+		}
+		if ui.LoginID.LoginIDType == wantType {
+			return ui.LoginID.LoginID
+		}
+	}
```
tung2744: If I understand it correctly, this only find the first login ID of the type but did not respect user's primary  email / phone number (standard attribute)?

pkg/lib/authenticationflow/declarative/intent_account_recovery_flow_step_select_destination.go
```diff
@@ -181,7 +270,22 @@ func deriveAccountRecoveryDestinationOptions(
...
+		// resolved to the user's actual email/phone at ReactTo time when the user picks one.
+		username := iden.IdentitySpec.LoginID.Value.TrimSpace()
+		for _, channel := range allowedChannels {
+			options = append(options, &AccountRecoveryDestinationOptionInternal{
+				AccountRecoveryDestinationOption: AccountRecoveryDestinationOption{
+					MaskedDisplayName: username,
+					Channel:           AccountRecoveryChannel(channel.Channel),
+					OTPForm:           AccountRecoveryOTPForm(channel.OTPForm),
+				},
+				TargetLoginID: username,
```
tung2744: I think resolving the TargetLoginID to the actual login id which receive the code here is better?
We already know which login id should be chosen when user choose which option here.
If no corresponding login id, `TargetLoginID` can be just nil.
In this way I think we do not need `accountRecoveryNoSendPrefix`.

carmenlau: Agree that resolving the actual login ID here is cleaner. However, I think `accountRecoveryNoSendPrefix` is still needed, not only to prevent message dispatch, but to preserve per-username rate limiting. The target login ID is the key for the rate-limit bucket: `generateDummyOTP(ctx, loginID, options)` at `forgotpassword/service.go` stores the OTP and charges the cooldown bucket keyed by loginID. If `TargetLoginID` is empty and we skip `SendCode` entirely, `generateDummyOTP` never runs and no rate limit is charged at all.

ref: https://github.com/authgear/authgear-server/blob/89a5cb393e2e7c93a5c2b535aef46799fa683670/pkg/lib/feature/forgotpassword/service.go#L120

pkg/lib/authenticationflow/declarative/intent_account_recovery_flow_step_select_destination.go
carmenlau: Yes, you are right. Let me update it.

pkg/lib/authenticationflow/declarative/intent_account_recovery_flow_step_select_destination.go
tung2744: I see. Thanks for explaining!

carmenlau:
Updated, thanks!

tung2744 (approved):

PR #5725, opened 2026-05-18
Setup staging auto deploy

tung2744:
ref DEV-3516

### DIFF
--- a/.github/workflows/oursky.yaml
+++ b/.github/workflows/oursky.yaml
@@ -97,3 +97,45 @@ jobs:
         ssh-agent -k
         echo "SSH_AUTH_SOCK=" >> "$GITHUB_ENV"
         echo "SSH_AGENT_PID=" >> "$GITHUB_ENV"
+
+  deploy-staging:
+    runs-on: [self-hosted, linux, x64, v1]
+    needs: [authgear-image-custom, portal-image-custom]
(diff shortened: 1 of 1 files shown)


.github/workflows/oursky.yaml
```diff
@@ -97,3 +97,45 @@ jobs:
...
+        AUTHGEAR_STAGING_REPO: ${{ secrets.AUTHGEAR_STAGING_REPO }}
+      run: |
+        IMAGE_TAG=$(make -C custombuild --no-print-directory --eval='print-%: ; @echo $($*)' print-GIT_HASH)
+        git clone "$AUTHGEAR_STAGING_REPO" staging-repo
+        cd staging-repo
+        git checkout master
+        git config user.email "github-actions[bot]@users.noreply.github.com"
+        git config user.name "github-actions[bot]"
+        ./scripts/update-authgear-image.sh "$IMAGE_TAG"
+        git push origin master
```
carmenlau, line 133: Would like to if the workflow tested? Since I have not sure where is the `AUTHGEAR_STAGING_REPO`, should it be `master` or `main`?

carmenlau: If it is tested and working, I think we could merge it

tung2744: I've tested it in another branch.

carmenlau (approved):

PR #5728, opened 2026-05-18
portal: Fix fraud protection cannot be turned off

tung2744:
ref DEV-3597

https://github.com/user-attachments/assets/5a25991c-5012-4e76-8db8-61f753ed89d4


carmenlau:
The controls on this page are not consistent with the other pages (auto-save vs save button at the bottom). Is this intentional?

I’ll merge this PR first so that the feature works, and we can revisit the consistency issue later if needed.

carmenlau (approved):

tung2744:
It was because it seems weird if we have a bottom footer in the overview / logs tab. So I decided to do it this way. Maybe see if @fungc-io have any comment after testing on staging.

carmenlau:
Sure! If we keep the auto-save behavior, I have one comment.
Right now, when we turn on fraud protection, switch to the Settings tab, and then turn it off, the save button bar still remains visible. Maybe we should hide it after turning it off.

PR #5734, opened 2026-05-21
portal: ProjectSelector in header

jimmycwc:
DES-234: Implement Project Selector in Navigation

截圖 2026-05-21 19 43 05

## Summary

- Add `ProjectSelector` dropdown in the portal screen header to show the current project, switch between accessible projects, copy the project ID, and create a new project.
- Unify mobile and desktop header app sections into `HeaderAppSection`; replace the previous static/copied app ID label and chevron.
- Extend `Logo` with optional `containerClassName` for tighter header spacing; add `projectQuota` helper and i18n strings for the selector.

## Test plan

- [x] Open a project at `http://portal.localhost:8000` and confirm the header shows the project selector trigger with the correct project ID.
- [x] Open the selector: verify Current Project, Your Projects list (checkmark on current), and Create Project (when not Authgear Once / quota not reached).
- [ ] Switch to another project and confirm navigation to getting-started; GTM `enteredProject` if applicable.
- [x] Copy project ID from the selector and confirm clipboard + tooltip feedback.
- [x] Click Create Project and confirm navigation to `/projects/create`.
- [x] Verify mobile header layout (hamburger + selector) and desktop layout (logo divider + selector).
- [x] Confirm Authgear Once mode hides Create Project and excludes the system app from the list.

Made with [Cursor](https://cursor.com)


jimmycwc:
@fungc-io,  I can’t set assignees on this repo. Please review this PR.

portal/src/components/header/ProjectSelector.tsx
```diff
@@ -278,10 +279,90 @@ const ProjectSelector: React.VFC<ProjectSelectorProps> = function ProjectSelecto
         { project_id: selectedAppID }
       );
       const typedID = toTypedID("App", selectedAppID);
-      navigate(`/project/${encodeURIComponent(typedID)}/getting-started`);
+      const newProjectBasePath = `/project/${encodeURIComponent(typedID)}`;
```
tung2744: This is too complicated and I do not think we will remember to update this part when adding new page. I also do not think it really covers all routes in our app. For example, audit logs paths seems to be missing.

We should have a more generic way to handle the redirect. What actually is the current logic? Why some pages stay in current page, some do not?

portal/src/components/header/ProjectSelector.tsx
```diff
@@ -0,0 +1,492 @@
...
+        }
+      };
+
+      document.addEventListener("pointerdown", onPointerDown, true);
+      document.addEventListener("keydown", onKeyDown);
+      return () => {
+        document.removeEventListener("pointerdown", onPointerDown, true);
+        document.removeEventListener("keydown", onKeyDown);
+      };
+    }, [isCalloutOpen, isInsideProjectSelector]);
```
tung2744: If I understand this code correctly, it is trying to close the callout when user click outside.
Please use `onDismiss` prop instead of implementing a fragile checker to check if user is clicking inside the callout.

The Keydown event listener is ok, but please check if `onDismiss` handled that as well.

portal/src/components/header/ProjectSelector.tsx
fungc-io: For example, if the admin is in the user details, `/users/[id]/details`, when they switch projects, the same id does not exist and will return error. And therefore it should go to `/users` for the user list instead.

fungc-io:
@tung2744 
New approach: a single list of section paths now lives in `util/projectPath.ts` and is shared by both ScreenNav and the switcher. resolveProjectSwitchPath keeps the longest matching section and drops any project-specific suffix (entity IDs, add/edit forms), falling back to getting-started:

- /user-management/users/<id>/details → /user-management/users
- /configuration/apps/<clientID>/edit → /configuration/apps
- /audit-log/<logID>/details → /audit-log

fungc-io:
### Update: feature-gated pages on project switch

Fixed a bug where switching to a project that lacks a feature-gated page (e.g. **Fraud Protection**) left a blank screen on the inaccessible URL. Root cause: `usePivotNavigation` re-added the URL hash and overrode the screen's redirect. App2App and Integrations had a related gap (hidden from nav but still reachable by URL/switch).

Added a route-level guard `RequireAppFeature` in `AppRoot` (wrapping fraud-protection, app2app, integrations) that redirects to **Getting Started** when the project lacks the feature, and renders the page via `<Outlet>` only once access is confirmed — so the redirect can't be overridden.

Verified: switch + direct-load on a project without the feature → Getting Started; project with the feature → renders normally. `typecheck` + `eslint` pass. (`4f1c1ee83b`)

portal/src/util/projectPath.ts
```diff
@@ -0,0 +1,107 @@
+// Single source of truth for the navigable "section" pages under a project,
```
tung2744: I still think having a function to handle this redirect is over complicated:
- If the logic is: "If the page you are viewing does not exist, return to a defined fallback page. " Isn't this can be handled in the screen itself? For example, I saw we already have related logic in UserDetailsScreen to redirect user to `/project/${appID}/user-management/users` when the user id does not exist. And I think all screens that could "disappear" after switching the project should use that pattern.

- The point is, there seems to be two patterns in the current code to handle the same problem: In `UserDetailsScreen`, we use useEffect to redirect user if the page does not exist. However, in `ProjectSelector`, we use this util to determine the "next path".

I think we should choose one from these two pattern. In my opinion, the pattern in `UserDetailsScreen` is better because we do not need to maintain another separated large mapping.

fungc-io: ok i see your point

fungc-io: Agreed — reworked in ce312f64a7 / 5a165eef91. The section mapping is gone: the project selector now just swaps the `/project/:id` prefix, and screens that load project-scoped entities redirect to their list page when the entity doesn't exist (the `UserDetailsScreen` pattern). For the user detail sub-pages this is done once with a route-level `RequireUser` guard (same shape as `RequireAppFeature`) instead of repeating the effect in every screen.

tung2744 (approved):

PR #5741, opened 2026-05-27
[OIDC] Specify https://authgear.com/claims/user/identities claim

carmenlau:
ref DEV-3606

### DIFF
--- a/docs/specs/sdk-settings-actions.md
+++ b/docs/specs/sdk-settings-actions.md
@@ -368,7 +368,7 @@ App developers might want to know which OAuth providers the user has linked to t
 const userInfo = await authgear.fetchUserInfo();
 const linkedOAuthProviders = userInfo.identities
   .filter((i) => i.type === "oauth")
-  .map((i) => i.providerAlias);
+  .map((i) => i.oauthProviderAlias);
 ```
 
 - Check whether a specific provider is linked
@@ -376,7 +376,7 @@ const linkedOAuthProviders = userInfo.identities
 ```typescript
 const userInfo = await authgear.fetchUserInfo();
 const isGoogleLinked = userInfo.identities.some(
-  (i) => i.type === "oauth" && i.providerAlias === "google"
+  (i) => i.type === "oauth" && i.oauthProviderAlias === "google"
 );
 ```
 
@@ -400,8 +400,8 @@ interface Identity {
   updatedAt: Date;
   loginIDKey?: string; // Present when type is "login_id", e.g. "email", "phone", "username"
   loginIDType?: "email" | "phone" | "username"; // Present when type is "login_id"
-  providerType?: string; // Present when type is "oauth", e.g. "google", "facebook"
-  providerAlias?: string; // Present when type is "oauth"
+  oauthProviderType?: string; // Present when type is "oauth", e.g. "google", "facebook"
+  oauthProviderAlias?: string; // Present when type is "oauth"
 }
 
 interface UserInfo {
@@ -456,8 +456,8 @@ interface UserInfo {
       "type": "oauth",
       "created_at": "2024-01-01T00:00:00Z",
       "updated_at": "2024-01-01T00:00:00Z",
-      "provider_type": "google",
-      "provider_alias": "google"
+      "oauth_provider_type": "google",
+      "oauth_provider_alias": "google"
     },
     {
       "type": "login_id",
--- a/docs/specs/sdk-settings-actions.md
+++ b/docs/specs/sdk-settings-actions.md
@@ -397,6 +397,7 @@ interface Identity {
   createdAt: Date;
   updatedAt: Date;
   loginIDKey?: string; // Present when type is "login_id", e.g. "email", "phone", "username"
+  loginIDType?: "email" | "phone" | "username"; // Present when type is "login_id"
   providerAlias?: string; // Present when type is "oauth"
 }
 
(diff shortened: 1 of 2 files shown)


docs/specs/sdk-settings-actions.md
```diff
@@ -374,11 +410,26 @@ interface UserInfo {
...
+      "created_at": "2024-01-01T00:00:00Z",
+      "updated_at": "2024-01-01T00:00:00Z",
+      "provider_alias": "google"
+    },
+    {
+      "type": "login_id",
+      "created_at": "2024-01-01T00:00:00Z",
+      "updated_at": "2024-01-01T00:00:00Z",
+      "login_id_key": "email"
+    }
```
carmenlau, line 468: I think it’s okay for different APIs to have different shapes as long as they are well documented and easy to understand, but we should keep the terminology consistent. (We could discuss)

Initially, I was wondering whether we should use the term `login_id`, since it feels more like an internal implementation detail to me. But after checking across different components, `login_id` has already been used in several places, so I think it’s fine to use it here as well.

For reference, how do the other APIs handle the `login_id` identity:

**Auth flow api**
https://docs.authgear.com/customization/custom-ui/authentication-flow-api

**User import api**
https://docs.authgear.com/reference/apis/user-import-api#initiate-import

**User search**
https://docs.authgear.com/reference/apis/admin-api/api-examples/search-for-users#search-users-by-email

docs/specs/user-profile/design.md
```diff
@@ -491,6 +491,20 @@ A boolean. True if the user can perform reauthentication.
...
+#### https://authgear.com/claims/user/identities
+
+A list of identities linked to the user. See [this doc](../sdk-settings-actions.md).
+
+Each element in the array has the following fields:
+
+- `type` (string): The identity type. One of `login_id`, `oauth`, `anonymous`, `biometric`, `passkey`, `siwe`, `ldap`.
+- `created_at` (string): RFC 3339 timestamp of when the identity was created.
+- `updated_at` (string): RFC 3339 timestamp of when the identity was last updated.
+- `login_id_key` (string, optional): The configured key name of the login ID (e.g. `"email"`, `"phone"`, `"username"`). Present only when `type` is `login_id`.
```
tung2744, line 503: I think we should follow the admin api model structure, i.e.
```jsonc
{
  "type": "login_id",
  "claims": {
    "https://authgear.com/claims/login_id/type": "email",
    "https://authgear.com/claims/login_id/key": "email"
  }
}
```

```jsonc
{
  "type": "oauth",
  "claims": {
    "https://authgear.com/claims/oauth/provider_alias": "google"
  }
}
```

I also think that login id type is more useful than login id key.

tung2744: But it seems `authenticators` did not follow the shape of admin api too.
Maybe ignore this comment

docs/specs/sdk-settings-actions.md
```diff
@@ -374,11 +410,26 @@ interface UserInfo {
...
+      "type": "oauth",
+      "created_at": "2024-01-01T00:00:00Z",
+      "updated_at": "2024-01-01T00:00:00Z",
+      "provider_alias": "google"
+    },
+    {
+      "type": "login_id",
+      "created_at": "2024-01-01T00:00:00Z",
+      "updated_at": "2024-01-01T00:00:00Z",
+      "login_id_key": "email"
```
tung2744: I think `login_id_type` is more useful than `login_id_key`, I suggest to expose both.

carmenlau: I think I’ve fixed them up. It seems that the `key` has the same value as `type`. Is there any case where the `key` would be different?

tung2744: `key` is a user defined value in `authgear.yaml`. See `identity.login_id.keys[].key`. And `type` is a enum, only `email`, `username`, `phone`.
Example:
```yaml
identity:
  login_id:
    keys:
    - create_disabled: false
      delete_disabled: false
      key: email <-- this is key, can be any value
      max_length: 320
      type: email
      update_disabled: false
    - create_disabled: false
      delete_disabled: false
      key: phone
      max_length: 40
      type: phone
      update_disabled: false
    - create_disabled: false
      delete_disabled: false
      key: username
      max_length: 40
      type: username
      update_disabled: false
```

docs/specs/user-profile/design.md
chpapa: Maybe we should review how many shapes we have now, and try to be consistent or pick one?

carmenlau: Yes, it is intended in userinfo. I added a design note here, see if it looks good

https://github.com/authgear/authgear-server/pull/5741/changes/BASE..7733c42f4c08f23593f03c391942560ab5503f9c#diff-299c7f857be8e61cccc8b09010cc5a0399272250bfeb8a4aeaff53292500b863R506

docs/specs/sdk-settings-actions.md
chpapa: It seems very different...

Can we map a data model behind that we can make it more consistent?

carmenlau: Let me draft something about it.

docs/specs/sdk-settings-actions.md
carmenlau: OK, they are the same in my case just because the config is created by the portal. Let me update it to expose both.

docs/specs/sdk-settings-actions.md
carmenlau: I've drafted a spec at #5758 as a guideline for future API consistency work.

One thing we didn't cover in the meeting, webhook events have the same issue as Admin GraphQL: identity data is exposed as a raw namespaced claims map rather than typed fields. We'll keep it as-is for now, same as Admin GraphQL.

tung2744 (approved):

tung2744:
I believe we have agreed on the latest design so I will merge this.

PR #5743, opened 2026-05-28
Add identities claim to userinfo endpoint

carmenlau:
ref DEV-3606

require:

- #5741


carmenlau:
Updated to rename provider_type/provider_alias to oauth_provider_type/oauth_provider_alias.

tung2744 (approved):

PR #5750, opened 2026-06-02
spec: Specify DCR, Audience Binding, Third Party App

tung2744:
ref DEV-3608


docs/specs/dcr.md
```diff
@@ -0,0 +1,533 @@
...
+  "userinfo_endpoint": "https://myapp.authgear.cloud/oauth2/userinfo",
+  "end_session_endpoint": "https://myapp.authgear.cloud/oauth2/logout",
+  "revocation_endpoint": "https://myapp.authgear.cloud/oauth2/revoke",
+  "jwks_uri": "https://myapp.authgear.cloud/oauth2/jwks",
+  "registration_endpoint": "https://myapp.authgear.cloud/oauth2/register", // Added
+  // ...
+}
+```
+
+## Initial Access Token
```
tung2744, line 266: Use scope to control first party app register

docs/specs/dcr.md
```diff
@@ -0,0 +1,533 @@
...
+
+```json
+{
+  "client_id": "dcrc_Xf2kLmNpQrStUvWx",
+  "client_id_issued_at": 1700000000,
+  "client_name": "PR #123 preview",
+  "redirect_uris": ["https://pr-123.preview.example.com/callback"],
+  "grant_types": ["authorization_code", "refresh_token"],
+  "response_types": ["code"],
+  "application_type": "spa",
```
tung2744: Should not be realted to our x_application_type

tung2744: Check if new defined application_type can be mapped to our internal x_application_type

docs/specs/dcr.md
```diff
@@ -0,0 +1,533 @@
...
+
+**Step 3 — Use the client in the authorization code flow**
+
+The PR preview app uses `client_id=dcrc_Xf2kLmNpQrStUvWx` as a normal SPA client for the lifetime of the PR.
+
+> Client deletion is not supported in this version. See [Future Works](#future-works).
+
+---
+
+### UC2. MCP (Model Context Protocol) clients
```
tung2744, line 132: Check usually what scope will be requested by mcp client

docs/specs/dcr.md
```diff
@@ -0,0 +1,533 @@
...
+- [Security Considerations](#security-considerations)
+  - [Access Token Audience Binding](#access-token-audience-binding)
+- [Admin API](#admin-api)
+- [Future Works](#future-works)
+
+## Glossary
+
+**Dynamic Client Registration (DCR)** — the process by which an OAuth client registers itself programmatically with an Authorization Server at runtime, rather than being statically configured in `authgear.yaml`.
+
+**Initial Access Token (IAT)** — a short-lived token presented to the registration endpoint that authorizes the caller to register a new client. Required when `initial_access_token_required: true` (the default). An IAT is a JWT signed with the Admin API auth key.
```
chpapa: in 2nd thought, I don't think IAT should be a JWT signed with the Admin API auth key. Instead, it could be a opaque (at least opaque to user) token, that's minted via the portal or Admin API.

Coz even we don't have the needs to support multiple resources binding for IAT, we still probably want different max number of clients and maybe in future some other setting like different rate limits etc.

docs/specs/dcr.md
```diff
@@ -0,0 +1,533 @@
...
+Per the [MCP Authorization specification](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization), each MCP client registers itself with the Authorization Server at first use. With open registration enabled, MCP clients self-register without any admin involvement per client.
+
+**Required configuration:**
+
+```yaml
+oauth:
+  dynamic_client_registration:
+    enabled: true
+    initial_access_token_required: false   # open registration — no IAT needed
+    allowed_resources:
```
chpapa: What we discussed yesterday actually reminds me, why we need a duplicated `allowed_resources` setting here, instead of reuse the existing resources? Maybe in our first version we can just have an `allow_dcr` bool columns, to white list the API resources allowed for DCR clients?

docs/specs/dcr.md
```diff
@@ -0,0 +1,533 @@
...
+
+```
+GET /oauth2/authorize
+  ?client_id=dcrc_AbCdEfGhIjKlMnOpQr
+  &response_type=code
+  &scope=openid+read:tools
+  &redirect_uri=https://mcp-client.example.com/callback
+  &code_challenge=<challenge>
+  &code_challenge_method=S256
+  &resource=https://mcp-server.example.com HTTP/1.1
```
chpapa, line 199: @tung2744 We discussed this yesterday in our use cases, maybe it make sense to:

1. if `resource` exists and mixes with OIDC profile scopes (openid, profile, email...), the request is rejected
2. if `resource` exists, the `aud` is the `resource` only, no project endpoint in aud (this is for UC2)
3. if `resource` not exists, than it is a "normal auth", aud = project end point (this is for UC1)

Does it make sense?

docs/specs/dcr.md
```diff
@@ -0,0 +1,533 @@
...
+
+The runtime behavior of a DCR client (authorization code flow, token endpoint, consent screen, etc.) is identical to that of a static client with the same `application_type` and metadata.
+
+DCR client secrets are stored hashed in the database.
+
+## Security Considerations
+
+### Access Token Audience Binding
+
+By default, all Authgear access tokens share `aud = [<project_endpoint>]`. A resource server that only validates `aud` cannot distinguish tokens intended for different services — this is the **audience confusion** risk.
```
chpapa: Given what you proposed here, I'm thinking maybe the right way to fix it is by asking users to validate jwt tokens for both:

1. Aud
2. Client ID

So a random token generated by DCR would have a different Client ID and won't got mixed up with others.

docs/specs/access-token-audience-binding.md
```diff
@@ -0,0 +1,266 @@
...
+Auth0's approach (opaque token by default) is the most principled but is a breaking change. Keycloak and Okta require explicit admin configuration with no safe default.
+
+### Authgear's decision: per-client default audience
+
+Authgear uses the client's own URI as the default audience, providing per-client token isolation without requiring any resource indicator configuration.
+
+**When no `resource` parameter is specified**, the issued JWT access token includes:
+
+```
+aud = ["<project_endpoint>", "<project_endpoint>/clients/<client_id>"]
```
chpapa: I wonder how is it different from asking users the best practices should be, validate both `aud` and `client_id`?

But maybe a major problem is it seems most open source library e.g. PyJWT don't check client id... ummm

chpapa (commented):
@tung2744 some after thought after yesterday discussion.

docs/specs/dcr.md
tung2744: Yes. This is actually another design I have considered. And this is similar to auth0's design. It have a "Default Permissions for Third-Party Applications"  config for api resources.
And the reason I finally put it in config is: This makes the config of dcr self explained. Do not depends on any config in other places and we know what resources are allowed for dcr clients.

However, when start considering CIMD, which is also registering third party clients dynamically. I found putting the config under dynamic_client_registration causes duplication (Because we also need that config for CIMD clients). So maybe I should just move it to existing API resources.

docs/specs/dcr.md
tung2744: I see. If it contains rate limits, maybe it shouldn't be decodable. Then, maybe I should change it to a JWE. There is no reason the user should read the claims of the token so I think making it opaque make sense.

docs/specs/access-token-audience-binding.md
tung2744: Actually, client_id is not expected to be validated (Although lots of people do that). The proper way is to validate `aud`.

And validating client_id has a problem in UC1: The client_id is dynamically assigned by authgear, how do your backend api server knows the expected value?
So in UC1, the only way to work is to validate `aud` with `resource=<YOUR_API>`.

docs/specs/dcr.md
tung2744: For 1 and 2, I think I will check what are the common practices.

For 3, I think the default `aud` need to be `"<project_endpoint>", "<project_endpoint>/clients/<client_id>"` suggested in the spec. This is to prevent developers to validate a jwt with `aud=<project_endpoint>` which poses a risk of audience confusion if they enabled open dcr later. So we ask them to validate against `<project_endpoint>/clients/<client_id>` from the beginning to eliminate that risk.

tung2744: I think for 1, it is very common to have `openid` scope together with `resource=<YOUR_API>`. Consider just a simple mobile app which have its own api server and integrate with authgear login:
- It needs id_token, so it need openid. (Probably also need `profile` or `email`)
- It needs to call its own api, so `resource=<YOUR_API>`.

So maybe it could work with mcp client which only need oauth, but it will break any app expect oidc integration.

For 2, I think it make sense, only if scope does not include any oidc scopes. And this seems to be the standard practice. In this case, Authgear act as a oauth authorization server which only grant access to `resource`. (And does not grant access to authgear, therefore cannot request userinfo, no id_token is issued)

docs/specs/dcr.md
chpapa: and we will need a db to store these kind of things I guess...

tung2744: If it is jwe, the content is still put in the token but encrypted. So it is opaque to user. And we do not need to store it in db. So we can still be stateless.

docs/specs/dcr.md
chpapa: BTW, for 2, I guess we already did that with M2M application types? Please also help double check if MCP clients works with that?

For 1, ummm, I agree half of it so maybe more research from you is needed:

If we allow openid scope with `resource`, and issue an token with multiple aud, that itself seems violate the spec?

At https://www.rfc-editor.org/info/rfc8707/#section-2, it said

> The authorization server SHOULD audience-restrict issued access tokens to the resource(s) indicated by the resource parameter.

Which basically should imply:

- `/authorize` could have multiple resources (and it is also common to have an consent screen that shows multiple requests), it returns authorization code that could request for multiple resources at `/token`
- `/token` should only return 1 resources for one access token.

Otherwise it falls into the same audience confusion problem you mentioned, Resources A can use token to replay the tokens at Resources B.

Now consider Authgear Project URL itself is **Resource 0**, I guess it make sense to disallow that? 

I'm discussing about these, to basically make open DCR safe instead of adding a new aud:

- By default, the **Resource 0** (need a better name) is one of the resources that have `allow_dcr` = false. Hence open DCR won't have the risk of audience confusion.
- If someone want to set Resource 0 `allow_dcr` to true, they should be warned to have IAT required true as well.
- And even for IAT enabled, we still DO NOT allow users to request a token that have both audience resource 0 and other resources

But I agree in your use case, the users will probably want both aud.... 

(I guess I don't have more time on this these two days.... See if it helps spark some more idea for you?)

docs/specs/access-token-audience-binding.md
chpapa: > The client_id is dynamically assigned by authgear, how do your backend api server knows the expected value?

Than in UC1, the server also can never validate if this token is valid or not, since it doesn't know the client_id?

docs/specs/dcr.md
chpapa: My point is this have to be stateful. Say if we want to store max number of clients, it is pointless to be stateless, we need a DB to store `[ <IAT>, <no of clients> ]` anyway

(plus in future we probably would be adding some more settings or things to track for each IAT, being stateless and have to work around that easily breaks backward compatibility)

Why we need it to be stateless?

docs/specs/access-token-audience-binding.md
tung2744: In oauth, the correct way to do it is to validate `aud=<API_SERVER>`. This is why we need to support `resource` parameter when we do DCR.

In current authgear, the validation validate `aud=<AUTHGEAR_ENDPOINT>`, which causes audience confusion if we allow open DCR.

Validate `client_id` works, but in UC1, client_id is dynamic so this cannot be done in UC1. So yes, without the `resource` parameter, API server in UC1 has no proper way to validate the token without audience confusion. This is the main reason why we need to support `resource` before support open DCR.

chpapa: Hence I asked why not we just make users care about this to validate client_id instead of adding another audience.

chpapa: Anyway, this discussion should mix with the broader discussion of how to fix the issue, let me resolve this thread and try to summarize and put up another better / more organized message.

docs/specs/dcr.md
tung2744: stateless is not a requirement. I do not against stateful token.
Stateless have some advantages, no need to do periodic database cleanup etc. But of course it can be stateful if there is a concern about backward compatibility. In such case, we will need an api to generate such token in admin api.

chpapa:
For this PR design, besides redundant allowed_resources and the stateless IAT, another major concern I see is the solution to fix the security issue of aud = <project_endpoint>

Thanks @tung2744 for bought up the discussion and proposed a solution. 

Yet my gut feeling of adding another `<project_endpoint>/clients/<client_id>` into the `aud` is probably not the right solution coz...

- `aud` as an array is allowed in spec, but less common in practices, that may create compatibile issue if it is the default and if developer do a simple string `aud == <aud>` check.
- it's literally equals to checking both `aud` and `client_id`, we could have opt for that
- RFC8707 specified while `/authorize`  can specify multiple resources, `/token` is not recommended to do it -- although semantically our two audience is the same so it is not a huge problem, yet still kinda werid.

It also won't 100% solve the problem since:
- Only static client application that update the code to check new aud would works
- UC1, the app won't know client_id hence can't validate it

Attack vector of our problems:
1. Open DCR: Audience-confusion, any one can register a new client application and get an access token with `aud` of `<project_endpoint>`
2. IAT DCR:
	1. If IAT is leaked, same problem with Open DCR
	2. Users do not know if they only consent to 1 Resources or `<project_endpoint>`

Some Mitigation Proposal:

(None of these are perfect, but let's list out what's there...)

Type A: Need Client-side code changes
- Validate both `aud` and `client_id` as best practices
- Add 2nd `aud` with `client_id` inside

Type B: Not 100% help:
- Validate both `aud` and `iss`, with `iss` between static client and DCR different (it is not 100% helpful as the audience-confusion for DCR client is still here for applications accept DCR)
- Make `<project_endpoint>` as a "default" Resources, and make the `allow_dcr_without_iat = false` by default (but when it is true, same problem happen; also, if IAT tokens is leaked, same problem as well)

Type C: Solution that prevent on backend side but with problems
- Do not allow mix of OpenID scope with `resource` attributes (it might breaks some existing clients, also failed the use cases of an app want to use the same tokens to access the app and API resources)

I might have missed some idea, so please help think of more... I somehow believed that the right solution might be a mix of multiple measures, e.g.

1. I think by default do not allow_dcr with or without iat for `project_endpoint` resources seems a good and attractive idea? I think it handle most use cases -- for MCP Client, it doesn't need `<project_endpoint>` at all; For UC1, it can allow_dcr_with_iat only; And the attack surface is if they leak the IAT tokens. (And we can revoke it at portal)
2. Do not allow mix of OpenID scope with resources attriburtes is interesting, but probably too harsh as it disallow the case of `aud = [<project_endpoint>, ResourceA]` as Tung pointed out. Any weaker alternatives?

tung2744:
@chpapa 

1. 
> aud as an array is allowed in spec, but less common in practices, that may create compatibile issue if it is the default and if developer do a simple string aud == <aud> check.

I have checked the existing access token already have a `"aud": ["<AUTHGEAR_ENDPOINT>"]` claim so it is already an array. But indeed, I cannot 100% sure no one will write `aud == ["<AUTHGEAR_ENDPOINT>"]`. However, common jwt validation library always accept the jwt if one of the `aud` claim in the jwt matches your required `aud`.

node:
https://github.com/auth0/node-jsonwebtoken/blob/master/verify.js#L194-L207

python:
https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L560-L561

golang:
https://pkg.go.dev/github.com/golang-jwt/jwt/v5#WithAudience

And this also matches what the spec says:
https://datatracker.ietf.org/doc/html/rfc9068#section-4

> > The resource server MUST validate that the "aud" claim **contains** a resource indicator value corresponding to an identifier the resource server expects for itself.

2.

> it's literally equals to checking both aud and client_id, we could have opt for that

Why I am against asking user to validate the client_id:

- It is non standard. No spec mentioned you should do it in this way.
- Most of the libraries does not expect you to validate `client_id`, as observed from their API.
  - https://pkg.go.dev/github.com/golang-jwt/jwt You can find `VerifyAudience`, but you cannot find `VerifyClientID` or similar method.
  - https://pyjwt.readthedocs.io/en/stable/api.html#jwt.decode You can find `audience` parameter but you cannot find `client_id`.
  - https://github.com/auth0/node-jsonwebtoken You can find `audience` option but you cannot find `client_id`.
  
So, my opinion is asking user to verify client_id is a non-standard way to do it and complicates the verify code. 

We should stick to the `aud` claim which is designed for this purpose.

3.

> Only static client application that update the code to check new aud would works

Correct. This is a real concern. However, I cannot think of any way to secure existing clients. There must be code change, no matter verify client_id or aud.

The best we can do is to make sure new setup apps are secure by default, which is the reason why I suggested adding a new aud value containing the client id.

I think we might need a warning on open DCR in portal for this.

---

And one point I've just discussed with @carmenlau as well:

> RFC8707 specified while /authorize can specify multiple resources, /token is not recommended to do it -- although semantically our two audience is the same so it is not a huge problem, yet still kinda werid.

This is a valid concern. Actually, carmen suggested:
There should be two access token, one for `<AUTHGEAR_ENDPOINT>`, one for `<API_RESOURCE>`. User should call `/token` endpoint two times with different `resource` to get two access token. Use them for different purpose.

Which, IMO the most "correct" solution. However, some concerns remain:

1. The default integration of authgear is still unsafe, because if `resource` is omitted, still `aud=[<AUTHGEAR_ENDPOINT>]`. And the api server probably still by default be exposed in the audience confusion risk if they validate against this issuer.
2. For a safe integration, the developer will need to register the API Resource first, associate it with the client. Then, the client can request a token with `resource=<API_SERVER>`. This seems to complicates the setup a lot compared to simply registering a client.

---

I think I should research more auth solutions and check how is their behavior on this problem. Let me post any update here.

chpapa:
@tung2744 And any thought on my recommendation of do not allow DCR without iat to request aud=project endpoint? I think that practically seems solved a lot of problem as well

tung2744:
@chpapa In my opinion, this seems to be a temporary solution. Once we got usecases which expect oidc integration, not oauth, using dcr or cimd, there will still be a problem.

Another usecase I found which could be relevant to this restriction but I did not put in the spec:

> Open Banking & Open Data Networks
> Regulatory frameworks (like the Financial-grade API / FAPI standard used in UK Open Banking or Australia's CDR) require banks to expose a DCR endpoint. The fintech app submits a software statement (signed by a central financial regulator) to the bank's authorization server to instantly obtain a client_id. It then launches an OIDC flow to authenticate the customer and obtain explicit consent to pull bank data.

And I am worried about if we handle this problem later it could be more difficult to solve as we have existing users using DCR already.

chpapa:
> @chpapa In my opinion, this seems to be a temporary solution. Once we got usecases which expect oidc integration, not oauth, using dcr or cimd, there will still be a problem.
> 
> Another usecase I found which could be relevant to this restriction but I did not put in the spec:
> 
> > Open Banking & Open Data Networks
> > Regulatory frameworks (like the Financial-grade API / FAPI standard used in UK Open Banking or Australia's CDR) require banks to expose a DCR endpoint. The fintech app submits a software statement (signed by a central financial regulator) to the bank's authorization server to instantly obtain a client_id. It then launches an OIDC flow to authenticate the customer and obtain explicit consent to pull bank data.
> 
> And I am worried about if we handle this problem later it could be more difficult to solve as we have existing users using DCR already.

@tung2744 Actually your banking use case inspired me to think clear about the problem. I think the underlying problem is we haven’t define 1st party and 3rd party client application clearly in the context of DCR.

Here is an idea that seems solve all problems:

1. We define 1st party app as: skip consent screen… blah blah blah, and most importantly, it have `aud=<project_endpoint>`
2. So 3rd party app should ALWAYS get a `aud=<resources>`

Currently we don’t do it, but that’s a problem anyway:

- 3rd party app shouldn’t have `aud=<project_endpoint>`, it is audience confusion already!
- our /userinfo endpoint now validate `aud=<project_endpoint>`, which is not necessary, we need to relax that. 

So first, we fix how Authgear handle 1st and 3rd party application.

(Some other details is needed, e.g. 3rd party app if not specified resources, the aud should be empty or limited or what? To avoid breaking existing apps, shall we just migrate existing 3rd party apps as 1st party apps — and create a new real 3rd party app?)

If the above is ok, than we can move on to DCR, everything is straight forward now.

I think at DCR, we should have 3 types of client application registration:

1. IAT for 1st party app, `aud=<project_endpoint>`
2. IAT for 3rd party app, `aud=<resource>`
3. DCR without IAT, always 3rd party.

So, no audience confusion, fits FAPI requirement. Clear cut.

Not sure if I missed anything?

tung2744:
@chpapa Based on your suggestion, something I would like to add:
- We need a default `aud` for third party app when `resource` is not given. I suggest `https://myapp.authgear.cloud/clients/<client_id>` so they are isolated by default. You are correct that our userinfo endpoint seems unnecessarily require `aud=[<AUTHGEAR>]`. This access token will still be able to call userinfo endpoint. We only check `scope` to determine what should be returned.

- resource indicator (`resource=<API>`) should be supported in first party app too to provide better isolation between clients. However it become not something we must support before DCR if third party app does not use same audience as first party app. In this case maybe first party app can simply use `aud=[<API>]` when `resource` is provided in authorize request. However this could be another feature we do in future,

The only remaining concern is the existing first party apps will remain not audience isolated by default. Developer need to opt-in to  resource indicator to isolate them. But this is the existing behavior and seems we accept.

chpapa:
>     * We need a default `aud` for third party app when `resource` is not given. I suggest `https://myapp.authgear.cloud/clients/<client_id>` so they are isolated by default. You are correct that our userinfo endpoint seems unnecessarily require `aud=[<AUTHGEAR>]`. This access token will still be able to call userinfo endpoint. We only check `scope` to determine what should be returned.

I have also thought about this, but not so sure should we go with default more limited that if resource is not given, the 3rd party app token is not for access at all (like auth0) or with sth like your suggestions. Maybe more research for pros and cons are needed.

On userinfo endpoint, I also want to look at more reference for others see why we added aud check in the first place.

>     * resource indicator (`resource=<API>`) should be supported in first party app too to provide better isolation between clients. However it become not something we must support before DCR if third party app does not use same audience as first party app. In this case maybe first party app can simply use `aud=[<API>]` when `resource` is provided in authorize request. However this could be another feature we do in future,

agree. and the /tokens exchange thing as suggested by Carmen (But I sort of remember we had it...? Or not?)

> The only remaining concern is the existing first party apps will remain not audience isolated by default. Developer need to opt-in to resource indicator to isolate them. But this is the existing behavior and seems we accept.

yea I think this seems ok for first party app definition. For advanced use case like they have multiple services and want each services authenticate correctly, the developer can implement with resources-based tokens?

tung2744:
@chpapa 

> I have also thought about this, but not so sure should we go with default more limited that if resource is not given, the 3rd party app token is not for access at all (like auth0) or with sth like your suggestions. Maybe more research for pros and cons are needed.

I believe they will be almost the same. With only a different which the backend of the third party app cannot validate the opaque access token. But in this case I believe using id token is more appropriate, and the third party app backend manage the session on their end, just like a standard oidc integration. If they need authgear to manage the access token, they need to use `resource=<API_SERVER>`.

Another finding during the research: Actually authgear has an option for clients to issue opaque access token. I checked the existing documentation, seems we did not mention what this token is for and how to use it. However in this doc: https://docs.authgear.com/get-started/backend-api/nginx
We use `/resolve` endpoint to auth the user in a backend. The `/resolve` endpoint did not tell which client or which api resources the user is authorized to access. So if `/resolve` accepts the new opaque token of 3rd party app, same "audience confusion" will occur. I suggest we just reject any access token of 3rd party app to avoid the problem in `/resolve`.

So, I think issuing opaque token for 3rd party app is ok. But this opaque token must be internally different from the existing opaque token and not accepted by `/resolve`.

> On userinfo endpoint, I also want to look at more reference for others see why we added aud check in the first place.

https://github.com/authgear/authgear-server/commit/37a5b5b1b8156b3f8ae18d41b40c4113af0549bf#diff-37d6601964baedf8cb4344ed4a118f2b25335cbdcd5a38661402254b62ce261cR83

It was added initially without specific reason. But I guess the answer is simple: The userinfo endpoint is technically a resource server of OAuth spec, thus it validates the aud according to the spec:

https://datatracker.ietf.org/doc/html/rfc9068
> The resource server MUST validate that the "aud" claim contains a resource indicator value corresponding to an identifier the resource server expects for itself.

On the other hand, I also researched some other auth solutions who supports oidc & resource indicator. And actually the behavior is different in each provider.

| Approach | Providers | How |
| :--- | :--- | :--- |
| Include userinfo URL in aud array | Auth0 | aud = [API_URI, USERINFO_ENDPOINT] — both are valid audiences on the same token |
| Don't validate aud at userinfo — only check scope | AWS Cognito | aud = API_URI only, but /userinfo accepts it as long as openid is in scope |
| Separate concerns: use ID token for identity, not access token | Duende, Connect2id, Curity | aud = API_URI only; /userinfo is not the intended path — use the ID token's claims directly |

The original design in this pr is similar to auth0. And if we relax the aud requirement in /userinfo the design will be similar to AWS Cognito. But basically, the effect is the same: The access token is accepted by both API server & Authgear.

> yea I think this seems ok for first party app definition. For advanced use case like they have multiple services and want each services authenticate correctly, the developer can implement with resources-based tokens?

Yes. I think lets make it opt-in as another feature.

tung2744:
Maybe let me summarize a bit:

1. We will deprecate the existing third_party_app. Introduce new third party app for dynamic clients.
2. The new third party app always issue opaque access token by default. The opaque token can only access userinfo. `/resolve` cannot be used with this token.
3. `resource` parameter will be supported for the new third party app. The allowed value of `resource` is defined by existing API resources. We will add a config on existing api resource and scopes to allow them to be used by new third party apps.
4. If `resource` is specifed, the issued access token always have `aud=[API_RESOURCE]`. The scope could still include openid scopes if requested. If scopes allowed, userinfo can be called. **This applies to both first party and third party app.** See Note 1.


---

And other changes:

1. IAT will be opaque. There could be per token configs. We need new admin api endpoint to issue an IAT.

---

Notes:

1. `resource` need to be supported for first party app too for UC1. In UC1, the backend of that dynamic client should validate the access token using a fixed `aud` value, so it must use API Resources.

Please see if anything I've missed.

chpapa:
@tung2744 Some notes / respond / idea.

> We will deprecate the existing third_party_app. Introduce new third party app for dynamic clients.

Last night when I think of it, I was thinking maybe third_party_app should be more like an attribute for Client Applications instead of "Type" of client application, maybe you can try to list them out and see which design is better.

> We will add a config on existing api resource and scopes to allow them to be used by new third party apps.

Not quite sure what it meant but maybe I can wait til your full design.

> We use /resolve endpoint to auth the user in a backend.

Just an idea / side question, what about introspection endpoint? Do we have that? XD

> Yes. I think lets make it opt-in as another feature.

Actually what they need to opt-in? I think as long as the developer do not accept `aud=<project_endpoint>` and try to define resources on their own and ask for the right resources in their authorize and tokens endpoint (maybe we need better SDK support for it?), than they have already "opt-in" it?

tung2744:
> Actually what they need to opt-in? I think as long as the developer do not accept aud=<project_endpoint> and try to define resources on their own and ask for the right resources in their authorize and tokens endpoint (maybe we need better SDK support for it?), than they have already "opt-in" it?

This is correct. But this is not how the existing user using authgear, so I called it "opt-in".

docs/specs/dcr.md
tung2744: Not related as it is opaque now

tung2744:
Updated. Here is a summary:

## Third-Party Clients

- Redefined third party client by listing out difference with first party client
- The existing x_application_type: third_party_app static YAML entry is deprecated; new third-party clients must be created via DCR and never exist in authgear.yaml
- By default (no resource parameter), third-party clients receive an opaque access token (no aud, userinfo only, not usable with /resolve)
- When resource is specified, a JWT is issued with aud = [resource_uri] only — the project endpoint is not included  


## API Resources and Scopes

- Extracted from m2m.md into a dedicated spec (api-resource.md)
- Added allow_any_client_access flag at both Resource and Scope level — when set on both, any client (including third-party DCR clients) can access that resource/scope without a per-client association
- Resource access for DCR clients is configured in the portal, not via authgear.yaml (allowed_resources config removed)

## Access Token Audience Binding

- First-party clients retain existing behavior: JWT with aud = [project_endpoint] when no resource is specified
- When resource is specified, aud contains only the resource URI — the project endpoint is never included (applies to both first-party and third-party)
- Removed the per-client URI (project_endpoint/clients/client_id) in previous version

## Initial Access Token (IAT)

- IAT is now an opaque token issued via the Admin API, replacing the previous JWT-based design
- Admin API mutations added: create, revoke, and list IATs

## DCR application_type Field

- Conforms to the OIDC DCR spec: web and native are the standard values, both creating third-party clients
  - web: HTTPS redirect URIs required, no localhost
  - native: custom URI schemes or http://localhost
- Added Authgear extension values for first-party clients: first_party_spa, first_party_native, first_party_traditional_webapp, first_party_confidential — all require an IAT
- Default is web
- x_application_type is not modified or renamed. It only appears in static clients in authgear.yaml and does not conflict with standard application_type which is in DCR which are in database.

DCR token_endpoint_auth_method Field

- Field removed from DCR — the auth method is now fixed by application_type. I think we do not need to support token_endpoint_auth_method.

chpapa:
@tung2744 haven't read the doc in details yet, just responding on summary:

>     * The existing x_application_type: third_party_app static YAML entry is deprecated; new third-party clients must be created via DCR and never exist in authgear.yaml

I still think we want to support creating 3rd party client application from portal (how to store it is another thing), but feature-wise "3rd parties client must be created via DCR" seems incorrect. (Implementation could be another ticket though)
 
>     * Added allow_any_client_access flag at both Resource and Scope level — when set on both, any client (including third-party DCR clients) can access that resource/scope without a per-client association

What's the intention on this? 

>     * When resource is specified, aud contains only the resource URI — the project endpoint is never included (applies to both first-party and third-party)

For first-party client, do we really want to do this? or we allow them to specify ....? 

And I assume even if you do this, we allow multiple resources + project endpoint at `authorize` but just `tokens` return 1 aud for 1 token?

>     * Added Authgear extension values for first-party clients: first_party_spa, first_party_native, first_party_traditional_webapp, first_party_confidential — all require an IAT

As previously commented, maybe we should have first_party vs 3rd_party as a attribute instead of app type? By making it app_type, we seems overloading a lot of attributes that could potentially conflict with other oidc spec on it...

tung2744:
@chpapa 

> I still think we want to support creating 3rd party client application from portal (how to store it is another thing), but feature-wise "3rd parties client must be created via DCR" seems incorrect. (Implementation could be another ticket though)

What I want to do is to deprecate the existing `third_party_app` in the config. I believe we can define another way to add third party app in config, or call DCR to register in portal. But I want to avoid designing this part in this spec. I think it is easy to add in the future.

> What's the intention on allow_any_client_access

It is the same purpose as `allowed_resources` in the initial design. Now we put this config on the api resources & scopes. It allows any client to request access to that resource.
Maybe you would ask, why is it "any client" but not third party client?
My reasons:
1. In UC1 which first party client is registered through DCR, if the project owner wants to use resource indicator, this will allow him to do so.
2. This design ensures third party client is always a subset of first party client (There should not be anything third party client can do but first party client can't).

 > For first-party client, do we really want to do this? or we allow them to specify ....?

I assume you mean we do `aud=[API_RESOURCE]` but not `aud=[AUTHGEAR_ENDPOINT, API_RESOURCE]`.
My reasons:
1. Basically the reason of using resource indicator is to narrow down the target audience. By using `resource=<API_RESOURCE>`, the user already tell us the intended consumer of the access token is `API_RESOURCE`.
    - The resource server (API_RESOURCE), can just verify the token with `aud=[API_RESOURCE]`, seems no reason to include `AUTHGEAR_ENDPOINT`.
2. If you are asking about why it is not for third party client only: In my opinion resource indicator should behave the same way in first or third party client. The `resource` parameter itself is already optional. If the developer do not want to narrow down the audience, just do not use it. IMO, in this way the `resource` parameter is also less confusing.

> And I assume even if you do this, we allow multiple resources + project endpoint at authorize but just tokens return 1 aud for 1 token?

1. I do not plan to allow using `<AUTHGEAR_ENDPOINT>` in `resource` parameter. This should only be the "default" audience when `resource` is not provided.
2. Multiple resources in `authorize` endpoint is allowed, as OAuth spec mentioned.
3. Multiple resources in `token` endpoint is also allowed. Reading https://www.rfc-editor.org/info/rfc8707/#name-access-token-request, although there isn't an example for it, it is actually allowed.
    > When requesting a token, the client can indicate the desired target service(s) where it intends to use that token by way of the resource parameter and can indicate the desired scope of the requested token using the scope parameter. The semantics of such a request are that the client is asking for a token with the requested scope that is usable at all the requested target services.
    - Although, for first party client, the exact mechanism for authgear to decide what `resource` or `scope` is allowed is still undefined. I believe that should be another feature. As a result, for now, first party app cannot really use `resource` parameter - We will always reject the request.

> As previously commented, maybe we should have first_party vs 3rd_party as a attribute instead of app type? By making it app_type, we seems overloading a lot of attributes that could potentially conflict with other oidc spec on it...

I would like to clarify to ensure we are talking about the same thing, the application_type here is referring to the parameter of the DCR /register endpoint.

We actually have two choice here:
1. Add additional options for the `application_type` parameter for first party app.
2. Add additional parameter on the /register endpoint, e.g. `x_is_first_party`.

Now it is 1. I did not do 2 because:
- It creates ambigious / invalid combinations. Like `application_type=web`, `x_is_first_party=true`. Should it be mapped to `traditional_webapp` or `spa`? Sure we can define a mapping in doc or do not allow registering `traditional_webapp` from DCR, and also block all invalid combination. But isn't a single `application_type` parameter simpler? We do not need to handle `x_is_first_party * application_type * token_endpoint_auth_method` combinations if application_type already handle all of them.
- I think it is easier to find the well known application_type parameter than a custom `x_is_first_party` parameter.
- I think the risk of value collision on `first_party_xxx` is low. If we want to be safe, we can further rename to `x_first_party_xxx`.

chpapa:
> What I want to do is to deprecate the existing `third_party_app` in the config. I believe we can define another way to add third party app in config, or call DCR to register in portal. But I want to avoid designing this part in this spec. I think it is easy to add in the future.

ok, but than we probably shouldn't say DCR is the only way for 3rd party app.

> It is the same purpose as allowed_resources in the initial design. Now we put this config on the api resources & scopes. It allows any client to request access to that resource.
> Maybe you would ask, why is it "any client" but not third party client?

I think this is werid and not obvious for client what this attributes does. We need both a better naming and description. (More specific for DCR open registration

> I do not plan to allow using <AUTHGEAR_ENDPOINT> in resource parameter. This should only be the "default" audience when resource is not provided.

Than what if a 1st party app want to authorize for `<AUTHGEAR_ENDPOINT>` + `resourcesA` (and call multiple `/token` to get tokens for each audience)

> We actually have two choice here: 
> 1. Add additional options for the application_type parameter for first party app.
> 2. Add additional parameter on the /register endpoint, e.g. x_is_first_party.

Actually I was thinking more like:

1. For static app, the client_id have an attribute to define is it a 1st party or 3rd party, so the client_id decide it.
2. For DCR app, the type of IAT define it.

I haven't thought through which approach is better. Yet I feel like the above seems less dangerous than what you proposed / we originally have (creating more application_type to signal it):

1. Actually I'm suggesting we should even remove the internal `x_application_type` thing, so we should just say app_type = web + auth_method = none is SPA; auth_method = client_secret_basic is web

2. So we don't run into, as you said, what if the app_type and auth_method attribute contradict 

(and we don't run into similar confusion when we need to adopt new OIDC protocol like this time)

---

BTW, it seems in your summary you didn't mention IAT for 1st party app vs 3rd party app, so how to limit it for now? Or IAT is always for 3rd party app only and 1st party app (UC1) is for future work?

---

See if you have think through all of the above, and let me know when you think I should read the whole spec again but not just your summary? (Yet I think via the summary we can discuss and align some design principle / use cases clarity)

p.s. I would be on leave til next Wed, maybe the best time to review it again is next Wed morning (or pls schedule with me separately.)

tung2744:
1. > Than what if a 1st party app want to authorize for <AUTHGEAR_ENDPOINT> + resourcesA (and call multiple /token to get tokens for each audience)

Do not allow. Reason:
- If you do not need audience isolated token, validate `aud=[AUTHGEAR_ENDPOINT]`.
- If you want audience isolated token, validate `aud=[API_RESOURCE]`.

There is also no reason for putting `AUTHGEAR_ENDPOINT` in `resource` parameter in authorize endpoint, because authgear server always accept access token issued by itself. (As discussed we will relax the aud requirement of userinfo endpoint).

So for this case just request `resource=API_RESOURCE` in both authorize and token endpoint.

2. > allow_any_client_access

Let me also think if we can have a better naming for allow_any_client_access.

3. > About application_type

Actually, I am trying to make it unrelated to the static config (And thus, unrelated to x_application_type).

Removal of `x_application_type` seems to be a redesign of our existing oidc client. I prefer not to do it here. There are quite many logic related to this field such as what portal configs are available, and how the clients are created, not only related to first/third party and client secret (m2m is actually also a x_application_type). I feel like this PR is already too many things.

So I suggest just to treat `application_type` a parameter of DCR. Map each possible DCR request to an existing x_application_type for now (If it is a first party app, new third party app has no relationship with x_application_type). Reject all other invalid combinations. In this way, migrate existing first party x_application_type to another model which compatible to all existing dynamic clients should be still possible.

I see some other services also overloading `application_type`, e.g. [Okta](https://developer.okta.com/docs/api/openapi/okta-oauth/oauth/client/createclient)

But it is indeed not very common.

If overloading `application_type` feels dangerous, I suggest we just add a new parameter `x_is_first_party`. And a mapping:
- x_is_first_party=true & application_type=web, map to x_application_type=spa of static client.
- x_is_first_party=true & application_type=native, map to x_application_type=native static client.
- x_is_first_party=false & application_type=ANY, new third party client. application_type not stored and only affect validation of redirect uris.

> For DCR app, the type of IAT define it.

I think the purpose of IAT should be limited to authorization. It can authorize the caller to create first party client, but whether to create a first party client should still be decided by the caller and be reflected in the request.

4. > About IAT for first / third party app

I think no change from the initial design. If there is an IAT, first party app creation is allowed. Third party app is allowed no matter you have a IAT or not.
You have a config `initial_access_token_required` to choose whether the registration endpoint is open or not.
Limitation per token (Such as number of client can be created, rate limits..) can be applied later because now the token is opaque and must be generated using an api. But I did not define what can be controlled in this version yet.

---

I think you have already get most of the idea from my summary. See if the above answers your question, or we further discuss later.

chpapa:
> Let me also think if we can have a better naming for allow_any_client_access.

Just wanna stress that I think it is not just naming, but the concept behind...

> If overloading application_type feels dangerous, I suggest we just add a new parameter x_is_first_party. And a mapping:
>I think the purpose of IAT should be limited to authorization. It can authorize the caller to create first party client, but whether to create a first party client should still be decided by the caller and be reflected in the request.

I would strongly oppose this. An IAT that can create 1st party vs 3rd party application have vastly difference security profile. In fact I always think the only decision for us is either DO NOT allow register of 1st party application via DCR at all (drop UC1), or the bottom line is make it explicit which IAT can do so.

(btw, hence it is better if we have some prefix on the IAT token, like prod vs test token in Stripe, so it is very obvious to users they're leaking a sensitive or non-sensitive tokens, e.g. `ag_iat_1st_xxxxxxxx` lol)

Since the decision of 1st party or 3rd party app is by type of IAT, hence the parameter of `x_is_first_party` is not a must (unless we want to let 1st party IAT user express this intent, but seems a bit unnecessary flexible)

tung2744:
@chpapa I've updated the spec according to our latest comments.

> allow_any_client_access

I've redesigned this part. Now, API Resource and Scope objects will have a `accessPolicy` field. Which, contains only one key:
`allowThirdPartyClientAccess: boolean`.
The meaning is:
> When true, all third-party clients may access this Resource or Scope
  without a per-client association.

- AccessPolicy is an object so we can extend it easily for other purpose in future, for example, `enableRBAC` which we can support associating scopes with roles & groups.
- I made this design referencing Auth0's "Default Permissions for third-party applications" concept. However, I did not reference their api design, which introduced another concept [Client Grant](https://auth0.com/docs/get-started/applications/application-access-to-apis-client-grants). I think a single `accessPolicy` object is easier to understand and enough.

An alternate design could be something similar to KeyCloak's Client Scope Policy in [Client Registration Policy](https://www.keycloak.org/securing-apps/client-registration#_client_registration_policies). I think this can be useful if we want to more fine grained control on DCR clients but for now I decide not to put it in the design.

> IAT

I see your concern on mixing 3rd & 1st party IAT. I have updated the spec to separate the two type of IATs.
First party IAT: `iat_fp_xxx`
Third party IAT: `iat_fp_xxx`

When creating IAT, caller must specify the type of IAT to create.

For details, see [Initial Access Token](https://github.com/tung2744/authgear-server/blob/016543a173237d9030ed932601cd61d2e0db23e1/docs/specs/dcr.md#initial-access-token)

Also, `application_type` now support `web` and `native` only. You can see the [`application_type`](https://github.com/tung2744/authgear-server/blob/016543a173237d9030ed932601cd61d2e0db23e1/docs/specs/dcr.md#application_type-optional) section for detail. Now it should be unrelated to `x_application_type` which remains a field in the static client config only.

tung2744:
1. We will design a model which should be able to handle all clients, including static config client & dcr clients.
2. allowThirdPartyClientAccess should be called more specific to open (Or publicly registered) third party client.

tung2744:
Updated, summary:
1. Added unified client model. Documented how existing static config clients and DCR requests map to that model.
2. Renamed allow_third_party_client_access → allow_dynamic_third_party_client_access.

tung2744:
Lets continue in https://github.com/authgear/authgear-server/pull/5847

PR #5766, opened 2026-06-10
Fix false conflict when secret config has non-standard YAML formatting

tung2744:
ref DEV-2378

How to test:
1. In authgear.secrets.yaml, add some indentation but do not break the syntax or meaning of the content.
2. In portal, try create a OIDC client. Before the fix, you will see no response. After the fix, you can create successfully.


carmenlau (commented):
The PR looks good. Could you fix `make check-tidy` please?

tung2744:
Updated, thanks!

carmenlau (approved):

PR #5769, opened 2026-06-11
Show human-readable error messages for known OAuth provider errors

carmenlau:
ref DEV-3649

Screenshot 2026-06-11 at 11 23 54


resources/authgear/templates/en/web/authflowv2/__error.html
```diff
@@ -288,7 +288,25 @@
...
+        <span>{{ include "v2.error.oauth-iamsmart-cancelled" nil }}</span>
+      {{ else if and (eq $oauth_provider_type "iamsmart") (eq $oauth_error_code "D40001") }}
+        <span>{{ include "v2.error.oauth-iamsmart-rejected" nil }}</span>
+      {{ else if and (eq $oauth_provider_type "iamsmart") (eq $oauth_error_code "D40002") }}
+        <span>{{ include "v2.error.oauth-iamsmart-failed" nil }}</span>
+      {{ else if and (eq $oauth_provider_type "iamsmart") (eq $oauth_error_code "D40003") }}
+        <span>{{ include "v2.error.oauth-iamsmart-timeout" nil }}</span>
+      {{ else }}
+        <span>{{ include "v2.error.oauth-error" (dict "error" $oauth_error_code "error_description" .Error.info.error_description) }}</span>
+      {{ end }}
```
tung2744: Instead of handling provider specific error code here, do you think we can have a error translation in the oauthrelyingparty code?

carmenlau: I think the translation and copywriting should be owned by the Auth UI layer. How about if I introduce an Error Category concept in `oauthrelyingparty`? And providers map their proprietary error codes to a set of defined categories, and the Auth UI template handles each category without any provider-specific logic.
 
Let me update the code to reflect that.

tung2744 (approved):

PR #5773, opened 2026-06-15
chore: Update deps

tung2744:
ref DEV-3645


carmenlau (commented):
For commit `b206155`, are we going to fix those warnings, or are we just trying to avoid introducing new warnings from the new component? If we plan to fix them, please open an issue to track the follow-up work.

The changes in this PR look good to me. I'll go ahead and merge it.

carmenlau (approved):

tung2744:
@carmenlau My thought is to avoid introducing new warnings. And fix these existing issues if we touch the related code. If they are not modified, then just suppress that specific warning, which is what I did in this PR. So we do not need to modify a lot of code every time upgrading eslint. What do you think?

carmenlau:
Sure, I prefer this approach as well. Let's keep it as is.

PR #5783, opened 2026-06-22
Simplify authgear setup in local

tung2744:
ref DEV-3657


carmenlau (approved):
This setup is really helpful for setting up the development environment! Maybe we could consider updating `CONTRIBUTING.md` to use it as well.

PR #5787, opened 2026-06-23
[Site Admin] Audit log for app plan and collaborator mutations

carmenlau:
ref DEV-3549

Emit audit log entries when site admin mutations are performed:
  - `site_admin.app.plan.updated`
  - `site_admin.app.collaborator.added`
  - `site_admin.app.collaborator.deleted`
  - `site_admin.app.collaborator.promoted`

All records are stored under the portal app ID (`SITEADMIN_AUTHGEAR_APP_ID`) in the global audit DB. The affected app ID is in `data.payload.app_id`. Log structure mirrors portal audit logs (audit_context, user_agent, seq, etc.).

Also filters the admin API audit query to known activity types by default to prevent crashes when unknown types are stored in the DB.


pkg/api/event/nonblocking/siteadmin_app_collaborator_added.go
```diff
@@ -0,0 +1,49 @@
+package nonblocking
```
tung2744: Can we put these siteadmin specific events into the site admin package?
Speaking on API level they are not a "event". They just shared the implementation of events.
So lets move them into another package and we won't be confused.

carmenlau: I've moved them to `pkg/siteadmin/auditlog`, see if it looks good.

tung2744 (approved):

PR #5788, opened 2026-06-24
spec: Specify select account flow with custom ui

tung2744:
ref DEV-3670

### DIFF
--- a/docs/specs/custom-ui-select-account.md
+++ b/docs/specs/custom-ui-select-account.md
@@ -1,0 +1,367 @@
+# Custom UI: Select Account
+
+- [Overview](#overview)
+- [Background](#background)
+- [Goals](#goals)
+- [Design](#design)
+  - [Security principle](#security-principle)
+  - [Phase 1: Authorization endpoint — detect existing session](#phase-1-authorization-endpoint--detect-existing-session)
+  - [Phase 2: Custom UI — display the account picker](#phase-2-custom-ui--display-the-account-picker)
+  - [Phase 3a: User continues with existing account](#phase-3a-user-continues-with-existing-account)
+  - [Phase 3b: User switches to a different account](#phase-3b-user-switches-to-a-different-account)
+- [HTTP API](#http-api)
+  - [GET /api/v1/accounts_hint/{token}](#get-apiv1accounts_hinttoken)
+  - [GET /api/v1/select_account](#get-apiv1select_account)
+- [End-to-end flow diagram](#end-to-end-flow-diagram)
+- [Edge cases](#edge-cases)
+- [Security analysis](#security-analysis)
+- [Backward compatibility](#backward-compatibility)
+- [Reference: x_ref](#reference-x_ref)
+---
+
+## Overview
+
+This document specifies how a Custom UI can present an account-selection screen when the end-user already has an active Authgear session, allowing them to continue as their existing account without re-entering credentials.
+
+---
+
+## Background
+
+When the built-in Auth UI handles a returning user, it routes the browser through `/authflow/v2/select_account`. This handler runs on Authgear's own domain, reads the session cookie, and—if the user clicks "Continue"—completes the OAuth authorization flow without creating an authentication flow at all.
+
+Custom UI is hosted on a different domain and communicates with Authgear via the [Authentication Flow HTTP API](./authentication-flow-api-reference.md). Two constraints make a direct port of the built-in behavior impossible:
+
+1. **Cross-domain cookies**: API calls from the Custom UI are cross-origin; the browser does not send Authgear's session cookie with them. The Custom UI cannot detect an existing session by calling the API.
+2. **Backward compatibility**: The Authentication Flow API must not inject new action types into existing flows, as that would break Custom UI implementations that do not know how to handle them.
+
+---
+
+## Goals
+
+- Allow a Custom UI to detect that the end-user has an existing Authgear session.
+- Allow the Custom UI to show user account information (display name) for the account-selection screen.
+- Allow the end-user to continue with the existing session without re-entering credentials.
+- Preserve security: an attacker who captures the redirect URL must not be able to complete authentication on behalf of the victim.
+- Keep existing Custom UI integrations working without modification.
+
+---
+
+## Design
+
+### Security principle
+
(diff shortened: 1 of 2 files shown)


docs/specs/custom-ui-select-account.md
```diff
@@ -0,0 +1,367 @@
...
+  │
+  ├─▶ GET /api/v1/accounts_hint/{x_accounts_hint}
+  │       ↓
+  ├─◀ { accounts: [{ display_name }, …] }
+  │
+  │   [Show "Continue as user@example.com / Use different account"]
+  │
+  │   User clicks "Continue" (selects account at index N)
+  │       ↓
+  ├─▶ GET /api/v1/select_account?x_ref=...&x_account_index=N
```
tung2744: We've discussed about the possibility of changing this to a POST request, but did not find a good way. The restriction is from the SameSite=Lax cookie which is not readable by POST request.

[ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#lax)

> Send the cookie only for requests originating from the same [site](https://developer.mozilla.org/en-US/docs/Glossary/Site) that set the cookie, and for cross-site requests that meet both of the following criteria:

> The request is a top-level navigation: this essentially means that the request causes the URL shown in the browser's address bar to change.

> This would exclude, for example, requests made using the [fetch()](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch) API, or requests for subresources from [](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img) or [<script>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script) elements, or navigations inside [<iframe>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe) elements.

> It would include requests made when the user clicks a link in the top-level browsing context from one site to another, or an assignment to [document.location](https://developer.mozilla.org/en-US/docs/Web/API/Document/location), or a [<form>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form) submission.

tung2744:
Updated the spec with the assumption of same-site custom ui.

docs/specs/custom-ui-select-account.md
```diff
@@ -97,25 +99,26 @@ If `x_select_account_token` is absent from the Custom UI URL, the Custom UI MUST
...
-window.location.href =
-  authgearEndpoint + '/api/v1/select_account'
-  + '?x_ref=' + encodeURIComponent(xRef)
-  + '&x_account_index=' + selectedAccountIndex;
+```html
+<form id="continueForm" method="POST" action="https://auth.example.com/api/v1/select_account">
+  <input type="hidden" name="x_ref" value="oauthsession_abc123">
+  <input type="hidden" name="x_account_index" value="0">
+</form>
+<script>document.getElementById('continueForm').submit();</script>
```
carmenlau: In the `authflow` API, we set the data in the JSON body and return the redirect URI in the response body (e.g. `finish_redirect_uri`). This API is different, so I'd like to understand the reason for this difference. Or do we not need to be consistent since this API is not part of the `authflow` API?

tung2744: 1. I think they do not need to be consistent because it is not part of authflow api.
2. This api must be triggered by browser form POST. `fetch` does not include Lax cookies even with include credential option, Thus, use the normal browser redirection mechanism.

carmenlau: Thanks for explaining.

tung2744:
Redesigned using authflow api with assumption on same site custom ui.
Key changes:
- Added `identification: select_account` to `login` and `signup_login` flows
- Specified CORS behavior of authentication flow api

docs/specs/custom-ui-select-account.md
```diff
@@ -0,0 +1,608 @@
...
+
+Opting in requires updating the Custom UI to recognize the new `identification` value, same as adding any other new identification method — plus, for a Custom UI with separate sign-in/sign-up screens, routing its first call through `login`/`signup_login` rather than `signup` (see [End-to-end sequence](#end-to-end-sequence)) — a real behavior change, not just a config one.
+
+---
+
+## Edge cases
+
+- **`prompt=login`**: option omitted (see [Session and account resolution](#session-and-account-resolution)).
+- **`prompt=none`**: not applicable — decided before any flow exists.
+- **`prompt=select_account`**: not supported — has no effect, despite sharing its name with this feature's `select_account` identification option (unrelated concepts).
```
tung2744, line 607: Research on prompt=select_account behavior

docs/specs/custom-ui-select-account.md
```diff
@@ -0,0 +1,608 @@
...
+- **`max_age`**: not a separate rule. An expired `max_age` is folded server-side into a synthesized `prompt=login`, so it's already covered by the `prompt=login` case above.
+- **`id_token_hint`**: option omitted when its `sub` claim identifies a different user than the session — the same rule as a mismatched `login_hint` (see [Session and account resolution](#session-and-account-resolution)).
+- **`acr_values`**: not supported — no interaction with `select_account` today.
+- **`login_hint` matching the session's user**: option is still offered — only a *mismatched* `login_hint` omits it.
```
tung2744: Check existing login_hint behavior on api / default auth ui

tung2744:
@carmenlau @fungc-io @chpapa 

Added a section on research of well known idp's behavior on prompt=select_account.

1. All tested Idps never returns account_selection_required error as the spec stated. All fallback to normal login.
2. `login_hint`, `id_token_hint` do not constraint login, they only prefill or provide choices.

So my take:
- We do not need to follow the spec strictly to return account_selection_required error when there are no choices. Fallback to normal login is a more common UX.
- `login_hint` and `id_token_hint` do not constraint the API. We should still pass them to the custom UI as a hint. The custom UI should decide how they should be used, select_account options on API level will not be affected.

tung2744:
I've made one more update:
`user_id` is added to the identification option data, because I found it necessary for the ui to know the user id so it can compare it with login_hint / id_token_hint.

rickmak:
Please be reminded for closing PR, press the approval button before merge. Otherwise, I will need to trace the conversation and confirm the peer review are in place in reality.

rickmak:
I notice you approved on another PR and it's platform problem on stacked PR. :sosad: Bypass should not possible without turn off the protection.

PR #5792, opened 2026-06-29
Portal: Various bug fixes

tung2744:
ref DEV-3679


carmenlau (commented):
The PR looks good, but the commit message for caa5b09 is a bit misleading. It says:

> Fix by removing the entire secret entry when no certificates remain after
deletion, consistent with how other secret types handle cleanup.

However, the code actually rejects the deletion with an error. If it's not too much trouble, could you update the commit message?

tung2744:
@carmenlau Fixed, thanks!

carmenlau (approved):

PR #5801, opened 2026-07-03
spec: Document rp-initiated logout in oidc spec

tung2744:
ref DEV-3642


tung2744:
1. Document that we cannot completely avoid id_token to appear on the url as it is a known oidc weakness. We will use POST in our SDK and avoid logging the query parameter in our server logs. And allow user to remove PII using hooks as a last resort.
2. We want to document the error behavior of post_logout_redirect_uri, which is a legacy behavior. And it should be documented as api debts

tung2744:
For 1 it is documented already. Added a commit for 2.

tung2744:
Merged in https://github.com/authgear/authgear-server/pull/5814

PR #5805, opened 2026-07-08
Support set idp session cookie as browser session cookie

tung2744:
ref DEV-3683


docs/specs/sessions.md
```diff
@@ -45,7 +45,18 @@ When the user authenticates successfully, an IdP session is created.
...
+IdP session token is stored in the user agent cookie storage. The cookie domain attribute is configurable. The default value is eTLD + 1. As long as the web application is under the same domain with Authgear, the IdP session is shared across between the two. The cookie is http-only and is not configurable. The cookie is SameSite=lax and is not configurable. The cookie is secure by default.
+
+By default, the cookie is a persistent cookie: its `Max-Age` attribute is set to the session lifetime, so the cookie survives browser restarts. The developer can set `session.is_session_cookie` to `true` to make the cookie a [session cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_the_lifetime_of_a_cookie) instead:
+
+```yaml
+session:
+  is_session_cookie: true
```
carmenlau: The rest of the config uses `<noun>_enabled` for boolean flags (e.g. `idle_timeout_enabled`). Could we rename this to
 `session_cookie_enabled` to stay consistent?

pkg/lib/config/session.go
```diff
@@ -39,6 +40,10 @@ type SessionConfig struct {
 	IdleTimeoutEnabled             *bool           `json:"idle_timeout_enabled,omitempty"`
 	IdleTimeout                    DurationSeconds `json:"idle_timeout_seconds,omitempty"`
 	Deprecated_CookieNonPersistent bool            `json:"cookie_non_persistent,omitempty"`
+	// IsSessionCookie indicates whether the IdP session cookie is a HTTP session cookie,
+	// i.e. it has no Max-Age, so the browser discards it when it is closed.
+	// This does NOT affect server-side session expiry.
+	IsSessionCookie bool `json:"is_session_cookie,omitempty"`
```
carmenlau: When I checked the deprecated `cookie_non_persistent` flag, I found that we're still referencing it on the portal anonymous user page. It should have been removed when we deprecated the flag. I think we should remove it so it doesn't cause confusion with this newly added flag.

docs/specs/sessions.md
tung2744: But I think `session_cookie_enabled` is a bit inaccurate: What is mean by enabling session cookie? Accurately speaking it is switching to use session cookie, or say storing it as session cookie. I have considered `use_session_cookie` too, see if any other suggestions.

carmenlau: I would prefer `use_session_cookie` because it's clearer. We also have some configuration names that start with a verb, such as `issue_jwt_access_token`.

tung2744: Let me rename to use_session_cookie

tung2744:
Updated, thanks!

tung2744:
ci fail with vulnerability check which is not related to this pr, I will fix it with another pr

carmenlau (approved):

PR #5810, opened 2026-07-14
Portal: Fix blocker dialog sometimes appears incorrectly

tung2744:
ref DEV-3699

### DIFF
--- a/hack/custom-resources/authgear.secrets.yaml
+++ b/hack/custom-resources/authgear.secrets.yaml
@@ -1,8 +1,8 @@
 secrets:
-  - key: db
-    data:
-      database_schema: app
-      database_url: 'postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable'
+  - data:
+      database_schema: public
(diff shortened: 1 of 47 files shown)


hack/custom-resources/authgear.secrets.yaml
```diff
@@ -1,8 +1,8 @@
 secrets:
-  - key: db
-    data:
-      database_schema: app
-      database_url: 'postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable'
+  - data:
+      database_schema: public
+      database_url: postgres://postgres:postgres@127.0.0.1:5432/app?sslmode=disable
+    key: db
```
carmenlau, line 5: I'd like to know which part is still using this file. I searched the codebase a bit but couldn't find anything.

carmenlau (approved):

hack/custom-resources/authgear.secrets.yaml
tung2744: Oh it was used by my .env only for CUSTOM_RESOURCE_DIRECTORY, In env example it seems pointed to ./var. So maybe I should just update my local .env and delete this.

PR #5814, opened 2026-07-20
Add id_token_hint support

tung2744:
requires
- https://github.com/authgear/authgear-server/pull/5808

ref DEV-3722

How to test:

1. Clone this repo: https://github.com/tung2744/test-oidc
2. Fill client id and client secret (You can test public client by leaving it empty)
3. Whitelist `http://localhost:15001/callback` as redirect uri, `http://localhost:15001/` as post_logout_redirect_uri.
4. Try login and logout
5. Change `OIDC_END_SESSION_METHOD` to `POST`  to test for POST method

---

Points to note:

- In this PR, I changed the spec a bit: Now id_token_hint do not only match the exact same session of sid, but the session in the same sso group
- For any offline grant (refresh tokens) that created from an idp session (By continue screen), now they will be treated as the same SSO group as the parent idp session. I guess this practically did not change anything: Offline grant can only be created with `idp_session_id` set is by the continue screen, and user only see continue screen if sso is enabled. But we can discuss on this if you see any concern.


carmenlau (approved):
Please rebase.

tung2744:
Rebased, thanks!

carmenlau (approved):

PR #5818, opened 2026-07-21
Fix access token not issued when offline_access scope is absent

tung2744:
The authorization_code token exchange only issued an access token without a refresh token when the client was confidential, so public clients (e.g. SPA) requesting authorization_code without offline_access got "cannot issue access token" instead of an access token.

ref DEV-3545


carmenlau:
Commit 5d36aa4b95cae05b06a3be0495fae49c155a4467 should be included in PR #5821. If so, could you please remove it from this PR? Thanks again for fixing the test.

carmenlau (approved):

PR #5819, opened 2026-07-21
[Portal] Fix wizard save failing when project name contains a quote

carmenlau:
ref DEV-3725

### DIFF
--- a/portal/src/graphql/portal/DesignScreen/form.ts
+++ b/portal/src/graphql/portal/DesignScreen/form.ts
@@ -48,6 +48,10 @@ import { nonNullable } from "../../../util/types";
 import { nullishCoalesce, or_ } from "../../../util/operators";
 import { deriveColors } from "../../../util/theme";
 import {
+  escapeMessageFormatText,
+  unescapeMessageFormatText,
+} from "../../../util/messageFormat";
+import {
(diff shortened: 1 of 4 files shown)


tung2744:
I want to know if this is a pure frontend problem? Because when you type a `'` in project name, you can actually pass the first step, but error in last step, but I guess the app was created in first step?

carmenlau:
I checked the current project flow. When a project is created, the translation JSON is generated by the server and populated with the app ID, so it doesn't have this problem. Ref: https://github.com/authgear/authgear-server/blob/019afa96e9aa8735fd565dd20450482d6c65d21c/pkg/portal/service/app.go#L610

The project name entered by the user is updated in the translation JSON when the user clicks "Done" to complete the wizard. The translation JSON file is modified on the client side, so this is a client-side issue.

Actually, before this fix, updating the project name on the Design page would also fail. Since both flows use the same method to update the project name, this fix resolves the issue there as well.

tung2744:
e2e tests should be failing because of a siteadmin test. See [be05390](https://github.com/authgear/authgear-server/pull/5821/commits/be05390950b238bba0bc967c2ada88fd6d25c350)

tung2744 (approved):

PR #5820, opened 2026-07-21
[OIDC] Support client_secret_basic at the token endpoint

tung2744:
Extract client_id/client_secret from an HTTP Basic Authorization header in TokenHandler.Handle, alongside the existing client_secret_post support, and advertise it in token_endpoint_auth_methods_supported.

ref DEV-3728

### DIFF
--- a/pkg/lib/oauth/handler/handler_token.go
+++ b/pkg/lib/oauth/handler/handler_token.go
@@ -261,26 +261,27 @@ func (h *TokenHandler) Handle(ctx context.Context, rw http.ResponseWriter, req *
 		return resultErr
 	}
 
-	if username, password, ok := req.BasicAuth(); ok {
-		// RFC 6749 section 2.3: The client MUST NOT use more than one
-		// authentication method in each request.
-		if r.ClientID() != "" || r.ClientSecret() != "" {
-			return errorResult(protocol.NewError("invalid_request", "client authentication must not be presented in both the Authorization header and the request body"))
-		}
+	// Prefer client authentication in the request body over the Authorization
+	// header: an Authorization header may come from an unrelated proxy (e.g.
+	// one protecting a staging environment with HTTP Basic auth) rather than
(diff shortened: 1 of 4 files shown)


pkg/lib/oauth/handler/handler_token.go
```diff
@@ -260,6 +261,28 @@ func (h *TokenHandler) Handle(ctx context.Context, rw http.ResponseWriter, req *
 		return resultErr
 	}
 
+	if username, password, ok := req.BasicAuth(); ok {
+		// RFC 6749 section 2.3: The client MUST NOT use more than one
+		// authentication method in each request.
+		if r.ClientID() != "" || r.ClientSecret() != "" {
+			return errorResult(protocol.NewError("invalid_request", "client authentication must not be presented in both the Authorization header and the request body"))
```
carmenlau: Could a proxy-level Basic Auth header, such as one used to protect staging environments, conflict with `client_secret_post` clients and cause an `invalid_request` error? This header was ignored before this change. It might be safer to prefer the body credentials when both are present instead of returning an error.
 
The spec says "the client MUST NOT use more than one authentication method in each request," but it doesn't define an error response for this case, and here, the Basic Auth header isn't for the token endpoint at all, it's for the proxy.

What do you think?

carmenlau: Or do you think the proxy should be responsible for removing the header in this case? If so, we can keep your change as it is.

tung2744: Right, let me change it to use body client_secret over basic auth instead of error

tung2744: > Or do you think the proxy should be responsible for removing the header in this case? If so, we can keep your change as it is.

I think the proxy should remove the header, but I also think it is good to implement it in defensive way.

tung2744:
Updated, thanks!

carmenlau (approved):

PR #5821, opened 2026-07-21
chore: Update deps

tung2744:
ref DEV-3713


carmenlau (approved):
Thanks for fixing the e2e test!

PR #5840, opened 2026-07-28
chore: Suppress gosec G710/G124 false positives

tung2744:
The golangci-lint bump to v2.11.3 (08c0cfb3b6) pulled in a newer gosec that added G710 (open redirect via taint analysis) and G124 (cookie missing Secure/HttpOnly/SameSite) checks not present in the previously pinned v2.5.0. Audited every flagged site: redirects are same-origin enforced, allow-listed against registered client URIs, or point at a server-configured host; cookie attributes are set via field assignment right after the literal gosec inspects. Added #nosec justification comments and synced .vettedpositions for the resulting line shifts.


tung2744:
@carmenlau See if you can prioritize merging this to unblock the ci, thanks!

carmenlau (approved):

PR #5837, opened 2026-07-28
Fix feature config merge to be field-level, not whole-section

carmenlau:
ref DEV-3738

Summary:

| # | Section | Merge behavior | Notes |
|---|---|---|---|
| 1 | `identity` | ✅ Field-level | 3-level cascade: `login_id`/`oauth`/`biometric` independent; within `oauth`: `maximum_providers`/`providers` independent; within `providers`: all 9 providers independent |
| 2 | `authentication` | ✅ Field-level | Cascades to `secondary_authenticators`'s one field (`oob_otp_sms`) — moot today (only one leaf), proactive fix |
| 3 | `authenticator` | ✅ Field-level | Cascades through `password` → `policy` to the 3 independent gates (`minimum_guessable_level`/`excluded_keywords`/`history`) |
| 4 | `custom_domain` | Whole-section replace | **Fine** — single field (`disabled`), nothing to clobber |
| 5 | `ui` | ✅ Field-level | `white_labeling`/`phone_input` independent |
| 6 | `oauth` | ✅ Field-level (already correct) | Cascades to `client`'s 4 independent fields — this was the reference pattern used for all the fixes above |
| 7 | `hook` | ✅ Field-level | `blocking_handler.maximum`/`non_blocking_handler.maximum` independent |
| 8 | `audit_log` | Whole-section replace | **Fine** — single field (`retrieval_days`) |
| 9 | `google_tag_manager` | Whole-section replace | **Fine** — single field (`disabled`) |
| 10 | `rate_limits` (top-level) | Whole-section replace | **Fine** — single field (`disabled`) |
| 11 | `messaging` | ✅ Field-level (already correct) + fixed nested gap | Top-level fields (`rate_limits`/`sms_usage`/etc.) independent; **and now** `rate_limits`'s 6 channels (`sms`/`sms_per_ip`/`sms_per_target`/`email`/`email_per_ip`/`email_per_target`) are independent too |
| 12 | `usage` | ✅ Field-level (already correct) | `hooks` appends across layers (deliberately, unique); `limits` merges per usage-name (`sms`/`email`/`whatsapp`/`user_import`/`user_export`) independently |
| 13 | `collaborator` | ✅ Field-level | `maximum`/`soft_maximum` independent |
[...]

PR #5842, opened 2026-07-29
Fix ui.phone_input.allowlist minItems:1 blocking explicit allow-all override

carmenlau:
ref DEV-3755 

require
- #5837


pkg/lib/config/feature_ui.go
tung2744: Should this be changed to omitzero? I think omitempty will omit the key if it is `[]`

tung2744:
Is this pr ready to merge? I remember we've discussed about removing `omitempty` from booleans. Will that be included in this pr?

carmenlau:
Yes, I've included them in the PR. Thank you!

carmenlau:
Rebased onto the latest `main`.

tung2744 (approved):

PR #5853, opened 2026-08-05
Reduce redundant queries

tung2744:
ref DEV-3760


pkg/lib/event/service.go
```diff
@@ -94,11 +94,15 @@ func (s *Service) DispatchEventOnCommit(ctx context.Context, payload event.Paylo
...
-		// then it is not possible to resolve the entity in DidCommitTx.
-		if err = s.Resolver.Resolve(ctx, payload); err != nil {
-			return
+		// A non-blocking payload is resolved again in WillCommitTx. Resolving it
+		// here as well is only observable when the later resolve cannot find the
+		// user: resolve.go swallows ErrUserNotFound and keeps whatever was
+		// resolved earlier. That happens exactly when the user row is gone by
+		// commit time, which the payload declares via DeletedUserIDs.
+		if len(typedPayload.DeletedUserIDs()) > 0 {
+			if err = s.Resolver.Resolve(ctx, payload); err != nil {
```
carmenlau: It seems this case can never occur, as we enforce the `resolve` tag and `DeletedUserIDs()` cannot coexist, as shown in the test case `no resolve:-tagged payload also declares a deleted user`.

tung2744: You are right, there are indeed no such case, maybe I will just remove this branch

tung2744: Added a commit for this, thanks!

carmenlau (approved):

PR #5858, opened 2026-08-12
Add script to add select_account to login flows

tung2744:
ref DEV-3785

### DIFF
--- a/cmd/portal/cmd/cmdinternal/migrate_resources_add_select_account.go
+++ b/cmd/portal/cmd/cmdinternal/migrate_resources_add_select_account.go
@@ -1,0 +1,135 @@
+package cmdinternal
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+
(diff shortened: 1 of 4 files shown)


cmd/portal/cmd/cmdinternal/migrate_resources_add_select_account.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	loginFlows, ok := mapGet[[]any](m, "authentication_flow", "login_flows")
+	if !ok {
+		return nil
+	}
+
+	updated := false
+	for _, loginFlow := range loginFlows {
+		flow, ok := loginFlow.(map[string]any)
```
carmenlau, line 70: Should we update all login flows, or just the default login flow only?

tung2744: All login flows, because there is a chance a client can be associated with a different flow using x_authentication_flow_allowlist. And I think adding select_account have no impact to the api response if no cookies is sent so it should be safe to add to all login flows.

carmenlau (approved):

PR #5866, opened 2026-08-18
 Fix authflow misconfiguration errors and messaging

carmenlau:
Promote `ErrStepNotFound`, `InvalidTargetStep`, and `InvalidFlowConfig` to non-recoverable errors with Error-level logging. Add `v2.error.flow-misconfigured` translation key wired into both `fatal_error.html` and `__error.html`. Add e2e test coverage for all three variants.

ref DEV-3759


fungc-io:
@carmenlau please update the translation

similar to 
```
"v2.page.oauth-provider-missing-credential.default.developer-warning": "[Developers] Missing configuration keys. Configure with your own keys in the Authgear Portal to enable {provider} login.",
```

These errors should be caught by the developer during their testing, so we use "[Developers]" at the beginning to indicate it's addressed to them

Change to:
```
"v2.error.flow-misconfigured": "[Developers] This Authentication Flow is misconfigured and can't be completed. Check your configuration in the portal."
```

Also, "Authentication Flow" should not be translated and should be kept as-is in other languages.

carmenlau:
Updated the copy!

tung2744 (approved):

PR #5870, opened 2026-08-19
Implement DCR

tung2744:
ref DEV-3792

How to test:

1. Run migrations: `go run ./cmd/authgear database migrate up`

2. Add the following config:

```yaml
oauth:
  dynamic_client_registration:
    enabled: true
    initial_access_token_required: false
```

3. Create resource and scope:

You can use graphiql to run the following mutations.

```
mutation {
  createResource(input: {
    resourceURI: "https://localhost:8090/"
    name: "test-mcp"
    accessPolicy: { allowDynamicThirdPartyClientAccess: true }
  }) { resource { id } }
}
```

```
mutation {
  createScope(input: {
    resourceURI: "https://localhost:8090/"
    scope: "read:tools"
    accessPolicy: { allowDynamicThirdPartyClientAccess: true }
  }) { scope { id } }
}
```

```
mutation {
  createScope(input: {
    resourceURI: "https://localhost:8090/"
    scope: "execute:tools"
    accessPolicy: { allowDynamicThirdPartyClientAccess: true }
  }) { scope { id } }
}
```

4. Clone and setup test-mcp https://github.com/tung2744/test-mcp:

```
npm install
npm run setup
AUTHGEAR_ENDPOINT=http://localhost:3100 npm start
```

5. Start mcp client:

```
export NODE_EXTRA_CA_CERTS={YOUR_DIRECTORY_TO_REPO}/test-mcp/certs/localhost.crt
npx @modelcontextprotocol/inspector
```

In the opened browser:

Add Servers -> Add manually ->

Select "Transport: streamable-http" and URL=`https://localhost:8090/mcp`

Then press Connect.

You should then see authgear login.


docs/specs/access-token-audience-binding.md
```diff
@@ -21,9 +22,19 @@ This is implemented via [RFC 8707 — Resource Indicators for OAuth 2.0](https:/
...
+- **Multiple `resource` values in one request.** Exactly one `resource` value is accepted; the `aud` claim therefore always contains a single URI. The multi-resource behaviour and the intersection-downscoping rule described below apply only once first-party multi-resource support is built.
+- **The `scope_by_aud` claim.** Not implemented. A resource-bound access token carries `aud` and the (unfiltered) `scope` claim only; there is no claim mapping individual scopes to individual audiences. This is safe today because at most one resource URI can ever appear in `aud` (see the point above) — every granted scope already applies to that single audience, so a per-audience breakdown carries no extra information. `scope_by_aud` becomes necessary once multiple resources can appear in one token; it is deferred alongside multi-resource support.
```
tung2744, line 31: In this pr, I did not implement `scope_by_aud`. Because I found it is not implemented in m2m as well, although it is in spec. And seems no real usecase need it at the moment.

docs/specs/access-token-audience-binding.md
```diff
@@ -21,9 +22,19 @@ This is implemented via [RFC 8707 — Resource Indicators for OAuth 2.0](https:/
...
 - [Relationship to M2M](#relationship-to-m2m)
 
+## Implementation Status
+
+Two parts of this specification are described in full below but are **not implemented**:
+
+- **First-party clients using `resource`.** Outside the `client_credentials` grant, a first-party client that sends `resource` receives `invalid_target`. Only **dynamic** third-party clients — DCR-registered today, CIMD-resolved later — can currently use the parameter with `authorization_code` / `refresh_token`. A static `third_party_app` client also receives `invalid_target`: it has no mechanism to be associated with a Resource for these grants (unlike an M2M client's explicit Client-Resource Association for `client_credentials`), so this is not a "not yet implemented" gap to close later — `allow_dynamic_third_party_client_access` is dynamic-only by design, per its name. First-party support is planned separately.
+- **Multiple `resource` values in one request.** Exactly one `resource` value is accepted; the `aud` claim therefore always contains a single URI. The multi-resource behaviour and the intersection-downscoping rule described below apply only once first-party multi-resource support is built.
```
tung2744, line 30: I did not support multiple resource at the moment, because it seems the current m2m client_credential flow did not support multiple resource as well. And no real usecase need it yet.

tung2744:
Rebased to main and added audit logs in d18ebc0...131dc10

e2e/tests/dcr/register_ratelimit_per_project.test.yaml
```diff
@@ -0,0 +1,53 @@
+name: POST /oauth2/register's per-project rate limit is project-configurable
+authgear.yaml:
+  override: |
+    oauth:
+      dynamic_client_registration:
+        enabled: true
+        initial_access_token_required: false
+        rate_limits:
```
carmenlau, line 8: I found that we have both rate limits and usage limits in `authgear.yaml`. How should we decide which one to use when introducing a new limit?

e2e/tests/dcr/resource_indicator.test.yaml
```diff
@@ -0,0 +1,814 @@
...
+    action: oauth_approve_consent
+    oauth_approve_consent_redirect_uri: "{{ .steps.authenticate.result.action.data.finish_redirect_uri }}"
+
+  - name: exchange_code
+    action: oauth_exchange_code
+    oauth_exchange_code_code_verifier: "{{ .steps.oauth_setup.result.code_verifier }}"
+    oauth_exchange_code_redirect_uri: "{{ .steps.approve_consent.result.redirect_uri }}"
+    oauth_exchange_code_client_id: "{{ .steps.register.result.http_json_body.client_id }}"
+
+  - name: refresh_without_resource
```
carmenlau, line 928: I'd like to know if there's any difference between refreshing without a resource and refreshing with the correct resource.

e2e/tests/dcr/resource_indicator.test.yaml
carmenlau, line 1: That would be great if the tests could also cover cases such as a first-party dynamic client and an invalid project endpoint in the resource, etc.

pkg/resolver/handler/resolve.go
```diff
@@ -66,6 +77,38 @@ func (h *ResolveHandler) resolve(ctx context.Context, r *http.Request) (*model.S
 
 	var info *model.SessionInfo
 	if valid && userID != nil && s != nil {
+		// A third-party client's opaque access token must not work here: it
```
carmenlau: I think it's worth adding an E2E test for the resolver endpoint if it's not too difficult.

pkg/lib/oauthclient/client_config.go
```diff
@@ -0,0 +1,67 @@
...
+		Name:                           c.DisplayName(),
+		ApplicationType:                appType,
+		RedirectURIs:                   c.RedirectURIs,
+		GrantTypes_do_not_use_directly: c.GrantTypes,
+		ResponseTypes:                  c.ResponseTypes,
+		ClientURI:                      derefOr(c.ClientURI, ""),
+		LogoURI:                        derefOr(c.LogoURI, ""),
+		TOSURI:                         derefOr(c.TOSURI, ""),
+		PolicyURI:                      derefOr(c.PolicyURI, ""),
+		IssueJWTAccessToken:            false, // fixed per client.md's DCR mapping table
```
carmenlau: I’d like to know which table in the spec you’re referring to.

I also think it would be clearer if we could have a summary table somewhere in the spec describing, under different conditions (e.g. first-/third-party client, dynamic/static, with/without resource):

1. Whether the token is JWT or opaque
2. Whether the resolver endpoint can be used
3. Whether the UserInfo endpoint can be used
4. Whether there is a consent screen
5. Anything else you think is worth documenting

What do you think?

carmenlau (commented):
The PR looks good to me so far. I only have a few comments on the tests and documentation. I haven't tested it yet, but I'll test it now.

e2e/tests/dcr/register_ratelimit_per_project.test.yaml
tung2744: I think rate limits and usage limits are handling different problems.
rate limits: How frequent an operation can be done.
usage limits: How many times of that operation can be done in a (billing) period.

And I think we do not have usage limits in authgear.yaml? They should be in authgear.features.yaml.

pkg/resolver/handler/resolve.go
tung2744: The problem is e2e tests did not start the resolver server. But I think it is doable.

pkg/lib/oauthclient/client_config.go
tung2744: Sure, let me add a comprehensive table for this behavior.

e2e/tests/dcr/register_ratelimit_per_project.test.yaml
carmenlau: Thanks for clarifying. 

We have usage limits in `authgear.yaml`, and they have a similar shape to `authgear.feature.yaml` by design.
https://github.com/authgear/authgear-server/blob/main/pkg/lib/config/usage.go

tung2744: oh yes. I missed this. The use case for this is to provide a way for project owner to monitor the usage of their own project, and get alert for a soft limit / set a billing cap.
So, same principle: Usage limit are things related to billing or quotas.

e2e/tests/dcr/resource_indicator.test.yaml
tung2744: There is no difference. In current design, omitting `resource` parameter will issue an access token with resource bound at authorization, which is same as providing the same resource parameter.

In future, if we support multiple resources, omitting `resource` parameter will mean ALL resources bound at authorization.

pkg/resolver/handler/resolve.go
tung2744: Turns out it was started with main server in the local setup.

e2e/tests/dcr/resource_indicator.test.yaml
tung2744: Modified the e2e tests to show this behavior

e2e/tests/dcr/resource_indicator.test.yaml
tung2744: Added cases for first part dynamic client and putting project endpoint in resource

pkg/lib/oauthclient/client_config.go
tung2744: Added a table in docs/specs/client.md for these behavior

carmenlau:
I tested the MCP flow and it works fine, with just one small thing. The MCP client opens http://127.0.0.1:xxxx for the inspector, while the client registration only accepts localhost for HTTP as the redirect_uri. I access the inspector through http://localhost instead, and the login works fine.

pkg/resolver/handler/resolve.go
```diff
@@ -66,6 +77,38 @@ func (h *ResolveHandler) resolve(ctx context.Context, r *http.Request) (*model.S
...
+			default:
+				// Every path that hands out an OfflineGrantSession as a
+				// ResolvedSession sets TokenType; an unset value means a new
+				// resolution path was added without setting it. Fail closed.
+				return &model.SessionInfo{IsValid: false}, nil
+			}
+		default:
+			return &model.SessionInfo{IsValid: false}, nil
+		}
+
```
carmenlau, line 122: - Need update resolver endpoint to allow first party client only and document the reason
- Update the summary table for the resolver endpoint behavior

tung2744: Added one more commit for this, thanks!
- Updated the table in document
- Updated implementation of resolver to reject all third party app tokens. First party app tokens are always accepted (The original implementation is correct for first party app).
- Added e2e tests for these cases.

carmenlau (approved):

PR #5874, opened 2026-08-19
[Site Admin] Fix bugs when an app has more than one owner

carmenlau:
An app can have more than one owner via a direct DB edit, which duplicated apps in search/list and left extra owners undemoted on promote. Fixes both, adds e2e regression test.

ref DEV-3757

### DIFF
--- a/pkg/siteadmin/auditlog/app_collaborator_promoted.go
+++ b/pkg/siteadmin/auditlog/app_collaborator_promoted.go
@@ -11,10 +11,28 @@ type AppCollaboratorPromotedPayload struct {
 	NewOwnerCollaboratorID string `json:"new_owner_collaborator_id"`
 	NewOwnerUserID         string `json:"new_owner_user_id"`
 	NewOwnerUserEmail      string `json:"new_owner_user_email"`
-	// DemotedEditor* fields are omitted when the app had no previous owner.
-	DemotedEditorCollaboratorID string `json:"demoted_editor_collaborator_id,omitempty"`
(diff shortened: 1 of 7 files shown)


pkg/siteadmin/auditlog/app_collaborator_promoted.go
```diff
@@ -11,10 +11,28 @@ type AppCollaboratorPromotedPayload struct {
...
-	DemotedEditorUserEmail      string `json:"demoted_editor_user_email,omitempty"`
+	// Deprecated_* fields: this event type already ships in production, so
+	// these are kept for older consumers/historical entries (same JSON keys
+	// as before -- only the Go names change), but new code should read
+	// DemotedEditors instead -- these three only ever describe the single
+	// (oldest) demoted owner, silently dropping any others. Omitted when
+	// the app had no previous owner.
+	Deprecated_DemotedEditorCollaboratorID string `json:"demoted_editor_collaborator_id,omitempty"`
+	Deprecated_DemotedEditorUserID         string `json:"demoted_editor_user_id,omitempty"`
+	Deprecated_DemotedEditorUserEmail      string `json:"demoted_editor_user_email,omitempty"`
```
tung2744: I think it is safe to just remove these fields? The fields will just be ignored in code and will not affect stored data.

carmenlau: Yes, removing them is harmless. Let me remove them.

carmenlau: Updated, thanks!

tung2744 (approved):

PR #5875, opened 2026-08-20
chore: Bump gomessageformat to upstream, drop the fork replace

carmenlau:
The double-apostrophe fix previously pulled in via the oursky/gomessageformat fork has been merged upstream into iawaknahc/gomessageformat. Point go.mod, custombuild/go.mod, and e2e/go.mod directly at the upstream commit and remove the replace directive, since replace directives do not propagate across module boundaries and custombuild/e2e need the fix independently.

ref DEV-3749


tung2744:
@carmenlau But I think we actually want to switch to the oursky fork?

tung2744:
I will merge this first for the fix, and create another issue for the switch

tung2744 (approved):

PR #5878, opened 2026-08-24
Add script to remove a project from database

tung2744:
ref DEV-3797

### DIFF
--- a/cmd/portal/cmd/cmddatabase/const_test.go
+++ b/cmd/portal/cmd/cmddatabase/const_test.go
@@ -1,0 +1,56 @@
+package cmddatabase
+
+import (
+	"sort"
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+
+	"github.com/authgear/authgear-server/pkg/util/sqlmigrate"
+)
+
+// tablesExcludedFromPrune mirrors the comment in const.go: these tables have
+// an app_id column but dumping/restoring/pruning them is deliberately not
+// supported (they mirror Stripe billing state, so bulk-deleting rows here
+// would desync it). A table only belongs in this list if that reasoning
+// still applies to it - do not add a table here just to silence this test.
+var tablesExcludedFromPrune = map[string]bool{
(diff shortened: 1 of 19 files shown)


cmd/portal/cmd/cmddatabase/const_test.go
```diff
@@ -0,0 +1,56 @@
...
+// tablesExcludedFromPrune mirrors the comment in const.go: these tables have
+// an app_id column but dumping/restoring/pruning them is deliberately not
+// supported (they mirror Stripe billing state, so bulk-deleting rows here
+// would desync it). A table only belongs in this list if that reasoning
+// still applies to it - do not add a table here just to silence this test.
+var tablesExcludedFromPrune = map[string]bool{
+	"_portal_historical_subscription": true,
+	"_portal_pending_domain":          true,
+	"_portal_subscription":            true,
+	"_portal_subscription_checkout":   true,
```
carmenlau, line 24: I’d like to know what happens if someone reuses a deleted project ID. Would this cause any problems in these tables?

tung2744: These records will be associated with the new project. Maybe I should stop the script and warn the user if there are something exist in these table instead.

tung2744: Per latest discussion we will:
1. Preserve the _portal_config_source row, such that no project with same app id can be created again.
1. Put a simple config as authgear.yaml & authgear.secrets.yaml, remove all other things in data.

tung2744:
@carmenlau Updated and rebased to main, thanks!

cmd/portal/cmd/cmddatabase/const_test.go
tung2744: Handled in [3fa5cca](https://github.com/authgear/authgear-server/pull/5878/commits/3fa5cca46376dfa55fc5603d3d578451a5665904)

carmenlau (approved):

PR #5879, opened 2026-08-26
Fix usage.limits override lost through deprecated-migration re-marshal

carmenlau:
An app-level `usage.limits.<name>: []` override (meant to clear an inherited quota) was silently discarded when a deprecated field like `messaging.sms_usage` was also enabled, the merge-fold's marshal/re-parse round trip dropped the empty slice due to `omitempty`, so `Migrate()` couldn't tell "explicitly cleared" from "never configured" and re-injected the legacy limit. Switches the tags to `omitzero` (matching the existing `phone_input.allowlist` fix) and adds a regression test covering the exact scenario.

ref DEV-3811

### DIFF
--- a/pkg/lib/config/feature_test.go
+++ b/pkg/lib/config/feature_test.go
@@ -335,6 +335,50 @@ collaborator:
 			Action: model.UsageLimitActionBlock,
 		}})
 	})
+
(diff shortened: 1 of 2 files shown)


tung2744 (approved):
lgtm, please rebase to resolve the conflicts

tung2744:
I will rebase it

tung2744 (approved):

PR #5881, opened 2026-08-27
[Portal] Add DCR configuration UI

fungc-io:
ref DEV-3801

Adds the portal UI for Dynamic Client Registration (#5870).

## What's included

**Client Applications › Dynamic clients (DCR) tab** — all DCR configuration as settings cards:
- Enable toggle (saves immediately; reveals the registration endpoint copy field)
- Registered clients count with usage against the plan's `oauth_client_dcr` cap, and **View all** → a paginated client list page (details dialog, delete with confirmation)
- Registration security: require-IAT toggle (confirmation dialog before allowing open registration) + Initial Access Token management — create (Third-party/First-party with high-privilege warning; preset expiries or a custom date-time down to the minute), one-time token reveal, revoke
- Default client configuration (token lifetimes)
- Allowed API resources: the shared resources/scopes flagged for dynamic third-party client access, linking to API Resources

**API Resources** — dynamic access policy controls:
- Resource › Applications tab: pinned "Allow dynamic third-party clients" toggle row (confirmation on turn-off)
- Scope create/edit: "Allow dynamic third-party clients" checkbox; "Dynamic" badge in the scopes list
- Fixes the raw `<ReactRouterLink>` tag rendering in the M2M applications empty state

## How to test

1. Enable registration on the Dynamic clients (DCR) tab; the registration endpoint appears.
2. Create a third-party IAT, copy it from the one-time reveal, then:
   ```sh
   curl -s http://localhost:3100/oauth2/register \
     -H "Content-Type: application/json" \
     -H "Authorization: Bearer iat_tp_..." \
     -d '{"client_name": "Test", "redirect_uris": ["https://example.com/callback"]}'
   ```
3. The client appears under View all; run the authorization code flow with `resource=`/scope after allowing dynamic access on a resource and scope.
4. Revoke the IAT → registration returns 401; delete the client → slot freed.


tung2744:
@fungc-io I've added some more commits for some issues I found, some changes the UI a bit. A summary:

  Bugs
  - Default-lifetime placeholders showed 30d/14d instead of the real 364d/30d — now read from effectiveConfig instead of duplicating server defaults
  - The registration toggle auto-saved, silently committing every other unsaved edit on the tab (and the sibling Applications tab, which shares the form)
  - M2M empty-state link rendered as plain grey text (bare <a> inside FluentUI Text + Tailwind preflight)
  - "Allowed API resources" silently truncated at 100 — now reports it, including in the empty case where the message was outright false
  - Require initial access token will be reset to turned on when you disable dcr automatically.
  
  Scope access policy UI
  - Create scope form brought to parity with the edit form: shared label, both description branches, checkbox moved below the fields
  - Scope list: unlabelled "Dynamic" chip → an "Access Policy" column, badge with the form's own wording on hover, policy-list shaped so future accessPolicy keys add a badge rather than a column 
  - Copy: "dynamic third-party" wherever the policy is described (5 strings) — "dynamically registered" wrongly includes dynamic first-party clients
  - Cards that only apply when DCR is on are now hidden when it's off

  Quality
  - Schema validation errors bind to the lifetime field that caused them
  - Invalid custom IAT expiry says why (past / unparsable / beyond the 32-bit ceiling) instead of just disabling Create
  - IAT create→reveal dialog handoff no longer leaves both modals open across a refetch


---

And there is a problem that I did not fix: When there is no applications in the project, no way to turn on or config dcr.

fungc-io:
@tung2744 
i've added a commit to fix the table header alignment
image

otherwise are ready to go.

tung2744 (approved):

PR #5887, opened 2026-09-02
Implement CIMD

tung2744:
ref DEV-3793

How to test:

1. Enable cimd. Add the following configs:

authgear.yaml

```
oauth:
  client_id_metadata_document:
    enabled: true
```

authgear.features.yaml

```
# These are needed to enable testing in local
oauth:
  client_id_metadata_document:
    insecure_http_allowed: true
    insecure_fetch_address_allowed: true
```

2. Create resource and scope for mcp server.  You can skip this step if you have already done it in DCR pr.

You can use graphiql to run the following mutations.

```
mutation {
  createResource(input: {
    resourceURI: "https://localhost:8090/"
    name: "test-mcp"
    accessPolicy: { allowDynamicThirdPartyClientAccess: true }
  }) { resource { id } }
}
```

```
mutation {
  createScope(input: {
    resourceURI: "https://localhost:8090/"
    scope: "read:tools"
    accessPolicy: { allowDynamicThirdPartyClientAccess: true }
  }) { scope { id } }
}
```

```
mutation {
  createScope(input: {
    resourceURI: "https://localhost:8090/"
    scope: "execute:tools"
    accessPolicy: { allowDynamicThirdPartyClientAccess: true }
  }) { scope { id } }
}
```

3. Clone and setup test-mcp https://github.com/tung2744/test-mcp:

Note, you need to pull the latest code.
The tls cert might need to be regenerated even if you have done it before.

```
npm install
npm run setup
```

After you run setup, you need to trust the cert in certs/localhost.crt. You can do it with the keychain app in mac.

```
AUTHGEAR_ENDPOINT=http://localhost:3100 npm start
```

4. Start inspector:

```
export NODE_EXTRA_CA_CERTS={YOUR_DIRECTORY}/test-mcp/certs/localhost.crt
npx @modelcontextprotocol/inspector
```

Go to http://127.0.0.1:6274/ to see the UI.

5. Enable CIMD in the inspector.

At the top right hand corner, there is a gear button. Press it to open a setting dialog. Set it like this:
Screenshot 2026-09-02 at 7 57 37 PM

Use `https://localhost:8090/oauth/client-metadata.json` as the metadata url.

6. Connect the MCP Server.

In the inspector UI:

Add Servers -> Add manually ->

Select "Transport: streamable-http" and URL=`https://localhost:8090/mcp`

Then press Connect.

You should then see authgear login.


pkg/lib/config/oauth_client_id_metadata_document.go
```diff
@@ -0,0 +1,86 @@
...
+
+// OAuthClientIDMetadataDocumentConfig carries no nullable tag, so
+// config.SetFieldDefaults force-allocates it (and ClientConfig below)
+// whenever the whole section is absent from authgear.yaml -- exactly like
+// OAuthDynamicClientRegistrationConfig. The nil-safe accessors below exist
+// only for code paths that read a config before SetFieldDefaults has run
+// (e.g. a test that yaml.Unmarshals a snippet directly); at runtime the
+// receiver is always non-nil.
+type OAuthClientIDMetadataDocumentConfig struct {
+	Enabled bool `json:"enabled,omitempty"`
```
carmenlau, line 49: - Would like to clarify the behaviour of the `enabled` flag. If the features (`dcr` / `cimd`) are disabled after some clients have already been registered, can those clients still be used?
-  I think the config should be designed to be secure by default to avoid misconfiguration. So:
    - When DCR is enabled, should `iat` be required by default?
    - When CIMD is enabled, should no domains be allowed unless explicitly configured (either allow all or specify the allowed domains)?

carmenlau (commented):
I’m still checking the E2E tests, so I’m still reviewing. Just one quick comment about the config, we can discuss it tmr.

Also, when I ran the tests locally, I got this error. I deleted the `certs` folder in `test-mcp` and re-ran `npm run setup`. Probably just an issue with my local setup, so we can check it tmr as well.

```
Get "https://localhost:8090/oauth/client-metadata.json": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

pkg/lib/config/oauth_client_id_metadata_document.go
tung2744: > Would like to clarify the behaviour of the enabled flag

I think for dcr, registered client continues to work. For cimd, I think fetched client should continue to work but will never be refetched. The current implementation disable fetched clients too which seems to be a bug.

> When DCR is enabled, should iat be required by default?

Yes, it is required by default.

> When CIMD is enabled, should no domains be allowed unless explicitly configured

Seems no, because no domain enabled actually means the feature is disabled? So actually enabled=true + allowed_domains=[] == enabled=false?

tung2744:
@carmenlau 

For the cert issue, you need to trust the cert in keychain.
Screenshot 2026-09-03 at 2 17 34 PM

pkg/lib/config/oauth_client_id_metadata_document.go
carmenlau: > I think for dcr, registered client continues to work. For cimd, I think fetched client should continue to work but will never be refetched. The current implementation disable fetched clients too which seems to be a bug.

Are there any use cases where developers may want to:

1. Turn off all dynamic clients without deleting them one by one.
1. Keep existing clients usable, but disable new client registration.

It seems like both are valid use cases. If so, maybe we should implement the stricter behaviour first and add the other option in the future if we need it? Thought?

> Yes, it is required by default.

Perfect!

> Seems no, because no domain enabled actually means the feature is disabled? 

Yes, you’re right. Maybe when we implement the Portal part, we should warn developers if the list is empty.

carmenlau:
> For the cert issue, you need to trust the cert in keychain.

It works, thank you!

pkg/lib/config/oauth_client_id_metadata_document.go
tung2744: > Turn off all dynamic clients without deleting them one by one.

I cannot think of one at the moment. I think even there is such usecase, we should not do it automatically by turning off the config. It should be a mutation to delete all dynamic clients instead.

> Keep existing clients usable, but disable new client registration.

I think this is the normal case, say, I developed a mcp server. I use dcr in the past, now I want to switch to cimd. So I will disable dcr, enable cimd. But I expect previously registered client continue to work.

e2e/tests/cimd/revoked_allowed_domains_keeps_existing.test.yaml
```diff
@@ -0,0 +1,64 @@
...
+    oauth:
+      client_id_metadata_document:
+        insecure_http_allowed: true
+before:
+  - type: custom_sql
+    custom_sql:
+      path: revoked_allowed_domains_keeps_existing_fixtures.sql
+steps:
+  # localhost is not *.example.com, so this host would never be allowed to
+  # onboard a NEW client -- but the seeded row is fresh, so
```
carmenlau: Is the client still usable after 1 hour? Will we use the stale record in this case?

tung2744:
1. The general rule is: What already registered should keep working. Disabling dcr / cimd only blocks new client being registered / client updates by refetch. Same rule apply to domain whitelist.
2. Portal might need alert text for the above rule, but not included in this pr.
3. Add e2e tests to show this behavior clearly.

tung2744:
About access token and refresh tokens, they should be same behavior as static clients:
- Refresh token always stop working immediately when client deleted (Because token endpoint always resolve client)
- Access token keep working until it expired

tung2744:
Updated:
1. Fixed the behavior of the enabled flag
2. Added a commit to fix a flaky test

Thanks!

carmenlau (approved):
The update looks good. There’s a small conflict in `pkg/lib/config/feature_usage.go`. Let me rebase it and merge.

carmenlau (approved):

PR #5893, opened 2026-09-04
Fix stale kernel CVE flagged in runtime image via linux-libc-dev

tung2744:
ref DEV-3817


rickmak (approved):

annaveniraviannavenikumaraswam-oss:
tung2744:dev-3817-osvuln se

PR #5900, opened 2026-09-08
[Portal] Add CIMD configuration UI

fungc-io:
ref DEV-3838

CIMD has backend support on main ([#5887](https://github.com/authgear/authgear-server/pull/5887)) but no way to configure it. This adds one, and reorganises the surface around it.

## What changed

**`Applications` becomes a nav group of two pages**, split by audience rather than by mechanism:

```
Applications
  Client Applications      configuration/apps          the static client list, unchanged
  AI Agents (Dynamic)      configuration/apps/agents    [ Overview | CIMD | DCR ]
```

- **Overview** — per-source client counts against their own quotas (`oauth_client_cimd` / `oauth_client_dcr` are counted separately), the allowed-API-resources card, and a way into the client list.
- **CIMD** — enable, Trusted domains ("Any domain" / "Only specific domains", the latter revealing a domain list), and token settings.
- **DCR** — unchanged in content: enable, registration endpoint, initial access tokens, token settings.

**Each tab owns its own `useAppConfigForm`** over just its config section. That is what makes the enable switches safe to save on toggle, which closes a real hole: an initial access token is created through the Admin API immediately, while `POST /oauth2/register` checks the *saved* config, so a token created before pressing Save came with a `curl` example that returned `403 access_denied`. The switch and the cards it governs both read the saved value, so what is on screen matches what the server will do.

**Client list** gains a Type column, a `?source=` filter and CIMD's `lastFetchedAt`, backed by a new `source` argument on the `dynamicClients` Admin API query which restricts both the page and `totalCount`.

**The `OAUTH_CLIENT_*` audit events** move out of the user audit log: every one of those payloads returns `""` from `UserID()`, so the Users Activities tab showed them with no user and its user-ID search could never match them.

## Reviewing

Two commits are worth knowing about before reading the diff:

- `9261cb783` adds two throwaway layout variants that `d55dbd82b` deletes — they existed so the three arrangements could be compared side by side in a running portal. Net effect on the tree is zero; the history keeps how the decision was made.
- `525957b09` is a behaviour-preserving extraction (`ApplicationsListSection`, `util/usageLimit`, the two mechanism content components) that the chosen layout builds on.

## Testing

Enable CIMD per [#5887](https://github.com/authgear/authgear-server/pull/5887)'s instructions, then configure it under **Applications → AI Agents (Dynamic) → CIMD** instead of by hand-editing `authgear.yaml`.

## Not in this PR

[DEV-3839](https://linear.app/authgear/issue/DEV-3839/portal-config-form-field-errors-persist-while-typing) — config-form field errors persist while typing, because `useAppConfigForm`'s `setState` never clears `updateError`. Portal-wide (~30 screens share the hook), so it wants its own change.


tung2744:
This PR makes the toggle save immediately again. Is this expected?

fungc-io:
Yes, this is to solve a UX quirk where if it doesn't save immediately 
1. Turn on DCR (Not saved)
2. Create token
3. Copy the curl code -> code does not work

tung2744:
@fungc-io Maybe lets do it this way:
off -> on: Immediately save as you implemented.
on -> off: Only make the form dirty, user need to press save to really disable.
What do you think?

tung2744:
And the action button menu "..." seems to be missing in the Dynamic client list, I think it should be a bug? Let me add it back.

tung2744:
Added a tooltip to show full client name on hover in case the name is long

tung2744 (approved):

PR #5905, opened 2026-09-09
[Portal] Sort project list and present it as a scrollable table

fungc-io:
ref [DEV-3841](https://linear.app/authgear/issue/DEV-3841/project-listing-should-be-sorted-by-alphabetical-order)

## What

The project list in the portal was rendered in the order the API returned, which made projects hard to find.

- Sort the list on the server: `AppService.GetAppList` now returns projects in ascending app ID order for every client, whichever config source produced the IDs, and the `appList` field description documents the order. The client-side sorts in the projects page and in the header's project selector are removed.
- Present the list as one bordered panel: title, search box and Create button, then tight two-column rows with the project ID on the left and its public origin on the right, divided by hairlines like the other v2 list tables.
- Cap the panel to the viewport and scroll only the rows, so the title and toolbar stay visible however long the list is.
- On mobile, the panel fills the area under the screen header edge to edge. The screen uses its own fixed-height container instead of `ScreenLayoutScrollView`, whose mobile override lets the page scroll instead.
- Fix a pre-existing mobile header bug this exposed: on pages without a hamburger button the logo kept its own left padding on top of the header's, landing 16px too far right. Also affects the project wizard and accept-invitation screens.

## Notes

- No change to the GraphQL shape or i18n strings; only the `appList` description changes, with the schema and generated types regenerated. Search still matches on both the ID and the origin.
- The page root prefers `100dvh` over `100vh` where supported so the fixed-height panel does not hide behind the iOS Safari toolbar; older browsers keep `100vh`. This needs a stylelint disable because the browserslist still includes iOS 14.5.

## Testing

- New unit test for the server-side sort in `pkg/portal/service`. `go test ./pkg/portal/... ./cmd/portal/...`, golangci-lint, and goanalysis pass on the touched packages.
- Full portal CI sequence passes locally: typecheck, eslint, stylelint, prettier, tests, gentype, build.
- Verified visually on desktop and mobile widths with 0, 2, and 50 projects, including the quota-reached callout.

## Screenshots

2 projects: 
SCR-20260909-mqfc

50 projects:
SCR-20260909-mqgv

Quota exceeded:
image

### DIFF
--- a/pkg/portal/service/app.go
+++ b/pkg/portal/service/app.go
@@ -10,6 +10,8 @@ import (
 	"net/url"
 	"path"
 	"regexp"
+	"slices"
+	"strings"
 	"time"
 
 	"github.com/lib/pq"
@@ -177,7 +179,18 @@ func (s *AppService) GetAppList(ctx context.Context, userID string) ([]*model.Ap
 			PublicOrigin: app.Context.Config.AppConfig.HTTP.PublicOrigin,
 		})
 	}
+	sortAppListItems(appList)
 	return appList, nil
+}
+
+// sortAppListItems orders the list by app ID in ascending byte order, so every
+// client shows projects in the same predictable order regardless of which
+// config source produced the IDs. App IDs are lowercase letters, digits and
+// hyphens, so byte order is alphabetical order.
+func sortAppListItems(items []*model.AppListItem) {
+	slices.SortFunc(items, func(a, b *model.AppListItem) int {
+		return strings.Compare(a.AppID, b.AppID)
+	})
 }
 
 // GetProjectQuota acquires connection.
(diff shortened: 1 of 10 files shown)


tung2744:
The sort should occur in server side. Maybe let me update the pr.

fungc-io:
Let me update that

fungc-io:
Moved the sort to the server: `AppService.GetAppList` now returns the list in ascending app ID order, the `appList` field description documents it, and both client-side sorts (projects page and the header's project selector) are removed. Covered by a unit test in `pkg/portal/service`.

pkg/portal/service/app.go
```diff
@@ -177,9 +179,20 @@ func (s *AppService) GetAppList(ctx context.Context, userID string) ([]*model.Ap
 			PublicOrigin: app.Context.Config.AppConfig.HTTP.PublicOrigin,
 		})
 	}
+	sortAppListItems(appList)
```
tung2744, line 182: Better to sort it in the query so we can do pagination. But that can be easily modified in future so I will merge the pr

tung2744 (approved):

PR #5907, opened 2026-09-09
Rate limit Admin API GraphQL mutations

carmenlau:
Adds two buckets, charged one token per top-level mutation field so batching doesn't buy writes at a discount:

- `admin_api.mutation.all.per_project` - 300/minute
- `admin_api.mutation.all.per_ip` - 150/minute

ref DEV-3835


pkg/lib/config/testdata/merge_feature.yaml
```diff
@@ -166,6 +166,26 @@ configs:
...
       sms:
         suppressed: true
+  # admin_api.rate_limits: one layer sets only per_project, a later layer sets
+  # only per_ip. Both must survive, and neither may reset create_session_enabled
+  # or the usage limits set by the earlier admin_api layers above -- that is the
+  # AdminAPIFeatureConfig -> RateLimits -> Mutation -> All cascade working.
+  - admin_api:
+      rate_limits:
+        mutation:
+          all:
```
tung2744, line 179: The word `general` is used in existing rate limits so I prefer call it general here.

pkg/lib/config/testdata/merge_feature.yaml
```diff
@@ -166,6 +166,26 @@ configs:
   - test_mode:
       sms:
         suppressed: true
+  # admin_api.rate_limits: one layer sets only per_project, a later layer sets
+  # only per_ip. Both must survive, and neither may reset create_session_enabled
+  # or the usage limits set by the earlier admin_api layers above -- that is the
+  # AdminAPIFeatureConfig -> RateLimits -> Mutation -> All cascade working.
+  - admin_api:
+      rate_limits:
+        mutation:
```
tung2744, line 178: Is there any reason we rate limit mutations only? I thought we are adding a rate limit for all admin api including query?

pkg/lib/config/testdata/merge_feature.yaml
carmenlau: The version in this PR counts **all** mutations together and uses the same `all` bucket, which is different from `authentication.general`. In authentication, `general` is the fallback for the other rate limits.

So my thought is that if we have `admin_api.mutation.create_group.per_project` in the future, `admin_api.mutation.general.per_project` could be the fallback.

I’d also like to confirm whether the version that counts all mutations together is fine. That means different operations will contribute to the same bucket. Or do you think different operations should be counted separately?

references: 

https://github.com/authgear/authgear-server/blob/50f7d7f1759479342383f83099c406a6e568e4c9/pkg/lib/ratelimit/ratelimits.go#L446

https://github.com/authgear/authgear-server/blob/50f7d7f1759479342383f83099c406a6e568e4c9/pkg/lib/ratelimit/ratelimits.go#L702

https://github.com/authgear/authgear-server/blob/50f7d7f1759479342383f83099c406a6e568e4c9/pkg/lib/ratelimit/ratelimits.go#L229

tung2744: 1. What I originally thought is rate limit counted per api. But the `all` bucket you added seems also make sense, but if all mutation shares the same bucket I think the default rate should be higher. Say 1000/minute per project. The default should be a value that really difficult to reach. We will tighten the limit for free projects.
2. I think per_ip rate limit is not needed. The per_ip rate limit in authentication flows are to prevent brute force on authenticators like password or account enumeration, in admin api there seems not such problem. It is likely called by the admin trusted server.

carmenlau: Updated to 1000/minute per project and dropped per_ip rate limit in commit [954a9bb](https://github.com/authgear/authgear-server/pull/5907/commits/954a9bb6661b1b093f176d9346c70bcc94aa7a3c) and [e1b894f](https://github.com/authgear/authgear-server/pull/5907/commits/e1b894fb33106fc1ebc711a995d7a6ed5026c4a5)

pkg/lib/config/testdata/merge_feature.yaml
carmenlau: Updated the spec to explain it in https://github.com/authgear/authgear-server/pull/5907/changes/5718da36db4db9c4050072a6b282bb961a635571, only `mutation.all` is implemented in this PR

carmenlau:
Added one more commit [adff51b](https://github.com/authgear/authgear-server/pull/5907/commits/adff51b82535335c053867625a6c4d79d99ee6ca) to handle rate limit error in the portal.

Thank you!

docs/specs/rate-limit.md
```diff
@@ -126,13 +127,189 @@ Some rate limits uses another rate limit config as a fallback if it is not set.
...
+identities, authenticators. None of those creations is individually capped, and
+before this limit none was rate limited either, so a caller's write rate was
+bounded only by how fast it could issue requests.
+
+**What is counted.** One token per **top-level mutation field** of the executed
+operation, taken before the operation is executed. A GraphQL document may carry
+several mutation fields, and aliases allow the same field to repeat, so charging
+per HTTP request would let batching buy writes at a discount. That discount is
+bounded but real: `graphqlutil` already refuses a document with more than
+`maxMutationFieldsPerRequest` (5) top-level mutation fields, so per-request
```
tung2744, line 165: Implementation detail should not be in spec. Spec should only describe the api.
Generally, I think this paragraph is too verbose. What need to be mentioned here seems to be:
1. The rate limit is counted per mutation operation, not per request.
2. When error, the rate limit name returned in the error.

pkg/lib/config/testdata/parse_feature_tests.yaml
```diff
@@ -327,3 +327,67 @@ config:
...
+  /admin_api/rate_limits/mutation/all/per_ip: 
+config:
+  admin_api:
+    rate_limits:
+      mutation:
+        all:
+          per_ip:
+            enabled: true
+            period: 1m
+            burst: 5
```
tung2744, line 393: No need these tests I guess.

carmenlau:
Simplified the spec and removed the unnecessary test, thanks!

tung2744 (approved):

PR #5908, opened 2026-09-10
Support loopback redirect URIs with arbitrary ports

tung2744:
ref DEV-3552


tung2744:
Included in https://github.com/authgear/authgear-server/pull/5913

PR #5911, opened 2026-09-10
[AuthUI] Restructure the consent screen's requested permissions

fungc-io:
Rebuilds the requested-permissions block on the OAuth consent screen.

The scopes were one flat `list-inside` list, so a wrapped item returned to the bullet's column instead of aligning under its own text, and the list inherited the centred alignment it then had to override. It also merged two unrelated things: what any sign-in to the project grants, and what this client asked for on top of it.

### What changed

**Structure.** Two labelled sections in a bordered panel. The border is what makes the block's start alignment read as deliberate — the title, identity and legal line follow the project's `alignment-content` setting, and a bulleted list cannot.

**Copy**, in the same commit because the labels and the bullets are one sentence: the label carries the verb ("This app will be allowed to access") and each bullet is its object, so `Access` leaves the bullets.

Three bullets went away:
- `openid` restated the sign-in the user was performing and appeared on every consent screen ever shown. With no other scope requested the whole list goes too, leaving a sign-in confirmation rather than "requesting access to:" above one uninformative line.
- `offline_access`, whose "Maintain access to data you have given it access to" no reader parsed.
- `profile`/`email`/`phone`/`address` when `full-userinfo` is granted — it allows every claim (`ScopeAllowsClaim` returns true unconditionally), so listing both stated the same access twice, and each narrower bullet read as a limit that wasn't there.

**Client website.** `client_uri` now appears beside the policy and terms links, which arrive through the same registration path and were already shown. The seven combinations of the three optional URIs are one ICU `select` over a template-computed variant, following `v2.component.toc-pp-footer.default.label`, rather than seven keys a translator would have to find separately.

**Translations** via `make generate-translations` for the five new keys and the six reworded ones. Worth noting the generator only fills *missing* keys, so a reworded en string leaves every locale stale — the third commit clears and regenerates those six. The results restructure rather than transliterate: `ja` separates the list with `、`, `de` moves the possessive to `… von {clientName}`, `el` declines the article per noun.

### Known gap, not addressed here

`$isSignInOnly` is computed from named and resource scopes only, so a request for `openid offline_access` renders as a bare sign-in with a "Continue" button and no list:

```
scopes=openid,offline_access   list=NO  button=Continue           -> Sign in to X | user@example.com
scopes=openid,email            list=yes button=Allow and continue -> ... | Your email address
```

Before this PR that request rendered two bullets and "Allow and continue". Dropping the `offline_access` bullet was deliberate; this interaction wasn't, and `offline_access` is the scope that grants a refresh token. `device_sso` has the same shape. Fixing it needs a copy decision about what the bullet should say, so I've left it — happy to do it here if you'd rather.

### Also worth a decision

Seven translation keys are removed. A project can override translations per app (`AppFS` layer), so an override of a removed key silently stops taking effect on upgrade — arguably a `docs/BREAKING-CHANGES.md` entry under `## [Unreleased]` by that file's own criterion ("a behaviour change a deployment could be relying on"). There's no precedent entry for key removal, so I haven't added one.

### Testing

Rendered every branch of the template against a local project: sign-in-only, one section, both sections, `full-userinfo` + resource scopes, the CIMD hostname banner, a client logo, and all eight combinations of website/policy/terms.

CI parity: `authui` typecheck, eslint, stylelint, prettier, test, build all pass; `gofmt`/`go vet`/`go test` clean on the touched package; `make check-tidy` exits 0; `lint-translation-keys` matches its expect file; `translationlinter` clean.

Two of those were failing mid-review and are fixed in the last commit — `gotemplatelinter` forbids `{{ template }}` on a non-`.html` name, and `subset_material_icons.py` had picked up `floor`, `group`, `groups`, `move` and `print` from my comment prose and a class name, adding five unused glyphs to the icon fonts shipped to every end user.


fungc-io:
@tung2744 
the layout has been updated

For CIMD, the screen would include the hostname like so:
image

tung2744 (approved):

tung2744 (approved):

PR #5914, opened 2026-09-11
[Migration Needed] Fix users list sorting by the previous login instead of the last login

fungc-io:
ref [DEV-3852](https://linear.app/authgear/issue/DEV-3852/portal-users-list-sorts-last-login-by-the-previous-login-not-the)

## Problem

In the portal's Users list, sorting by **Last Login** produced an order that did not match the displayed dates, and a user who had just logged in did not rise to the top.

`_auth_user` has two login columns. `login_at` is the most recent login and is what the Admin API exposes as `lastLoginAt`. `last_login_at` is the login **before** that: `UpdateLoginTime` copies `login_at` into `last_login_at` before writing the new time. The list without a search keyword or filter is served from `_auth_user`, and the sort helper used the `SortBy` value (`last_login_at`) verbatim as the ORDER BY column. So the list sorted by the previous login while showing the latest one.

The search-backed list (a keyword or a role/group filter set) reads `_search_user`, whose `last_login_at` holds the latest login, and was already correct. That is why typing a keyword made the order right.

## Fix

- `SortOption` gains `ApplyWithColumns`, which resolves the sort key through a per-table column map; `Apply` keeps its previous behaviour for the search store.
- The `_auth_user` store maps `SortByLastLoginAt` to `login_at`.
- Migration: index `(app_id, login_at DESC NULLS LAST)` to support the corrected sort, and drop `_auth_user_app_id_last_login_at`, which nothing sorts by any more.

No change to the GraphQL API: the `LAST_LOGIN_AT` enum and the `lastLoginAt` field are unchanged. Bug fix, so no breaking-changes entry.

## Notes for deployment

The migration creates one index and drops one on `_auth_user`, in the same style as the 2024 index migration. As with that migration, the index build holds a write lock on `_auth_user` for its duration on large tables.

## Testing

- New unit tests for the column resolution and the cursor filter in `pkg/lib/authn/user`.
- Migration verified with an up, down, up round trip on a database with the current schema; the planner uses the new index for the corrected sort.
- `go build ./pkg/... ./cmd/...`, `go test` for the user, search, admin, and elasticsearch packages, golangci-lint, goanalysis, and the goimports check all pass.

### DIFF
--- a/cmd/authgear/cmd/cmddatabase/migrations/authgear/20260911161002-add_index_auth_user_app_id_login_at.sql
+++ b/cmd/authgear/cmd/cmddatabase/migrations/authgear/20260911161002-add_index_auth_user_app_id_login_at.sql
@@ -1,0 +1,11 @@
+-- +migrate Up
+-- The Admin API sorts users by their most recent login, which _auth_user
+-- stores in login_at (last_login_at holds the login before that). Index the
+-- column actually sorted on, and drop the index on the column no longer used
+-- for sorting.
+CREATE INDEX _auth_user_app_id_login_at ON _auth_user (app_id, login_at DESC NULLS LAST);
+DROP INDEX _auth_user_app_id_last_login_at;
+
+-- +migrate Down
+CREATE INDEX _auth_user_app_id_last_login_at ON _auth_user (app_id, last_login_at DESC NULLS LAST);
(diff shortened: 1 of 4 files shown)


tung2744:
Added commit for more index, and do not drop existing index.

tung2744 (approved):

PR #5921, opened 2026-09-17
Allow all clients to use resource indicator

tung2744:
ref DEV-3856

PR #5922, opened 2026-09-17
[Portal] Redesign the API resource details page for the four-category access policy

fungc-io:
Portal UI for the four-category access policy that #5921 added to Resources and Scopes. Rebased onto `main` now that #5921 has merged, so this PR is just its own 7 commits.

## What changes

- **API resource details page** gets Settings and Scopes tabs; the Scopes tab uses the full-width layout. The header shows an icon, the resource name, and the identifier with a copy button.
- **Application access policy card** replaces the separate Applications and Dynamic clients cards. It has two blocks that follow the two grant types: **User-delegated access**, a table with one row per client category and an Allowed toggle (saves on toggle, confirms when turning off, toasts the setting written), and **Machine-to-machine access**, the existing list of authorized M2M applications.
- **Scope forms** (the add panel and the Edit Scope dialog) choose categories with pill toggles. A new scope starts from the resource's current policy. A warning names any selected category the resource itself does not allow, since the two-level check makes such a scope unreachable.
- **Scope list** shows one badge per allowed category, flags unreachable ones with a warning icon, and shows an Admin-API-set static third-party allowance as a read-only badge.
- **Bulk change**: select scopes in the list, then "Change access" opens a dialog whose chips reflect the selection's aggregate (on / off / mixed). Applying issues one `updateScope` per scope that actually changes, sending only the changed keys; failures keep those scopes selected for a retry.
- **Shared `FieldLabelWithTooltip`** component replaces two local copies (SSO widget, Anonymous Users) and the ghost-button variant used in API resources.
- **DCR switch** now writes `initial_access_token_required: false` when enabling, so registration starts open.
- Test card copy reworded around machine-to-machine applications; scope Description label explains that the text is shown on the consent screen.

## Deliberate deviations from `docs/specs/api-resource.md`

Reviewers will notice these; each is a product decision, not an oversight, and the spec should be updated if they stand.

1. Enabling DCR from the portal starts as open registration (spec default keeps the initial access token required).
2. A new scope copies the resource's access policy (spec: "no such pre-selection at the Scope level").
3. Create API Resource pre-selects dynamic first-party clients as well as static first-party (spec: static first-party only).
4. Static third-party clients have no control in the portal, because the `third_party_app` type cannot be created on the Client Applications page. The scope list still shows the key read-only when set via the Admin API; the resource-level table hides it.

## Verification

Portal CI sequence run locally on the rebased tree: `typecheck`, `eslint`, `stylelint`, `prettier`, `test` (35 suites, 279 tests), `gentype`, `build`, all passing, with `gentype` and `build` leaving the tree clean. No Go, GraphQL, or config changes; no `BREAKING-CHANGES.md` entry needed.


fungc-io:
@tung2744 this is ready for review

tung2744 (approved):

PR #5928, opened 2026-09-18
[Portal] Add light, dark and device appearance modes

fungc-io:
Adds an appearance preference to the portal — Light, Dark or Device — and makes the existing UI theme-aware.

## What it does

A three-icon segmented control in the header, moving into the nav drawer on mobile so the project name keeps its width. The choice is stored per device in `localStorage` and defaults to Device, which follows the OS setting and keeps following it live. Two tabs stay in sync.

The resolved mode is applied as a `light-theme` / `dark-theme` class on `<html>`, which Radix Themes and Tailwind both cascade from. An inline script in `index.html` applies it before the bundle loads, so a returning user never sees a flash of the wrong theme.

## Notes for review

- **The custom palette needed a dark half.** It shipped a custom slate scale under `:root` only. `:root` matches `<html>` regardless of the class and outranks Radix's own `.dark` by source order, so without the new block dark mode rendered almost entirely light. The wide-gamut block is converted colorimetrically from the same values; omitting it would leave P3 displays on the old colours with no visible signal. A test fails if any light token lacks a dark counterpart — nothing else in CI catches that.

- **`dark` was already taken.** `components/v2` uses a bare `dark` class to mark a component as sitting on a dark island (the `darkMode` props, and the onboarding survey's wrapper). Putting the page-level appearance on the same class made every stepper in the portal pick up the treatment meant for the blue gradient. The page class is `dark-theme` for that reason, and a test pins the distinction.

- **Step 9 is a fill, not a foreground.** ~46 declarations across 39 files used the solid accent as an icon or text colour, which glares on a dark card. They use step 11, the readable-on-background accent, which is what the rest of the portal already did. `Stepper` keeps step 9 because its glyph sits on a white fill it paints itself.

- **Both logos are now SVG.** `logo-inverted.png` carried an opaque white background that drew a box on anything but a white header. Framed to the old ink-to-padding ratio so nothing changes size. Note these assets were never served in development — nginx sends everything but `/api` to Vite — which is why the component used to hardcode the production CDN; a dev-only middleware now serves them from the resource directory.

- **Deliberately left light:** the AuthUI branding previews, the Get Started hero cards, `WhiteButton`, and the onboarding gradient, which now renders identically in both appearances.

- **One thing to decide:** the built-in logo PNGs are deleted. Those paths are overridable per deployment through the system-config translations, so a deployment that pointed a logo key at the built-in PNG path would get a broken image. Narrow, and nothing in the repo references them, but it may warrant a `BREAKING-CHANGES.md` entry.

Conventions are recorded in `portal/docs/FRONTEND.md` under Colour tokens: which accent step belongs to which job, the `Stepper` exception, and how to write a dark-only rule. Nothing in CI catches a wrong step or a stray hex.

## Testing

`typecheck`, `eslint`, `stylelint`, `prettier`, 296 tests, `gentype` (no drift) and `build` all pass. Verified by hand in both appearances across Getting Started, Users, Applications, Audit Log, the login-method settings, the project wizard, the onboarding survey and the invitation flow, at desktop and mobile widths.

## Screenshots
image
image
image


tung2744:
@fungc-io Is this ready for review?

fungc-io:
@tung2744 yes, this is ready for review

PR #5932, opened 2026-09-21
[AuthUI] Centre settings dialogs in the viewport on long pages

fungc-io:
ref [DEV-3855](https://linear.app/authgear/issue/DEV-3855/dialog-in-settings-moved-down-if-the-list-is-long)

## Problem

In `/settings`, when the Signed-in devices list is longer than the screen, the Terminate session dialog opens below the fold instead of centred in the viewport.

The dialog backdrop is `position: absolute` inside `.layout--default`, the page wrapper the dialog controller marks `relative` on open. That wrapper grows with the page, so the backdrop spans the whole document and `justify-center` centres the dialog in the document, not in the viewport. On a short page the two coincide, which is why it looked fine until the list got long.

`position: fixed` is not an option: DEV-3003 moved away from it because iOS 26 in-app web sessions render a fixed element under the browser's bottom bar.

## Fix

- On open, the controller covers only the part of the host that is on screen: it sets the backdrop's `top` and `bottom` inline from the host's bounding rect and the viewport height. The stylesheet keeps `inset-0` as the resting rule.
- Page scroll is locked on `<html>` while any dialog is open, so that box stays put without a scroll listener. The lock is counted, so a dialog opening while another fades out does not unlock early.
- The lock is released as soon as the dialog is asked to close, not on `transitionend`, and again on disconnect, because a successful submit swaps the page and removes the `data-turbo-temporary` dialog before its closing transition ends. The inline offsets are reset when the fade finishes.

Mobile behaviour is unchanged: the bottom sheet still slides up from the bottom of the visible area, which is what DEV-3003 wanted.

## Testing

- Verified on the local dev server with 25 seeded sessions: the dialog centres at any scroll position on desktop width, the page cannot scroll while it is open, and scrolls again after cancel or after terminating a session.
- Full AuthUI CI sequence passes: typecheck, eslint, stylelint, prettier, tests, build.
- Not verified here: iOS 26 in-app web session. QA please check the passkey removal dialog from DEV-3003 still sits above the browser bottom bar.


fungc-io:
Screenshot:
image

tung2744 (approved):

PR #5933, opened 2026-09-21
Show authorized third-party apps and their granted permissions in the portal and Auth UI

fungc-io:
ref [DEV-3863](https://linear.app/authgear/issue/DEV-3863/authorized-third-party-clients-shown-as-blank-in-the-portal)

## Problem

A grant to a DCR- or CIMD-registered client rendered as "-" in the portal's user details Authorized apps table, because the table looked clients up in `authgear.yaml` only. Its Permission column knew one string (full profile access) or "-", so resource scopes such as `read:tools` were invisible. The Auth UI's `/settings/authorized-apps` had the same blind spot: its per-row summary covered the four identity scopes and nothing else.

## Portal

- Resolve app names through the existing `dynamicClients` query after the static client list, falling back to the raw client ID rather than a dash. The sessions table on the same tab shares the resolution.
- Show granted scopes as badges, capped at two with a "+N more" badge; `openid` and `offline_access` are hidden as protocol scopes. A grant with nothing else reads "Sign-in only".
- A row opens a details dialog: client ID with copy, client type and kind, metadata links, the full scope list, when access was authorized and last extended, and Revoke behind the existing confirmation. The user query now selects `Authorization.updatedAt`, which the Admin API already exposed.

## Auth UI

- Each row on `/settings/authorized-apps` links to a new page, `/settings/authorized-apps/view?q_authorization_id=…`, with the client name as the page title.
- The page lists what the app was granted in the consent screen's own layout and words: identity scopes from the consent strings, resource scopes as the project's configured description with the raw scope name as fallback. When access was given is shown below, and Remove access sits behind the existing confirmation. A grant that no longer exists (typically the back button after removing it) redirects to the list.
- Backend: `Store.ListScopesByNames` maps stored scope names back to configured scopes in one query; the list handler becomes GET-only and a new detail handler owns the remove action, both gated by the same third-party filter the list applies.
- Navbar titles and opted-in settings-item labels clip with an ellipsis instead of wrapping, so arbitrary app names fit. The Material Symbols subset drops a stray `join` glyph.

## Notes

- No GraphQL schema or config changes. Bug fix plus UI, so no breaking-changes entry.
- Eight new English strings; the non-English locales are generated in their own commit per the translation workflow.
- A grant does not record which resource each scope came from, so scopes from two APIs appear in one list; grouping by API would need the grant to store the resource and is a follow-up. Exposing `clientName` on the Admin API `Authorization` type, and third-party grants' token activity, are also follow-ups.

## Testing

- New unit tests for the scope-to-permission mapping.
- Full CI sequences pass locally for `portal` (typecheck, eslint, stylelint, prettier, tests, gentype, build), `authui` (same plus build), and Go (golangci-lint, goanalysis with updated vetted positions, template linter, translation linter, tests, build, wire check).
- Verified end to end on the local dev server with seeded DCR grants: name resolution and badges in the portal, the details dialog, the Auth UI list and detail page at desktop and mobile widths including a long generated client name, and removing a grant through the confirmation, landing back on the list.


fungc-io:
screenshots:

SCR-20260921-ntlh

SCR-20260921-okvj

tung2744:
@fungc-io It seems we did not show the corresponding resource of the scope. Say, now permission show `execute:tools` but I do not know what it means if I do not know it is belongs to the mcp server. It can be solved by a detail description tho.
Screenshot 2026-09-24 at 4 14 00 PM
Screenshot 2026-09-24 at 4 13 48 PM

tung2744:
@fungc-io I've updated the PR for a few bug fixes. You can work on top of those commits, thanks!

fungc-io:
@tung2744 , seems this will need an API change,
the authorizations store scope names only, so the portal can't tell which resource they came from. I've added the resource in the permission list, but all resources that share the same scope names will be shown, which is not ideal.

PR #5934, opened 2026-09-21
[Portal] Fix the scope list clipping and the misaligned delete button in TextFieldList

fungc-io:
ref [DEV-3864](https://linear.app/authgear/issue/DEV-3864/the-trash-icon-mis-positioned-in-the-cimd-config-page)

Two portal layout fixes, one commit each.

## Scope list clipped its right-hand columns

On the API resource details page, the Scopes table was cut off on the right with no way to reach the Access Policy and actions columns. `CardTable`'s frame was always exactly the viewport width, so once the columns could not shrink further the row overflowed it and the border-radius clip hid the rest, leaving the scroll wrapper nothing to scroll.

`CardTable` now lets a caller declare the total of its column minimums as the table's own minimum width, so the card scrolls horizontally on a narrow viewport instead. The scope list declares that total, and its access policy column grows from 8rem to 12rem to fit the longest badge with its warning icon, since the badges do not truncate.

## Delete button riding above the input in TextFieldList

A ghost `IconButton` has no height of its own, because Radix cancels its padding with equal negative margins, so top-aligning it against the row left it centred 8px above the size-2 input. The button now sits in a box as tall as the input. Rows stay top-aligned so a field showing a validation message keeps its delete button beside the input rather than below it.

Affects the OAuth client edit form, the SAML application form, and the CIMD metadata documents tab.

## Testing

- Verified on the local dev server: the Scopes table scrolls inside its card at a narrow width with whole badges, and the delete icon sits level with its input on the three pages that use `TextFieldList`.
- Portal typecheck, eslint, stylelint, and prettier pass.


fungc-io:
Screenshots: 

image

image

tung2744 (approved):

